PR #17, opened 2014-12-02
README.md: additional noncore platforms

smikes:
A problem that prevents `node` releases getting packaged for debian and ubuntu is the need to compile on all required debian platforms.  Currently new versions of `node` are breaking on MIPS and GNU hurd, mainly due to minor libuv problems (eg missing headers, undefined constants).

Aiming to at least test those platforms would help keep fresh `io.js` versions coming to debian/ubuntu users through "official" channels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ The following platforms will eventually be included in the build and **test-all*
   * MinGW 32-bit
   * MinGW 64-bit
   * POWER8 (pending V8 changes and acceptance by the core team)
+  * GNU Hurd
+  * MIPS
 
 
 CI Software


kenperkins:
What's the status of this PR? The last explicit comment was @bnoordhuis with a :-1: on Hurd.

bnoordhuis:
I think we can close this.  Bug reports or pull requests for the MIPS build issues never materialized and no one's going to commit to Hurd.

smikes:
Yes, if I get back to the debian thing I will circle back and create a new issue/PR.

PR #25, opened 2015-01-01
Added iojs.org web setup Ansible playbook

rvagg:
This PR adds an Ansible playbook to set up an iojs.org web server. It's already been run against host `104.236.136.193` which just needs an SSL certificate and an nginx restart.

The server is listening to a GitHub Webhook for [iojs/iojs.github.io](https://github.com/iojs/iojs.github.io) and any commits to master will trigger a sync to the served files, so it should be pretty autonomous.

@indutny you have root access to 104.236.136.193. the certificate chain goes in */etc/nginx/ssl/iojs_chained.crt* and the key in */etc/nginx/ssl/iojs.key*, nginx can be restarted with `service nginx restart`. If you want to change the server setup at all can you either edit these setup files or at least provide details in this repo about what you've done so we can have the setup scripted properly.

### DIFF
--- a/setup/www/README.md
+++ b/setup/www/README.md
@@ -1,0 +1,25 @@
+# io.js Website Setup
+
+This directory contains an Ansible playbook to set up the iojs.org website from a fresh Ubuntu host (Upstart configuration assumes 14.04 or similar).
+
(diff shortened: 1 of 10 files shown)


rvagg:
done, now live on iojs.org with webhook active for the iojs.github.io repo

PR #26, opened 2015-01-01
Dockerfile & instructions for generating debs for GCC 4.9 on weezy

rvagg:
This is a Dockerfile to make a Docker image based on `debian:weezy` that installs the source of GCC 4.9 from jessie but all of the dependencies from weezy. I've made a patch for the `debian/control` file to make it compile.

When you run the Docker image and pass it an `/opt/debs` volume (i.e. just something local) it'll do the full compile & test of GCC 4.9 and dump all the debs into that volume.

This only works with ***gcc-4.9-4.9.2*** which is what's in jessie right now but it's given me binaries which I've hosted in a repo at https://deb.nodesource.com/weezy-gcc49 and can be installed on a weezy server something like this:

```
apt-get install -y apt-transport-https curl
echo 'deb https://deb.nodesource.com/weezy-gcc49 weezy-gcc49 main' >> /etc/apt/sources.list
curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
apt-get update
apt-get install -y g++-4.9 gcc-4.9 cpp-4.9
ln -sf /usr/bin/gcc-4.9 /usr/bin/gcc
ln -sf /usr/bin/gcc-4.9 /usr/bin/cc
ln -sf /usr/bin/g++-4.9 /usr/bin/g++
```

Which will compile the current io.js/v0.12 (with the addition of `python` & `make` packages) and pass _most_ of the tests (failures are unrelated to compile afaik).

### DIFF
--- a/setup/debian-wheezy-gcc/Dockerfile
+++ b/setup/debian-wheezy-gcc/Dockerfile
@@ -1,0 +1,19 @@
+FROM debian:wheezy
+
+MAINTAINER Rod Vagg <rvagg@nodesource.com>
+
+RUN echo 'deb-src http://http.debian.net/debian/ jessie main non-free contrib' >> /etc/apt/sources.list
+
+RUN apt-get update
+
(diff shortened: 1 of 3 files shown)


piscisaureus:
:+1:

PR #28, opened 2015-01-03
Added manual Windows setup instructions

rvagg:
I'm not going to bother even attempting to do something automatic for this, Ansible could probably do it but the downloads make it tricky.

Manual setup instructions instead. Two new servers are now running according to these instructions and appear to be working just fine for both iojs and libuv.


bnoordhuis:
LGTM FWIW

PR #30, opened 2015-01-07
Added Ubuntu host setup

rvagg:
* 14.10 64
* 14.04 64 & 32 + GYP version for libuv
* 12.04 64
* 10.04 64 & 32

### DIFF
--- a/setup/ansible-inventory
+++ b/setup/ansible-inventory
@@ -7,6 +7,20 @@ iojs-build-centos6-1
 [iojs-build-centos5]
 iojs-build-centos5-1
 
+[iojs-build-ubuntu1410]
+iojs-build-ubuntu1410-64-1
+
+[iojs-build-ubuntu1404]
+iojs-build-ubuntu1404-64-1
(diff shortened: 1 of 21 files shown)


rmg:
The multiple copies of the readme and host variables is clearly intentional.. But I don't know the reason. 

Other than my own curiosity, LGTM.

rvagg:
@rmg yeah, there's lots of duplication here and I'd love someone to help combine it into a single "setup all ubuntu hosts" directory with a single ansible script that just does things a little different for them all. I really just took the easiest path, which was a `cp -a` to get this all working. There's small variations on each of these.

The host variables need to be different because each one has a secret that's unique.

rmg:
@rvagg merge as is and mark the improvement as low hanging fruit for any new contributors who happen to be good at DRYing up Ansible scripts?

rvagg:
yeah, sorry, got lost in so many other things, and I already have some ansible updates locally that I need to push!

PR #31, opened 2015-01-12
www: sync from `public/` for iojs/website#26

snostorm:
For https://github.com/iojs/iojs.github.io/pull/26

If this was going out live, I'd recommend merging https://github.com/iojs/iojs.github.io/pull/26 first as it will handle redirecting to the new public/ directory.


snostorm:
Note: switched this to `public/` to match the change in the sibling PR.

rvagg:
live

PR #33, opened 2015-01-13
www: add .xz & fix .gz mime-types

Fishrock123:
Hopefully this will let .xz files be downloaded properly.

The gzip one isn't super important, but nginx is just outdated on it.


rmg:
LGTM.

rvagg:
thanks! I think I already added gz on the server but forgot to add it here. will sync now.

Fishrock123:
Works now. Cheers. :tada:

PR #34, opened 2015-01-14
Renames iojs/iojs.github.io -> iojs/website

snostorm:
To happen in sync with a rename of the iojs/iojs.github.io project as per iojs/iojs.github.io#57

### DIFF
--- a/setup/www/README.md
+++ b/setup/www/README.md
@@ -2,7 +2,7 @@
 
 This directory contains an Ansible playbook to set up the iojs.org website from a fresh Ubuntu host (Upstart configuration assumes 14.04 or similar).
 
-This setup listens to the [iojs/iojs.github.io](https://github.com/iojs/iojs.github.io) repository for new commits on the *master* branch. The changes are pulled locally and copied (sans the *.git* directory) into a publicly served directory.
+This setup listens to the [iojs/website](https://github.com/iojs/website) repository for new commits on the *master* branch. The changes are pulled locally and copied (sans the *.git* directory) into a publicly served directory.
 
 The inventory uses the host `iojs-www`, the easiest way to link this is to add it to your *~/.ssh/config*, something like:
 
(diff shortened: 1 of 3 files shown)


setup/www/README.md
```diff
@@ -12,7 +12,7 @@ Host iojs-www
   User root
 ```
 
-This setup assumes a GitHub Webhook is configured ([here](https://github.com/iojs/iojs.github.io/settings/hooks)) to point to port `9999` of the server with the path `/webhook` plus an additional secret.
+This setup assumes a GitHub Webhook is configured ([here](https://github.com/iojs/website/settings/hooks)) to point to port `9999` of the server with the path `/webhook` plus an additional secret.
```
snostorm: Unrelated to my PR, but this webhook doesn't exist as far as I can tell. Can we tweak the README? (Not sure where this is supposed to be though.)

setup/www/host_vars/iojs-www.tmpl
```diff
@@ -1,4 +1,4 @@
 ---
 github_secret: "INSERT SECRET FROM WEBHOOK HERE"
-clone_command: "cd /home/iojs/ && rm -rf iojs.github.io && git clone https://github.com/iojs/iojs.github.io.git"
-update_command: "cd /home/iojs/iojs.github.io/ && git checkout master && git reset --hard && git clean -fdx && git pull origin master && rsync -avz --delete --exclude .git /home/iojs/iojs.github.io/ /home/iojs/www/"
\ No newline at end of file
+clone_command: "cd /home/iojs/ && rm -rf iojs.github.io && rm -rf website && git clone https://github.com/iojs/webite.git"
```
snostorm: I intentionally left a reference to `iojs.github.io` here so that we'd have it `rm -fr` on the first run (as the folder would otherwise be left behind post merge.)

setup/www/README.md
rvagg: you probably can't see it without full admin privs

PR #38, opened 2015-01-26
program to calculate and update download stats

rvagg:
parse last few log files each night and update /home/iojs/download-stats.json with new data

not the most elegant of scripts but it gets the job done (using 5 "programming" languages no less)

### DIFF
--- a/dist/update-download-stats.sh
+++ b/dist/update-download-stats.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+outfile=/tmp/dld.txt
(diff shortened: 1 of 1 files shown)


rvagg:
and yes I recognise the sheer hackiness of this, but it kind of _evolved_ and I don't have the inclination to refactor into a pure Node (awk, or whatever) program

Fishrock123:
So long as it being really hacky isn't going to bite us terribly in the future... :s

PR #54, opened 2015-02-20
[WIP] Cleaning up the www setup to match production

kenperkins:
This is a work in progress; the plan is to get the anisble playbook up to date with production such that we can re-build an identical box in moments should the need arise.

- [x] Figure out how to setup `dist` and `staging` users
- [x] Figure out how to setup crontab jobs
- [x] deploy `dist-indexer` and `npm install` in `/home/dist/dist-indexer`
- [ ] get `promote_nightly.sh` into `/home/staging/promote/`
- [x] get `update-download-stats.sh` into `/home/iojs`

#### Current Changes
- Detected if `dhparam.pem` exists, saving a boatload of time
- adding a bunch of paths into the nginx config from `/home/dist`
- Added @kenperkins as an `authorized_keys` member
- reworked `authorized_keys` setup to be a bit cleaner
- added `dist`, `staging` users with ssh keys
- added crontab jobs for `promote_nightly.sh` and `update-download-stats.sh`

#### Open Questions
- What is `/root/mkdldtxt.sh`?
- What is `/root/countdldtxt.sh`
- Who is responsible to create `/home/dist/public` and `/home/dist/public-test` directories?
- What is `/home/dist/dist-indexer/ls-types.js`?
- What is `/home/dist/dist-indexer/transform-filename.js`?
- What are all of the files in `/home/staging/promote`?
- Who creates `/home/staging/staging` directory?


setup/www/ansible-playbook.yaml
```diff
@@ -24,8 +24,12 @@
...
       tags: general
 
-    - name: User | Add {{ server_user }} user
-      user: name="{{ server_user }}" shell=/bin/bash
+    - name: User | Add system users
+      user: name="{{ item }}" shell=/bin/bash
+      with_items:
+        - "{{ server_user }}"
+        - "{{ staging_user }}"
+        - "{{ dist_user }}"
```
kenperkins: @rvagg we're now creating all three of the utility accounts we use, you can see that here. I have a couple of questions lower in the playbook.

setup/www/ansible-playbook.yaml
```diff
@@ -39,9 +43,23 @@
...
 
-    - name: General | Create authorized_keys for {{ server_user }}
-      authorized_key: user="{{ server_user }}" key="{{ lookup('file', '/tmp/' + item + '.keys') }}"
-      with_items: ssh_users
+    - name: General | authorized_keys for users
+      authorized_key: user="{{ item[0] }}" key="{{ lookup('file', '/tmp/' + item[1] + '.keys') }}"
+      with_nested:
+        - [ '{{ dist_user }}', '{{ server_user }}', '{{ staging_user }}' ]
+        - ssh_users
+      tags: user
```
kenperkins: We now set all of the `ssh_users` to have access to `authorized_keys` for all of our utility accounts.

setup/www/ansible-playbook.yaml
```diff
@@ -39,9 +43,23 @@
...
+
+    - name: General | place ssh keys
+      copy: owner="{{ item[0] }}" group="{{ item[0] }}" src=resources/keys/{{ item[0] }}/{{ item[1].name }} dest=/home/{{ item[0] }}/.ssh/{{ item[1].name }} mode={{item[1].perms}}
+      with_nested:
+        - [ "{{ dist_user }}", "{{ staging_user }}" ]
+        - [ { name: 'id_rsa', perms: '0600' }, { name: 'id_rsa.pub', perms: '0644' } ]
```
kenperkins: This requires placing the `id_rsa` for each of `dist` and `staging` in the appropriate folders. They're currently `.gitignore`d.

setup/www/ansible-playbook.yaml
```diff
@@ -39,9 +43,23 @@
...
+
+    - name: General | add id_rsa.pub to staging, dist authorized_keys
+      authorized_key: user="{{ item }}" key="{{ lookup('file', 'resources/keys/' + item + '/id_rsa.pub') }}"
+      with_items:
+        [ "{{ staging_user }}", "{{ dist_user }}" ]
```
kenperkins: I determined we needed the `id_rsa.pub` added to `authorized_keys` for `dist` and `staging` based on inspecting the current state of production. @rvagg I'd love confirmation on this.

setup/www/ansible-playbook.yaml
```diff
@@ -78,20 +102,36 @@
...
 
+    ### configure cron and utility capabilities
+
+    - name: Configure | setup cron for nightly
+      cron: name="promote_nightly" minute="*/5" job="/home/staging/promote/promote_nightly.sh" user="{{ dist_user }}"
+      tags: setup
+
+    - name: Configure | setup cron for download-stats
+      cron: name="update_download_stats" minute="30" hour="1" job="/home/iojs/update-download-stats.sh" user="root"
+      tags: setup
```
kenperkins: both of these crontab jobs go into the user specific crontab (i.e. `crontab -e`) and not into `/etc/crontab`. Not sure if that matters.

kenperkins:
This is getting close, but I'm unable to find `promote_nightly.sh` in source control. @rvagg are you able to assist on this? should I just take it from the current production webserver?

kenperkins:
Additional open questions:

- What is `/root/mkdldtxt.sh`?
- What is `/root/countdldtxt.sh`
- Who is responsible to create `/home/dist/public` and `/home/dist/public-test` directories?
- What is `/home/dist/dist-indexer/ls-types.js`?
- What is `/home/dist/dist-indexer/transform-filename.js`?
- What are all of the files in `/home/staging/promote`?
- Who creates `/home/staging/staging` directory?

kenperkins:
Reminder @rvagg that I'm blocked on finishing this without some input.

rvagg:
Really sorry @kenperkins, I'm a terrible bottle-neck for these things, way too overloaded. Trying to get on top of some of these things now.

### What is /root/mkdldtxt.sh?

An initial version of what's in `/home/iojs/update-download-stats.sh`, it's redundant now. It only ever was meant to be a temporary "let's see what these numbers look like" thing but that script is now run via cron and exposed via https://iojs.org/download-stats.json because ppl asked for it.

### What is /root/countdldtxt.sh

Part of the above, redundant now too. There were two scripts to do the job that one is now doing.

## Who is responsible to create /home/dist/public and /home/dist/public-test directories?

The promote script(s) are responsible for `/home/dist/public/`. There is a `mkdir -p` in there but the root directories should be created by the script I think since they are served by nginx. Currently there needs to be `next-nightly`, `nightly`, `release` (aliased ad https://iojs.org/dist/) and there is also a `test` directory but that's not controlled by any tool, it's been a place to share some test builds, manually, so unimportant.

### What is /home/dist/dist-indexer/ls-types.js?
### What is /home/dist/dist-indexer/transform-filename.js?

Parts of dist-indexer https://github.com/iojs/build/tree/master/tools/dist/dist-indexer, _but_ also:

1. There seems to be an incomplete refactor going on there and I'm wondering if a commit or two got lost somewhere, transform-filename.js should be used by dist-indexer.js as well as ls-types.js but it's only used by the latter with some duplicated code in dist-indexer.js.
2. ls-types.js is used by the scripts in `/usr/local/bin`, you'll see there are 3 `dist-*` scripts there now and they are used for the release process so that authorised release people can run (tools/release.sh](https://github.com/iojs/io.js/blob/v1.x/tools/release.sh) to promote and sign a release without needing to log in to the server. Those files need to become part of a scripted setup of the server too.

### What are all of the files in /home/staging/promote?

I've put them in a PR including a README, see https://github.com/iojs/build/pull/76 and note that they are also invoked by scripts in `/usr/local/bin` which also need to be part of this setup

### Who creates /home/staging/staging directory?

It should be created by server setup. Each of the release build slaves has a jenkins config which makes it `ssh` in to the server as staging and `mkdir -p` and then `scp` in the build artifacts, but as with `/home/dist/` we should probably set the basics of it up including the `next-nightly`, `nightly` and `release` subdirectories.

setup/www/ansible-playbook.yaml
rvagg: off the top of my head I don't think these are actually required on there, they are just where I created them, I can't think of anything that these users do that needs to have them available

setup/www/ansible-playbook.yaml
rvagg: `staging` is used by all the jenkins release build slaves so that's needed in there but I can't think of why `dist` has one, perhaps a historical legacy, we use that account now by putting public keys of authorised releasers into it - you can find those people on the io.js readme, the ones who have GPG key sigs (3 of us) and I'd just be grabbing those from github/user.keys, it might be worth at least hard-wiring those into this script so they can go into dist/.ssh/authorized_keys

setup/www/ansible-playbook.yaml
rvagg: I can't think of any reason this would matter so :+1:

kenperkins:
Wow, it's like Christmas morning! Thanks @rvagg I'll try to resurrect this as soon as I can find some engineering time.

jbergstroem:
This unfortunately looks stale. Going to close it but feel free to reopen with a rebased version seeing how a lot of the changes has gone in one way or the other.

PR #60, opened 2015-02-28
dist: Support SmartOS

geek:
This only includes the basic setup for now.  Still need to work on build

### DIFF
--- a/setup/smartos/ansible-playbook.yaml
+++ b/setup/smartos/ansible-playbook.yaml
@@ -17,8 +17,12 @@
       with_items: packages
       tags: general
 
+    - name: User | Add {{ server_user }} group
+      command: groupadd {{ server_user }}
+      tags: user
+
     - name: User | Add {{ server_user }} user
-      user: name="{{ server_user }}" shell=/bin/bash append=yes groups=staff
+      user: name="{{ server_user }}" shell=/bin/bash append=yes groups=staff,{{ server_user }}
       tags: user
 
     - name: User | Unlock {{ server_user }} user
--- a/setup/smartos/ansible-playbook.yaml
+++ b/setup/smartos/ansible-playbook.yaml
@@ -13,7 +13,7 @@
       tags: general
 
     - name: General | Install required packages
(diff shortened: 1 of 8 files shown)


setup/smartos/ansible-playbook.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+    - include_vars: ansible-vars.yaml
+      tags: vars
+
+    - name: General | pkgin Update
+      command: pkgin up
+      tags: general
+
+    - name: General | Install required packages
+      command: pkgin in {{ item }}
```
jbergstroem: I had to add `-y`: `pkgin -y in {{ item }}`

setup/smartos/ansible-playbook.yaml
```diff
@@ -0,0 +1,59 @@
...
+      authorized_key: user="{{ server_user }}" key="{{ lookup('file', '/tmp/' + item + '.keys') }}"
+      with_items: ssh_users
+      tags: user
+
+    - name: Jenkins | Download Jenkins' slave.jar
+      command: curl -sL https://jenkins-iojs.nodesource.com/jnlpJars/slave.jar -o /home/{{ server_user }}/slave.jar
+      tags: jenkins
+
+    - name: Jenkins | Copy start.sh script
+      copy: src=./resources/start.sh dest=/home/{{ server_user }}/start.sh owner={{ server_user }} group={{ server_user }} mode=0755
```
jbergstroem: This will fail if the group `iojs` is missing which doesn't seem to be added above.

geek: Good catch, this is now fixed

jbergstroem:
LGTM

jbergstroem:
Another minor thing: Since path doesn't seem to be provided through `/etc/profile`we might want to add it as a envvar in the manifest (otherwise jenkins won't find path to git).

jbergstroem:
(btw, confirmed that adding path fixed Jenkins)

geek:
Closing in favor of #64

PR #65, opened 2015-03-19
smartos: remove extended network privileges

jbergstroem:
Using net_privaddr caused two tests assuming regular users cannot use privileged port ranges to fail. We should probably not be testing in such an environment anyway.

Passing suite here: https://jenkins-iojs.nodesource.com/job/iojs+any-pr+multi/nodes=iojs-smartos14-64/329/console

R=@geek, @rvagg

### DIFF
--- a/setup/smartos/resources/jenkins_manifest.xml
+++ b/setup/smartos/resources/jenkins_manifest.xml
@@ -15,7 +15,7 @@
         </dependency>
 
         <method_context working_directory="/home/iojs">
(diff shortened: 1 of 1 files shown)


rvagg:
nice! lgtm

geek:
LGTM, I would merge if I had access :/

jbergstroem:
Merged in edac962b32d28dd65ee74119bce2cf88818a1b5f. Thanks for reviewing, Wyatt and Rod.

PR #67, opened 2015-03-19
dist: Support FreeBSD

jbergstroem:
Add setup scripts for two builders behind a vpn at Voxer. The configuration assumes you're deploying in a FreeBSD jail; we look up the default routed ip-adress through ansible and pass it to the init script.

/R=@rvagg

### DIFF
--- a/setup/freebsd/README.md
+++ b/setup/freebsd/README.md
@@ -1,0 +1,26 @@
+# io.js Build FreeBSD Setup
+
+The current FreeBSD lives behind a VPN at [Voxer](http://voxer.com).
+Should you need access, speak to [https://github.com/rvagg](rvagg) for
+further information.
+
+To set up hosts, make sure you add them to your ssh config first:
+```
+Host iojs-build-freebsd-64-1
+  HostName 172.16.31.2
+
+Host iojs-build-freebsd-32-1
+  HostName 172.16.31.3
+```
+
+Note that these hostnames are also used in the ansible-inventory file. The IP addresses will need to be updated each time the servers are reprovisioned.
+
+To set up a host, run:
+
+```text
+$ ansible-playbook -i ../ansible-inventory ansible-playbook.yaml
+```
+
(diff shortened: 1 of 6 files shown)


setup/freebsd/README.md
```diff
@@ -0,0 +1,28 @@
+# io.js Build FreeBSD Setup
+
+The current FreeBSD lives behind a VPN at [Voxer](http://voxer.com).
+If you want access, speak to [rvagg][rvagg] for further information.
```
rvagg: s/want/need, don't want to give Voxer the impression we're overly liberal about handing out access

jbergstroem: Ok, I'll elaborate.

setup/freebsd/resources/jenkins
```diff
@@ -0,0 +1,71 @@
...
+
+jenkins_jnlpurl="https://jenkins-iojs.nodesource.com/computer/iojs-voxer-freebsd101-{{id}}/slave-agent.jnlp"
+jenkins_secret="{{secret}}"
+jenkins_jar="/home/iojs/slave.jar"
+jenkins_log_file="/home/iojs/jenkins_console.log"
+jenkins_args="-jnlpUrl ${jenkins_jnlpurl} -secret ${jenkins_secret}"
+jenkins_user="iojs"
+jenkins_group="iojs"
+
+pidfile="/var/run/jenkins/jenkins.pid"
```
rvagg: does the jenkins user need access to this file?

rvagg:
lgtm

setup/freebsd/resources/jenkins
jbergstroem: It's good practise to own the pidfile with the same user as the daemon. In this case it's not needed but I tried to stay close to [freebsd's own init script](https://svnweb.freebsd.org/ports/head/devel/jenkins/files/jenkins.in?revision=368098&view=markup).

jbergstroem:
Thanks for the feedback. Merged in d8c0f1d4df143dac2e3cea05fe2e936eb3e75d00. Closing.

PR #69, opened 2015-03-23
Dist indexer v8 update

rvagg:
V8 has changed where they are putting version macros, which is annoying but the new location make so much more sense. This change makes dist-indexer look in both locations.


rmg:
LGTM

PR #71, opened 2015-04-11
added initial nightly-builder app

rvagg:
Probably R=@Fishrock123 but input welcome from others.

This app is to be used via cron to trigger nightly builds for both `nightly` and `next-nightly`. It finds the latest build and gets the commit sha from it, then it compares that commit sha of the appropriate git branch (`v1.x` for `nightly` and `next` for `next-nightly`). If they don't match then it triggers a build on Jenkins using all the appropriate parameters for the iojs+release+nightly including an authorization token that's in the job configuration.

Used it just now to build https://jenkins-iojs.nodesource.com/job/iojs+release+nightly/139/

### DIFF
--- a/tools/nightly-builder/build-required.js
+++ b/tools/nightly-builder/build-required.js
@@ -1,0 +1,49 @@
+"use strict"
+
+const strftime     = require('strftime').timezone(0)
+    , after        = require('after')
+    , latestBuild  = require('./latest-build')
+    , latestCommit = require('./latest-commit')
+
+const dateFormat = '%Y%m%d'
+
+
+function timeString () {
+  return strftime(dateFormat)
+}
+
+
+function buildRequired (type, callback) {
+  let done = after(2, onData)
+    , build
+    , commit
+
+  function onData (err) {
+    if (err)
+      return callback(err)
+
+    let buildCommit = build.commit.substr(0, 10)
+      , buildDate   = strftime(dateFormat, build.date)
+      , nowDate     = timeString()
+
+    commit = commit.substr(0, 10)
+
+    if (buildCommit != commit) // only need to compare commit
+      return callback(null, { type: type, commit: commit, date: nowDate })
+
+    return callback()
+  }
+
+  latestBuild(type, function (err, data) {
+    build = data
+    done(err)
+  })
+
+  latestCommit(type, function (err, data) {
+    commit = data
+    done(err)
+  })
+}
+
+
+module.exports = buildRequired
--- a/tools/nightly-builder/latest-build.js
+++ b/tools/nightly-builder/latest-build.js
@@ -1,0 +1,45 @@
+"use strict"
+
+const jsonist = require('jsonist')
+
+const indexUrl = 'https://iojs.org/download/{type}/index.json'
+
+
+function latestBuild (type, callback) {
+  let url = indexUrl.replace(/\{type\}/, type)
+
+  function onGet (err, data) {
+    if (err)
+      return callback(err)
+
+    let i = -1
+
+    while (data[++i]) {
+      let m      = data[i].version.match(/nightly(20\d\d)(\d\d)(\d\d)(\w{10,})/)
+        , date   = new Date(m && `${m[1]}-${m[2]}-${m[3]}` || data[i].date)
+        , commit = m && m[4]
+
+      if (m && commit) {
+        return callback(null, {
+            version : data[i].version
+          , date    : date
+          , commit  : commit
+        })
+      }
+    }
+
+    return callback(new Error(`no latest version for "${type}"`))
+  }
+
+  jsonist.get(url, onGet)
+}
+
+
+module.exports = latestBuild
+
+latestBuild('nightly', function (err, data) {
+  if (err)
+    throw err
+
+  console.log('data', data)
+})
--- a/tools/nightly-builder/latest-commit.js
+++ b/tools/nightly-builder/latest-commit.js
@@ -3,7 +3,7 @@
 const ghrepos = require('ghrepos')
 
 const typeRefs = {
-          'nightly'      : 'heads/v1.x'
+          'nightly'      : 'heads/master'
         , 'next-nightly' : 'heads/next'
       }
 
--- a/tools/nightly-builder/latest-commit.js
+++ b/tools/nightly-builder/latest-commit.js
@@ -1,0 +1,30 @@
+"use strict"
+
+const ghrepos = require('ghrepos')
+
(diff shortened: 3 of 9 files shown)


rvagg:
probably needs readme ..

cjihrig:
Thoughts on using the same coding style as node core?

rvagg:
running on iojs-www in /etc/crontab:

```
0 18    * * *   dist    node /home/dist/nightly-builder/ --type nightly --token redacted
0 19    * * *   dist    node /home/dist/nightly-builder/ --type next-nightly --token redacted
```

this server is running as UTC-4 (unfortunately)

rvagg:
> Thoughts on using the same coding style as node core?

1cda5be0f1bc1ac709847305a1565c72__tumblr_inline_n0510cwkrd1r3yh7o

cjihrig:
Sorry :fearful:

tools/nightly-builder/trigger-build.js
```diff
@@ -0,0 +1,53 @@
...
+            , {
+                  name  : 'buildtype'
+                , value : options.type
+              }
+          ]
+      }
+    , post = qs.encode({
+          token : token
+        , json  : JSON.stringify(data)
+      })
```
Fishrock123: using a new `let` / `const` per line would be a little more readable :)

tools/nightly-builder/build-required.js
```diff
@@ -0,0 +1,49 @@
...
+const dateFormat = '%Y%m%d'
+
+
+function timeString () {
+  return strftime(dateFormat)
+}
+
+
+function buildRequired (type, callback) {
+  let done = after(2, onData)
```
Fishrock123: This is fine once understood, but it's definitely a bit more arcane compared to `async.parallel`.

tools/nightly-builder/latest-build.js
```diff
@@ -0,0 +1,36 @@
+"use strict"
+
+const jsonist = require('jsonist')
+
+const indexUrl = 'https://iojs.org/download/{type}/index.json'
+
+
+function latestBuild (type, callback) {
+  let url = indexUrl.replace(/\{type\}/, type)
```
Fishrock123: Maybe just use a template string?

tools/nightly-builder/latest-build.js
```diff
@@ -0,0 +1,36 @@
...
+
+  function onGet (err, data) {
+    if (err)
+      return callback(err)
+
+    let i = -1
+
+    while (data[++i]) {
```
Fishrock123: The loop functionality doesn't appear to be used, shouldn't this be an if?

tools/nightly-builder/latest-build.js
```diff
@@ -0,0 +1,36 @@
...
+        , commit = m && m[4]
+
+      return callback(null, {
+          version : data[i].version
+        , date    : date
+        , commit  : commit
+      })
+    }
+
+    return callback(new Error(`no latest version for "${type}"`))
```
Fishrock123: More concise to bail to this before the above if there isn't data. i.e:

`if (!data[0]) return callback(new Error(...))`

tools/nightly-builder/latest-commit.js
```diff
@@ -0,0 +1,27 @@
+"use strict"
+
+const ghrepos = require('ghrepos')
+
+const nightlyRef     = 'heads/v1.x'
+    , nightlyNextRef = 'heads/next'
+
+
+function latestCommit (type, callback) {
+  let ref = type == 'nightly' ? nightlyRef : type == 'next-nightly' ? nightlyNextRef : null
```
Fishrock123: Using a switch here might be cleaner and more extensible

tools/nightly-builder/build-required.js
```diff
@@ -0,0 +1,49 @@
...
+    , build
+    , commit
+
+  function onData (err) {
+    if (err)
+      return callback(err)
+
+    let buildCommit = build.commit.substr(0, 10)
+      , buildDate   = strftime(dateFormat, build.date)
```
Fishrock123: unused, I guess the intention was originally to check both, but that seems unnecessary as per below.

tools/nightly-builder/test.js
```diff
@@ -0,0 +1,47 @@
...
+  t.plan(8)
+  var other = null
+
+  function verify (type) {
+    return function (err, sha) {
+      t.error(err, 'no error')
+      t.equal(typeof sha, 'string', 'got sha')
+      t.equal(sha.length, 40, `sha looks good (${sha})`)
+      t.notEqual(sha, other, 'sha not the same as other type of sha')
+      other = sha // who's gonna be first??
```
Fishrock123: yay safe parallelism XD

Fishrock123:
Seems fine. Some style meh, but no big deal I suppose. Mostly just leftover nits.

tools/nightly-builder/latest-build.js
rvagg: well the idea here is that I want to keep these kinds of fixed strings out of the logic body, template strings don't really help with that because you have to piece multiple bits together

tools/nightly-builder/latest-build.js
rvagg: heh, yeah, what's going on here?? thanks for picking this up - I believe it was actually meant to search through for a _valid_ release as it's plausable that there are things other than nightlies listed in the index. changed to actually use the loop for validating.

tools/nightly-builder/latest-commit.js
rvagg: changed to an object property lookup instead

tools/nightly-builder/build-required.js
rvagg: Rright, so this is something I wanted sanity check on - I originally made it compare commit _and_ date but then rolled back to just commit check. But this means two things can happen:

1. if the program is run manually and the head commit has changed it'll just make a new nightly, this is not a big problem because I don't expect it to be run manually but it could be useful
2. if you manually create a nightly earlier in the day and the head commit has moved on, this program is going to trigger another nightly build, and perhaps that's undesirable.

We could enforce one-nightly-per-day with this but at the moment it's not.

Fishrock123: > We could enforce one-nightly-per-day with this but at the moment it's not.

Hard to bother at this point. Anyone that works with the nightlies will already have to deal with more than one a day.

rvagg: agreed

PR #76, opened 2015-04-15
added promotion scripts for web server

rvagg:
including documentation

### DIFF
--- a/tools/dist/promote/_promote.sh
+++ b/tools/dist/promote/_promote.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+if [ -z ${srcdir+x} ]; then
+  echo "\$srcdir is not set"
+  exit 1
+fi
+
+if [ -z ${dstdir+x} ]; then
+  echo "\$dstdir is not set"
+  exit 1
+fi
+
+version=$1
+
+for subdir in $(cd $srcdir && ls); do
+
+  if [ -d "${srcdir}/${subdir}" ]; then
+    resha=no
+
+    if [ "X${version}" != "X" ] && [ "X${version}" != "X${subdir}" ]; then
+      continue
+    fi
+
+    for donefile in $(cd ${srcdir}/${subdir} && ls *.done 2> /dev/null); do
+      doneref=$(echo $donefile | sed 's/\.done$//')
+
+      if [[ -f "${srcdir}/${subdir}/${doneref}" || -d "${srcdir}/${subdir}/${doneref}" ]]; then
+        echo "Promoting ${subdir}/${doneref}..."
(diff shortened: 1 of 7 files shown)


jbergstroem:
Do we possibly want to add a routine for removing older directories? I see no point in storing nightlies forever.

rvagg:
open for discussion, I don't like deleting stuff in general so it's difficult for me to embrace a full cleanup strategy, throwing them in S3 (or similar) would be great though

jbergstroem:
I'm meh but obviously prefer storage since it's the easy way out. The nightlies are easy to reproduce and have no purpose for anything older than $latest_release. People using them as a tool should use git anyway.

tools/dist/promote/_promote.sh
```diff
@@ -0,0 +1,46 @@
...
+    fi
+
+    for donefile in $(cd ${srcdir}/${subdir} && ls *.done 2> /dev/null); do
+      doneref=$(echo $donefile | sed 's/\.done$//')
+
+      if [[ -f "${srcdir}/${subdir}/${doneref}" || -d "${srcdir}/${subdir}/${doneref}" ]]; then
+        echo "Promoting ${subdir}/${doneref}..."
+        mkdir -p "${dstdir}/${subdir}"
+        rm -rf "${dstdir}/${subdir}/${doneref}"
+        cp -dr "${srcdir}/${subdir}/${doneref}" "${dstdir}/${subdir}/${doneref}"
```
jbergstroem: `cp -a` would be an improved cross-platform choice (`-d` being specific to linux, but preserving symlinks accessible through `-a`on most systems)

jbergstroem:
I'm a bit scared about `rm -rf`:ing with slashes, but since you guard against some empty variables I think its good enough.

Nothing to remark about. LGTM; thanks for adding these.

tools/dist/promote/_promote.sh
rvagg: I'm sure there was a good reason I used `-dr` here, I'd normally reach for `-a` in almost any situation and am not sure why I'm not using it here .. I'm a tiny bit afraid to find out what I was thinking!

jbergstroem: Windows perhaps?

rvagg: I think the main difference between `-a` and `-dr` is that `-d` only does a `--preserve=links` while `-a` does a `--preserve=all` and the implication here is that the mode isn't copied - which is 664/775 for staging but ends up being 644/755 for dist. _but_ that doesn't really matter because there aren't any other users in the dist group.. So I'm still not sure what I was thinking! Will `-a` and see what breaks, if anything.

PR #80, opened 2015-04-21
added raspberry pi config

rvagg:


jbergstroem:
I'd prefer having an init script, but I can add it later since I'm already on to that. The hosts file seems to suggest a user called `pi` while ansible contains iojs. Not sure if intentional but LGTM either way as long as it works.

rvagg:
* yes init scripts would be awesome for these machines in particular because java regularly dies in the constrained environment, I just haven't proritised yet but am happy to have assistance!
* `pi` user is the default user on raspbian and has sudo access, it's used to set up the boxes, the `iojs` user is unprivileged and is created by the setup procedure to run the slave.

jbergstroem:
Thanks for the reply. Happy to merge.

PR #85, opened 2015-04-26
added Ubuntu 15.04 config

rvagg:


jbergstroem:
Perhaps add `ccache`?

jbergstroem:
I can possibly provide an init script if you haven't got one already.

rvagg:
systemd, I have one for something else, fedora perhaps .. or did I dump it in an issue here somewhere?

jbergstroem:
Probably this no? https://github.com/nodejs/build/blob/master/setup/fedora21/resources/jenkins.service

rvagg:
ah yes, but actually, this is probably closer to Ubuntu: https://github.com/nodejs/build/tree/master/setup/debian8

PR #86, opened 2015-04-26
added Fedora 21

rvagg:


rvagg:
I've just added a systemd config to this as well as a "start" at the end for it so this is the first fully setup & start slaves we have, no setup.sh script. Critique welcome.

rvagg:
also puts ccache into the build path, first time we've done this on a non-arm machine

PR #89, opened 2015-04-28
freebsd: enable building with ccache

jbergstroem:
Add ccache to dependencies and inject it to $PATH. Also, some minor edits to init script/grammar.

For this to "work", after pulling in production we need to [restart the jenkins master](https://issues.jenkins-ci.org/browse/JENKINS-27739).

PR #90, opened 2015-04-28
smartos: enable building with ccache

jbergstroem:
We need to symlink ourselves since the package manager doesn't do it for us.

Edit: This PR also requires jenkins master to be restarted (See #89 for link)

### DIFF
--- a/setup/smartos/ansible-playbook.yaml
+++ b/setup/smartos/ansible-playbook.yaml
@@ -47,6 +47,17 @@
       authorized_key: user="{{ server_user }}" key="{{ lookup('file', '/tmp/' + item + '.keys') }}"
       with_items: ssh_users
       tags: user
+      
+    - name: CCache | Create directory that hold symlinks to ccache
+      file: path=/opt/local/libexec/ccache state=directory mode=0755
+      tags: ccache
+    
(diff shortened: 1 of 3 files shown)


rvagg:
Cool, lgtm.

Does the install process for ccache on smartos not make symlinks itself? I'm finding for newer linux distros it's all taken care of automatically.

re restarting Jenkins, I think you have access to do that via the Jenkins interface, otherwise ping me and I can do it manually on the server.

jbergstroem:
@rvagg nope -- symlinking has to be taken care of manually. I've updated to the newest pkg repo on my local smartos machines and it seems like we might have to do some path changes forward as well. No biggie though.

As for Jenkins, thanks for letting me know. I'd rather have you do it this time since it's somewhat voodoo. He's probably already annoyed with me.

I'll merge this and #89 if you think that looks good as well and then we'll update/restart.

jbergstroem:
Merged in 46d79036f09a488e13422300b30599fe007117cb . Thanks for reviewing (and doing it so quickly).

jbergstroem:
Found a typo. Crap. `s|/usr/local/libexec|/opt/local/libexec|g`..

PR #93, opened 2015-05-05
companies & people changes

rvagg:
Just some quick changes to the companies and people listed on the README, the list hasn't changed since we started this in node-forward and @mikeal reminded me that the hardware providers list was wrong so I've fixed that now and updated the people list while I was at it. I think I have it right ..?

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,27 +91,21 @@ Hardware Sponsors
 
 The following companies are contributing hardware to this project:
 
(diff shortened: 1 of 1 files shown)


jbergstroem:
I think part of the reason it never got updated was because parts of this text was in the governance PR. Anyhow, LGTM if you need it.

rvagg:
yeah, that too, I might take over that PR from Will cause I know he's busy with school atm

PR #111, opened 2015-06-13
freebsd: improve service handling in playbook

jbergstroem:
Use lineinfile instead of echo and use the standard "start" directive instead of "enable" which isn't available everywhere.


rvagg:
nice, lgtm

jbergstroem:
Landed in 72a7038ca868e6879c15f2034b81014402bbb40e. Thanks for the review.

PR #112, opened 2015-06-13
add fedora22 config

rvagg:
This is live in CI now, the major change here is the removal of `yum` so we need to use `dnf` but the ansible dnf module is 1.9+ only and there's a bug in 1.9.1 which will be fixed in the next release but I've patched my local version to make it work. Documented in the readme.


jbergstroem:
LGTM. Perhaps open a new issue/PR that edits the root readme about ansible requirements?

PR #117, opened 2015-06-17
dist: use ccache on ubuntu1404

jbergstroem:
upstart script shortly


rvagg:
the apt install does the symlinks properly for ccache with no further action needed on 14.04? if so, then lgtm

jbergstroem:
Ye man, all ubuntu's seems to install everything (including symlinks) to `/usr/lib/ccache`

jbergstroem:
Merged in 6131d5cac044e52958c3b2da7597bf3cc34ce2e9. Thanks.

PR #120, opened 2015-06-23
added initial server resources sheet

rvagg:
I figured that since GitHub has a nice display format for .csv files that this would be the way to go.

I believe @mhdawson has some PPC machines to add to this soon as well.


rvagg:
see https://github.com/nodejs/build/blob/server-resources/doc/server-resources.csv for rendered version, quite nice

Starefossen:
Does the CSV spec require the file to be all on one line?

Also, there is no Mac OS X 10.12 version (yet).

rvagg:
thanks for picking up 10.12 @Starefossen, an Excel error.

re it being on one line, no there is nothing dictating that, it's just that I'm using Excel on OSX right now so if you're seeing it as one line then that'll be a linebreaking thing, I'll try to make it betterer

Starefossen:
Never mind the one line thing, I was just GitHub's view acting up that wouldn't let me comment on the Mac OS X version thingy here: https://github.com/nodejs/build/pull/120/files.

jbergstroem:
Looks good! The only feedback I have is that a markdown table would be more readable when editing (unless Excel) - but I fear that'd render much worse on github.

rvagg:
@jbergstroem yeah, Github is going to cap the width of it unfortunately, I do like the idea of it being readable as text though

another alternative would be to use CSV but space it out so it's easily readable as text too, I'm just not sure how GitHub is going to render it with extra spaces

rvagg:
Check that out, best of both worlds? https://raw.githubusercontent.com/nodejs/build/server-resources/doc/server-resources.csv the GitHub formatter strips extra padding whitespace.

Also, this is why I love Node:

```js
> csv2 = require('csv2')
> columnify = require('columnify')
> listStream = require('list-stream')
> fs.createReadStream('./doc/server-resources.csv')
  .pipe(csv2())
  .pipe(listStream.obj(function (err, list) {
    fs.writeFileSync('doc/server-resources.csv',
      columnify(list, {
        dataTransform: trans = function (data) { return (data.indexOf(',') > -1 ? `"${data}"` : data) + ',' },
        headerTransform: trans, showHeaders: false }))
  }))
```

jbergstroem:
looks great! columnify is my new best friend.

jbergstroem:
Suggesting we land this.

rvagg:
holding off till I update the rackspace ones, we're missing some non-windows servers there

jbergstroem:
Lets bring this up to date and land it.

jbergstroem:
In some distant dream I was hoping to be able to generate this as part of the ansible setup stack. This comment might remind our future selves to do just that.

jbergstroem:
How about we write a playbook for the refactor that just connects to each host and returns minimal facts for each? (cores, ram, disk)

BridgeAR:
Should this stay open? There was no progress for more than 1.5 years now.

gdams:
I think that https://github.com/nodejs/build/blob/master/ansible/inventory.yml contains most of this information anyway

PR #126, opened 2015-06-30
extraced decode-ref.js, add tests, add rc builds

rvagg:
allow for `v3.0.0-rc1` type build numbers, this means we have to tag RC builds as such in the repo but that's a good idea anyway.

### DIFF
--- a/tools/dist/dist-indexer/decode-ref.js
+++ b/tools/dist/dist-indexer/decode-ref.js
@@ -1,0 +1,28 @@
+const assert = require('assert')
+    , dirre  = /^(v\d+\.\d+\.\d+(?:-rc\.\d+)?)(?:-(?:next-)?nightly\d{8}(\w+))?$/ // get version or commit from dir name
+
+
+function decodeRef (dir) {
+  var m = dir.match(dirre)
+  return m && (m[2] || m[1])
+}
+
+
+module.exports = decodeRef
+
+
+if (module === require.main) {
+  var tests = [
+      { dir: 'v1.0.0'                                , ref: 'v1.0.0'          }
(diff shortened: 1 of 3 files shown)


Starefossen:
Sticking to the [semver specification](http://semver.org); RC versions should be labeled `v3.0.0-rc.1`.

rvagg:
Thanks for pointing that out @starefossen, I hadn't noticed the dot separation of tags in the spec before tbh. Will adjust for rc.3.

rvagg:
Actually, getting feedback from @ljharb here would be useful before jumping straight in.

ljharb:
I'm fine with `vX.Y.Z-rc.N` as a format, as long as it stays consistent - if so, can that be updated on the existing iojs.org rc site?

rvagg:
the existing builds have `rcN` baked in so can't be changed, I could remove them from index.tab and index.json by changing the regexes in this PR, would that help? the next RC can be the first one supported by nvm then

ljharb:
hmm - if both formats are there it's going to be a *huge* pain. As long as either the archive files themselves are all the same convention, across both existing and new releases, or the old releases don't exist in `index.tab`, I think I should be alright.

Basically I want to avoid having to code in support for more than one convention format.

rvagg:
I've gone ahead and adopted the `vx.y.z-rc.n` format in Jenkins, I've updated this PR and this is live on the server now. I also deleted the older RCs so now there's only rc.3 @ https://iojs.org/download/rc/

Unfortunately node-gyp also needed the same treatment in `next`, I've proposed that here and there will be an rc.4 for this: https://github.com/nodejs/io.js/pull/2171

mzgol:
Just a quick update - it will be https://nodejs.org/download/rc/ now.

PR #130, opened 2015-07-07
dist: use ccache for ubuntu15.04 slaves

jbergstroem:
This will be refactored soon but might as well add it for consistency.


joaocgreis:
LGTM

The machine is using ccache, but it's good to keep this in sync. @jbergstroem Hope it's not too late!

PR #131, opened 2015-07-08
dist: add ccache and systemd support to centos7

jbergstroem:
This also introduces a minor change to the init script so that it starts on boot (the [Install] section).


rvagg:
argh, start on boot is so frustrating with systemd

lgtm

jbergstroem:
Merged in d5a44ca1a24efcc37fb2e876dedfde175dc2ab6a.

PR #139, opened 2015-07-21
dist: add PPC BE machine

mhdawson:
Add PPC LE machine to inventory

	modified:   ansible-inventory
	new file:   fedora20/README.md
	new file:   fedora20/ansible-playbook.yaml
	new file:   fedora20/ansible-vars.yaml
	new file:   fedora20/resources/jenkins.service

### DIFF
--- a/setup/ansible-inventory
+++ b/setup/ansible-inventory
@@ -1,3 +1,6 @@
+[iojs-build-fedora20]
+iojs-ibm-ppcbe-fedora20-64-1
+
 [iojs-build-fedora21]
 iojs-build-fedora21-1
 
--- a/setup/fedora20/README.md
+++ b/setup/fedora20/README.md
@@ -1,0 +1,19 @@
+# io.js Build Fedora 20 Setup
+
+For setting up a Fedora 20 box
+
+```text
+Host iojs-ibm-ppcbe-fedora20-64-1
+  HostName 140.211.168.172
+  User root
+```
+
+Note that these hostnames are also used in the *ansible-inventory* file. The IP addresses will need to be updated each time the servers are reprovisioned.
+
+To set up a host, run:
+
+```text
+$ ansible-playbook -i ../ansible-inventory ansible-playbook.yaml
+```
+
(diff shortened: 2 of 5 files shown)


jbergstroem:
The compiler packages dependencies looks specific; is that necessary?

mhdawson:
The level of the compiler which you get on Fedora by default is not recent enough so I needed to specify at least a 4.8 version.  If you know there is a better way to specify at least a 4.8 version which will still match that dependency I can it a try.

mhdawson:
@jbergstroem any suggestions ?

jbergstroem:
@mhdawson from the looks of it, would referencing `gcc.ppc64` from the updates repo work?
```bash
$ yum info gcc.ppc64
Installed Packages
Name        : gcc
Arch        : ppc64
Version     : 4.8.3
Release     : 7.fc20
Size        : 36 M
Repo        : installed
From repo   : updates
```

mhdawson:
I'm heading off for vacation but I'll try it out when I get back.

mhdawson:
@jbergstroem   Finally go back to looking at this.

So I think using gcc.ppc64 would work. I did a yum install on another machine and it installed the right compiler.  

However, I think that would then make the config ppc specific.  I was thinking I might rename fedora20 to fedor20ppc but since I don't see any others like that I thought I'd ask if we have dealt with this issue before.

jbergstroem:
Perhaps override the gcc package for fedora somehow. I can have a look (or someone with more ansible experience could suggest how!).

mhdawson:
@jbergstroem I read up a bit on ansible and found one way to only install the ppc specific gcc update for ppc.  Can you take a look and let me know what you think.

mhdawson:
I had figured out how to conditionally install gcc.ppc64 and g++.ppc64 (see second commit) however, after seeing the ansible output and additional testing it seems we don't need that as 4.8 seems to get installed on fedora20 by default when just gcc and g++ are specified so I removed that part and the result is generic.

I think we should be good to go

jbergstroem:
LGTM. Great that we could avoid ppc-specific gcc stuff.

mhdawson:
Agreed

Landed as https://github.com/nodejs/build/commit/55a2a398f7d635b56a0c532857eedb8e9dd9a224

Issue #145, opened 2015-08-01
Missing which command on fedora 22 machine


targos:
Tests are failing because of it:
```
+ which git
/tmp/hudson5087374295565828274.sh: line 3: which: command not found
```

https://jenkins-iojs.nodesource.com/job/node-test-commit-linux/62/nodes=fedora22/console
https://jenkins-iojs.nodesource.com/job/node-test-commit-linux/63/nodes=fedora22/console

orangemocha:
Thanks for reporting this. I removed `which git` from the scripts a few days ago.

joaocgreis:
Fixed by the removal of that line from the script. If we ever need to add it again, 0aecff5b897bd3241468190554608e90725317e1 enables it.

PR #149, opened 2015-08-05
fedora22: Add `which`

jbergstroem:
Our build scripts uses `which` that is missing from the default fedora22 distribution.

Fixes/refs: https://github.com/nodejs/build/issues/145

/cc @targos

### DIFF
--- a/setup/fedora22/ansible-vars.yaml
+++ b/setup/fedora22/ansible-vars.yaml
@@ -14,3 +14,4 @@ packages:
   - libtool
   - ntp
   - ccache
+  - which


thefourtheye:
LGTM

targos:
LGTM

joaocgreis:
LGTM

The `which git` call was there only temporarily while debugging a problem with Ubuntu 12.04. We never know when we might need it again for some other problem, so it is good to have it.

orangemocha:
LGTM

joaocgreis:
Landed in 0aecff5b897bd3241468190554608e90725317e1.

PR #154, opened 2015-08-14
Add joaocgreis to the ssh_users list in Jenkins slaves

joaocgreis:
I added myself to the `ssh_users` list in almost all Jenkins slaves.

Left out `armv7-wheezy`, `freebsd` and `raspberry-pi` because they are not listed in Digital Ocean or Rackspace and I don't have their IPs. Should I include them anyway?

I already have access to all machines but `debian8`.

cc @nodejs/build


rvagg:
added your keys to the debian8 machine, it's currently @ 104.130.166.132

Issue #155, opened 2015-08-14
DNS configuration for the SmartOS slaves


joaocgreis:
As seen recently in https://github.com/joyent/node/issues/25858 , DNS configuration is sensitive for the tests. The test machines should be configured to use Google DNS servers, and using other servers may cause tests to fail, as discussed in https://github.com/joyent/node/issues/8056 .

I changed the DNS configuration in the SmartOS slaves by editing `/etc/resolv.conf`. I am now looking for a way to make this permanent and documented.

I have very little experience with SmartOS, by my research it seems the proper place to change the DNS configuration in SmartOS is in the zone (VM) configuration: https://wiki.smartos.org/display/DOC/How+to+create+a+zone+%28+OS+virtualized+machine+%29+in+SmartOS . Can we access this configuration?

Another way would be to add the servers to `/etc/resolv.conf` in the ansible script, with something like (untested):
```
    - name: Remove 4.2.2.2 from the list of DNS servers
      lineinfile: dest=/etc/resolv.conf state=absent regexp='^(\s*)nameserver(\s+)4.2.2.2(\s*)$'
      tags: general

    - name: Add 8.8.8.8 to the list of DNS servers
      lineinfile: dest=/etc/resolv.conf state=present regexp='^(\s*)nameserver(\s+)8.8.8.8(\s*)$' line='nameserver 8.8.8.8'
      tags: general

    - name: Add 8.8.4.4 to the list of DNS servers
      lineinfile: dest=/etc/resolv.conf state=present regexp='^(\s*)nameserver(\s+)8.8.4.4(\s*)$' line='nameserver 8.8.4.4'
      tags: general
```
However, I don't know if/when this gets overwritten.

So, should I test the ansibles changes above carefully and open a PR, can we access the zone configuration or should this be done in some other way?

chorrell:
If this is a machine on the Joyent Public Cloud you would not be able to modify the zone configuration. However, I'm /etc/resolv.conf shoul only be populated at provision time so any modifications you make should persist after a reboot. Also note that depending on which data center you provision too I think the 4.2.2.2 entry might not always be added. I'm not 100% on that though. It's prolly easier to just ensure you have the Google DNS servers and no other entries.

joaocgreis:
@chorrell that's what I needed to know, thank you! PR it is then.

joaocgreis:
Fixed by 18b3608f45a11dabff8bcf6f6558c706528c73f1.

PR #156, opened 2015-08-17
smartos: use Google DNS servers

joaocgreis:
As seen recently in https://github.com/joyent/node/issues/25858 , DNS configuration is sensitive for the tests. The test machines should be configured to use Google DNS servers, and using other servers may cause tests to fail, as discussed in https://github.com/joyent/node/issues/8056 .

I changed the DNS configuration in the SmartOS slaves by editing `/etc/resolv.conf`. This PR makes this change permanent by adding it to the ansible playbook.

Fixes: https://github.com/joyent/node/issues/25858
Fixes: #155

### DIFF
--- a/setup/smartos/ansible-playbook.yaml
+++ b/setup/smartos/ansible-playbook.yaml
@@ -20,6 +20,15 @@
       with_items: packages
       tags: general
 
+    - name: Remove all DNS servers from resolv.conf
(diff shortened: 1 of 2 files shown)


joaocgreis:
cc @nodejs/build

rvagg:
rubber-stamp lgtm but I'd like to see input from @jbergstroem if he has time

geek:
LGTM

jbergstroem:
LGTM

PR #162, opened 2015-08-23
WIP: config for new nodejs.org server

rvagg:
This is a work in progress PR, only part of the way there. You can see the nginx conf for the current nodejs.org in resources/ which I'm trying to merge in to keep consistency. We now have certificates and a server to test on so I'll be trying to make quick progress on this for testing purposes.


setup/www/resources/nodejs.org
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    location /api {
+      alias /home/dist/public/release/latest/doc/api;
+      autoindex on;
+      default_type text/plain;
+    }
+
+    location /download-stats.json {
+      alias /home/iojs/download-stats.json;
```
hoxoa: @rvagg Shouldn't that be `/home/nodejs/download-stats.json`

rvagg: ahh, yes, actually I need to remove that because it's not workable with CloudFlare anyway

rvagg:
this is the current state of the new web server so I'm going to merge this, I expect there to be changes this week as we fix for node.js builds but that can be a separate PR

PR #170, opened 2015-08-29
Raspbian ccache upgrade

rvagg:
Grabbing ccache 3.2 from source, compiling and installing it on the raspberry pi machines

Also, I've changed the ccache config for all of these to be:

```
max_size = 10.7G
sloppiness = file_macro,include_file_ctime,include_file_mtime,pch_defines,time_macros
temporary_dir = /tmp
umask = 002
```

It did just have `max_size` previously but I'm tinkering with other options to speed things up.

Please review @orangemocha, @jbergstroem et. al. This is set on the RPi machines as of right now if you want to test it out and get some numbers.

### DIFF
--- a/setup/raspberry-pi/ansible-playbook.yaml
+++ b/setup/raspberry-pi/ansible-playbook.yaml
@@ -94,6 +94,11 @@
         ./configure --prefix=/usr --with-gitconfig=/etc/gitconfig && \
         make && \
         sudo make install && \
+        sudo mkdir -p /usr/lib/ccache && \
+        sudo ln -sf $(which gcc) /usr/lib/ccache/gcc && \
+        sudo ln -sf $(which cc) /usr/lib/ccache/cc && \
+        sudo ln -sf $(which g++) /usr/lib/ccache/g++ && \
+        sudo ln -sf $(which c++) /usr/lib/ccache/c++ && \
         rm -rf git-2.5.0*
       tags: git
 
--- a/setup/raspberry-pi/ansible-playbook.yaml
+++ b/setup/raspberry-pi/ansible-playbook.yaml
@@ -96,3 +96,14 @@
         sudo make install && \
         rm -rf git-2.5.0*
       tags: git
+
+    - name: git | Download, compile and install ccache 3.2
+      shell: rm -rf ccache-3.2.3 && \
+        curl -sL http://samba.org/ftp/ccache/ccache-3.2.3.tar.xz | tar -Jxv && \
+        cd ccache-3.2.3/ && \
+        ./configure --prefix=/usr --sysconfdir=/etc -q && \
+        make && \
+        sudo apt-get remove ccache -y && \
+        sudo make install && \
+        rm -rf ccache-3.2.3
+      tags: git
--- a/setup/raspberry-pi/ansible-vars.yaml
+++ b/setup/raspberry-pi/ansible-vars.yaml
@@ -13,4 +13,3 @@ packages:
   - curl
   - gcc-4.8
   - g++-4.8
-  - ccache


rvagg:
Also FYI I discovered that pi1p-4 and pi1p-5 didn't have their NFS shares mounted so they were working on local caches only, this could explain some recent slowness; fixed now

rvagg:
left out the symlinks for /usr/lib/ccache so my test build has been doing it raw, ooops

orangemocha:
<3

Going to look at the next few builds to see if things get better.

rvagg:
I'm suspecting that my config tweaks are basically invalidating the existing cache so it might take some time

jbergstroem:
Did the earlier version not work as intended? I haven't really tinkered too much with ccache options but as long as it "works".

How about using the generic ccache installer I created a while ago? It lives in `ansible-tasks/ccache.yaml` -- feel free to suggest improvements such as adding a stub so we can specify options (guessing you're adding to `~/.ccache`)?

setup/raspberry-pi/ansible-playbook.yaml
```diff
@@ -94,5 +94,21 @@
...
         rm -rf git-2.5.0*
       tags: git
+
+    - name: git | Download, compile and install ccache 3.2
+      shell: rm -rf ccache-3.2.3 && \
+        curl -sL http://samba.org/ftp/ccache/ccache-3.2.3.tar.xz | tar -Jxv && \
+        cd ccache-3.2.3/ && \
+        ./configure --prefix=/usr --sysconfdir=/etc -q && \
+        make && \
+        sudo apt-get remove ccache -y && \
```
joaocgreis: This will fail if ccache is not installed (like if this is run twice). What about `(sudo apt-get remove ccache -y || true) && \` ?

setup/raspberry-pi/ansible-playbook.yaml
```diff
@@ -94,5 +94,21 @@
...
         sudo make install && \
+        sudo mkdir -p /usr/lib/ccache && \
+        sudo ln -sf $(which gcc) /usr/lib/ccache/gcc && \
+        sudo ln -sf $(which cc) /usr/lib/ccache/cc && \
+        sudo ln -sf $(which g++) /usr/lib/ccache/g++ && \
+        sudo ln -sf $(which c++) /usr/lib/ccache/c++ && \
         rm -rf git-2.5.0*
       tags: git
+
+    - name: git | Download, compile and install ccache 3.2
```
joaocgreis: `name: ccache | `

setup/raspberry-pi/ansible-playbook.yaml
```diff
@@ -94,5 +94,21 @@
...
+      shell: rm -rf ccache-3.2.3 && \
+        curl -sL http://samba.org/ftp/ccache/ccache-3.2.3.tar.xz | tar -Jxv && \
+        cd ccache-3.2.3/ && \
+        ./configure --prefix=/usr --sysconfdir=/etc -q && \
+        make && \
+        sudo apt-get remove ccache -y && \
+        sudo make install && \
+        rm -rf ccache-3.2.3
+      tags: git
```
joaocgreis: `tags: ccache`

setup/raspberry-pi/ansible-playbook.yaml
```diff
@@ -94,5 +94,21 @@
         ./configure --prefix=/usr --with-gitconfig=/etc/gitconfig && \
         make && \
         sudo make install && \
+        sudo mkdir -p /usr/lib/ccache && \
+        sudo ln -sf $(which gcc) /usr/lib/ccache/gcc && \
+        sudo ln -sf $(which cc) /usr/lib/ccache/cc && \
+        sudo ln -sf $(which g++) /usr/lib/ccache/g++ && \
+        sudo ln -sf $(which c++) /usr/lib/ccache/c++ && \
```
joaocgreis: This block went in the wrong section. Merge? :smile:

joaocgreis:
The ccache installer by @jbergstroem looks good, please ignore my comments above if you decide to use it.

rvagg:
aw crap, sorry @jbergstroem, your ccache.yaml looks really good, can you point me to a use of it so I can integrate here instead of my hax please?

rvagg:
baaaaah! 

```
       sudo ln -sf $(which gcc) /usr/lib/ccache/gcc && \
       sudo ln -sf $(which cc) /usr/lib/ccache/cc && \
       sudo ln -sf $(which g++) /usr/lib/ccache/g++ && \
       sudo ln -sf $(which c++) /usr/lib/ccache/c++ && \ 
```

In other words, linke `gcc`, `cc`, `g++`, `c++` to themselvs, not `ccache`, hence no ccache since this was changed, which is why the Pi's have been exceptionally slow recently .. I've fixed and will revisit this and use @jbergstroem's ansible module instead for ccache. I'll put a few runs through to warm up the cache again and let's monitor to see if we're back to decent speeds.

PR #180, opened 2015-09-02
Roll-up of Raspberry Pi changes

rvagg:
Introduced 2 new Pi2's into the mix while I was at it, including one donated by @ceejbot and one donated by [miniNodes](http://mininodes.com/)

### DIFF
--- a/setup/raspberry-pi/ansible-playbook.yaml
+++ b/setup/raspberry-pi/ansible-playbook.yaml
@@ -22,18 +22,23 @@
 
     - name: General | update-alternatives for gcc
       alternatives: name=gcc link=/usr/bin/gcc path=/usr/bin/gcc-4.8
+      tags: general
 
     - name: General | update-alternatives for cc
       alternatives: name=cc link=/usr/bin/cc path=/usr/bin/gcc-4.8
+      tags: general
 
     - name: General | update-alternatives for g++
       alternatives: name=g++ link=/usr/bin/g++ path=/usr/bin/g++-4.8
+      tags: general
 
     - name: General | update-alternatives for c++
       alternatives: name=c++ link=/usr/bin/c++ path=/usr/bin/g++-4.8
+      tags: general
 
     - name: General | update-alternatives for cpp
       alternatives: name=cpp link=/usr/bin/cpp path=/usr/bin/g++-4.8
+      tags: general
 
     - name: General | Increase file descriptor limits
       lineinfile: dest=/etc/security/limits.conf line="{{ item }}"
@@ -50,7 +55,6 @@
 
     - name: User | Download pubkey(s)
       get_url: url=https://github.com/{{ item }}.keys dest=/tmp/{{ item }}.keys
-      delegate_to: 127.0.0.1
       with_items: ssh_users
       tags: user
 
@@ -94,11 +98,6 @@
         ./configure --prefix=/usr --with-gitconfig=/etc/gitconfig && \
         make && \
         sudo make install && \
-        sudo mkdir -p /usr/lib/ccache && \
-        sudo ln -sf $(which gcc) /usr/lib/ccache/gcc && \
-        sudo ln -sf $(which cc) /usr/lib/ccache/cc && \
-        sudo ln -sf $(which g++) /usr/lib/ccache/g++ && \
-        sudo ln -sf $(which c++) /usr/lib/ccache/c++ && \
         rm -rf git-2.5.0*
       tags: git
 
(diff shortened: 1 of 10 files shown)


setup/raspberry-pi/ansible-playbook.yaml
```diff
@@ -96,3 +100,27 @@
...
+        sudo mkdir -p /usr/lib/ccache && \
+        sudo ln -sf $(which ccache) /usr/lib/ccache/gcc && \
+        sudo ln -sf $(which ccache) /usr/lib/ccache/cc && \
+        sudo ln -sf $(which ccache) /usr/lib/ccache/g++ && \
+        sudo ln -sf $(which ccache) /usr/lib/ccache/c++ && \
+        rm -rf ccache-3.2.3
+      tags: ccache
+
+    - name: Monit | Copy monit config
+      copy: src=./resources/monit-jenkins.conf dest=/etc/monit/conf.d/jenkins owner=root group=root mode=0644
```
joaocgreis: forgotten `git add ./resources/monit-jenkins.conf` ?

setup/raspberry-pi/ansible-playbook.yaml
```diff
@@ -96,3 +100,27 @@
...
+        make && \
+        sudo apt-get remove ccache -y && \
+        sudo make install && \
+        sudo mkdir -p /usr/lib/ccache && \
+        sudo ln -sf $(which ccache) /usr/lib/ccache/gcc && \
+        sudo ln -sf $(which ccache) /usr/lib/ccache/cc && \
+        sudo ln -sf $(which ccache) /usr/lib/ccache/g++ && \
+        sudo ln -sf $(which ccache) /usr/lib/ccache/c++ && \
+        rm -rf ccache-3.2.3
+      tags: ccache
```
joaocgreis: Not using @jbergstroem 's task? This whole section could be just
```
    - include: ../ansible-tasks/ccache.yaml version=3.2.3
```

rvagg: yeah, should be, I'm trying that out now on the armv8 machines to see how it goes then I'll replace this if it's all good there

setup/raspberry-pi/ansible-playbook.yaml
rvagg: thanks

setup/raspberry-pi/ansible-playbook.yaml
```diff
@@ -50,7 +55,6 @@
 
     - name: User | Download pubkey(s)
       get_url: url=https://github.com/{{ item }}.keys dest=/tmp/{{ item }}.keys
-      delegate_to: 127.0.0.1
```
joaocgreis: I've also had a problem with this, because sudo is yes and it tries to sudo in the machine that's running ansible. I've changed to
```
      delegate_to: 127.0.0.1
      become: no
```
and it's ok. Well, either way should be good as long as the lookup below works.

rvagg: no, the lookup isn't working

rvagg: thanks for the tip!

rvagg:
Added the new batch of Pi's, also added the full NFS configuration, used the shared monit task (imported @joaocgreis PR), also made the git and ccache tasks idempotent(ish) in that they test the version that's installed and skip if it's the one that's been requested to be installed.

SO, the upshot of this latest batch of changes is that I can run this playbook against the Pi cluster and it's relatively quick (in Pi terms) and doesn't mess up existing config for Pi's that are already set up. How it _should_ be.

jbergstroem:
2 questions:
 - Is it possible to check for an existing binary once and bail instead of at every task? (`ccache_version.stdout.find('{{ version }}') == -1`, etc)
 - can we improve the monit check to look for slave name? That way we could use it out of the box for jobs where we run multiple slaves (centos5-gcc41 comes to mind)

I'm guessing the checks are for packages that are already installed through a package manager or similar? How about attempting to uninstall them?

rvagg:
> Is it possible to check for an existing binary once and bail instead of at every task? (ccache_version.stdout.find('{{ version }}') == -1, etc)

I looked for this but didn't find it, perhaps you can find something better?

> can we improve the monit check to look for slave name? That way we could use it out of the box for jobs where we run multiple slaves (centos5-gcc41 comes to mind)

```
$ ssh iojs-ns-pi1p-1 ps auxww | grep java
iojs      2090  1.0  9.2 394412 41208 ?        Sl   Sep10  18:07 java -jar slave.jar -jnlpUrl https://jenkins-iojs.nodesource.com/computer/iojs-nodesource-raspbian-wheezy-pi1p-1/slave-agent.jnlp -secret secret_omitted_from_paste
```

So yes, it could be done for multiple slaves, just match the jenkins id, in this case `iojs-nodesource-raspbian-wheezy-pi1p-1`, currently we're just doing a check for `matching slave.jar`.

For the purpose of the Pi's this is fine, perhaps the next round of changes can make it more flexible.

jbergstroem:
Regarding version check;  the simplest improvement would be using ansibles [`register`](http://docs.ansible.com/ansible/playbooks_conditionals.html#register-variables) and check against that. The other option would be combining above result with a [fail check](http://docs.ansible.com/ansible/fail_module.html). 

Anyway, above suggestions are improvements, not requirements; LGTM for progress.

PR #181, opened 2015-09-02
Update nodejs.org

No9:
So looking at this a little more it turns out 443 request works  
https://www.nodejs.org successfully rewrites to https://nodejs.org but http://www.nodejs.org does not. 
Looking at line 150 I see the order of the server_name is different than line 17
Totally guessing as I can't test but this seems a reasonable try


rvagg:
Thanks @no9! I've fixed this in 631fda6 but you were right that this is a `server_name` issue, I've just opted for being explicit in the redirect instead.

PR #186, opened 2015-09-04
added armv8 ubuntu 14.04 support

rvagg:
not quite ready, I need to turn one of the machines into a release machine and that's not reflected here yet

### DIFF
--- a/setup/ansible-inventory
+++ b/setup/ansible-inventory
@@ -39,6 +39,9 @@ test-softlayer-ubuntu14-x64-1
 test-softlayer-ubuntu14-x86-1
 test-digitalocean-ubuntu14-x86-1
 test-digitalocean-ubuntu14-x64-1
+iojs-ns-xgene-1
+iojs-ns-xgene-2
+iojs-ns-xgene-3
 
 [iojs-build-ubuntu1204]
 test-digitalocean-ubuntu12-x64-1
--- a/setup/ubuntu14.04/ansible-playbook.yaml
+++ b/setup/ubuntu14.04/ansible-playbook.yaml
@@ -1,7 +1,6 @@
 ---
 - hosts: iojs-build-ubuntu1404
-
-  remote_user: root
+  sudo: yes
 
   tasks:
 
@@ -28,6 +27,7 @@
     - name: User | Download pubkey(s)
       get_url: url=https://github.com/{{ item }}.keys dest=/tmp/{{ item }}.keys
       delegate_to: 127.0.0.1
+      become: no
       with_items: ssh_users
       tags: user
 
--- a/setup/ubuntu14.04/ansible-vars.yaml
+++ b/setup/ubuntu14.04/ansible-vars.yaml
@@ -2,10 +2,10 @@
 server_user: iojs
 ssh_users:
   - rvagg
-  - wblankenship
-  - ryanstevens
   - mhdawson
   - joaocgreis
+  - jbergstroem
+  - orangemocha
 packages:
   - openjdk-7-jre
   - git
@@ -15,4 +15,3 @@ packages:
   - libtool
   - curl
   - ntp
-  - ccache


jbergstroem:
ohh! can we get a `cat /proc/cpuinfo`?

rvagg:
revisiting this, I've rebased and removed the start.sh and monit stuff if we're just going to jump in on systemd, I have some questions inline

setup/ubuntu14.04/ansible-playbook.yaml
```diff
@@ -1,7 +1,6 @@
 ---
 - hosts: iojs-build-ubuntu1404
-
-  remote_user: root
+  sudo: yes
```
rvagg: @jbergstroem remote user on these machines isn't root, I'd like it to log in and `sudo` but I know other servers we're using log in as `root`, any ideas on how we can reconcile the two?

setup/ubuntu14.04/ansible-vars.yaml
```diff
@@ -15,4 +15,3 @@ packages:
   - libtool
   - curl
   - ntp
-  - ccache
```
rvagg: @jbergstroem are we doing ccache from source or putting up with what comes in by default?

setup/ubuntu14.04/ansible-playbook.yaml
jbergstroem: We have a few similar cases in other os/provider combos. In my `ansible 2.0` refactor I've solved this by doing exceptions based on above pair. We can either solve it through the inventory file or through logic in handlers. As for now; you should be able to override with `host_vars`?

setup/ubuntu14.04/ansible-vars.yaml
jbergstroem: I've rarely run into issues with ccache, especially versions `>=3`. If that's provided I don't see the point in using our hand baked stuff.

One benefit _could_ be consistent paths to the compiler symlinks, but since most init scripts vary per os it's of less significance.

setup/ansible-inventory
```diff
@@ -39,6 +39,9 @@ test-softlayer-ubuntu14-x64-1
 test-softlayer-ubuntu14-x86-1
 test-digitalocean-ubuntu14-x86-1
 test-digitalocean-ubuntu14-x64-1
+iojs-ns-xgene-1
+iojs-ns-xgene-2
+iojs-ns-xgene-3
```
jbergstroem: If you're rebaking these machines, could you perhaps use the new naming convention? In this case it would be
`test-nodesource-ubuntu14-arm-{1,2,3}`. Or is it perhaps `arm64`? (should match with what's passed to `--dest-cpu`).

setup/ubuntu14.04/ansible-playbook.yaml
```diff
@@ -28,6 +27,7 @@
     - name: User | Download pubkey(s)
       get_url: url=https://github.com/{{ item }}.keys dest=/tmp/{{ item }}.keys
       delegate_to: 127.0.0.1
+      become: no
```
jbergstroem: I've just ignored these parts of the playbook since we nowadays use the shared pubkeys (Ignored while running the playbook by stepping it through). I don't have a problem with this change but it's probably pointless.

setup/ubuntu14.04/ansible-vars.yaml
```diff
@@ -2,10 +2,10 @@
 server_user: iojs
 ssh_users:
   - rvagg
-  - wblankenship
-  - ryanstevens
   - mhdawson
   - joaocgreis
+  - jbergstroem
+  - orangemocha
 packages:
```
jbergstroem: Same as above (downloading pubkeys). We don't really use per-user keys any longer. We could always walk through all files and remove all things related to pubkeys but that means we'd have to replace it with the new system which I didn't want to spend too much time on outside of the `2.0` refactor.

PR #189, opened 2015-09-07
more tweaks for new nodejs.org server

rvagg:
WIP, so far this just fixes nightly & rc auto-promote, nightly creation is still not happening and that'll be one of the next things


rvagg:
Just added config for http://dist.libuv.org/ which we are now hosting off the main web server, it has a user account `libuv` that core members of that team have access to and a `dist/` directory that is exposed via nginx. We're also getting a wildcard ssl cert for *.libuv.org and I'll update the config and secrets repo when we have that in place.

jbergstroem:
I can't see the relevant nginx config for libuv/dist stuff. Is that coming once ssl certs are in place?

PR #187, opened 2015-09-04
update armv7-wheezy config

rvagg:

PR #192, opened 2015-09-10
setup: add monit to ansible-tasks

joaocgreis:
Ref: https://github.com/nodejs/build/pull/187#issuecomment-137994003

Also changed the cycle to 15 seconds (was 2 minutes).

I don't like the way the conf file is found, but didn't find a better one. Suggestions?

@jbergstroem 

> (re suggestion above) Config probably needs to live elsewhere though since some machines are into init scripts. Also, newer upstarts and systemd afaik handles restarting/keepalive for you.

For now this is only apt-get, but we could generalize. Should not be a problem on current machines.

### DIFF
--- a/setup/ansible-tasks/monit.yaml
+++ b/setup/ansible-tasks/monit.yaml
@@ -1,0 +1,23 @@
+---
+# Installs and configures monit for Debian based systems.
+# @requires: monit_conf_file=../ansible-tasks/resources/monit-jenkins.conf
+
+- name: Monit | Install deb package
+  apt: name=monit update_cache=yes state=latest
+  tags: monit
+
+- name: Monit | Set the cycle length to 15 seconds
+  lineinfile: dest=/etc/monit/monitrc state=present regexp="^(\s*)set(\s+)daemon" line="set daemon 15"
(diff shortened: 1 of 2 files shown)


rvagg:
:+1: I'm going to try working this into the raspbian setup scripts I'm working on right now, I have to set up all these new Pi's so am taking the opportunity to clean some stuff up

jbergstroem:
Perhaps abstract the config slightly since it will be a useful combo on deployments where we launch jenkins from sysvinit as well?

rvagg:
I'm finding that it's just not necessary to add sysvinit when you're also doing monit because monit will bring the process up when it does its first check anyway, the Pi's are set up this way and happily restart when I reboot them

jbergstroem:
Ok, cool.

jbergstroem:
FWIW, exploring putting `mon`  (instead of monit) into our sysvinit script over at #292.

joaocgreis:
I don't think this is still needed, with `refactor-the-world`.

jbergstroem:
@joaocgreis correct, it's aleady in there.

Issue #191, opened 2015-09-10
Create test script to verify redirects don't break stuff


fhemberger:
With the huge amount of redirects from the old nodejs.org website, it can become quite tedious to check for possible side effects when changing the server config.

Current examples:
- https://github.com/nodejs/new.nodejs.org/issues/124
- https://github.com/nodejs/new.nodejs.org/issues/125

Cases which need to be tested:
- http/https links to dist tarballs
- http/https links to old windows x32/x64 builds
- http/https links to new windows x32/x64 builds (different file name/directory)
- http/https links to dist/latest/, dist/latest-v0.10.x/, dist/latest-v0.12.x/  
- http/https links to previous versions of the api docs
- http/https links for [subdomain].nodejs.org
- …

What would be the preferred way to create such a test suite? Shell script with wget/curl checking for response codes and redirect locations? Something like [supertest](https://www.npmjs.com/package/supertest)?

leachiM2k:
+1

fhemberger:
Maybe it would also be a good idea to put that long list of redirects into an external file and include it into  nginx.conf, to keep the main config short and clean.

Trott:
@nodejs/build @fhemberger Should this remain open?

fhemberger:
@Trott I still think this would be useful in the light of the planned relaunch, but in the end it depends on the needs of the build team.

maclover7:
Closing for now as something that would be nice to have, but not currently within our means. If someone wants to tackle this, please feel free to reopen.

PR #193, opened 2015-09-10
Add testing for nodejs.org

fhemberger:
First idea for implementing #191, a mocha test-suite issuing HEAD requests.
- [x] http/https links to dist tarballs
- [ ] http/https links to old windows x32/x64 builds
- [ ] http/https links to new windows x32/x64 builds (different file name/directory)
- [x] http/https links to dist/latest/, dist/latest-v0.10.x/, dist/latest-v0.12.x/
- [ ] http/https links to previous versions of the api docs
- [ ] http/https links for [subdomain].nodejs.org

One redirect test deliberately failing: https://blog.nodejs.org/ redirects to http://blog.nodejs.org/ (which redirects again) instead of redirecting directly to https://nodejs.org/en/blog/.

### DIFF
--- a/tools/nodejs.org-test/.editorconfig
+++ b/tools/nodejs.org-test/.editorconfig
@@ -1,0 +1,17 @@
+# http://editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
+
+[package.json]
+indent_style = space
+indent_size = 2
--- a/tools/nodejs.org-test/.eslintrc
+++ b/tools/nodejs.org-test/.eslintrc
@@ -1,0 +1,24 @@
+{
+  "parser": "babel-eslint",
+  "env": {
+    "node": true,
+    "mocha": true
+  },
+  "rules": {
+    "complexity": [1, 8],
+    "consistent-return": 0,
+    "eol-last": 1,
+    "key-spacing": 0,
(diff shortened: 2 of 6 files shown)


fhemberger:
@nodejs/build Is this something useful for you to work with?

rvagg:
@fhemberger yes, this is great, I probably wouldn't have chosen mocha but that's just personal bias and whoever starts the work gets to choose in my books

fhemberger:
What would you prefer instead of mocha? It was just the first thing that came to my mind. I'd just like to use a test runner with simple, readable output, where everybody can add tests easily.

rvagg:
it's fine, proceed as you prefer and we'll adapt

jbergstroem:
This could alternatively be written as part of the nginx test tool (It's perl and annoying though). Benefit being it actually running against the web server config instead of testing the live stuff.

DavidTPate:
I've been using [RunScope](https://www.runscope.com) for some of my Projects. They have a good method of configuration for tests against live systems as well as for running tests like CRON and also from multiple regions in the world with performance metrics and such.

They have a markup for importing tests which could be added to a repo and used to configure RunScope.

fhemberger:
Hmm, not sure that RunScope is the right match. @jbergstroem If you want to port it, please go ahead (sorry, I don't do perl).

jbergstroem:
@fhemberger was just thinking that it'd be great being able to reuse the nginx config as part of the test suite. I won't pursue this and since I'm not driving I don't get to decide where to turn :)

fhemberger:
Closing this, there seems to be no interest in pursuing this idea further.

PR #195, opened 2015-09-16
freebsd: assorted updates

jbergstroem:
Few things improved:
 - switch to openjdk-jre which has less dependencies
 - introduce `-r` to daemon(8) which restarts child on exit
 - Pass BROADCAST since `255.255.255.255` isn't always available (ref: https://github.com/nodejs/node/pull/2851)


jbergstroem:
Any feedback? The reason java path is different between the linter machines and our builders is because packages available on the linter (through do) seems slightly different.

jbergstroem:
FWIW, this is already in production to improve auto-restarts. Unless anyone have objections I'll merge this to keep in sync with production in a few days.

PR #208, opened 2015-10-07
build: Add ubuntu14 benchmarking machine

mhdawson:
Add iojs-softlayer-benchark to the inventory
Update template for start script so that it points
to the right CI and does not assume a fixed naming of the
machines


mhdawson:
@ jbergstroem can you take a look

jbergstroem:
The only question would be whether we need that gyp slave, but I reckon we can refactor that later.

LGTM

mhdawson:
Landed as 0fea5dcccef0273d81fb70b623e75d7c80147ccd

PR #214, opened 2015-10-09
Update redirects for legacy images

fhemberger:
See https://github.com/nodejs/new.nodejs.org/pull/227


rvagg:
should be live now, please test and confirm

fhemberger:
@rvagg Thanks, works.

PR #215, opened 2015-10-10
build: add Windows Ansible playbook

joaocgreis:
Replaces the text instructions in README.md with a ansible playbook that can be used like other OSs, after configured. The new README.md includes instructions to run ansible with the correct usernames and passwords.

~~Two new machines running Windows 10 with Visual Studio 2015 were added to CI, this was used to set them up. Hence, this was tested (only) with Windows 10, but should be usable for other versions.~~


joaocgreis:
Updated. Now also tested in Windows 2008R2 and 2012R2.

joaocgreis:
This has had plenty of testing over the last year, was improved several times, and is now a very stable version. Even if we update to Ansible 2 soon, it would be good to land this version to keep it in history. @nodejs/build do you mind if I go ahead and land this?

jbergstroem:
@joaocgreis All for it. I have a few suggestions but lets focus our work on the 2.x branch instead. LGTM

joaocgreis:
Landed in https://github.com/nodejs/build/commit/23d3246014732b9325a687c3c9d10f1b11bf9ccf

Issue #217, opened 2015-10-14
Add field for LTS in /dist/index.json


phillipj:
Would it be possible to distinguish LTS from stable releases as mentioned in [new.nodejs.org #200 comment](https://github.com/nodejs/new.nodejs.org/issues/200#issuecomment-146051237)? Or possibly  a generic `policy` field containing `stable | lts | maintenance` as mentioned in [node PR #3343](https://github.com/nodejs/node/pull/3343) for each release?

rvagg:
thanks for the reminder, see #218

PR #218, opened 2015-10-14
add lts field and support older versions in index.*

rvagg:
See https://nodejs.org/download/release/index.json and https://nodejs.org/download/release/index.tab for the output of these changes. The "lts" field is "Argon" for `^0.4.2` and otherwise `"-"` in index.tab and `false` in index.json.

I'm not doing anything other than the LTS name, #217 suggests adding "stable", "lts" and "maintenance" and currently I don't see a need for this, nor is there easy logic to fetch this full information from the repos.

I've also added SHASUMS256.txt (tho not a signed version) and therefore support for the index.* files for all of the older directories, back to v0.5.1. @ljharb may be interested in this.

### DIFF
--- a/setup/www/tools/dist-indexer/dist-indexer.js
+++ b/setup/www/tools/dist-indexer/dist-indexer.js
@@ -354,7 +354,7 @@ function inspectDir (dir, callback) {
     , date
 
   if (!gitref) {
-    return fs.stat(dir, function (err, stat) {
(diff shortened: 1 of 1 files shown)


ljharb:
Thanks - I'd need `index.tab` to list every version down to v0.1.14 for me to be able to throw out my legacy scraping code :-/ any chance of that happening?

(this unblocks https://github.com/creationix/nvm/issues/870 as well, so thanks!)

PR #223, opened 2015-10-16
remove hard-coded git refs for nightly types, use cmdline

rvagg:
for nightly-builder


rvagg:
also added a change from iojs to nodejs for nightly-builder, not going to add an option to make it switchable since I don't see io.js releases on our horizon

PR #228, opened 2015-10-26
freebsd: improve init resiliency and jenkins environment

jbergstroem:
We have to sunset our freebsd vm's at Voxer. This commit replaces The FreeBSD ones and then some:

- resolve java path which makes changes to openjdk "just work"
- remove unnecessary environment variables

With this, we will unfortunately not test for 32-bit freebsd any longer (because DO only have 64-bit freebsd vm's). On the other hand, we now run two 64-bit vm's in parallel, allowing for slightly higher throughput.

### DIFF
--- a/setup/freebsd/README.md
+++ b/setup/freebsd/README.md
@@ -1,6 +1,12 @@
 # io.js Build FreeBSD Setup
 
-The current FreeBSD lives at Digitalocean.
+The current FreeBSD VMs lives at Digitalocean.
+
+You have to install `python` before running ansible. Log in to the VM
+and run the following:
+```
+sudo pkg update && sudo pkg install -y python
+```
 
 To set up hosts, make sure you add them to your ssh config first:
 ```
@@ -13,14 +19,12 @@ Host test-digitalocean-freebsd10-x64-2
   User freebsd
 ```
 
-Note that these hostnames are also used in the ansible-inventory file. The IP addresses will need to be updated each time the servers are reprovisioned.
+Note that these hostnames are also used in the ansible-inventory file.
+The IP addresses will need to be updated each time
+the servers are reprovisioned.
 
 To set up a host, run:
 
 ```text
 $ ansible-playbook -i ../ansible-inventory ansible-playbook.yaml
 ```
-
-**Users**: The ansible-vars.yaml file contains a list of users who's GitHub public keys are pulled and placed into
-authorized_keys for both root and iojs users. This file should be updates when new users are added to the build project
-who are able to help maintain the containerized builds.
--- a/setup/freebsd/README.md
+++ b/setup/freebsd/README.md
@@ -28,3 +28,7 @@ To set up a host, run:
 ```text
 $ ansible-playbook -i ../ansible-inventory ansible-playbook.yaml
 ```
+
+If you have to update settings related to how jenkins is deployed,
+the configuration lives in `/usr/local/etc/rc.conf.d/jenkins`.
+
--- a/setup/freebsd/README.md
+++ b/setup/freebsd/README.md
@@ -2,12 +2,6 @@
 
 The current FreeBSD VMs lives at Digitalocean.
 
-You have to install `python` before running ansible. Log in to the VM
-and run the following:
-```
-sudo pkg update && sudo pkg install -y python
-```
-
 To set up hosts, make sure you add them to your ssh config first:
 ```
 Host test-digitalocean-freebsd10-x64-1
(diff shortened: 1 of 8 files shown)


jbergstroem:
This will also "fix" a few of our flaky tests related to routing since there's no magic concerning `LOCALHOST` and `BROADCAST` any longer (side effect of freebsd jails).

jbergstroem:
We could explore setting up a jail in one of the freebsd slaves. Thinking this would hurt automating through ansible.

setup/freebsd/README.md
```diff
@@ -1,26 +1,30 @@
 # io.js Build FreeBSD Setup
 
-The current FreeBSD lives behind a VPN at [Voxer](http://voxer.com).
-Should you need access, speak to [https://github.com/rvagg](rvagg) for
-further information.
+These machines lives at DigitalOcean, running as default VM's.
```
rmg: s/lives/live/

setup/freebsd/ansible-vars.yaml
```diff
@@ -4,8 +4,9 @@ init_script_path: /usr/local/etc/rc.d/jenkins
 ssh_users:
   - rvagg
   - jbergstroem
+  - joaocgreis
```
rmg: Is this intentional? Seems like the kind of thing that should get a mention in the commit message if so.

setup/freebsd/ansible-vars.yaml
```diff
@@ -4,8 +4,9 @@ init_script_path: /usr/local/etc/rc.d/jenkins
...
 packages:
-  - openjdk-jre
+  - openjdk
```
rmg: Mostly for my own curiosity, is this just because the JDK package has saner defaults? Or is the compiler actually required?

setup/freebsd/README.md
```diff
@@ -1,26 +1,30 @@
...
+
+You  have to install `python` before running ansible. Log in to the VM
```
rmg: Extra space.

rmg:
Some _very_ minor nits...

Assuming the decision has already been made and this is just the code for it, LGTM.

setup/freebsd/ansible-vars.yaml
jbergstroem: Copied from our linter config (which is also freebsd). I reckon this list should be expanded; probably with @orangemocha.

setup/freebsd/ansible-vars.yaml
jbergstroem: I've seen discrepancies between the different package sources of openjdk (the stuff on DO is slightly different from what I've seen at voxer). OpenJDK had less dependencies for some reason.

jbergstroem:
> @rmg said:
> Assuming the decision has already been made and this is just the code for it

We were told by Voxer to sunset the current resources. We've now got two 64-bit FreeBSD bots running at DO and I've asked [twitter](https://twitter.com/eatingfoodbrb/status/658457815346905089) for more.

joaocgreis:
I did not run this, but the changes LGTM.

jbergstroem:
Update: we're going to go with one vm at joyent and one at digital ocean. We're still short on 32-bit vm's which is unfortunate.

Trott:
Is there an expected/planned date for this? (I have no idea how these cutovers work.)

jbergstroem:
@Trott sorry, pretty much based on effort which varies. I'll see if I can squeeze this tomorrow -- just need to free resources on DO ahead.

Trott:
@jbergstroem Thanks. I don't mean to rush you or anything. There's no urgency.

jbergstroem:
@trott all good, I just agree that "as soon as possibly" is a shitty eta :)

jbergstroem:
I've now retired the remaining bots that are either run in jails or having different network setups.

jbergstroem:
sorry, didn't mean to close.

jbergstroem:
Rebasing since I already did a few things [here](3a65bc81b9228eb1de921fb078921dca2925aa32).

jbergstroem:
I've retired our Voxer freebsd bots as of today. This PR kind of took another direction but I just want to make sure we get all changes in there.

jbergstroem:
Updating topic since this issue floated a bit more south than expected.

jbergstroem:
I'm going to close this since I'm venturing further away from the original topic. I'll create a new PR which will illustrate a few of the simplified things we have in the pipe for our ansible stack.

PR #231, opened 2015-10-28
stop download directory being crawled

rnsloan:
See https://github.com/nodejs/new.nodejs.org/issues/240

Will update the robots.txt file on nodejs.org after this is merged.


rvagg:
merged and deployed, thanks @rnsloan

PR #270, opened 2015-11-26
doc: add Softlayer to list of Hardware Sponsors

mhdawson:
Add another line to identify IBM is also contributing
through Softlayer

Not sure if this is the best way but I would like to call out Softlayer and the OSU Open Source Lab separately.


mhdawson:
@jbergstroem @rvagg

jbergstroem:
LGTM, overdue :+1:

rvagg:
aye, lgtm

we need to remove linaro and put something about ARM Holdings, also we're getting FreeBSD from somewhere new right @jbergstroem?

jbergstroem:
@rvagg correct. lets add the freebsd stuff once its in.

mhdawson:
Landed as 28a1a5e6062ea7b50139f652c1dbd8d83139612b

PR #273, opened 2015-12-01
centos5/6: multiple slave refactor

jbergstroem:
We previously used to run multiple jenkins on centos5 and 6 to handle gcc41/gcc44 jobs. This logic is now handled by passing labels in jenkins (more specifically centos6-64-gcc44, centos5-32-gcc41). Doing so, we can remove all the extra server stuff.

While at it, a minor fix to the centos6 init script for ccache, simplifying server_id and using our custom ccache installer.


joaocgreis:
Did not run this, LGTM if it works.

setup/centos6/resources/jenkins.initd
```diff
@@ -21,19 +21,21 @@ PIDFILE=/var/run/jenkins/$NAME.pid
 # Read config
 [ -f "$JENKINS_CONFIG" ] && . "$JENKINS_CONFIG"
 
-# Add scl-specific path as well as regular path
-JENKINS_PATH="$JENKINS_PATH:$PATH"
+# Add scl-specific path as well as regular path.
+# We add /usr/local/bin since that's where ccache symlinks live
+JENKINS_PATH="/usr/local/bin:$JENKINS_PATH:$PATH"
```
joaocgreis: Isn't this needed for centos5 as well?

jbergstroem: yes, invoking our ccache script installs/symlinks into `/usr/local/bin`, and I just switched the python symlink path from our custom `/home/iojs/bin/` to `/usr/local/bin`.

PR #286, opened 2015-12-23
redirect for Windows Environment wiki page

joaocgreis:
This is necessary to land https://github.com/nodejs/node/pull/4111

@rvagg is this the correct way to add the redirect?

### DIFF
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -380,4 +380,6 @@ server {
     rewrite ^/images/yahoo-logo.png                               https://$server_name/static/legacy/images/yahoo-logo.png permanent;
 
     rewrite ^/images/(.*)                                         https://$server_name/static/images/$1 permanent;
+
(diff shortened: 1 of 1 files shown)


rvagg:
yes, I think remove the first one though, let the direct go through from http to https and we'll always just link to the https version when we refer to it publicly anyway

jbergstroem:
+1 to https only. Redirects like these will stay in a config forever; the less rules the better

joaocgreis:
Thanks, updated

orangemocha:
LGTM

joaocgreis:
@rvagg what more needs to be done here to update the server config?

rvagg:
Sorry! Someone with access to the web server just needed to drop this in /etc/nginx/sites-available/ and reload nginx. I've done that now, https://nodejs.org/windows-environment is active.

joaocgreis:
Thanks!

jbergstroem:
@joaocgreis (you should have access with the infra key)

PR #298, opened 2016-01-07
Add metrics data generation and publication tooling

rvagg:
As agreed by the security group, we're ready to push forward on exposing this data.

This PR is the raw tooling, it's not wired up in ansible to be installed, tbh that makes me tired just thinking about it (that makes me a bad person, I know, I'm sorry).

All of the scripts are currently in /root/metrics/ on the main web server and there's a crontab entry:

```
0 */4   * * *   root    /root/metrics/download-counts.sh
```

I'm not checking in the GeoLite City database here, but there's a script in here to manually install and update that.


rvagg:
I'm going to take no further discussion here as an implicit vote of confidence in this approach!

jbergstroem:
Post-lgtm

PR #314, opened 2016-01-29
nodejs.org: remove SSL-config for new.nodejs.org

phillipj:
Its been a while since we launched nodejs.org and thereby moved from new.nodejs.org.
This minor cleanup removes the last evidence of new.nodejs.org from the nginx config.

### DIFF
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -150,7 +150,7 @@ server {
     listen *:443 default_server ssl spdy;
     listen [::]:443 default_server ipv6only=on ssl spdy;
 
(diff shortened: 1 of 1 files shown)


rvagg:
lgtm, but again I'm going to defer to @jbergstroem, I have nginx-config fatigue

jbergstroem:
I need to look further into this before I can LGTM.

jbergstroem:
Did some googling; couldn't find any references to `new.nodejs.org` (besides the git repo) so lets just get rid of it. I'll create a PR that'll also take care of the https redirect.

jbergstroem:
See #389.

jbergstroem:
Ok, I'll kill it; rationale being less cruft.

PR #315, opened 2016-01-29
nodejs.org: changed www host check to its own server directive

phillipj:
Fixes the [server name (if)](https://www.nginx.com/resources/wiki/start/topics/tutorials/config_pitfalls/#server-name-if) pitfall mentioned by nginx.

> When NGINX receives a request no matter what is the subdomain being requested, be it www.example.com or just the plain example.com this if directive is **always** evaluated. Since you’re requesting NGINX to check for the Host header for **every request**. It’s extremely inefficient. You should avoid it.

### DIFF
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -14,7 +14,7 @@ log_format nodejs    '$remote_addr - $remote_user [$time_local] '
 server {
     listen *:80 default_server;
     listen [::]:80 default_server ipv6only=on;
-    server_name www.nodejs.org nodejs.org;
+    server_name nodejs.org;
 
     access_log /var/log/nginx/nodejs.org-access.log nodejs;
     error_log /var/log/nginx/nodejs.org-error.log;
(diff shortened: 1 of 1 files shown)


jbergstroem:
I've been meaning to do this for a while as well. Thanks for sorting it -- LGTM.

rvagg:
Imma leave it to you @jbergstroem, it seems that whenever I change anything at the top of the `server` sections I break everything, no matter how trivial.

jbergstroem:
I have a vm somewhere with a ton of changes, this being one of them. I'll spin another one up tomorrow and verify the change.

phillipj:
Thanks @jbergstroem. Guess I should try getting ansible up-n-running myself

jbergstroem:
Small note: you cannot redeclare `ipv6only=on`, but it's not really needed. Will merge/land without it.

jbergstroem:
Merged in e0880ae7cb5da7ec4270d805e9c131c043409506 without the extra `ipv6only=on`.

PR #316, opened 2016-01-29
nodejs.org: falling back to english when page hasn't been translated

phillipj:
Refs https://github.com/nodejs/nodejs.org/pull/490

Currently translation working groups will have to translate all pages to their respective
language, if not most pages would end in displaying a (english) 404 page. That's a massive
requirement on translation groups, and keeping up with future changes are probably near impossible.

Lowering the barrier and letting translation groups start on the frontpage and work their way
down a couple of page depths, should be alot simpler and more rewarding.

NB! This only affects local development as nodejs.org runs nginx serving strictly static content.
Getting the same behaviour in production requires changes to the nginx config.

Example of the current 404 issue; navigate any of the headers links in the current Korean translation: https://nodejs.org/ko/. I haven't done my dues when it comes to translating (yet), this is one of the main reasons I've decided not to.

### DIFF
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -40,7 +40,11 @@ server {
     default_type text/plain;
     index index.html;
 
+    error_page 404 @localized_404;
+
     location / {
+        try_files $uri $uri/ @english_fallback;
+
         location ~ \.json$ {
             add_header access-control-allow-origin *;
         }
@@ -54,6 +58,19 @@ server {
         location ~ \.json$ {
             add_header access-control-allow-origin *;
         }
+    }
+
+    # instead of serving a 404 page when a page hasn't been translated
+    location @english_fallback {
+        if ($uri ~* ^/(it|ko)/) {
+            set $lang $1;
+        }
+        rewrite ^/(it|ko)/(.*)$ http://nodejs.org/en/$2;
+    }
+
+    # serve a localized 404 page if we've got $lang set from @english_fallback
+    location @localized_404 {
+        try_files /$lang/404.html /en/404.html;
     }
 }
 
@@ -190,7 +207,7 @@ server {
     gzip_disable "MSIE [1-6]\.";
     gzip_types text/plain text/css application/x-javascript text/xml application/xml application/xml+rss text/javascript;
 
-    error_page 404 /en/404.html;
+    error_page 404 @localized_404;
 
     root /home/www/nodejs;
     default_type text/plain;
@@ -199,9 +216,24 @@ server {
     location / {
         rewrite ^/$ /en/ redirect;
 
+        try_files $uri $uri/ @english_fallback;
+
         location ~ \.json$ {
             add_header access-control-allow-origin *;
         }
+    }
+
+    # instead of serving a 404 page when a page hasn't been translated
+    location @english_fallback {
+        if ($uri ~* ^/(it|ko)/) {
+            set $lang $1;
+        }
+        rewrite ^/(it|ko)/(.*)$ https://nodejs.org/en/$2;
+    }
+
+    # serve a localized 404 page if we've got $lang set from @english_fallback
+    location @localized_404 {
+        try_files /$lang/404.html /en/404.html;
     }
 
     location /documentation/ {
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -62,10 +62,10 @@ server {
 
     # instead of serving a 404 page when a page hasn't been translated
     location @english_fallback {
-        if ($uri ~* ^/(it|ko)/) {
+        if ($uri ~* ^/(it|ko|zh-cn)/) {
             set $lang $1;
         }
-        rewrite ^/(it|ko)/(.*)$ /en/$2;
+        rewrite ^/(it|ko|zh-cn)/(.*)$ /en/$2;
     }
 
     # serve a localized 404 page if we've got $lang set from @english_fallback
@@ -225,10 +225,10 @@ server {
 
     # instead of serving a 404 page when a page hasn't been translated
     location @english_fallback {
-        if ($uri ~* ^/(it|ko)/) {
+        if ($uri ~* ^/(it|ko|zh-cn)/) {
             set $lang $1;
         }
-        rewrite ^/(it|ko)/(.*)$ /en/$2;
+        rewrite ^/(it|ko|zh-cn)/(.*)$ /en/$2;
     }
 
     # serve a localized 404 page if we've got $lang set from @english_fallback
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -65,7 +65,7 @@ server {
         if ($uri ~* ^/(it|ko)/) {
             set $lang $1;
         }
-        rewrite ^/(it|ko)/(.*)$ http://nodejs.org/en/$2;
+        rewrite ^/(it|ko)/(.*)$ /en/$2;
     }
 
     # serve a localized 404 page if we've got $lang set from @english_fallback
@@ -228,7 +228,7 @@ server {
         if ($uri ~* ^/(it|ko)/) {
             set $lang $1;
         }
-        rewrite ^/(it|ko)/(.*)$ https://nodejs.org/en/$2;
+        rewrite ^/(it|ko)/(.*)$ /en/$2;
     }
 
     # serve a localized 404 page if we've got $lang set from @english_fallback


jbergstroem:
This will make all 404's redirect to `/en/` for the nodejs.org scope (besides dist) - right? Not sure that is the desired effect.

phillipj:
Good question. Got to admit I didn't think about URLs other than the actual
website. I'll dig more into that as soon as I'm back from one week vacation.

phillipj:
> This will make all 404's redirect to /en/ for the nodejs.org scope (besides dist) - right?

I assume you would rather see it check the /en/ equivalent exist before doing a redirect, correct? Haven't been able to get that working... Got a trick up your sleeve @jbergstroem?

phillipj:
To elaborate on what I've tried to accomplish without luck so far:

```
location @english_fallback {
    set $english_filename '';

    if ($request_filename !~* /en/) {
        set $english_filename ...; # need string replace??
    }

    if (-e $english_filename) {
        rewrite ^(/\w+/)(.*)$ http://nodejs.org/en/$2;
    }

    return 404;
}
```

The biggest challenge here is doing a string replace on the current `$request_filename` to get the english equivalent need to do the existence check. AFAIK the kind of manual string replace functionality needed, would require `nginx_substitutions_filter` to be installed.

Anyway the use `if` is strongly discouraged by the nginx team and community...

jbergstroem:
How about going at it the other way? redirecting all other 'known' languages that doesn't exist to en?

phillipj:
@jbergstroem something like this?

```
location @english_fallback {
    rewrite ^/(it|ko)/(.*)$ http://nodejs.org/en/$2;
    return 404;
}
```

That surely removes the complexity and no if's in sight :+1:

phillipj:
@jbergstroem just updated the PR with the simplified version posted above ^^ with explicit known languages as you suggested. Also added the english fallback to the `https` server directive.

phillipj:
@jbergstroem any thoughts on this after it got updated?

Fishrock123:
Didn't we have something like this for iojs.org?

Looks like it does work on iojs.org: https://iojs.org/fa/es6.html

Ideally, however, it would not re-write to /en/ so that any future links can still go to the native language? I think? I'm not sure.

Fishrock123:
Hmm, looks like we just made redirect static files for each page: https://github.com/nodejs/iojs.org/blob/master/source/static/es6.html

Fishrock123:
As a note, 404's don't even work correctly at the current time: https://nodejs.org/ko/404/ - yet we clearly have one for korean: https://github.com/nodejs/nodejs.org/blob/master/locale/ko/404.md

setup/www/resources/config/nodejs.org
```diff
@@ -45,6 +45,8 @@ server {
     }
 
     location / {
+        try_files $uri $uri/ @english_fallback;
```
Fishrock123: Maybe we could do `try_files $uri $uri/ @english_fallback @localized_404;`?

Where `@localized_404` is just `return 404;` without `rewrite`?

Fishrock123: Or even `try_files $uri $uri/ @english_fallback =404;`?  See the docs at http://nginx.org/en/docs/http/ngx_http_core_module.html#try_files

phillipj: Ahhh, that looks smooth. That probably means we could remove the `return 404` line in `@english_fallback` as well?

Fishrock123: Right, so what I'm trying to suggest is a way to get the [localized 404 page](https://github.com/nodejs/nodejs.org/blob/master/locale/ko/404.md)s. The rewrite in `@english_fallback` prevents that I think, but at the same time, it doesn't currently work either and I'm not so sure why.

Then again if the localized 404 doesn't exist, all bets are off but I don't think that is priority and 404 pages are super easy to make.

Fishrock123:
Ok I spent some time trying to get the localized 404 pages working today.

What I've come up with:

```nginx
location @english_fallback {
    if ($uri ~* ^/(it|ko)/) {
        set $lang $1;
    }
    rewrite ^/(it|ko)/(.*)$ /en/$2; # I've used a relative url here for testing, not sure if it should be absolute or not
    # No `return 404;` here, I think that makes our page actually return that status code ontop of serving..
}
```

Where we *currently* declare `error_page 404 /en/404.html;`:
```nginx
error_page @404;
# use localized 404 pages if possible
location @404 {
    try_files /$lang/404.html /en/404.html;
}
```

Fishrock123:
@phillipj Hopefully you have time to look at this, I'd definitely like to get it live!

jbergstroem:
I will follow up here shortly too.

phillipj:
@Fishrock123 awesome, thanks for pitching in here! Just pushed an update which contains your localized 404 hackery 👍

setup/www/resources/config/nodejs.org
```diff
@@ -219,11 +236,26 @@ server {
...
             add_header access-control-allow-origin *;
         }
     }
 
+    # instead of serving a 404 page when a page hasn't been translated
+    location @english_fallback {
+        if ($uri ~* ^/(it|ko)/) {
+            set $lang $1;
+        }
+        rewrite ^/(it|ko)/(.*)$ /en/$2;
```
phillipj: @jbergstroem do you think relative redirect paths here are okey, or should we use absolute URLs?

Fishrock123:
@nodejs/build could some people look at this please? This is really important for website translation and I'd like to get it live ASAP.

phillipj:
Rebased to fix merge conflict.

Also decided to not go for relative rewrites to do things as before, and not raising more questions than necessary.

Starefossen:
Looks good to me 👍

jbergstroem:
Sorry for the delay -- I just need to do some performance testing (if's are evil yada yada)

Fishrock123:
`if`s might be evil but I don't know another way to do it. :(

Fishrock123:
@jbergstroem Can we merge this and if it really turns out to be a problem, undo it and have a single, multi-language 404 page? This is badly blocking translation groups as I understand... I know you're busy. :(

(I don't really know how to benchmark this... if you just want me to run `ab` or `wrk` against it I can do that though!)

jbergstroem:
@Fishrock123 I usually replay traffic with logs. Didn't know this was badly blocking translation -- I'll try and get this done shortly then.

Fishrock123:
Without this, unlocalized pages simply 404, meaning either you translate everything, or don't really bother. :/

jbergstroem:
Performance wise this looks ok but I'm not sure it works as intended; I don't get served /en/about/ if I visit /it/about/ and it 404's.

Fishrock123:
Oops didn't see that. Looking...

Fishrock123:
@jbergstroem can you check if it works with `/it/about.html`?

My testing env (not this full ngnix) works with `.html` but not `about` or `about/`.. Maybe that's the problem? If not I can try using the node.js nginx with this patch.

Fishrock123:
See https://github.com/nodejs/nodejs.org/issues/764 for a bug report on this.

Will try to clone this config asap and try it...

phillipj:
I've tried setting up nodejs.org the ansible way wo/luck. Any pointers to a good ansible introduction which relates to this ansible setup would be awesome!

Fishrock123:
cc @rvagg

jbergstroem:
@phillipj have you read the instructions at https://github.com/nodejs/build/tree/master/setup/www ?

jbergstroem:
@phillipj you can always just copy the configs over to your own nginx setup - its relatively straightforward.

phillipj:
> @phillipj have you read the instructions at https://github.com/nodejs/build/tree/master/setup/www ?

@jbergstroem yupp, the biggest issue is my lack of prior ansible experience. I probably need a good for-dummies-guide.

I'd really like to get ansible setup rather than copying the config to my local nginx instance. Both because I've got several dev machines and I want to see what ansible brings compared to other similar projects.

jbergstroem:
@phillipj I'd suggest having a ubuntu14 machine available and override hosts `iojs-www` and `node-www` when you run the playbook then. You'll have to add those hosts to your ssh config.

rvagg:
I've been messing with setting up a test machine for you to play with this on, knowing that our ansible scripts are not entirely solid for the web server this was a good chance to try and address that.

@phillipj I've put your ssh keys from GitHub into root@107.170.115.122 and it's set up almost identically to the nodejs.org server and you can do what you like with it. Stick `107.170.115.122 nodejs.org` into your /etc/hosts to make it work like the real thing for you. /home/www/nodejs/ is where the web files are. Run `/home/nodejs/build-site.sh nodejs` if you want to rebuild the web site from the nodejs.org/node repo. `/etc/nginx/sites-available/nodejs.org` is where the nginx config for the site is that you can play with. This machine is not entirely devoid of production secrets so don't go liberally sharing access to it, in fact if you want to give someone else access you might want to ask here about that.

For the record: the playbook in this repo for www _should_ be in a better state now, it can be run with `ansible-playbook -i <path to inventory file with a [node-www] block and a host listed under it> ansible-playbook.yaml` but the tricky bit is that you'll need some production secrets available to do a proper deploy, faking them out as things fail is one way to go, some info about what's needed can be gleaned from resources/secrets/makesecrets.sh too.

rvagg:
@Fishrock123 I put your keys in there too if you're interested in shaving this yak

Fishrock123:
Noted, looking at it right now.

Fishrock123:
Hmm, Indeed it does not appear to be working.

Aside: why do we have 2 server configs for both blog.nodejs.org and nodejs.org? It's quite confusing.

rvagg:
@Fishrock123 legacy, you'll note that all of the blog redirects are contained within the blog.nodejs.org part of the config, which leave the main nodejs.org config with much fewer redirects in its flow.

I think that's what you're referring to. Or perhaps you're seeing `:80:` vs `:443`. Again, legacy, the ideal situation is to have a minimal `:80` but we've yet to fully make that cutover because we've been allowing http access to /dist/ for legacy reasons. We'll get there soon, I think in the next few months in fact.

rvagg:
Also, you're welcome to propose improvements to our nginx config, for both readability and actual efficiency. Just be aware of the legacy constraints, things need to continue to work as they do now. Mostly I'm pretty afraid to mess with nginx config too much because it's a bit like playing wac-a-mole.

https://github.com/nodejs/build/blob/master/setup/www/resources/config/nodejs.org

phillipj:
@rvagg thanks for adding my keys!

After some trial-n-error, it seems like using relative URLs in the fallback rewrites made a positive impact. I  just pushed a commit. Those changes has been made on `107.170.115.122` if some of you want to do a quick check, by adding it into `/etc/hosts` as mentioned in https://github.com/nodejs/build/pull/316#issuecomment-222710235.

phillipj:
Added `zh-cn` in addition to `it|ko` cause of recently added translations (e.g. https://github.com/nodejs/nodejs.org/pull/771 & https://github.com/nodejs/nodejs.org/pull/776)

setup/www/resources/config/nodejs.org
```diff
@@ -40,7 +40,11 @@ server {
     default_type text/plain;
     index index.html;
 
+    error_page 404 @localized_404;
+
     location / {
+        try_files $uri $uri/ @english_fallback;
```
rvagg: can you explain `$uri $uri/`?

rvagg: fwiw this directive is only processing `^/dist/?.*$` and `^/.*\.json$` so I don't think it needs to be localized because even the .json are only in the docs which are not translated (afaik). Also, the whole `http` section is going away _soon_.

rvagg:
sgtm but I'm way too gun shy with this nginx config to be comfortable signing off on it. Can others who are comfortable hacking nginx config have a look and try it out on the test host too? @jbergstroem?

setup/www/resources/config/nodejs.org
jbergstroem: `$uri $uri/` would test /en followed by /en; usually good if the filename is actually a folder but otherwise unnecessary.

jbergstroem:
I will follow up shortly to confirm that it works as intended. I wasn't a big fan of how variable scopes were used; but it was indeed nifty.

setup/www/resources/config/nodejs.org
Fishrock123: Hmmm, might have been from my investigations. Can't recall why/if it would be necessary.

Fishrock123:
At some point we're going to have to just try it.

setup/www/resources/config/nodejs.org
phillipj: @rvagg `try_files $uri $uri/` is really what you would expect from a standard static server, and probably the most used nginx `try_files` pattern.

1. Try matching the current request URI as a filename: `https://nodejs.org/en/index.html` will match  `/home/www/nodejs/en/index.html` directly, no need to try the second `try_files` option.
2. If a file wasn't matched, try the URI as a directory instead: `https://nodejs.org/en/about` doesn't match a filename directly, meaning nginx will try `/en/about/` where it finds `/home/www/nodejs/en/about/index.html`.

The third option provides as fallback if none of the above matched something in the filesystem. Really the essence of what we're trying to do with this PR; try to find an english version of the requested file, instead of serving a 404 directly.

Fishrock123:
... ping again ...

rvagg:
I'm OK with giving this a try, but only if @jbergstroem is +1 too. We should probably do it during a low traffic period, the weekend I guess, as long as we have people around to revert quickly if need be.

jbergstroem:
Not sure which timezone digitalocean runs by, but least amount of traffic occurs in ~13h from now. I can attempt to give it a test around then.

jbergstroem:
This has been running for a bit on nodejs.org now and seems to be working as intended.

Fishrock123:
Has been running for about 3 days now? I think this can be merged

jbergstroem:
I'll merge this!

phillipj:
@jbergstroem any progress on merging this? nodejs.org just got a new locale (es / spanish), so it would be nice to get this merged before opening a new PR adding spanish as well.

jbergstroem:
@phillipj I'll add es/spanish.

jbergstroem:
Merged (with the addition of `es`) in 9488a616ba9fb3627a55394e886a6ede529f6c55.

phillipj:
Awesome 👍

PR #323, opened 2016-02-04
Add ansible configuration for AIX machines

mhdawson:
Add initial ansible configuration for AIX machines and include instructions for manual bootstrap.


mhdawson:
@jbergstroem ready for your review

mhdawson:
@jbergstroem  ping

jbergstroem:
@mhdawson sorry, just busy with work. Others -- feel free to review as well; otherwise I'll do it as quick as possible.

setup/aix61/resources/S20jenkins
```diff
@@ -0,0 +1,26 @@
...
+# purpose: script that will start or stop jenkins
+##################################################
+
+case "$1" in
+start )
+        su - {{user}} -c 'ulimit -c 0; \
+              export LIBPATH=/opt/freeware/lib:$LIBPATH; \
+              export PATH=/usr/java71_64/jre/bin:$PATH; \
+              export HOME=/home/{{user}}; \
+              export NODE_COMMON_PIPE="$HOME/test.pipe"; \
```
jbergstroem: also set `NODE_TEST_DIR` (usually $HOME/tmp)

mhdawson:
@jbergstroem added setup of NODE_TEST_DIR

setup/aix61/README.md
```diff
@@ -0,0 +1,22 @@
+# io.js Build AIX61 Setup
+
+For setting up an AIX 61 box
+
+```text
+Host test-ibm-aix61-ppc64-1:18822
+  HostName cloud.siteox.com
+  User root
+  This machine is a temporary one until we get additional resources at osusol. It is some
```
jbergstroem: Could you move this text out of the code? The pattern seems to be using code for copy/paste into ssh config.

jbergstroem:
not sure how `/etc/security/login.cfg` looks, but could perhaps [lineinfile](http://docs.ansible.com/ansible/lineinfile_module.html) simplify this?

setup/aix61/README.md
mhdawson: Ok, updated, removed text and added port which is required for ssh config

mhdawson:
@jbergstroem in respect to lineinfile I search on bullfreeware and I don't see it ported there so I don't think its available for AIX.

The line in question looks like this:

 shells = /bin/bash,/bin/sh,/bin/bsh,/bin/csh,/bin/ksh,/bin/tsh,/bin/ksh93,/usr/bin/sh,/usr/bin/bsh,/usr/bin/csh,/usr/bin/ksh,/usr/bin/tsh,/usr/bin/ksh93,/usr/bin/rksh,/usr/bin/rksh93,/usr/sbin/uucp/uucico,/usr/sbin/sliplogin,/usr/sbin/snappd


were I need to add /bin/bash since it is not there by default.

mhdawson:
@jbergstroem wondering if you have more comments ?

jbergstroem:
Not sure what init system aix is running but is output from java being picked up? That'd be my only remaining feedback as of now.

mhdawson:
When you say "is output from java being picked up" can you clarify ?  Do you mean do we capture it and send it to a log etc ?  If so I'll look at what we do for other platforms.

jbergstroem:
@mhdawson on other platforms we usually pipe to a logfile; check other examples.

Btw, I just compiled ccache manually; if you could add /usr/lib/ccache preceding $PATH in your init script and possibly add the ccache compilation to that manual step document it'd be great.

jbergstroem:
Come to think of it; I recall the slave having a `-slaveLog` option. On init systems that doesn't redirect stdout/err we should probably implement that instead.

mhdawson:
I had looked at some other platforms and added the redirect to a log file like they had.  Hope that is good for now.  @jbergstroem if thats ok then I'm hoping this is ready to go in

jbergstroem:
Is there a global cert store in aix? It's a bit unfortunate that we can't validate certs in the playbook.

jbergstroem:
Other than that, LGTM.

mhdawson:
I'll look into the cert side, but as discussed unless somebody else objects I'll squahs land this  tomorrow or Friday.

mhdawson:
Landed as f5a5dd95887ab0827a3adf03c9462e80fe619cc0

PR #332, opened 2016-02-16
Add nodejs.org metrics generation and hosting

rvagg:
Generates all of the data and assets located at https://nodejs.org/metrics/, opened in consultation with the Node.js security group


rvagg:
bleh! I already opened this PR @ https://github.com/nodejs/build/pull/298, how about I go and resolve that one?

PR #349, opened 2016-03-07
www: remove cloudfuse

jbergstroem:
I recall @rvagg mentioning this being a failed experiment. Might as well remove.


rvagg:
lgtm

PR #369, opened 2016-03-28
build: add step for enabling aha fs

mhdawson:
aha fs is needed on AIX  to pass fs watch tests both in libuv and node.js 

This adds the instructions on how to enable for AIX in order to support the work
being done under https://github.com/libuv/libuv/pull/776


jbergstroem:
LGTM

mhdawson:
Landed as https://github.com/nodejs/build/commit/3bca161439037b895da0f25fbdd8e99059313378

PR #370, opened 2016-03-29
build: relax firewall on rhel72-linuxone

mhdawson:
The initial firewall settings were way too
restrictive to be able to run our tests.  These
additional to the ansible playbook relax the
default rules.

both libuv and node tests pass with this change:

https://ci.nodejs.org/job/libuv+any-pr+multi-mdawson-test390/
https://ci.nodejs.org/job/node-test-commit-linuxone-mdawson/nodes=rhel72-s390x/15/console

### DIFF
--- a/setup/rhel72-linuxonecc/ansible-playbook.yaml
+++ b/setup/rhel72-linuxonecc/ansible-playbook.yaml
@@ -41,6 +41,18 @@
       with_items: ssh_users
       tags: user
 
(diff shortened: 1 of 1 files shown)


jbergstroem:
LGTM. A note for the future: once we require ansible 2.0 we should use the iptables module.

mhdawson:
Landed as f1af626a20ce24cac94cd96cb3b89c1d1cf8b658

PR #380, opened 2016-04-05
setup: remove github/pubkey mechanism

jbergstroem:
The majority of our machines now use shared keys that are accessed through a encrypted github repo. Additionally, the list of users were out of date.

There's a catch 22 when it comes to adding pubkeys to newly deployed servers. Optimally, this PR should replace what is removed with adding a test* key, but since this is in most of the cases already added through the VM provider (most often you choose what key to deploy an image with), I'd like to see if there's a better solution.

### DIFF
--- a/setup/aix61/ansible-playbook.yaml
+++ b/setup/aix61/ansible-playbook.yaml
@@ -16,22 +16,6 @@
       user: name="{{ server_user }}" shell=/bin/bash
       tags: user
 
-    - name: User | Download pubkey(s)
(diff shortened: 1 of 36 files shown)


mhdawson:
LGTM

jbergstroem:
@rvagg, @joaocgreis: since its only us three that mainly will be affected by this I'd appreciate if you weighed in any opinions.

joaocgreis:
LGTM, thanks for the direct mention. I've only created linux machines in Azure (the CC server), there the key is added when creating the machine so I'm ok with just removing this.

rvagg:
fine by me

PR #382, opened 2016-04-07
build:  changes to support ubuntu 14 for PPC BE

mhdawson:
ubuntu PPC BE defaults to 32 bit compiles as opposed to
the rest of the distros (including PPC LE) which default to 64
bit.  This is the case even though the OS is 64 bit. The
net is that since we want to ship/support 64 bit binaries we
need to install the multi-lib package and only do that
for PPC BE.  Looking at the ansible vars there does not
seem to be one that can be used to identify between
BE versus LE so I first check that we are on PPC from the
vars which are available, and if we are PPC then use
the node name (which includes PPC64_be for be machines)
to figure out if it is BE or LE.

This change adds the required changes to support BE
ubuntu machines as well as the definition of the
initial ubuntu BE machine

### DIFF
--- a/setup/ansible-inventory
+++ b/setup/ansible-inventory
@@ -43,6 +43,8 @@ test-digitalocean-ubuntu14-x86-1
 test-digitalocean-ubuntu14-x64-1
 release-osuosl-ubuntu14-ppc64_le-1
 test-osuosl-ubuntu14-ppc64_be-1
+test-osuosl-ubuntu14-ppc64_be-2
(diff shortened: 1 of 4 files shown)


mhdawson:
@jbergstroem I added the 2nd test and release machine so I think this is ready to go

jbergstroem:
LGTM

mhdawson:
Landed as b8a9401b5d5e384b21f09d4a6da71cfb1df69946

Issue #381, opened 2016-04-07
AIX machine timing out doing git checkout


mhdawson:
Latest AIX runs seem to be timing out during git checkout.

Investigating

mhdawson:
Seems to be consistent, also don't anything that looks like it changed in terms of config or timeouts although I do see that the timeout configured for git only seems to apply to some of the operations but that is not new.

mhdawson:
Will try a reboot to see if that helps

mhdawson:
Still timed out even after reboot.  Will contact David (at IBM) who provided this temporary machine to see if he can ask the provider if there are issues.  

@jbergstroem in the mean time do you think it would be ok for me to bump the global git timeout to see if that helps as the local one does not seem to apply to the git step that is timing out.

jbergstroem:
@mhdawson you mean on the machine? Feel free (you're referring to `git config --global`, no?) imo.

mhdawson:
I was not thinking about git config --gobal but instead the jenkins level config for timeouts but I'll see if the --global works

mhdawson:
Ready more I was pretty sure it was jenkins/plug-in timing out.  Tried adding this when starting the slave:

<PRE>
-Dorg.jenkinsci.plugins.gitclient.Git.timeOut=30
</PRE>

and also upped the memory for the jvm (hoping this speeds things up)

The combination seems to have resolved the issue.  Will validate and then update the ansible files

jbergstroem:
Did you test them separately? It'd be nice to avoid having to bump ram since we likely want to use the same over all machines (so we can put it at a shared variable file at some stage).

mhdawson:
I'll try to do that next week

mhdawson:
Build with heap size changed back to 128m https://ci.nodejs.org/job/node-test-commit-aix/120/

mhdawson:
The build with the smaller heap size was ok, so removed that change.  See PR for rest of discussion.

mhdawson:
Have extended the timeout under https://github.com/nodejs/build/pull/383 closing

PR #383, opened 2016-04-08
build: extend timeout for git for AIX

mhdawson:
We were seeing timeouts in git checkouts. Extended
the timeout through a parameter passed to the plugin
as a command line option when starting the jenkins
agent

Increased the memory available to the JVM in hopes
that might speed up actions like the checkout
which are being handled by a java plugin

Addresses https://github.com/nodejs/build/issues/381

### DIFF
--- a/setup/aix61/resources/S20jenkins
+++ b/setup/aix61/resources/S20jenkins
@@ -17,7 +17,7 @@ start )
               export NODE_COMMON_PIPE="$HOME/test.pipe"; \
               export OSTYPE=AIX61; \
               export GIT_SSL_CAINFO="$HOME/ca-bundle.crt"; \
-        java -Xmx256m -Dorg.jenkinsci.plugins.gitclient.Git.timeOut=30 -jar "$HOME/slave.jar" \
+        java -Xmx128m -Dorg.jenkinsci.plugins.gitclient.Git.timeOut=30 -jar "$HOME/slave.jar" \
           -secret {{secret}} \
           -jnlpUrl https://ci.nodejs.org/computer/{{id}}/slave-agent.jnlp >$jenkins_log_file 2>&1 &'
         ;;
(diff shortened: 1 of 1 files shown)


rvagg:
perhaps you should just include a bare git mirror like we do for arm machines, located in /home/iojs/git/io.js.reference, when a clone is done, any refs found in there will be linked straight to it and any not found in there are pulled from the remote. I occasionally update the mirror as upstream gets more bloated. The reference location is specified in the jenkins job setup under advanced git options, or something like that.

jbergstroem:
I left my comment in the originating issue. I'd prefer the following triage:
1. Is `-mx256m` required to solve issues?
2. Can we test setting the timeout locally or even globally [still machine only] to confirm that's the actual issue?
3. What @rvagg said.

mhdawson:
Doing 1) now https://ci.nodejs.org/job/node-test-commit-aix/120/ build without 12m

mhdawson:
@rvagg which jobs uses the bare mirror ? I looked at https://ci.nodejs.org/job/node-test-commit-arm/ and I did find references to io.js.reference but maybe I just did not look hard enough.

mhdawson:
Ok was fine without change to heap size so removed that.  Change pushed to pr.

mhdawson:
Ok found this:https://randyfay.com/content/reference-cache-repositories-speed-clones-git-clone-reference

So looks like I create the mirror using

git clone --mirror  https://github.com/nodejs/node.git   /home/iojs/git/io.js.reference

And I also found that "Path of the reference repo to use during clone"  is already set

rvagg:
bingo, and occasionally you can `git fetch` inside it to update it, but I suspect that won't be needed for a while in this case

mhdawson:
I added the mirror but the total execution did not seem to be any faster.  I'm thinking the problem may not be the remote traffic but slowness on the disk I/O.  At this point I think extending the timeout still makes sense.

mhdawson:
@jbergstroem you ok with landing this now or was there something else I should try out first ?

jbergstroem:
@mhdawson sure, LGTM

mhdawson:
landed as c7e244bcc73b2dc373bab973ebd725255e487f4f

PR #389, opened 2016-04-15
www: simplify ssl and redirects

jbergstroem:
Minor refactor of the www config surrounding redirects:
 - avoid `if {}` when possible
 - share ssl defaults
 - avoid extra http://www.foo.com -> http://foo.com -> https://foo.com when possible

Note: this is not tested; don't have access to my testing environment at the moment -- but I just wanted more eyes in case there are additional improvements to be made.

### DIFF
--- a/setup/www/resources/config/ssl-defaults.conf
+++ b/setup/www/resources/config/ssl-defaults.conf
@@ -1,0 +1,17 @@
+ssl_certificate ssl/nodejs_chained.crt;
+ssl_certificate_key ssl/nodejs.key;
+ssl_trusted_certificate ssl/nodejs_chained.crt;
+ssl_dhparam ssl/dhparam.pem;
+
+ssl_ciphers ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS;
+ssl_prefer_server_ciphers on;
+ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
+
+ssl_session_cache shared:nodejs:376m;
+ssl_session_timeout 24h;
+
+ssl_stapling on;
+ssl_stapling_verify on;
+
+spdy_keepalive_timeout 300;
+spdy_headers_comp 9;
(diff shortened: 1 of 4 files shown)


rvagg:
ok I guess, stretching my nginx fu a little too much so other reviews would be appreciated here

setup/www/resources/config/ssl-defaults.conf
```diff
@@ -0,0 +1,17 @@
...
+ssl_certificate_key ssl/nodejs.key;
+ssl_trusted_certificate ssl/nodejs_chained.crt;
+ssl_dhparam ssl/dhparam.pem;
+
+ssl_ciphers ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS;
+ssl_prefer_server_ciphers on;
+ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
+
+ssl_session_cache shared:SSL:10m;
+ssl_session_timeout 24h;
```
Starefossen: Should we ad  `ssl_session_tickets off;` ([source](https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=nginx-1.9.5&openssl=1.0.1e&hsts=yes&profile=intermediate))

Starefossen:
LGTM

setup/www/resources/config/ssl-defaults.conf
jbergstroem: We're still on 1.4.x (ubuntu 14 stable something)

fhemberger:
@jbergstroem FYI: This change needs to land as well, once v6 is out: https://github.com/nodejs/build/pull/394

jbergstroem:
Just a comment: Yes, we're raising the shared cache to 376M (~1.5m unique ip's) but since we have a cache timeout of 24h at least 200M (~800k) seems to be too low. We could always lower the timeout but we've got nothing better to spend memory on anyway.

jbergstroem:
Ok, some post-review commits here I've found while testing it out locally. I've begun to deploy this in production and it looks good; load is going down (just slightly though).

jbergstroem:
FYI; we just plowed through 376M @24h. I'm testing 512M @12h at the moment.

PR #391, opened 2016-04-16
freebsd: refactor init scripts

jbergstroem:
Improve the daemon monitoring facility by avoiding `su` as part of starting the script.
Also:
- modernise playbooks
- update readme


rvagg:
lgtm, I'm liking the .j2 templates for these.

Also, I've been putting a download of slave.jar in my recent startup scripts (currently all the ARM servers do this but I haven't finished the changes so they are not in this repo yet), that way we can keep it updated. Not important now but something to keep in mind for future iterations.

jbergstroem:
@rvagg in the "duke nukem" refactor I have locally I made a task that updates all slave.jar's.

PR #392, opened 2016-04-18
www: redirect new.nodejs.org => nodejs.org

fhemberger:
See https://github.com/nodejs/nodejs.org/issues/571


jbergstroem:
I have a PR that aims to simplify redirects and I couldn't find any traces of `new.nodejs.org`. Seems I was wrong after all. I'll update the PR https://github.com/nodejs/build/pull/389.

PR #396, opened 2016-04-22
setup: install lldb-3.6/lldb-3.6-dev on ubuntu

indutny:
Required for [llnode](https://github.com/indutny/llnode) smoke testing.

cc @nodejs/build

### DIFF
--- a/setup/ubuntu12.04/ansible-vars.yaml
+++ b/setup/ubuntu12.04/ansible-vars.yaml
@@ -8,6 +8,8 @@ packages:
   - python-software-properties
   - automake
   - libtool
+  - lldb-3.6
+  - lldb-3.6-dev
   - curl
   - gcc-4.8
   - g++-4.8
--- a/setup/ubuntu14.04/ansible-vars.yaml
+++ b/setup/ubuntu14.04/ansible-vars.yaml
@@ -7,6 +7,8 @@ packages:
   - python-all
   - automake
   - libtool
+  - lldb-3.6
+  - lldb-3.6-dev
   - curl
   - ntp
   - ccache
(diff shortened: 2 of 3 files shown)


indutny:
NOTE: I'm not sure if the lldb-3.6 is available on all of these platforms. Please re-check!

jbergstroem:
Does it specifically require 3.6 or is it the lowest req?

indutny:
I don't think it is the lowest required version, but I know for sure that it works starting from it.

indutny:
Additional note: while not absolutely necessary, it will make things easier if the version of lldb will be the same across the machines. I'll need to supply path to `lldb` when running `llnode`'s build script, and this path depends on the version.

TheAlphaNerd:
@indutny why is only ubuntu included? Smoke testing also runs on fedora, osx, and ppc

indutny:
@TheAlphaNerd I'm sure that this thing works on OS X and Ubuntu, and OS X doesn't need any additional packages. I may take a look at fedor, but don't have access to ppc.

jbergstroem:
@indutny that's kind of why I asked. Can only assume that mileage will vary between the different systems. I know FreeBSD only have 3.7 or 3.8.

indutny:
So anything actionable on this?

jbergstroem:
@indutny yeah; likely finding out all packages on the platforms we support and adding them to the playbooks. I can look at this, likely end of week. Is something held up by this at the moment?

maclover7:
ping @indutny -- is this still needed? If yes, this will need to be updated to change `ansible`, not `setup`

BridgeAR:
This should likely be closed. There was no progress for more than 1.5 years now.

mmarchini:
Closing since this has been inactive for quite some time now and https://github.com/nodejs/build/pull/1025 downloads `lldb` before testing llnode. Feel free to reopen if you believe this should still land though.

PR #403, opened 2016-05-01
www: add a open file cache

jbergstroem:
Avoid re-reading file handles when serving downloads (well, most things). This lessens the strain on nginx.


rvagg:
LGTM but why turn errors off? Wouldn't caching be just as good for those?

jbergstroem:
I don't see a point in getting a lot of "file not found" from the cache backend.

PR #405, opened 2016-05-03
www: Store gzip representations of common files on deploy

fhemberger:
Refs:
- https://github.com/nodejs/nodejs.org/issues/712
- https://github.com/nodejs/nodejs.org/pull/713

### DIFF
--- a/setup/www/resources/scripts/build-site.sh
+++ b/setup/www/resources/scripts/build-site.sh
@@ -25,7 +25,7 @@ if [ ! -d "${clonedir}" ]; then
 fi
 
 if [ "$site" == "nodejs" ]; then
-  build_cmd="npm run build"
+  build_cmd="npm run deploy"
   rsync_from="build/"
 else
   build_cmd="node_modules/.bin/gulp build"


jbergstroem:
LGTM

fhemberger:
@nodejs/build So can we merge this?

jbergstroem:
Happy to merge, just wanted to give it some time seeing how I was the only one ack:ing.

rvagg:
it probably should be `&&` not `;`, perhaps it should just be one npm script so you control it all in node/nodejs.org? `npm run deploy`

fhemberger:
Updated to `npm run deploy`.

rvagg:
lgtm

rvagg:
```
> find -E build -type f -regex '.*(html|css|js|xml|json)$' -exec gzip -kf9 {} \;

find: unknown predicate `-E'
```

tested on osx and not linux? reverting on the server until you give the :+1: here that it's redy

jbergstroem:
Urgh. Sloppy reviewing on my end. I'll revisit too.

jbergstroem:
Shouldn't that really just be `find build/ -type f ...`?

fhemberger:
Sorry for that!

Replaced the command with `find build -type f -name '*.html' -o -name '*.css' -o -name '*.js' -o -name '*.xml' -o -name '*.json' -exec gzip -kf9 {} \;` – a bit longer, but no incompatible regex switches. Also tested this on Linux as well.

https://github.com/nodejs/nodejs.org/commit/ff6c50f8841032d072b567a0dc3fc8e38fac47b4

rvagg:
@fhemberger OK, I've switched it over to `deploy` now but I don't think it's quite working right yet, my guess is that it's only `-name '*.json'` that's getting picke up cause I see site.json.gz but the others are plain.

rvagg:
Indeed ...

```
find . -type f -name '*.html' -o -name '*.css' -o -name '*.js' -o -name '*.xml' -o -name '*.json'
```

prints them all but


```
find . -type f -name '*.html' -o -name '*.css' -o -name '*.js' -o -name '*.xml' -o -name '*.json' -exec echo '{}' \;
```

prints only the .json files, so `-exec` must apply only to the last `-name`

this works tho:

```
find . -type f \( -name '*.html' -o -name '*.css' -o -name '*.js' -o -name '*.xml' -o -name '*.json' \) -exec echo '{}' \;
```

sort of like a grouped `||`

fhemberger:
@rvagg Okay, tested `find build -type f \( -name '*.html' -o -name '*.css' -o -name '*.js' -o -name '*.xml' -o -name '*.json' \) -exec gzip -kf9 {} \;` again, seems to work fine.

https://github.com/nodejs/nodejs.org/commit/3716408c02e5a59cd99730333ee82df644a0302c

rvagg:
neat, appears to be working: https://nodejs.org/en/index.html.gz

@jbergstroem can we confirm somehow that nginx is actually using these files? or do we just take it for granted?

jbergstroem:
@rvagg I've used it previously -- more info here: http://nginx.org/en/docs/http/ngx_http_gzip_static_module.html. I guess we could set up a test environment to prove it works?

rvagg:
Nah, let's not go overboard, it's just a shame we can't easily run a command and confirm that all this work actually achieved something.

jbergstroem:
We consume less cpu than before at least.

rvagg:
well that's certainly something, good work!

PR #411, opened 2016-05-09
qualify /blog url with ^

rvagg:
I happened to notice that (some of?) our blog content wasn't redirecting from http to https and I _believe_ this was the culprit. It's live now and the redirects appear to be working now. I can't find anything else to explain it, although this one doesn't really make sense to me either!


jbergstroem:
LGTM

PR #414, opened 2016-05-11
ubuntu16: fix environment declaration

jbergstroem:
ccache wasn't working because a newer version of systemd didn't accept multiline Environment declarations. Fix this by declaring it multiple times instead.


rvagg:
systemd ...
change lgtm, systemd doesn't lgtm

Issue #420, opened 2016-05-19
test-osuosl-ubuntu14-ppc64_be-2 - not responding (Resolved ?)


mhdawson:
So that we have a record in case this occurs with other machines as well

test-osuosl-ubuntu14-ppc64_be-2 was not responding. Turns out machine was ok but just not responding to its public ip, restart had no effect.

Raised ticket ([support.osuosl.org #28499] ) with osuosl and answer was that:

<PRE>
It looks like that floating IP is no longer attached/allocated to that VM or
the project; I'm currently looking into why this happened. I did notice that
140.211.168.4 is not within the range of IPs that are normally allocated to
OpenPOWER machines.
</PRE>

I allocated a new IP to resolve the issue, and @joaocgreis added to the firewall so its  back and working again.  

Did response to osuosl asking if they can figure out why  as 2 of our other machines don't shown the association to the public ip in the console anymore and I'm a bit worried we might see the same problem with those machines in the future. 

Will leave this open at least until we hear back from them

mhdawson:
Well I could not ssh in either from the public or private ip so I tried to soft reboot which failed, then shutdown which worked but now it will not start up.  Sent osuosl an update on the ticket.

mhdawson:
Been an ongoing saga. Has been brought online and then failed again a few times.

Currently down.

mhdawson:
Back up, latest update sent to osuosl:

Following up further based on a suggestion from Johan I checked the key that was being reported during the outage and I believe it was :  ECDSA key fingerprint is e5:c1:e0:a0:a9:ba:e1:38:f1:96:f9:35:e9:b5:4c:c5 
which does not look like it matches the one reported when the machine is working properly:  RSA key fingerprint is d9:62:25:3e:2f:32:15:6d:74:c2:02:93:7a:19:53:68.

mhdawson:
After lots of discussion back and forth the machine has been moved to a different node and it looks ok after having run a number of jobs through it.  Closing

PR #421, opened 2016-05-19
build: update ip - test-osuosl-ubuntu14-ppc64_be-2

mhdawson:
As per issue https://github.com/nodejs/build/issues/420
updated ip for test-osuosl-ubuntu14-ppc64_be-2


jbergstroem:
LGTM. Perhaps keep this open and add the other floating ip's which is likely to cause issues shortly?

mhdawson:
@jbergstroem  believe we have not lost/fixed all that will happen.

Updated PR to include all that needed updates as well as renaming of machines for le to use the standard naming convention.  

Should be good to land now.

jbergstroem:
Lets get it in!

mhdawson:
Landed as 96ba7214cb996a9396b06c1e6eae5484551f2d3f

PR #424, opened 2016-05-28
add basic osx instructions

rvagg:
I added ssh tunnels to all of the OSX build machines, connected to ci.nodejs.org, authenticated with _test and _release keys so y'all can get in and mess around. The slave process is started via `./start.sh` on a local terminal but we should probably change that to something that can be more easily managed remotely. Not sure what yet tho, open to suggestions.

### DIFF
--- a/setup/osx/README.md
+++ b/setup/osx/README.md
@@ -1,7 +1,5 @@
 # Node.js Build OS X Setup
 
-_Documentation pending ..._
-
 Add this to your ssh config:
 
 ```text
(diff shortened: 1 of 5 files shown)


rvagg:
added more instructions to this and scripts and launchd configs with instructions on how to manage that, should be a bit more resilient and easier to manage via ssh.

TheAlphaNerd:
LGTM

perhaps there could be a bit of copy included at the beginning to describe what these instructions are for, dropping right into the ssh config without much context makes it a bit harder to grok

jbergstroem:
I'd suggest making start.sh a template and use variables to replace things like secrets or hostname. Also, perhaps replace the static `JOBS=` with either `$(getconf _NPROCESSORS_ONLN)` or `{{ ansible_processor_vcpus }}`.

rvagg:
@jbergstroem none of this is ansibleised, I don't even know if this is even possible on osx. `getconf _NPROCESSORS_ONLN` is not really appropriate for the machines we're using atm, the VMs are allocated 6 cores each but we're limiting builds to using `-j2` because we're only working with 2 bare machines and usually in parallel. I don't really want to reduce core allocation to 2 to make this work. Perhaps we can revisit this when we get new machines, it's all very hacky.

jbergstroem:
Thanks for elaborating; lets add it to our todos for next setup. LGTM then.

PR #425, opened 2016-05-29
ubuntu12: add new slave

jbergstroem:
We were runnning short of ubuntu12 slaves which made the build slow down, especially during smoketests.


TheAlphaNerd:
LGTM!

mhdawson:
LGTM

PR #429, opened 2016-05-31
add support for Odroid C2 (ARMv7) with Ubuntu 16.04

rvagg:
A few fixups and some compatibility things for 16.04 to support the new C2's that are now in CI ("ubuntu1604-armv8"). Note that I'm using `server_jobs: 2` for these, they have 8 cores but not a ton of RAM and we've hit OOMs compiling with `-j4` already.

R=@jbergstroem

### DIFF
--- a/setup/ubuntu16.04/resources/jenkins.service.j2
+++ b/setup/ubuntu16.04/resources/jenkins.service.j2
@@ -11,7 +11,7 @@ Type=simple
 User={{ server_user }}
 
 Environment="USER={{ server_user }}" 
-Environment="JOBS={{ ansible_processor_vcpus }}"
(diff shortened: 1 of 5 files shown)


setup/ubuntu16.04/resources/jenkins.service.j2
```diff
@@ -11,7 +11,7 @@ Type=simple
 User={{ server_user }}
 
 Environment="USER={{ server_user }}" 
-Environment="JOBS={{ ansible_processor_vcpus }}"
+Environment="JOBS={{ server_jobs | default(ansible_processor_vcpus) }}"
```
jbergstroem: Perhaps mention `server_jobs` in readme.md?

Issue #301, opened 2016-01-12
Install docker on subset of test machines


mhdawson:
As discussed in latest build WG meeting, in order to be able to run tests on certains OS's (like Alpine - see https://github.com/nodejs/build/issues/75) we need docker to be installed on a subset of the test machines.

We should likely start by adding it as a dependency in Ansible for our ubuntu 14 machines which are fairly recent and which run docker well.

jbergstroem:
quick question: how do we then automate installing alpine/similar on top of docker? is it part of the same build script?

Vanuan:
@jbergstroem distribution installation is always a part of Dockerfiles. No ansible scripts needed. Here's a pull request that passes locally, but fails due to resource constraints:
https://github.com/nodejs/docker-node/pull/156

Vanuan:
The problematic part is how to publish binaries from docker to dist/ archives.

Vanuan:
So there are 4 dockerfiles needed:
1. Dockerfile that include build tools to build distributed binaries for each version (release delivery  to https://nodejs.org/dist/).
2. Dockerfile that include build tools to build distributed binaries for latest version (continuous delivery to  https://nodejs.org/dist/).
3. Dockerfile that include build tools and test tools to run tests (continuous integration)
4. Dockerfile that include binaries from https://nodejs.org/dist/ and deploy them to Docker Hub

Vanuan:
Number 4 in that list is this: https://github.com/nodejs/docker-node/pull/156
But it should be triggered either regularly by the publish hook to https://nodejs.org/dist/

Number 3 is what this issue is for.

Number 1 & 2 should be handled in https://github.com/nodejs/build/issues/75

Vanuan:
Number 3 is presumably being fixed by #437

mhdawson:
I think this issue can be closed as the motivation was to get alpine runs into the ci and it looks like they are there.

Issue #75, opened 2015-04-14
Alpine Linux / Docker Build


tracker1:
Would really be appreciated if some help could be rendered to getting an alpine image for iojs passing the test suite.
- https://github.com/iojs/docker-iojs/issues/44
- https://github.com/iojs/docker-iojs/issues/44#issuecomment-89814475

If there could be two build images, one for just node, and another with everything needed for node_gyp and nan to run.

Also, would be nice to see tests for building common binary modules (sqlite3, expat and a few others).

jbergstroem:
@tracker1 I think one challenge will be fully supporting musl as a libc replacement. I'm not against it but have very little knowledge about it. It doesn't look like its in too bad a shape. I'm keen on setting up a system based on musl, so I might give this a go down the road, but don't count on me for the moment.

tracker1:
@jbergstroem understandable... :-)  Really wish I had more time (and a more intimate working relationship with C/C++).

springmeyer:
> Also, would be nice to see tests for building common binary modules (sqlite3, expat and a few others).

As the maintainer of node-sqlite3, :+1: 

> I'm keen on setting up a system based on musl, so I might give this a go down the road, but don't count on me for the moment.

Same here. I'm also interested in musl.

mhart:
The failures in the tests (there are only two) are due to assumptions in the tests themselves. One being that the platform has glibc (which Alpine isn't), the other being that the `ps` command supports the `-p` flag (which Alpine doesn't).

You can read more here: https://github.com/iojs/docker-iojs/issues/44#issuecomment-95969892

Starefossen:
I am putting this on the agenda for #300

jbergstroem:
I don't think anyone objects to get this part of the ci, but work needs to be done. Lets get an outline going for whats required - feel free to add:
 - we lack a vm provider that can run alpine (unless we automate docker stuff)
 - we lack ansible scripts to set a host up

Starefossen:
Regarding version of Alpine linux it looks like the Docker Library guys are using `alpine:3.3` for some of their official images on Docker Hub.

jbergstroem:
@Starefossen how often is stable updated? How does that affect toolchain, etc? (thinking if they have a LTS mindset)

Starefossen:
I had a chat with one of the Alpine Linux maintainers today. This is what they told me:

* They have Node.js in main repository
* Releases are supported for 2 years since initial release with security backports or critical bugfixes
[...]

PR #437, opened 2016-06-14
Preliminary Alpine support for CI

rvagg:
This adds a config for Ubuntu 16.04 + Docker that builds and manages an Alpine Linux 3.3 container which has the Jenkins slave running inside it.

After some digging I discovered that the preferable way to run Node in Alpine is to build it in alpine rather than reusing binaries that we build for generic Linux. Initially I started down the track of emulating our nodejs.org binaries by building in a CentOS 5 container before transferring to the Alpine container for testing. I scrapped that and went for everything in Alpine, assuming that we're going to ship images with binaries created inside them specifically for Alpine.

I'm using 3.3 because I had trouble getting 3.4 to run. It doesn't seem to like something about the combination of the mount point and the `CMD` and gives a weird `sh` error (`su: can't execute ',,,:/home/': No such file or directory` if anyone wants to chase this up).

It's temporarily in CI, you can see an example run @ https://ci.nodejs.org/job/node-test-commit-linux/nodes=ubuntu1604_docker_alpine34-64/3796 (the host is misnamed btw, it's "alpine34" but should be "alpine33").

Failures that we have are:
- parallel/test-net-better-error-messages-port-hostname
- parallel/test-net-connect-immediate-finish
- parallel/test-os
- parallel/test-process-getgroups
- parallel/test-setproctitle

If we don't have a path to resolving these fairly quickly then we should remove it from CI so collaborators don't get upset at all their runs failing. ATM I don't think we have a way of marking these tests flaky _just_ for Alpine, it's done by operating system, not distro.

Where to next?

/cc @nodejs/docker @nodejs/testing @mhart

### DIFF
--- a/setup/ubuntu16.04_docker_alpine3.4/resources/Dockerfile.j2
+++ b/setup/ubuntu16.04_docker_alpine3.4/resources/Dockerfile.j2
@@ -1,0 +1,47 @@
+FROM alpine:3.4
+
+ENV LC_ALL C
+ENV USER {{ server_user }}
+ENV JOBS {{ server_jobs | default(ansible_processor_vcpus) }}
+ENV HOME /home/{{ server_user }}
+ENV NODE_TEST_DIR /home/{{ server_user }}/tmp
+ENV PATH /usr/lib/ccache/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+ENV NODE_COMMON_PIPE /home/{{ server_user }}/test.pipe
+ENV OSTYPE linux-gnu
+ENV DESTCPU x64
+ENV ARCH x64
+
+RUN apk add --no-cache \
+        libstdc++ \
+    && apk add --no-cache --virtual .build-deps \
+        binutils-gold \
+        curl \
+        g++ \
+        gcc \
+        gnupg \
+        libgcc \
+        linux-headers \
+        make \
+        paxctl \
+        python \
+        tar \
+        ccache \
+        openjdk8 \
+        git \
+        procps \
+        openssh-client
+
+RUN addgroup -g 1000 {{ server_user }}
+
+RUN adduser -G {{ server_user }} -D -u 1000 {{ server_user }}
+
+VOLUME [ "/home/{{ server_user }}/" ]
+
+USER iojs
+
+CMD cd /home/iojs \
+  && curl https://ci.nodejs.org/jnlpJars/slave.jar -O \
+  && java \
+      -jar slave.jar \
+      -jnlpUrl https://ci.nodejs.org/computer/test-digitalocean-ubuntu1604_docker_alpine34-x64-1/slave-agent.jnlp \
+      -secret efbfadbf4807af0f1084a1e120ef5467b49432113661e11ce2ecc1ab71b96f38
--- a/setup/ubuntu16.04_docker_alpine3.4/resources/Dockerfile.j2
+++ b/setup/ubuntu16.04_docker_alpine3.4/resources/Dockerfile.j2
@@ -29,7 +29,8 @@ RUN apk add --no-cache \
         openjdk8 \
         git \
         procps \
-        openssh-client
+        openssh-client \
+        bash
 
 RUN addgroup -g 1000 {{ server_user }}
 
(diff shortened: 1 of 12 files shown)


mhart:
I'd keep striving to get Alpine 3.4 working tbh – it adds a bunch of very useful stuff for ppl in environments like Kubernetes and it'd be easier to deal with just one version of Alpine. Node.js definitely builds on it, I guess it's just a matter of finding out what's going on in this env.

I know very little about the nodejs build environment, so it'll take some digging to figure out how things are setup

mhart:
Instead of the `CMD su -c ...` gymnastics, I think you should be able to do:

```
USER {{ server_user }}
CMD cd /home/{{ server_user }} ...
```

setup/ubuntu16.04_docker_alpine3.4/resources/Dockerfile.j2
```diff
@@ -0,0 +1,45 @@
...
+        openjdk8 \
+        git
+
+RUN addgroup -g 1000 {{ server_user }}
+
+RUN adduser -G {{ server_user }} -D -u 1000 {{ server_user }}
+
+VOLUME [ "/home/{{ server_user }}/" ]
+
+CMD su {{ server_user }} -c '\
```
mhart: Can this be changed to:
```
USER {{ server_user }}
CMD cd /home/{{ server_user }} ...
```

mhart:
I explored the tests a while ago and found some reasons they were failing:

https://github.com/nodejs/docker-iojs/issues/44#issuecomment-95969892

cjihrig:
Two of the failures (parallel/test-net-better-error-messages-port-hostname and parallel/test-net-connect-immediate-finish) are `AssertionError: 'EAI_AGAIN' === 'ENOTFOUND'`, which are being discussed in https://github.com/nodejs/node/pull/5099 and https://github.com/nodejs/node/pull/5196.

mhart:
I'd also be interested to see if, instead of having to build inside of Alpine, you can just build with musl instead of glibc – technically I think that should work. (I believe this is what rust does, for example)

mhart:
(actually, that idea doesn't really address the testing-in-Alpine aspect, so it's probably of only limited value)

setup/ubuntu16.04_docker_alpine3.4/resources/Dockerfile.j2
rvagg: THANKYOU! I didn't know this was a Dockerfile thing, must be new I suppose. This has solved the 3.4 problem (plus I'll be changing a bunch of other Dockerfiles now I know this exists).

mhart: Sure mate, you tell yourself that 😉  https://github.com/docker/docker/blob/master/CHANGELOG.md#builder-20

rvagg:
Updated to Alpine 3.4 thanks to @mhart's pointer.

Here's a run on `master`: https://ci.nodejs.org/job/node-test-commit-linux/3803/nodes=ubuntu1604_docker_alpine34-64

Only 3 failures:

parallel/test-net-better-error-messages-port-hostname
parallel/test-net-connect-immediate-finish
parallel/test-setproctitle

mhart:
@rvagg the `setproctitle` one could possibly be fixed by adding `procps` to the list of `apk` packages to install – apparently that allows support for the `-p` flag. Whether or not that's a good enough solution, I dunno. I'm not sure why the test has to specifically rely on that flag being supported though – perhaps it could be changed to something else?

rvagg:
great, procps has done it, updated Dockerfile and we're down to just the `EAI_AGAIN' != 'ENOTFOUND'` errors https://ci.nodejs.org/job/node-test-commit-linux/3804/nodes=ubuntu1604_docker_alpine34-64

Vanuan:
It's not possible to build Alpine binaries in CentOS unless you use cross compiler. CentOS (and almost any other generic linux) uses `glibc`, Alpine uses `musl`.

Vanuan:
So this is a continuous integration part (point 3 in https://github.com/nodejs/build/issues/301#issuecomment-220574960), right?
Would it be trivial to port this to other versions of node?

jbergstroem:
I think we need bash as requirement

jbergstroem:
I guess its worth mentioning that we now have full alpine passes through both docker instances and joyent's virtualization.

jbergstroem:
We should probably update to Alpine 3.5. There's also a few new Ansible modules for managing (building/running) docker containers we can replace the shell commands with. I see both as optional before merging though.

SimenB:
>We should probably update to Alpine 3.5.

Alpine 3.6 is released, FWIW.

rvagg:
alpine is in the new ansible/ stuff so I'm assuming it's good there, I haven't used it myself though

SimenB:
Is there a generic issue for tracking a build of node for alpine? Or should I just watch #989?

rvagg:
New version of this in #992, that's what's active in CI now with Alpine 3.4, 3.5 and 3.6 running in node-test-commit-linux. I've taken the old test-digitalocean-ubuntu1604_docker_alpine34-x64-1 host out of rotation, it was set up according to this PR and has been running our Alpine 3.4 tests till now.

mhart:
@SimenB this might be what you're looking for? https://github.com/nodejs/build/issues/75

PR #446, opened 2016-07-05
fedora: refactor and introduce fedora 24

jbergstroem:
A few things going on here:
- Remove configs for fedora 20 and 21.
- Update playbooks; most of them look exactly the same now
- Disable SELinux on F24 since they've changed strategy to enforcing
  and I don't know how to let java execute through systemd

Open to suggestions on how to adapt a rule to SElinux -- I ran audit2allow but ran into the next obstacle


setup/fedora23/host_vars/.gitignore
```diff
@@ -0,0 +1 @@
+iojs-*
```
rvagg: test-*

or just remove it, I think it's covered in /.gitignore

jbergstroem: yeah, this is probably a relic.

rvagg:
My primary experience with selinux has been in disabling it so I don't think I can help beyond that! I'm fine with disabling it here but I guess it's something we'll want in our test environment at some point if it's coming out enabled like this on systems.

Since the playbooks are looking the same, perhaps they could be collapsed into a single "fedora" playbook with some rules to differentiate between them where needed? I did this in the Raspberry Pi one to support both wheezy and jessie.

rvagg:
this lgtm for me though, merge as is if you want, the above are just suggestions

jbergstroem:
Just so I don't forget: fedora23 will bail since it has python but not python-libselinux. I reckon we should start splitting up in proper ansible roles here and then parse `inventory_hostname` to figure out all the cases (if os is fedora23..)

rvagg:
I think `ansible_distribution_release` will probably do what we need

jbergstroem:
@rvagg: problem being we don't have facts at that stage. I'm writing a filter plugin that parses our naming so we can extract os and use that. Will probably be applicable at a lot of places!

jbergstroem:
complete support can be found in my branch: https://github.com/jbergstroem/build/tree/feature/refactor-the-world/ansible

jbergstroem:
Closing in favor of the refactor.

PR #461, opened 2016-08-04
doc: fix probs found in install of new AIX machine

mhdawson:
- added instructions to increase size of /opt
- replace install with cp/chmod for ccache install as it
  looks like install does not run the same on the new
  AIX machine
- fix ln commands for ccache install which were incorrect
- add instructions to create /aha before mounting.
- Add numbering
- Add step to create git mirror

### DIFF
--- a/setup/aix61/ansible-playbook.yaml
+++ b/setup/aix61/ansible-playbook.yaml
@@ -21,7 +21,7 @@
       tags: jenkins
 
     - name: Jenkins | Copy S20jenkins
-      copy: src=./resources/S20jenkins dest=/etc/rc.d/rc2.d mode=755
+      copy: src=./resources/S20jenkins dest=/etc/rc.d/rc2.d mode=0755
       tags: jenkins
 
     - name: Jenkins | Copy certs required to access jenkins/git
--- a/setup/aix61/ansible-playbook.yaml
+++ b/setup/aix61/ansible-playbook.yaml
@@ -21,7 +21,7 @@
       tags: jenkins
 
     - name: Jenkins | Copy S20jenkins
-      copy: src=./resources/S20jenkins dest=/etc/rc.d/rc2.d
+      copy: src=./resources/S20jenkins dest=/etc/rc.d/rc2.d mode=755
       tags: jenkins
 
     - name: Jenkins | Copy certs required to access jenkins/git
--- a/setup/aix61/manualBootstrap.txt
+++ b/setup/aix61/manualBootstrap.txt
@@ -11,8 +11,8 @@ chfs -a size=+1300000 /opt
 
 ftp://ftp.software.ibm.com/aix/freeSoftware/aixtoolbox/RPMS/ppc/wget/wget-1.9.1-3.aix6.1.ppc.rpm
 rpm -i wget-1.9.1-3.aix6.1.ppc.rpm
-ftp://ftp.software.ibm.com/aix/freeSoftware/aixtoolbox/RPMS/ppc/unzip/unzip-5.51-2.aix6.1.ppc.rpm
-rpm -i unzip-5.51-2.aix6.1.ppc.rpm
+ftp://ftp.software.ibm.com/aix/freeSoftware/aixtoolbox/RPMS/ppc/unzip/unzip-6.0-3.aix6.1.ppc.rpm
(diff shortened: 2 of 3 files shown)


setup/aix61/manualBootstrap.txt
```diff
@@ -28,24 +39,28 @@ http://www.bullfreeware.com/affichage.php?id=2048
...
 sh install-libtool-rpm.sh
 
+4) install ccache
 # install ccache based on these instructions ./setup/ansible-tasks/ccache.yaml
 # we should update that script so that it can be used on AIX as well
 # also currently requires the fix outlined here: https://bugzilla.samba.org/show_bug.cgi?id=11519
 # to be applied manually before the compile
+# install on the new linux machines did not seem to work so use cp and chmod instead
 cd /tmp
 wget http://samba.org/ftp/ccache/ccache-3.2.4.tar.gz
```
jbergstroem: perhaps bump to 3.2.7 while at it

setup/aix61/manualBootstrap.txt
```diff
@@ -28,24 +39,28 @@ http://www.bullfreeware.com/affichage.php?id=2048
...
-install -c -m 755 ccache /usr/lib
-ln -s /usr/lib/ccache/cc /usr/lib/ccache
-ln -s /usr/lib/ccache/g++ /usr/lib/ccache
-ln -s /usr/lib/ccache/gcc /usr/lib/ccache
+cp ccache /usr/lib
+chmod 755 /usr/lib
+
+ln -s /usr/lib/ccache /usr/lib/cc
+ln -s /usr/lib/ccache /usr/lib/g++
+ln -s /usr/lib/ccache /usr/lib/gcc
```
jbergstroem: Not sure about base layout on AIX but having these in `/usr/lib` seems a bit off to me. Is `/usr/libexec` a known convention?

mhdawson:
I think you did the initial ccache build/install and that's were they went, I'll have to see what is most common for AIX.

mhdawson:
Talked to your AIX expert, sounds like the right place would be /opt/freeware/bin will move to there

mhdawson:
Fixed up ccache issues both in doc and updated ansible resource to use new ccache location.

setup/aix61/ansible-playbook.yaml
```diff
@@ -21,7 +21,7 @@
       tags: jenkins
 
     - name: Jenkins | Copy S20jenkins
-      copy: src=./resources/S20jenkins dest=/etc/rc.d/rc2.d
+      copy: src=./resources/S20jenkins dest=/etc/rc.d/rc2.d mode=755
```
jbergstroem: `mode=0755`

setup/aix61/manualBootstrap.txt
```diff
@@ -1,24 +1,60 @@
...
+rpm -hUv gettext-0.19.7-1.aix6.1.ppc.rpm
+
 
 Java7r1_64.jre.7.1.0.200.tar.gz
 gunzip Java7r1_64.jre.7.1.0.200.tar.gz
 tar -xvf Java7r1_64.jre.7.1.0.200.tar
 installp -agXYd . Java71_64.jre 2>&1 | tee installp.log
 
-wget http://www.oss4aix.org/download/everything/RPMS/make-3.82-1.aix5.3.ppc.rpm
+curl --insecure http://www.oss4aix.org/download/everything/RPMS/make-3.82-1.aix5.3.ppc.rpm -O
```
jbergstroem: do you need insecure on http urls? does it redirect?

mhdawson: Without the --insecure it complains about the cert and does not download.  There is probably a way to specify more certificates as well but not sure what that is without additional investigation.

mhdawson:
Agree with both general points that we should work to automate as much as possible with ansible, it is on the list just have not had time yet.  Also agree with updating to md, but I think doing that along with the ansible work makes sense.

mhdawson:
ok pushed fix for 755 -> 0755 I'll plan to land tomorrow unless I hear any objections.

mhdawson:
Landed as https://github.com/nodejs/build/commit/650fd384e17a1de970856b4bd9e69e060c4e1a74

PR #466, opened 2016-08-10
ubuntu16.04: add new slave

jbergstroem:
Also, minor nits to the playbook to fix remote_user (lets look at handling non-root users later) as well as enabling ntp.


rvagg:
lgtm

Ansible doesn't complain on use of "remote_user"? I recall having to switch to using the "become" stuff because of complaints. Or maybe it was "sudo" that was being deprecated. I can't keep up.

jbergstroem:
`sudo` is deprecated, replaced by `become`. Since we pass user through the config in the refactor I think we should just avoid writing extra logic. I guess the question is what hosts running ubuntu16 uses non-root ssh usernames?

PR #469, opened 2016-08-22
github-bot: ansible setup

phillipj:
Sooo, here's my first shot at creating an ansible setup for the bot. Has been run on the test host @jbergstroem created for this purpose, seems to be working nicely.

Primarly installs necessary binaries, checks out the application code from github.com, before starting the created systemd service which runs (and supervises) `npm start`.

As there's a big ansible refactor just around the corner, these changes doesn't need to be merged asap, but would be nice to get some feedback about obviously missing/wrong parts.

### DIFF
--- a/setup/github-bot/README.md
+++ b/setup/github-bot/README.md
@@ -1,0 +1,19 @@
+# Node.js build Debian setup for github-bot
+
+Add this to your ssh config:
+
+```text
+Host infra-rackspace-debian8-x64-1
+  HostName 23.253.100.79
+```
+
+Run playbook:
+
+```bash
+$ ansible-playbook -i ../ansible-inventory ansible-playbook.yaml
+```
+
+If you need to restart github-bot:
+```bash
+$ systemctl restart github-bot
+```
--- a/setup/github-bot/ansible-playbook.yaml
+++ b/setup/github-bot/ansible-playbook.yaml
@@ -1,0 +1,71 @@
+---
+- hosts: github-bot
+  remote_user: root
+  gather_facts: False
+
+  tasks:
+
+    - name: Bootstrap | Install baseline packages
+      raw: apt-get install -y aptitude python-apt apt-transport-https
+      tags: bootstrap
+
+- hosts: github-bot
+  remote_user: root
+
+  tasks:
+
+    - include_vars: ansible-vars.yaml
+      tags: vars
+
+    - name: General | Add the NodeSource GPG key
+      apt_key: url=https://deb.nodesource.com/gpgkey/nodesource.gpg.key state=present
+      tags: general
+
+    - name: General | Add the NodeSource Node.js repo
+      apt_repository: repo='deb https://deb.nodesource.com/node_6.x jessie main' state=present
+      tags: general
+
+    - name: General | APT Update and upgrade
+      apt: update_cache=yes upgrade=full
+      tags: general
+
+    - name: General | Install required packages
+      apt: name="{{ item }}" state=latest
+      with_items: "{{ packages }}"
+      tags: general
+
+    - name: General | Enable time syncing
+      service: name=systemd-timesyncd state=started enabled=yes
+      tags: general
+
+    - name: User | Add {{ server_user }} user
+      user: name="{{ server_user }}" shell=/bin/bash
+      tags: user
+
+    - name: Init | Generate and copy init script
+      template: src=./resources/github-bot.service dest=/lib/systemd/system/github-bot.service
+      tags: init
+
+    - name: Init | Create directory for EnvironmentFile
+      file: path="/home/{{ server_user }}/environment" state=directory mode=0755 owner="{{ server_user }}"
+      tags: init
+
+    - name: Init | Generate and copy systemd EnvironmentFile
+      template: src=./resources/environment-file dest="/home/{{ server_user }}/environment/github-bot"
+      tags: init
+
+    - name: Init | Clone github-bot repo
+      become: yes
+      become_user: "{{ server_user }}"
+      git: repo=https://github.com/nodejs/github-bot.git dest="/home/{{ server_user }}/github-bot"
+      tags: init
+
+    - name: Init | Install npm dependencies
+      become: yes
+      become_user: "{{ server_user }}"
+      npm: path="/home/{{ server_user }}/github-bot"
+      tags: init
+
+    - name: Init | Start github-bot
+      service: name=github-bot state=started enabled=yes
+      tags: init
--- a/setup/github-bot/resources/github-bot.service
+++ b/setup/github-bot/resources/github-bot.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=github-bot
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+ExecStart=/usr/bin/node server.js
+EnvironmentFile=/home/{{server_user}}/environment/github-bot
+WorkingDirectory=/home/{{server_user}}/github-bot
+Restart=always
+RestartSec=10
+User={{server_user}}
--- a/setup/ansible-inventory
+++ b/setup/ansible-inventory
@@ -149,3 +149,6 @@ test-ibm-aix61-ppc64-1:18822
 [iojs-build-rhel72-linuxonecc]
 test-linuxonecc-rhel72-s390x-1
 test-linuxonecc-rhel72-s390x-2
+
+[github-bot]
+infra-rackspace-debian8-x64-1
--- a/setup/github-bot/ansible-vars.yaml
+++ b/setup/github-bot/ansible-vars.yaml
@@ -1,0 +1,5 @@
+---
+server_user: iojs
+packages:
+  - git
+  - nodejs
--- a/setup/github-bot/resources/environment-file
+++ b/setup/github-bot/resources/environment-file
@@ -1,0 +1,6 @@
+NODE_ENV=production
+PORT={{envs.port}}
+TRAVIS_TOKEN={{envs.travis_token}}
+GITHUB_TOKEN={{envs.github_token}}
+GITHUB_WEBHOOK_SECRET={{envs.github_webhook_secret}}
+LOGIN_CREDENTIALS={{envs.login_credentials}}


setup/github-bot/resources/github-bot.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=github-bot
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+ExecStart=/usr/bin/npm start
```
Starefossen: I tend to avoid starting node services using `npm start` in production and using `node server.js` instead because npm heavily (and some times indeterministic) changes environment variables depending on what user it is run under. Not sure if this will be a problem for the bot though.

setup/github-bot/README.md
```diff
@@ -0,0 +1,25 @@
...
+To set up a host, run:
+
+```text
+$ ansible-playbook -i ../ansible-inventory ansible-playbook.yaml
+```
+
+Start-up of the github-bot service:
+
+```text
+$ systemctl start github-bot
```
jbergstroem: This will probably already be made by the sensible script; perhaps nitpick but I would suggest how to restart it instead.

setup/github-bot/README.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+**Users**: The ansible-vars.yaml file contains a list of users who's GitHub public keys are pulled and placed into authorized_keys for both root and iojs users. This file should be updates when new users are added to the build project who are able to help maintain the containerized builds.
```
jbergstroem: This is old. I suggest you use a newer `README.md` template such as this: https://github.com/jbergstroem/build/blob/65e5ac8e7dbf8a35dce3a78c0bba679e2d295a7f/setup/fedora/README.md

setup/github-bot/ansible-playbook.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+  remote_user: root
+
+  tasks:
+
+    - include_vars: ansible-vars.yaml
+      tags: vars
+
+    - name: General | Add the NodeSource Node.js repo
+      command: "bash -c 'curl -sL https://deb.nodesource.com/setup_6.x | bash -'"
```
jbergstroem: I'd use http://docs.ansible.com/ansible/apt_repository_module.html here.

setup/github-bot/ansible-playbook.yaml
```diff
@@ -0,0 +1,59 @@
...
+      file: path=/home/nodejs/environment state=directory mode=0755 owner=nodejs
+      tags: init
+
+    - name: Init | Generate and copy systemd EnvironmentFile
+      template: src=./resources/environment-file dest=/home/nodejs/environment/github-bot
+      tags: init
+
+    - name: Init | Clone github-bot repo
+      become: yes
+      become_user: "nodejs"
```
jbergstroem: I would use `{{ server_user }}` here and put it in vars. Although not entirely current, we use `iojs` almost everywhere -- I reckon consistency over correctness is the way here.

setup/github-bot/ansible-playbook.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - name: General | Install baseline packages
+      apt: name={{ item }} state=latest
+      with_items: "{{ baseline_packages }}"
+      tags: general
+
+    - name: General | Install required packages
+      apt: name={{ item }} update_cache=yes state=latest
+      with_items: "{{ packages }}"
+      tags: general
+
```
jbergstroem: No need to split up packages into categories.

setup/github-bot/resources/github-bot.service
```diff
@@ -0,0 +1,13 @@
...
+EnvironmentFile=/home/nodejs/environment/github-bot
+WorkingDirectory=/home/nodejs/github-bot
+Restart=always
```
jbergstroem: `{{ server_user }}`

setup/github-bot/resources/github-bot.service
```diff
@@ -0,0 +1,13 @@
...
+RestartSec=10
+User=nodejs
```
jbergstroem: Same as above

setup/github-bot/ansible-playbook.yaml
```diff
@@ -0,0 +1,59 @@
...
+      become_user: "nodejs"
+      git: repo=https://github.com/nodejs/github-bot.git dest=/home/nodejs/github-bot
+      tags: init
+
+    - name: Init | Install npm dependencies
+      command: "npm install"
+      args:
+        chdir: /home/nodejs/github-bot
+      tags: init
+
```
jbergstroem: Perhaps this should use `become_user` too?

setup/github-bot/ansible-playbook.yaml
```diff
@@ -0,0 +1,59 @@
...
+      apt: update_cache=yes upgrade=full
+      tags: general
+
+    - name: General | Install baseline packages
+      apt: name={{ item }} state=latest
+      with_items: "{{ baseline_packages }}"
+      tags: general
+
+    - name: General | Install required packages
+      apt: name={{ item }} update_cache=yes state=latest
```
jbergstroem: No need to update cache again.

setup/github-bot/ansible-playbook.yaml
phillipj: When the server was "empty", installing packages would fail w/o baseline packages installed. I don't have the displayed error available anymore ofc..

phillipj: On second thought, IIRC baseline packages was required *before* updating apt.

jbergstroem: oh, yeah -- you probably want that as part of bootstrapping then. See for instance [this](https://github.com/nodejs/build/pull/446/files#diff-54a20c9aff06229996d3d614d66b5425R4) where we run raw commands to avoid calling python/apt.

phillipj:
Thanks for excellent reviews, just what I needed!

jbergstroem:
You accidentally checked in `ansible-playbook.retry`

setup/github-bot/ansible-playbook.yaml
phillipj: @jbergstroem not sure I totally understand what you mean with running raw commands/avoid python. PTAL at https://github.com/nodejs/build/pull/469/commits/72f5ad9a710425c58b2970b6e43f01a6ca8b69bb seems to be what's needed on a local vagrant box I've got.. How would that compare to what you were thinking of?

setup/github-bot/ansible-playbook.yaml
```diff
@@ -0,0 +1,64 @@
...
+  tasks:
+
+    - include_vars: ansible-vars.yaml
+      tags: vars
+
+    - name: General | Install baseline packages
+      apt: name="{{ item }}" state=latest
+      with_items: "{{ baseline_packages }}"
+      tags: general
+
```
jbergstroem: I would bootstrap similarly to how we do it in other packages -- sometimes python isn't available and I reckon this would break if you don't have aptitude installed.

Check the fedora example I gave where we set `get_facts` to `False` and run `raw` commands to install the baseline stuff.

setup/github-bot/ansible-playbook.yaml
```diff
@@ -0,0 +1,64 @@
...
+      become_user: "{{server_user}}"
+      git: repo=https://github.com/nodejs/github-bot.git dest=/home/{{server_user}}/github-bot
+      tags: init
+
+    - name: Init | Install npm dependencies
+      become: yes
+      become_user: "{{server_user}}"
+      command: "npm install"
+      args:
+        chdir: /home/{{server_user}}/github-bot
```
jbergstroem: Quote because variable. In a few other places too!

setup/github-bot/ansible-playbook.yaml
```diff
@@ -0,0 +1,73 @@
+---
+- hosts: github-bot
+  remote_user: root
+  gather_facts: False
+
+  tasks:
+
+    - name: Bootstrap | Install baseline packages
+      raw: apt-get install -y aptitude python-apt apt-transport-https
```
phillipj: @jbergstroem something like this, or would you rather see some fancy `when`-checks like what you put in the [fedora setup](https://github.com/nodejs/build/pull/446/files#diff-54a20c9aff06229996d3d614d66b5425R15)?

jbergstroem: no no thats perfect but if I'm hallucinating and its not needed then we can just skip it -- I'm just sure I encountered a box that didn't have it installed. How about we stick to facts and you just install it as you chose to. Sorry for the noise :/

jbergstroem: up to you I guess :) Happy with both!

setup/github-bot/ansible-playbook.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - name: Init | Clone github-bot repo
+      become: yes
+      become_user: "{{ server_user }}"
+      git: repo=https://github.com/nodejs/github-bot.git dest=/home/"{{ server_user }}"/github-bot
+      tags: init
+
+    - name: Init | Install npm dependencies
+      become: yes
+      become_user: "{{ server_user }}"
+      command: "npm install"
```
jbergstroem: You can use the [npm module](http://docs.ansible.com/ansible/npm_module.html) here.

phillipj:
Just pushed a commit removing some unwanted quotes. I thought *every* variable should be quoted, that didn't work as expected. Tried this playbook on the server running the bot in production ATM, but couldn't get the bot started afterwards. Suddenly realised a strange directory had been made in `/home`:

```bash
/home# ls
drwxr-xr-x 4 iojs   root   4096 Aug 25 13:41 "iojs"
drwxr-xr-x 5 iojs   iojs   4096 Aug 25 13:47 iojs
```

jbergstroem:
@phillipj you should quote the full string, not the variable. That way it doesn't get escaped.

jbergstroem:
Ok, if you've tested that quotes work this is LGTM. Thanks for enduring my nitpick :) Great to have you on board!

phillipj:
My pleasure, now I almost know how ansible works and your nitpicks was perfect for doing things right™.

I'll squash 14 commits so far, but let you decide when the time is right to merge -- maybe land your big ansible refactoring first for all I know.

jbergstroem:
Squash it and see if anyone else in the build group wants to LGTM, then merge at will!

mhdawson:
LGTM

PR #477, opened 2016-08-26
iptables/dgram: Allow local origin traffic on the test port

lostnet:
Some Fedora/Redhat versions have overly restrictive firewall rules
that prevent local tests involving the system's real IP or
multicast IPs.

While making this [putback with a dgram multicast test](https://github.com/nodejs/node/pull/7855), I encountered issues with several systems that stem from their default iptables rules. I have done some testing on fedora 22 to determine a correct configuration to allow local tests on the test port without opening it for remote traffic. I have used these changes to make changes that I think will work for redhat 7.2, but I don't have access to one of these systems.

### DIFF
--- a/setup/fedora22/ansible-playbook.yaml
+++ b/setup/fedora22/ansible-playbook.yaml
@@ -42,6 +42,32 @@
       user: name="{{ server_user }}" shell=/bin/bash
       tags: user
 
+    - name: Firewall | check basic rule to allow test port communication
+      command: iptables -C INPUT -p udp --dport 12346 -j ACCEPT
+      register: iptables_find_basic
+      ignore_errors: True
+      tags: firewall
+
+    - name: Firewall | add basic rule to allow test port communication
+      command: firewall-cmd --permanent --direct --passthrough ipv4 -I INPUT 1 -p udp --dport 12346 -j ACCEPT
+      when: iptables_find_basic|failed
+      tags: firewall
+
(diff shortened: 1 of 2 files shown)


mhdawson:
@jbergstroem what you do think of adding configuration of the iptables in this way

mcollina:
@mhdawson @rvagg can we move this forward? it's blocking https://github.com/nodejs/node/pull/7855.

maclover7:
ping -- is this still needed?

mhdawson:
@jbergstroem any objections ?

rvagg:
I'm not seeing _any_ firewall rules on the fedora hosts that I'm testing. Does it make a difference now that we only have Fedora 23+ and not 22?

lostnet:
@rvagg I think most newer distributions were setup to accept all locally generated traffic by default, so 23 might have been the cut off for fedora.. Still needing rules for 127.0.0.2 for a host like [rhel72-s390x](https://github.com/nodejs/build/commit/d2942fe66f6b4bd2dba2b7ee37557ea7c3c8b608#diff-4bbb7b4f3a118086659d01bec9c83d28R62 ) implies that distribution doesn't have a setup to let all locally generated traffic pass which means a local test with either multicast and/or the nodes own external addresses selected would fail limiting what tests could be done.

sam-github:
CI is green, so this doesn't seem to be needed anymore, or was worked around in other ways. We can reopen if that is not the case.

PR #483, opened 2016-08-31
rpi: update docs, add apt proxy config

rvagg:
Mainly adds docs on how to set this up as well as all of the SSH config stuff that you can use to get in. anyone with nodejs_build_test can get in to all of these except the two release- pi's but the jump host etc. will work so @nodejs/build should feel free to administer these as any other test host.

Also adds APT proxy config to the Ansible setup, it's been a manual step since I added it.

I have two more ideals items to add here:
- Set up custom image for the Wheezy-based boards which does the first partition setup as per the docs so I just have to burn it in. It'd also be really nice to have some of the setup baked into it too, like ssh keys and some of the items I have to set with `raspi-config`
- Ansiblise the `raspi-config` stuff that can't be baked in, such as expanding the filesystems .

With those two additions it'd be soooo simple to refresh SD cards for failed machines and fairly simple to add new machines.


jbergstroem:
Looks good. Thanks for adding the rpi installation instructions.

Nitpick: could you also rename the xgene's?

rvagg:
@jbergstroem yes I can get around to renaming them sometime, but so could you now!

jbergstroem:
Ok, merge this and I'll rename them separately!

PR #493, opened 2016-09-15
Created a small working with Ansible guide

phillipj:
As briefly mentioned in the last WG meeting, while working in the github-bot setup, I got a local Vagrant setup working for running playbooks against. I thought that was really valuable when trying to learn Ansible basics, and it made it trivial to start from scratch after having tested a lot of different approaches.

Any feedback would be more than welcome!

### DIFF
--- a/setup/README.md
+++ b/setup/README.md
@@ -1,0 +1,82 @@
+# Ansible setup
+
+Servers in the Node.js infrastructure is provisioned with Ansible. This directory contains all
+the related Ansible files.
+
+For more information and getting started with Ansible, please read [getting started on ansible.com].
+
+# Working with Ansible
+
+While working with these ansible playbooks, there's great value in testing them locally first, before running
+them on one or more remote servers. The following section tries to describe how you would do just that
+on a Vagrant virtual machine running debian for the [github-bot](github-bot/ansible-playbook.yaml).
+
+Good luck, have fun!
+
+## Prerequisites
+
+- Vagrant: https://www.vagrantup.com/downloads.html
+
+## Create a Vagrantfile
+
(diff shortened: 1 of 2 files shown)


jbergstroem:
This is awesome. Will check soon!

phillipj:
Not sure how / where to mention the required vars from the secrets repo tho

jbergstroem:
That's fine, we can add that somewhere else.

mhdawson:
My first thought after reading was that we might want to split into 2 docs.  The first which has the links and general information which would go in the readme.md under setup and then a separate one with the specific example of how to do it with Vagrant.  This might avoid any confusion about Vagrant being used/required.

jbergstroem:
I'm more interested in the Vagrant guide seeing how that will make more people able to test the playbooks! I have a 'general' ansible guide (at least in terms of using it) at my branch: https://github.com/jbergstroem/build/tree/feature/refactor-the-world/ansible#ansible-scripts-for-the-nodejs-build-group-infrastructure

phillipj:
Thanks for reviewing, sorry about the late reply!

In an attempt to make this guide more Vagrant related, I removed what could be seen as an intro to ansible and the use of ansible in the project.

PTAL

mhdawson:
I think the content looks good now, but should probably be named something other than README.md.  I'm thinking the README.md for the setup directory should be an overview and point to docs like the "general" ansible guide and this one about testing locally.

jbergstroem:
I'm with @mhdawson -- let's reference it from README.

phillipj:
I won't push the merge button yet, as there hasn't been a clear LGTM from anyone yet, tho feedback has been positive and there's no outstanding issues afaik.

PR #510, opened 2016-10-07
build: add SSL_CERT_FILE env variable

mhdawson:
When building v4.x ICU is downloaded and without
setting SSL_CERT_FILE the download fails on AIX
because by default the required certificates are not
found.  Add the SSL_CERT_FILE environment variable
to point to the certs we install as part of the
ansible configuration  in a manner similar to what
we already do for GIT_SSL_CAINFO.


jbergstroem:
LGTM

mhdawson:
Landed as https://github.com/nodejs/build/commit/30ed21140f9680fdecba7728171646e0fbe1a2ee

Issue #550, opened 2016-11-21
https://nodejs.org/metrics/ is/was borken 


Fishrock123:
https://nodejs.org/metrics/

cc @nodejs/build, @rvagg 

If it no longer is it would good to know what happened. :D

jbergstroem:
what does "broken" mean? 404?

targos:
Looking at https://nodejs.org/metrics/summaries/total.csv
It seems that there is an issue collecting the metrics. There is no data between 2016-10-13 and today apart from one point on 2016-11-01.

rvagg:
logrotate wasn't working, fixed in https://github.com/nodejs/build/pull/552, the effect of this is that the unrotated access.log file was getting too big to process effectively—there were a few processes running on the server trying to complete it to no avail.

I've fixed it all up now and split up the big file manually into one per day and it's now processing and should look kind of normal again soon.

PR #552, opened 2016-11-22
Fix nginx logrotate script for www

rvagg:
fixes https://github.com/nodejs/build/issues/550


rvagg:
apparently logrotate doesn't support globs ... so there you go

jbergstroem:
Wow. I've been "using" globs for I don't know how long..

rvagg:
perhaps it was the commas? or perhaps this is a linux thing and you're just used to some dodgy *bsd hack

jbergstroem:
Ah, I think it was the commas. Was just baffled by the fact that it might have been broken and I've never noticed on my servers either.

PR #553, opened 2016-11-23
github-bot: redirect stdout to file

phillipj:
Fixes an issue we have currently, where all current logs are deleted whenever the bot is redeployed because of two things:

1. We have the github-bot repo cloned and write logs to a sub-directory of repo root (`logs`)
2. Upon deploy `git clean -fdx` deletes any untracked files in the repo

Simply writing out bunyan logs to stdout would also simplify our current log related code in the bot.

Is `logrotate` the way to go for rotating this?

### DIFF
--- a/setup/github-bot/resources/github-bot.service
+++ b/setup/github-bot/resources/github-bot.service
@@ -5,7 +5,7 @@ Description=github-bot
 WantedBy=multi-user.target
 
 [Service]
-ExecStart=/usr/bin/node server.js
+ExecStart=/usr/bin/node server.js &> /home/{{server_user}}/logs/github-bot.log
 EnvironmentFile=/home/{{server_user}}/environment/github-bot
 WorkingDirectory=/home/{{server_user}}/github-bot
 Restart=always


phillipj:
Refs https://github.com/nodejs/github-bot/issues/88#issuecomment-262628736

jbergstroem:
Wouldn't stdout be collected by systemd (when deployed on debian8, ubuntu, etc) which is already managed by logctl?

phillipj:
I'm all ears to better / simpler ways. I'd say it's important to be able to
expose this log over HTTP like we currently have. Using bunyan to filter
and trace logs as been really valuable so far.

jbergstroem:
I'm personally not a "log over http" guy (grep, awk, tail, ..) but if has been a big part of the teams debugging lets opt for that then.

If you're going that route, look at systemd's `StandardOutput=` directive and pass that to file. Another alternative could be using console+syslog and use the system logger (with an identifier) to control output. That way, the system logger could probably handle rotating as well.

jbergstroem:
By the way, does stdout follow the same format bunyan expects?

phillipj:
> By the way, does stdout follow the same format bunyan expects?

Yepp. Everything we log is through bunyan, which writes both to stdout and a log file ATM. When running locally via `npm start` the stdout output gets prettified for easier reading tho. That's not the case on the server, as the systemd service executes the server script directly.

phillipj:
On second thought, doing this seemed more complex than simply writing logs outside the directory where the github-bot repo is cloned into.

Refs https://github.com/nodejs/build/pull/602

phillipj:
Closing this in favor of #602

PR #557, opened 2016-12-02
www: enable fallback for German locale

lpinca:
This enables English fallback for files not translated to German.

Refs: https://github.com/nodejs/nodejs.org/issues/1039


jbergstroem:
Merged in 964af46ceb62850ac9138efc9d01a793422bad02.

Issue #561, opened 2016-12-08
set git author for the github-bot


Fishrock123:
See https://github.com/nodejs/github-bot/issues/100

I wasn't able to get it to work by setting it globally, but setting it in the cloned node repo worked.

```
git config user.name "nodejs-github-bot"
git config user.email "github-bot@nodejs.org"
```

phillipj:
I'll take a look asap. Any idea why author needs be set when the bot
doesn't try to commit anything?

Fishrock123:
@phillipj Technically it does "commit" stuff -- using `git am` applies with the user as the commiter.

phillipj:
Ah right! I'll try to get it done tonight. Setting that config locally to
that repo rather than global seems fine to you @Fishrock123?

Fishrock123:
@phillipj Either or should be fine. On-repo might be simpler since I could not get globally to work on that machine.

phillipj:
Opened a PR fixing this permanently. It seems to have been set correctly on the server tho, I assume someone has logged in and fixed it by hand.

Fishrock123:
@phillipj (That was me)

phillipj:
Fixed in https://github.com/nodejs/build/commit/857f5900f3cde69080f4f64b2c725bca2d24728f

PR #567, opened 2016-12-09
github-bot: config git author name and email

phillipj:
Needed for the attempt-backport functionality of the bot since it does
some git actions requiring author name and email, such as `git am`.

Seeing this when running locally:

```
TASK [Runtime dependencies | Config node repo author email] ***********************
changed: [github-bot]
 [WARNING]: Consider using git module rather than running git
```

But as far as I see, manipulating config isn't possible with the [git module](http://docs.ansible.com/ansible/git_module.html). Found [git_config module](http://docs.ansible.com/ansible/git_config_module.html) tho, if we're fine with requiring ansible v2.1?

Refs: https://github.com/nodejs/build/issues/561


phillipj:
For the record:

@jbergstroem you approving the changes as is, I assume we don't want to require ansible v2.1.

Issue #551, opened 2016-11-21
AIX: Backlog


mhdawson:
We have a few tasks outstanding in respect to the AIX configuration. @gibfahn was asking if we had a list somewhere so this issue will act as that list:

- [x] install version of python 2.x that supports PIP - https://github.com/nodejs/build/issues/533
- [x] enable jenkins agent auto start - https://github.com/nodejs/build/issues/534
- [x] submit PR to install instructions to cover disk io optimization. - https://github.com/nodejs/build/issues/537
- [x] make binaries needed for AIX install available through scp (or other appropriate method) from 
      jenkins masker
- [ ] update Ansible scripts so that they pull binaries instead of requiring manual install

mhdawson:
@gibfahn as requested.  If there are others I'll add to this list.  The installation of the python version that supports PIP is the most urgent.  Is that something you can look at this week ?   We'll want to look at the dependencies carefully as I ran into problems in previous updates and stage to one machine at a time.

gibfahn:
I'm pretty sure that there's an issue with curl on the AIX boxes, not sure if it's a LIBPATH issue or if something's wrong with the install. We don't have it on IBM machines, so it's definitely fixable, just annoying (it means things have to be scp'd rather than just curl'd.

```bash
$ curl
exec(): 0509-036 Cannot load program curl because of the following errors:
        0509-022 Cannot load module /usr/lib/libcurl.a(libcurl.so.4).
        0509-150   Dependent module /opt/freeware/lib/libcrypto.a(libcrypto.so) could not be loaded.
        0509-152   Member libcrypto.so is not found in archive 
        0509-022 Cannot load module curl.
        0509-150   Dependent module /usr/lib/libcurl.a(libcurl.so.4) could not be loaded.
        0509-022 Cannot load module .
```

cc/ @sxa555 who is looking at the AIX Ansible stuff.

sxa555:
It should work if LIBPATH is pointing at /usr/lib (although that could well cause other issues of course!) - the libcrypto.a in /opt/freeware/lib doesn't have a non-version-suffixed version of libcrypto.so in it but the one in /usr/lib does. I'm not sure why the one in /opt/freeware/lib doesn't ... You'd expect it to work with the version of curl in there. You could just override it for the execution of curl for now.

gibfahn:
@sxa555 Yep, that works. I'll add the LIBPATH setting to the manual install instructions for now (it's not a problem in CI AFAIK).

Issue #533, opened 2016-11-14
Install latest python 2.7 with pip on AIX


mhdawson:
From @jbergstroem 

Since we will be requiring pip to install tap2junit at some stage (junit xml output from test runners) I hope the install won't have any issues. According to http://www.aixtools.net/index.php/python, recent releases should have pip installed?

gibfahn:
@gdams has used pip on AIX before, he's going to try getting it working on a temporary machine so we can work out the process.

edelsohn:
I have Python pip installed on my AIX systems. I was able to bootstrap with get-pip.py.  There is a typo / bug in the sysconfigdata built on AIX for some Python releases.

gdams:
okay so I got pip working on aix. Firstly download `https://bootstrap.pypa.io/get-pip.py` and then just run `python get-pip.py` and then just ensure that `/opt/freeware/bin` is in the path

jbergstroem:
great news! can we get this in as part of the ansible task

gibfahn:
@jbergstroem So, my plan is:

 1. Install pip manually on one of the test machines and document in [setup/aix61/manualBootstrap.txt](https://github.com/nodejs/build/blob/master/setup/aix61/manualBootstrap.txt).
 2. If nothing is broken after a week, install on the other test machine and ask you to install on the release machine.
 3. Once Ansible is setup on AIX (see https://github.com/nodejs/build/issues/551) add it to the Ansible scripts.

gibfahn:
I've installed pip on `power8-nodejs3.osuosl.org` (which I'm pretty sure is `test-osuosl-aix61-ppc64_be-1`). Pip itself seems to be working fine, my issue with curl is documented in #551.

What I did:

```bash
curl -O https://bootstrap.pypa.io/get-pip.py 
scp get-pip.py test-osuosl-aix61-ppc64_be-1:/home/gib

ssh test-osuosl-aix61-ppc64_be-1
bash
python /home/gib/get-pip.py
export PATH=/opt/freeware/bin:$PATH
pip
```

sxa555:
I thought I'd added this comment yesterday but apparently not - might be better to symlink /opt/freeware/bin/pip to /usr/bin/pip to avoid having to tweak the path - that would make it consistent with how a lot of the other freeware tools get installed.

gibfahn:
Makes sense, I'll update the script (and the docs)

gibfahn:
[...]

PR #598, opened 2017-01-25
AIX: Update manual setup instructions

gibfahn:
doc: update AIX manual install instructions
- Convert txt -> md
- Remove obsolete sections
- Document curl LIBPATH fix
- Document pip install instructions

Refs: #551
Refs: #533

To view the markdown (if the diff doesn't work): https://github.com/gibfahn/build/blob/aix-pip/setup/aix61/manualBootstrap.md

cc/ @mhdawson @sxa555


setup/aix61/manualBootstrap.md
```diff
@@ -85,60 +122,59 @@ ln -s /opt/freeware/bin/ccache/ccache /opt/freeware/bin/ccache/g++
 ln -s /opt/freeware/bin/ccache/ccache /opt/freeware/bin/ccache/gcc
 cd ../..
 rm -rf tmp
+```
+
+## Install svn
 
-5) install svn, only needed for v8 testing
+*N.B. Only needed for v8 testing*
```
gibfahn: @mhdawson do we still need to install svn? I thought we'd moved to git for the v8 testing. If so I'll remove this as well.

mhdawson: I think you're right, we had to switch over a while back.

jbergstroem: Does this mean we can drop svn from our global dependencies for all builders?

mhdawson: Likely yes.  We'd want to remove on one set of machines fist to validate but if we are right we don't need it anymore.

Issue #548, opened 2016-11-18
Jenkins connection issue on test-osuosl-ubuntu14-ppc64_be-2


gibfahn:
Fixed by restarting the Jenkins service

https://ci.nodejs.org/job/node-test-commit-plinux/5208/nodes=ppcbe-ubuntu1404/console

```
[EnvInject] - Loading node environment variables.
Building remotely on test-osuosl-ubuntu14-ppc64_be-2 (ppcbe-ubuntu1404) in workspace /home/iojs/build/workspace/node-test-commit-plinux/nodes/ppcbe-ubuntu1404
java.io.IOException: remote file operation failed: /home/iojs/build/workspace/node-test-commit-plinux/nodes/ppcbe-ubuntu1404 at hudson.remoting.Channel@57fe741f:test-osuosl-ubuntu14-ppc64_be-2: java.io.IOException: Remote call on test-osuosl-ubuntu14-ppc64_be-2 failed
	at hudson.FilePath.act(FilePath.java:1014)
	at hudson.FilePath.act(FilePath.java:996)
	at org.jenkinsci.plugins.gitclient.Git.getClient(Git.java:137)
	at hudson.plugins.git.GitSCM.createClient(GitSCM.java:739)
	at hudson.plugins.git.GitSCM.createClient(GitSCM.java:731)
	at hudson.plugins.git.GitSCM.checkout(GitSCM.java:1076)
	at hudson.scm.SCM.checkout(SCM.java:495)
	at hudson.model.AbstractProject.checkout(AbstractProject.java:1278)
	at hudson.model.AbstractBuild$AbstractBuildExecution.defaultCheckout(AbstractBuild.java:604)
	at jenkins.scm.SCMCheckoutStrategy.checkout(SCMCheckoutStrategy.java:86)
	at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:529)
	at hudson.model.Run.execute(Run.java:1720)
	at hudson.matrix.MatrixRun.run(MatrixRun.java:146)
	at hudson.model.ResourceController.execute(ResourceController.java:98)
	at hudson.model.Executor.run(Executor.java:404)
Caused by: java.io.IOException: Remote call on test-osuosl-ubuntu14-ppc64_be-2 failed
	at hudson.remoting.Channel.call(Channel.java:805)
	at hudson.FilePath.act(FilePath.java:1007)
	... 14 more
Caused by: java.lang.OutOfMemoryError: PermGen space
[...]

PR #605, opened 2017-01-30
build: update jenkins agent to 256m

mhdawson:
We seemed to be having out of memory issues on some machines.
For example:  https://github.com/nodejs/build/issues/548#issuecomment-269974066.
Bumping up to 256m seems to have resolved so modifying the
ansible config to reflect that.

Fixes: https://github.com/nodejs/build/issues/548

### DIFF
--- a/setup/ubuntu14.04/resources/jenkins.conf.j2
+++ b/setup/ubuntu14.04/resources/jenkins.conf.j2
@@ -20,7 +20,7 @@ setgid iojs
 setuid iojs
 
 script
-  exec java -Xmx128m -jar "$HOME/slave.jar" \
+  exec java -Xmx256m -jar "$HOME/slave.jar" \
 	-secret {{ secret }} \
 	-jnlpUrl https://ci.nodejs.org/computer/{{ inventory_hostname }}/slave-agent.jnlp
 end script


jbergstroem:
The problem is that some vm's are running on 2G; meaning the more memory we have available for compilation and testing, the better. An idea I've been playing around with in my refactor is passing memory as a variable to the template and overriding it - if needed - for specific providers, operating systems or architectures (all of those are available to test against in Ansible).

mhdawson:
Should I just close this given that we are moving to the refactor ?

gibfahn:
>Should I just close this given that we are moving to the refactor ?

Maybe we should change the default `server_ram` from 128M to 256M in the refactor?

```bash
➜  ansible git:(master) ❯ rg server_ram                                                                                                                                             ~/wrk/com/build/ansible
host_vars/README.md
20:- `server_ram`: how much memory the slave should assign to java-base

roles/jenkins-worker/templates/centos.initd.j2
20:JENKINS_SLAVE_ARGS="-Xmx{{ server_ram|default('128m') }} -jnlpUrl {{ jenkins_url }}/computer/{{ inventory_hostname }}/slave-agent.jnlp -secret {{ secret }}"

roles/jenkins-worker/templates/freebsd.initd.j2
36:jenkins_args="-Xmx{{ server_ram|default('128m') }} -jar ${jenkins_jar} \

roles/jenkins-worker/templates/jenkins_manifest.xml.j2
31:                   exec='{{ java_path[os] }} -Dsun.security.pkcs11.enable-solaris=false -Xmx{{ server_ram|default('128m') }} -jar /home/{{ server_user }}/slave.jar -jnlpUrl {{ jenkins_url }}/computer/{{ inventory_hostname }}/slave-agent.jnlp -secret {{ secret }}'

roles/jenkins-worker/templates/openrc.initd.j2
7:command_args="-Xmx{{ server_ram|default('128m') }} \

roles/jenkins-worker/templates/systemd.service.j2
25:ExecStart=/usr/bin/java -Xmx{{ server_ram|default('128m') }} \

roles/jenkins-worker/templates/upstart.j2
26:  exec java -Xmx{{ server_ram|default('128m') }} \
```

jbergstroem:
server_ram can be overridden from the host vars. The idea was to the specific hosts that needs it allow override.

gibfahn:
Sure, my point was that if we think they'll mostly need 256M rather than 128M then we should consider updating the default. If it's just the odd one then I agree we don't want to.

jbergstroem:
My experience is that they - at least so far - doesn't need more than 128m.

gibfahn:
Might as well close then.

jbergstroem:
I think the issue mutated after the branch landed -- it's much easier to override per host or even per group via config.

PR #624, opened 2017-02-13
doc: Updated README.md for restart instructions

kunalspathak:
Updated `README.md` to be clear on restart instructions after
`ansible-playbook` command is run.

### DIFF
--- a/setup/windows/README.md
+++ b/setup/windows/README.md
@@ -48,7 +48,6 @@ To run the Ansible playbook to setup the machines run:
 $ ansible-playbook -i ../ansible-inventory ansible-playbook.yaml -vv
 ```
 
-The servers should logon automatically at boot and start the Jenkins slave, but there is no automatic restart procedure for crashes (for now).
-Hence, to start Jenkins after running `ansible-playbook`, the server must be restarted manually so Jenkins script finds the required tools (`java` and `curl`) in the path.
+The servers should logon automatically at boot and start the Jenkins slave.
 
 The release servers need to have the WiX Toolset and 7-Zip installed and in the path (not part of this script).
--- a/setup/windows/README.md
+++ b/setup/windows/README.md
@@ -49,6 +49,6 @@ $ ansible-playbook -i ../ansible-inventory ansible-playbook.yaml -vv
 ```
 
 The servers should logon automatically at boot and start the Jenkins slave, but there is no automatic restart procedure for crashes (for now).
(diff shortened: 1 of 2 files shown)


setup/windows/README.md
```diff
@@ -49,6 +49,6 @@ $ ansible-playbook -i ../ansible-inventory ansible-playbook.yaml -vv
 ```
 
 The servers should logon automatically at boot and start the Jenkins slave, but there is no automatic restart procedure for crashes (for now).
-Hence, to start Jenkins after running `ansible-playbook`, the server must be restarted or the slave must be manually started with RDP.
```
gibfahn: Is there no other way on windows to update the PATH than restarting the machine?

Presumably you only have to restart the server the first time you run the script, as if you run it again to update the machines later `curl` and `java` will already be in the path.

Also, does Ansible not have a way to restart the server once it's finished (@joaocgreis @piccoloaiutante ) ?

piccoloaiutante: Yes there is a command for doing this (https://github.com/nodejs/build/blob/master/setup/windows/vs2015-ansible-playbook.yaml#L29). Perhaps we could add another reboot after `common-ansible-playbook.yaml` has been played?

joaocgreis: @gibfahn I like to restart the machine at the end to make sure that autologon and starting Jenkins work. Sometimes the machine does not logon automatically for reasons unknown and running that step again fixes it.

We could add the reboot at the end of the Ansible script. @kunalspathak can you add it to this PR and update the text here accordingly?

kunalspathak: I was under the impression that auto logon never worked if we restart through playbook. 

Also I am a bit confused reading the following line in `README.txt`
```
but there is no automatic restart procedure for crashes (for now).
```

What crashes does this line talk about? Once we add `win_reboot`  step in `common-ansible-playbook.yaml`, I think it will be advisable to just say `"The servers should logon automatically at boot and start the Jenkins slave."` and remove other text/line related to restarting the server? Thoughts?

By the way, I verified adding `win_reboot` step and the auto logon and path resolution works.

joaocgreis: @kunalspathak that line is outdated now, I wrote it before I added the loop in `jenkins.bat` and never removed it. You suggestion is perfect, please go ahead and update this! Thanks for noticing!

kunalspathak: Thanks @joaocgreis . Updated the PR.

joaocgreis:
Landed in https://github.com/nodejs/build/commit/d2d5dd2506951d0dfdd2ffa83aeba8b37b3e6cbc

PR #638, opened 2017-03-01
update releasers list

rvagg:
updated to match releases list

### DIFF
--- a/setup/www/ansible-vars.yaml
+++ b/setup/www/ansible-vars.yaml
@@ -1,12 +1,12 @@
 ---
 server_user: nodejs
 dist_users:
-  - chrisdickinson
   - cjihrig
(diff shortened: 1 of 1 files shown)


cjihrig:
@rvagg I'm not sure if it matters, but I've done the past few libuv releases. Do I need to be listed in the `libuv_users` portion of this file?

rvagg:
Yes @cjihrig, good catch, if I add you is that libuv list correct or does it need additional adjustments?

cjihrig:
If this is just for releases, then yes, that is enough. If you're looking for a list of all of the maintainers, then there are a few missing from https://github.com/libuv/libuv/blob/v1.x/MAINTAINERS.md.

PR #650, opened 2017-03-14
ubuntu: bump gcc to 4.9 on 12.04

targos:
The latest version of 4.8 available for 12.04 is 4.8.4.
It is not able to build V8 from version 5.7 and does not
satisfy the requirements to build Node.js.

Refs: https://github.com/nodejs/node/pull/11840
Refs: https://github.com/nodejs/v8/issues/5

/cc @nodejs/build @bnoordhuis @rvagg

### DIFF
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -104,5 +104,12 @@ packages: {
     'g++',
     'gcc',
     'git',
+    ],
+
+  ubuntu12: [
+    'ccache',
+    'g++-4.9',
+    'gcc-4.9',
+    'git'
     ]
   }
--- a/setup/ubuntu12.04/ansible-vars.yaml
+++ b/setup/ubuntu12.04/ansible-vars.yaml
@@ -9,7 +9,7 @@ packages:
   - automake
   - libtool
   - curl
-  - gcc-4.8
-  - g++-4.8
+  - gcc-4.9
+  - g++-4.9
   - ccache
   - subversion


targos:
ping. I haven't tested either but I think it's important to test this (and update the machines) soon.
It is blocking https://github.com/nodejs/node/pull/11752.

jbergstroem:
Ah. Well,  seeing how at least one other build member agrees I'll log in to one of our testers and replace it. Will update this issue.

jbergstroem:
Edit: all three builders now runs gcc 4.9. Looks like there were symlinks missing for ccache so i fixed that as well. Please let me know how testing goes and if we need to do something else.

targos:
I started a build with V8 5.7 and it's fully green on Linux: https://ci.nodejs.org/job/node-test-commit-linux/8603/

targos:
This seems to cause a problem with clang 3.4.1. https://ci.nodejs.org/job/node-test-commit-linux/8680/nodes=ubuntu1204-clang341-64/

bnoordhuis:
@targos The ubuntu1204-clang341-64 buildbot was taken out of commission.  Does that mean this pull request can go through now?

targos:
Yes, I think so.

targos:
ping @nodejs/build

gibfahn:
I'm okay with this landing, but we're trying to move to the scripts in [`ansible/`](https://github.com/nodejs/build/tree/master/ansible), so this change probably needs to be made there as well.

targos:
I can't find the place where Ubuntu 1204 is configured in `ansible/`

gibfahn:
>I can't find the place where Ubuntu 1204 is configured in ansible/

It might not have been added yet, I haven't had a chance to check out the new scripts, that's more a question for @jbergstroem @joaocgreis @piccoloaiutante

piccoloaiutante:
@targos I think you could add a new section at the end of https://github.com/nodejs/build/blob/master/ansible/roles/baselayout/vars/main.yml like this:

```
ubuntu12: [
   'ccache',
   'gcc-4.9',
   'g++-4.9',
   'git',
]
```
is it correct @jbergstroem ?

targos:
@piccoloaiutante OK thank you. I added it. Let me know if something else is missing.

Issue #648, opened 2017-03-13
HTTP forward and CNAME for newsletter


mikeal:
Heya,

We'd like `newsletter.nodejs.org` to point to http://us14.campaign-archive2.com/home/?u=c7c2e114a827812354112c23b&id=f006b61f29 . We need something setup similar to what we have for `interactive.nodejs.org`. It's just a few clicks and copy pastes in the Cloudflare dashboard :)

Original issue is here https://github.com/nodejs/nodejs.org/issues/1142

jbergstroem:
I can set it up unless anyone from the build group has opinions on how it should be done. I'll wait around for a day or three.

mikeal:
The initial request has been sitting for like 20 days already, so the quicker the better ;)

jbergstroem:
So, I looked at it (over at cloudflare) and it's not handled through there -- rather in our [nginx config](https://github.com/nodejs/build/blame/master/setup/www/resources/config/nodejs.org#L163).

We should probably add SSL redirects for both as well.

mikeal:
Unfortunately the newsletter listing page isn't available in TLS. However, the link to signup which actually accepts personal information you may not want to leak is TLS.

jbergstroem:
@mikeal was just referring to capturing `https://newsletter.nodejs.org` in our redirect rule as well. It would land in `default_server` otherwise; likely implying `nodejs.org`.

mikeal:
ahhhhh

mikeal:
poke :)

rvagg:
pls review @mikeal @jbergstroem https://github.com/nodejs/build/pull/652

PR #652, opened 2017-03-17
www: add newsletter.nodejs.org redirect

rvagg:
Fixes: https://github.com/nodejs/build/issues/648

### DIFF
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -168,6 +168,13 @@ server {
 }
 
 server {
+    listen *:80;
+    server_name newsletter.nodejs.org;
+
+    return 301 http://us14.campaign-archive2.com/home/?u=c7c2e114a827812354112c23b&id=f006b61f29;
+}
+
+server {
     listen *:443 ssl http2;
     listen [::]:443 ssl http2;
     server_name www.nodejs.org;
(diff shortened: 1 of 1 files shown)


setup/www/resources/config/nodejs.org
```diff
@@ -184,6 +191,14 @@ server {
 }
 
 server {
+    listen *:443 ssl http2;
+    listen [::]:443 ssl http2;
+    server_name newsletter.nodejs.org;
```
jbergstroem: this needs to include the ssl partial include

mikeal:
I am +1 on this happening. However, I'm not familiar enough with the technical bits here to do a proper "review."

setup/www/resources/config/nodejs.org
rvagg: @jbergstroem can you explain this? I don't get it.

jbergstroem: @rvagg: As always; I goofed. When I refactored how we add SSL certificates I put it in [a partial](https://github.com/nodejs/build/blob/master/setup/www/resources/config/ssl-defaults.conf), so my thought was that we had to include it. What I had forgotten was that we chose to include it as default and override it [where needed](https://github.com/nodejs/build/blame/master/setup/www/resources/config/iojs.org#L13..L15).

rvagg:
@mikeal this is done now and is working, prefer the https version when handing it around or linking, sorry it's taken so long when it's actually very simple!

PR #654, opened 2017-03-17
Metrics: Update download-counts.awk to correctly parse Windows zip and IBM platforms

Paul-Ivinson:
The download-counts.awk is skipping win 7z and zip files so these are not included in the metrics csv files. Also the csv records for the IBM binaries do not have the os and arch fields set.

Added zip and 7z file extensions to the 'not excluded' regex.
Added parsing of the fileType 'win' and the IBM platforms (aix-ppc, linux-ppc64, linux-ppc64le and linux-s390x) to set os and arch.

### DIFF
--- a/setup/www/tools/metrics/download-counts.awk
+++ b/setup/www/tools/metrics/download-counts.awk
@@ -8,12 +8,14 @@ BEGIN {
   printf fmt, "ip", "day", "path", "version", "os", "arch", "bytes"
 }
 
+#{print $0}
+
 (!($6 == "\"GET" || $6 == "GET")) { next } # non-GET requests
 
 ($10 < 1000) { next } # unreasonably small download
 
-!/\.(tar\.gz|tar\.xz|pkg|msi|exe)(\?[^ ]+)? HTTP\/[12]\.[10][" ]/ {
-  #print "Skipping:", $0
+!/\.(tar\.gz|tar\.xz|pkg|msi|exe|zip|7z)(\?[^ ]+)? HTTP\/[12]\.[10][" ]/ {
+  print "Skipping:", $0
   #check we're not missing anything with: grep Skipping /tmp/_var_log_nginx_nodejs.org-access.log | awk '{print $10 $8}'| grep -v '/$\|html$\|png$\|svg$\|json\|jpg$\|xml$\|txt$\|jar$\|js$\|pom$\|css$\|ico$\|zip$\|lib$\|exp$\|^40\|^30\|tab$\|eps$\|asc$\|gpg$\|pdf$\|tgz$\|\?\|\#\|pdb$\|rtf$\|md$\|SHASUMS'
   next
 }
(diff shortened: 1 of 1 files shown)


Paul-Ivinson:
@gibfahn Gibson - can you take a look at this PR please? The nginx access.log records for win 7z and zip downloads were being skipped. Also the IBM platform file names were not parsed to populate the os and arch fields in the metrics csv files.

richardlau:
@Paul-Ivinson Did you mean to close this?

Paul-Ivinson:
Doh - no, I didn't. Thanks for spotting that.

setup/www/tools/metrics/download-counts.awk
```diff
@@ -63,7 +65,9 @@ BEGIN {
       os = "osx"
     } else if (match(fileType, /^sunos-/)) {
       os = "sunos"
-    } else if (match(fileType, /msi$/) || match(file, /node\.exe$/)) {
+    } else if (match(fileType, /^aix-/)) {
+      os = "aix"
+    } else if (match(fileType, /msi$/) || match(file, /node\.exe$/) || match(fileType, /^win-/)) {
```
gibfahn: Does this cover the `.7z` and `.zip` packages?

setup/www/tools/metrics/download-counts.awk
```diff
@@ -8,11 +8,13 @@ BEGIN {
...
 }
 
+#{print $0}
+
 (!($6 == "\"GET" || $6 == "GET")) { next } # non-GET requests
 
 ($10 < 1000) { next } # unreasonably small download
 
-!/\.(tar\.gz|tar\.xz|pkg|msi|exe)(\?[^ ]+)? HTTP\/[12]\.[10][" ]/ {
+!/\.(tar\.gz|tar\.xz|pkg|msi|exe|zip|7z)(\?[^ ]+)? HTTP\/[12]\.[10][" ]/ {
```
gibfahn: cc/ @joaocgreis

setup/www/tools/metrics/download-counts.awk
```diff
@@ -8,11 +8,13 @@ BEGIN {
   printf fmt, "ip", "day", "path", "version", "os", "arch", "bytes"
 }
 
+#{print $0}
```
gibfahn: Could you just delete this?

joaocgreis:
The changes look good (if `m[10]` starts with `win-` and does not include everything under the `win-*` folders). I have no experience with this, is there an easy way to do a test run?

cc @rvagg @jbergstroem

setup/www/tools/metrics/download-counts.awk
```diff
@@ -8,11 +8,13 @@ BEGIN {
...
+
```
jbergstroem: ..and this :)

setup/www/tools/metrics/download-counts.awk
jbergstroem: Thinking yes seeing how path is prepended by `win-{x86|x64}` -- the variable (`fileType`) is just poorly named.

Paul-Ivinson:
Don't have access to the logs so I took a log file from my own nginx server and mocked up some [test data](https://github.com/nodejs/build/files/858358/test.txt). This highlighted a problem with parsing ppc64le - it needs to be handled before ppc64. 

That's now fixed. It would be good if this change could be run against the real logs @jbergstroem

rvagg:
good to go! I've merged a single-commit version of this and am currently reprocessing all of 2017 logs, we could go back further than that but this'll do for now.

gibfahn:
For posterity, this landed as https://github.com/nodejs/build/commit/ae39829c4f3a8e70c0e5e6f96ed7439bb7521a6a

PR #658, opened 2017-03-22
remove the NODE_TEST_DIR env variable on Ubuntu 16.04

mcollina:
Solves some recent problems with `readable-stream` and CITGM.

`NODE_TEST_DIR` is not present in the other environments.

For some context: 

https://github.com/nodejs/node/pull/11760#issuecomment-287990227
https://github.com/nodejs/node/pull/11760#issuecomment-288376079.

cc @MylesBorins @jasnell


jbergstroem:
We added this to solve cross-filesystem and disk size issues. Although it might not be needed for ubuntu16, I think consistency across the fleet is desired. We haven't rolled this out on all machines (since we don't redeploy on every change to playbooks), but it will likely happen sooner or later. We didn't bother creating the directory as part of a playbook since the nodejs test suite added it. Should we not just add the directory instead?

mcollina:
@jbergstroem I'm happy with adding the directory. I thought that given it was not on the full fleet, it was a mistake.

I'm happy with having the directory. The whole point is making sure that the directory exists if the env variable is there.

jbergstroem:
@mcollina perhaps citgm can create the tmp folder if it doesn't exist, similarly to the node test suite? I can add the folder as part of the playbook in parallel.

mcollina:
@jbergstroem it's not a problem of CITGM for that. It's in https://github.com/nodejs/readable-stream/blob/master/test/common.js#L45, which is lifted from core https://github.com/nodejs/node/blob/master/test/common.js#L35-L36. So, it is the node test suite.

I would rather prefer not to do a `readable-stream` release just to change `test/common.js`

piccoloaiutante:
@mcollina @jbergstroem how about adding [here](https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/tasks/partials/tap2junit/ubuntu.yml) this:

```
- name: Creates NODE_TEST_DIR directory
  file: path=/home/{{ server_user }}/tmp state=directory
```

Instead of this pr change?

jbergstroem:
@piccoloaiutante i'd suggest adding in a more generic role, like a java worker. The tap2junit python script is more related to the conversion of output; not so much just running a java worker.

piccoloaiutante:
@jbergstroem shouldn't be better to add it as part of an already existing role like`bootstrap`? We just only need to create the missing folder for ubuntu machine, so we could add it in ubuntu partials [here](https://github.com/nodejs/build/tree/master/ansible/roles/bootstrap/tasks/partials).

jbergstroem:
@piccoloaiutante exactly. It should live with the jenkins client role.

piccoloaiutante:
should we close this in favour of https://github.com/nodejs/build/pull/785  @mcollina  ?

mcollina:
Closed!

gibfahn:
>should we close this in favour of #785 @mcollina ?

In general I'd say leave it open, and add the `Fixes:` line in #785, that way the issue stays open even if the PR doesn't land for some reason. Doesn't really matter either way though.

PR #659, opened 2017-03-24
Redirect 'Advisory Board' links to TSC

fhemberger:
The outdated page `/en/about/organization` is about to be removed (see https://github.com/nodejs/nodejs.org/pull/1076).
Redirects should point to the TSC page.


jbergstroem:
Merged in d3e72bca5. nodejs.org updated and reloaded.

PR #670, opened 2017-03-31
doc: initial version of npm management doc

mhdawson:


gibfahn:
Refs: https://github.com/nodejs/TSC/issues/211

gibfahn:
@mhdawson I had a bunch of really minor nits/suggestions, so I just committed them directly to your branch. If you don't agree with any of them feel free to revert!

doc/process/npm_management.md
```diff
@@ -0,0 +1,34 @@
...
+We need to make sure that we have continuity in terms of being able to publish
+and update these modules.
+
+We have decided to manage these modules as follows:
+
+* Create a user called [`nodejs-foundation`][] who we always add as one of the
+  collaborators with admin rights and for which the password is maintained by
+  the build workgroup.
+* We would then add individuals as collaborators who can also publish.
+  Generally, a module push can be done by the additional collaborators.
```
gibfahn: `Generally, a module push can be done` -> `A module publish will be done`

As below, if this is used as a backup method to add collaborators for continuity, we should emphasize that it's not to override the autonomy of Node.js Foundation module teams.

doc/process/npm_management.md
```diff
@@ -0,0 +1,34 @@
...
+* In the cases where collaborators other than `nodejs-foundation`
+  cease to be active, the build workgroup would provide continuity by using the
+  `node-foundation` user to add additional collaborators who would have the
+  ability to push the module. The `node-foundation` user could also be used to
+  remove collaborators if that was ever necessary.
```
gibfahn: It might be good to add another bullet point:

> The purpose of the `node-foundation` user is not to enable Build Workgroup members to publish npm modules, that should be left to the module collaborators.

I think we need to state that explicitly, to be clear on the purpose of this.

gibfahn:
We should probably get feedback from @nodejs/TSC and module authors, e.g. @nodejs/citgm , @nodejs/post-mortem (`node-report`), and @nodejs/diagnostics (`node-inspect`).

mhdawson:
@gibfahn pushed commit to address comments.

jkrems:
+1 (from `node-inspect` side)

rnchamberlain:
+1 from node-report side
Can you add nodejs/llnode to the first bullet list? thanks

gibfahn:
>When doing an npm publish, the proper thing to do* is to update the package.json version and tag the master branch of the respective github project, and update its changes file. So I think the nodejs-foundation userid will need push access on the projects.

I'm not sure this is necessary right now, as we don't expect `nodejs-foundation` to actually publish modules. If we have to (down the line) then I'm pretty sure any [`nodejs/` org owner](https://github.com/orgs/nodejs/people?utf8=%E2%9C%93&query=%20role%3Aowner) can give the bot access.

mhdawson:
Ok, will plan to land this tomorrow.

joaocgreis:
Sorry for asking this so late. Should the credentials be in the `test` folder? Given that this account is only to be used to add or remove users and that someone with the credentials could cause a lot of harm, I believe we should reduce the scope. We could create a new folder `tsc` and give access to people who are simultaneously Build WG and TSC members (and perhaps more on a case-by-case basis if it makes sense).

Actually, I'm inclined to say this is completely outside of the Build WG purpose, and the only reason we're doing this here is because of access to the secrets repo. The TSC team could be given access to secrets and the new folder `tsc` created on the root, instead of inside `build`.

mhdawson:
@joaocgreis, @jbergstroem did have the discussion around whether it would be ok in the test repo.  There will be other collaborators with access to each of the npm modules as well who would be able to do the same things.  I'm ok with creating a new one if we think its needed though. @jbergstroem what's your take ?

joaocgreis:
@mhdawson I have no intention of blocking this forever, feel free to land if that's your intention. This can always be moved later if somebody else agrees with my concerns above.

mhdawson:
@joaocgreis Since the key is already there, it might make sense to land the doc, and then we can move/update the doc when we complete the discussion.

mhdawson:
@joaocgreis is it ok if I land as is and we revise if we decide to so something later on.

joaocgreis:
@mhdawson sure, we can revise later.

mhdawson:
landed as 85b33e3f42988261afea99da2879482a927afea5

PR #679, opened 2017-04-10
www: add log-collect for managing CloudFlare logs

rvagg:
Use the CloudFlare log fetching functionality to store download logs locally so we can get CF to cache _everything_ and not have to manage logs on the server itself—the primary cause of our lack of resilience, we have a SPOF.

* Store logs in /home/logs/nodejs.org/
* One log per hour of traffic, maximum xz compressed, log file name format is <start timestamp>-<end timestamp>.log.xz
* Use the CloudFlare [logshare](https://github.com/cloudflare/logshare) Go client to fetch logs, CF only makes the last 72 hours of logs available so we have a small window to ensure we get what we want
* Pipe to filter through a simple Node.js script that parses the line-delimited JSON and discards entries that don't relate to our downloadable assets that we want to use for our metrics generation. This makes the compressed files go from around ~20Mb or more to 3-4Mb.
* Unfortunately it's not a perfect API and occasionally results in zero-sized results or seemingly truncated results, but a retry is usually successful.
* Fetch on the hour, discard any log file that looks suspiciously small and always re-fetch the last two because the most recent was always incomplete and the second-most recent could potentially be incomplete depending on the timing of the fetch (just to be sure!)
* Every 12 hours, run a sanity checker across the recent batch of logs, check the last entry of each file and if that entry isn't near the top of the hour then discard the log and the fetcher will redo it again and hopefully get a better version

Yes it's quite convoluted but I'm trying to be careful that we get the highest quality logs since CF doesn't store them beyond 72 hours.

I have this in place now and I'm going to run it for a few days at least and check the quality of what it's got - whether we have a complete set of files and whether those files look complete in their content. I'll then try running metrics from them and see how they compare to what we are currently generating to make sure that the data is good.

### DIFF
--- a/setup/www/ansible-playbook.yaml
+++ b/setup/www/ansible-playbook.yaml
@@ -32,3 +32,6 @@
 
     - include: tasks/metrics.yaml
       tags: metrics
+
+    - include: tasks/log-collect.yaml
+      tags: log-collect
--- a/setup/www/ansible-vars.yaml
+++ b/setup/www/ansible-vars.yaml
@@ -19,4 +19,5 @@ packages:
   - git
   - python-pip
   - gnuplot
+  - golang-go
 benchmark_host: 50.97.245.4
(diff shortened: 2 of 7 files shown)


rvagg:
moving to #987

Issue #689, opened 2017-04-18
meeting planning: Thumbs up / thumbs down


jbergstroem:
(we should really have a template for creating meeting issues)

Lets use thumbs up/thumbs down for noting attendance to the meeting via the issue. 👍🏻 for "I will attend" and  👎🏻 for "not attending". This gives us the benefit of also seeing build member activity.

gibfahn:
There were no objections to this at the meeting. Next steps are for someone to write up the meeting template and PR it to the build WG, and then we'll start doing this for future meetings.

@jbergstroem I can do the PR if you want (I'm already doing the minutes).

rvagg:
I have a request re meetings too -- can we consider adjusting our time for it? DST has made it go back to 6am for me and that sucks so deeply (for me at least). I don't want to exclude folks but perhaps we could do a spreadsheet similar to what the CTC uses to find good meeting times and work out if there is a better time of day? @mhdawson do you have time to facilitate that? Perhaps copy the CTC one and leave you and I on it and have others fill it in. Alternatively perhaps we just expand the CTC one and include Build, I don't know if there's anything particularly sensitive there, what do you think @trott?

Trott:
> I don't know if there's anything particularly sensitive there, what do you think @trott?

I wouldn't object, but I guess if you're asking my preference: separate spreadsheet. I don't know that sharing the link (and thus the data in the spreadsheet) is a violation of anyone's privacy, but it *feels* wrong to me. So just make a new one. But if you want to instead just add a Build tab, I'm not going to stop you.

Trott:
Also: I've not been terribly active in the WG. I'd like to maintain some elevated privileges in Jenkins to take problematic hosts offline and to be able to log in to the machines to terminate stalled tests and stuff like that. If I can somehow do that without having the guilt of missing pretty much every WG meeting and not really participating, that would be awesome. On the other hand, if the two are tied closely together--like, you don't want someone who's not paying attention to Build WG matters to be able to mess things up--I get it.

piccoloaiutante:
@gibfahn if you want i can go ahead with the PR for template. 

@rvagg yes 6AM sucks a lot, lets try to find a better timeframe.

mhdawson:
Opened issue here to work on new time: https://github.com/nodejs/build/issues/698

maclover7:
ping -- can this be closed?

gibfahn:
Yep

PR #692, opened 2017-04-19
Add minutes and issue template for wg meetings

gibfahn:
First draft of templates for meetings, feedback on any part of it welcome. I didn't want to leave placeholder URLs so it was obvious which URLs needed to be added, but 🤷‍♂️ . We could also have more `<!--` style comments, but then we have to delete them which is annoying

Refs: https://github.com/nodejs/build/issues/689

### DIFF
--- a/doc/templates/build-wg-meeting-0000-00-00.md
+++ b/doc/templates/build-wg-meeting-0000-00-00.md
@@ -1,0 +1,31 @@
+# Node.js Foundation Build WG Meeting YYYY-MM-DD
+
+[GitHub issue]()
+[Meeting video]()
+[Previous meeting]()
+
+Next meeting: DD MMM YYYY
+
+## Present
+- <!-- Add yourself if present-->
+- Member (@member)
+
+## Standup
+
+- Member
+  - Work done
+
+## Agenda
+
+- Issue name [#000](https://github.com/nodejs/build/issues/000)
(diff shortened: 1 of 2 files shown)


doc/templates/build-wg-meeting-0000-00-00.md
```diff
@@ -0,0 +1,33 @@
...
+
+Next meeting: DD MMM YYYY
+
+## Present
+- Michael Dawson (@mhdawson)
+- Johan Bergström (@jbergstroem)
+- João Reis (@joaocgreis)
+- Gibson Fahnestock (@gibfahn)
+- Michele Capra (@piccoloaiutante)
+- Kyle Farnung (@kfarnung)
```
piccoloaiutante: How about leaving just one name as place holder as the group will change from one standup to another?

gibfahn: That's reasonable, especially if people fill in their own names in the attendance. To be honest the key for me is avoiding messing up spelling people's names!

I think I'll remove the names and add a comment saying "add your names here", I'm hoping people can spell their own names 😁 .

gibfahn:
Landed in https://github.com/nodejs/build/commit/388499d15ff2666a2cc7cb820ac58511e4a96a42

PR #694, opened 2017-04-22
Rewrite guide

fhemberger:
URL for the debugging guide changed to match the other guides.

### DIFF
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -426,4 +426,6 @@ server {
 
     # Rename 'Stable' release to 'Current'
     rewrite ^/(.*)/download/stable$                               https://$server_name/$1/download/current permanent;
+    
+    rewrite ^/en/docs/guides/debugging_getting_started/           https://$server_name/en/docs/guides/debugging-getting-started/ permanent;
 }
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -426,6 +426,8 @@ server {
 
     # Rename 'Stable' release to 'Current'
     rewrite ^/(.*)/download/stable$                               https://$server_name/$1/download/current permanent;
-    
+
+    # Fix underscores vs. dashes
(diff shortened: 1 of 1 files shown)


setup/www/resources/config/nodejs.org
```diff
@@ -426,4 +426,6 @@ server {
 
     # Rename 'Stable' release to 'Current'
     rewrite ^/(.*)/download/stable$                               https://$server_name/$1/download/current permanent;
+    
+    rewrite ^/en/docs/guides/debugging_getting_started/           https://$server_name/en/docs/guides/debugging-getting-started/ permanent;
```
refack: why just `/en/`?

fhemberger: You're right. Just saw there's a Korean translation as well.

fhemberger:
@refack Fixed both URLs.

Both renamed files are already online.

setup/www/resources/config/nodejs.org
refack: Wont a regex be better? Or are you concerned only with specific old URLS?

fhemberger: The guide only exists in those two languages. New translations will use the correct file name. So no regex is needed.

refack: 👍 You got my LGTM

refack:
LGTM (I'm not write allowed in this repo, so I don't know if it's enough for you)

rvagg:
This is now deployed, but there was a problem on the two lines, @fhemberger I don't quite know what it is but there's some special space character at the begining of both lines, maybe a unicode whitespace of some kind. nginx borked at it and died. Gave us a good chance to exercise our failover solution though and it worked fine!

rvagg:
See https://github.com/nodejs/build/commit/f347666a3a1db760356aa31177ff5521272904a5 FYI, note the weird diff for those two lines.

fhemberger:
Uh, sorry for that! Used the GitHub web editor for this PR … glad it worked out, thanks!

PR #697, opened 2017-04-24
www: redirect for calendar

rvagg:
@mikeal @williamkapke pls review, is this what we're after for calendar access?

### DIFF
--- a/setup/ansible-inventory
+++ b/setup/ansible-inventory
@@ -77,6 +77,10 @@ iojs-release-armv7-wheezy-1
 release-scaleway-debian7-arm-1
 
 [iojs-raspbian]
+test-requireio_joeyvandijk-debian7-arm_pi2-1
+test-requireio_svincent-debian7-arm_pi2-2
+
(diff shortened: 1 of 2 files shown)


mikeal:
Yup.

Also, how do I get the .ics url for this?

gibfahn:
If you follow [the link](https://calendar.google.com/calendar/embed?src=nodejs.org_nr77ama8p7d7f9ajrpnu506c98%40group.calendar.google.com) there's a `+google calendar` button.

image

If you click on the `ICS` button in the calendar settings after adding it, you get this:

https://calendar.google.com/calendar/ical/nodejs.org_nr77ama8p7d7f9ajrpnu506c98%40group.calendar.google.com/public/basic.ics

I feel like there's probably an easier way...

rvagg:
added `$` to the regex and also added one for https://nodejs.org/calendar.ics to take you to the .ics file, is that going to be helpful?

rvagg:
deployed, we now have those two URLs active

PR #714, opened 2017-05-07
ansible: install logrotate package on nginx role

santigimeno:


ansible/roles/nginx/tasks/main.yml
```diff
@@ -7,7 +7,10 @@
 #
 
 - name: install
-  package: name=nginx state=present
+  package: name={{ item }} state=present
```
jbergstroem: quote please

ansible/roles/nginx/tasks/main.yml
```diff
@@ -7,7 +7,10 @@
 #
 
 - name: install
```
jbergstroem: nitpick: seeing how we're going from only installing nginx to installing multiple packages perhaps change to "install required packages"

ansible/roles/nginx/tasks/main.yml
santigimeno: Changing it but, is it necessary?

santigimeno:
Updated. PTAL. Thanks!

PR #715, opened 2017-05-07
ansible,docs: a couple of doc improvements

santigimeno:


ansible/README.md
```diff
@@ -43,7 +43,8 @@ These playbooks are available to you:
   - **jenkins/linter.yml**: Sets up the code linters (flavour of a worker).
 
   - **create-webhost.yml**: Configures the server(s) that host nodejs.org,
-                            iojs.org and dist.libuv.org among other things.
+                            iojs.org and dist.libuv.org among other things
+                            (only works on Ubuntu boxes).
```
gibfahn: Maybe add the same comment to [the file itself](https://github.com/nodejs/build/blob/master/ansible/playbooks/create-webhost.yml#L4)?

ansible/README.md
```diff
@@ -23,7 +23,7 @@ running one (or multiple) of below playbooks. If you're adding a new host,
 limiting ansible to just running on that host is probably quicker:
 
 ```console
-$ ansible-playbook playbooks/create-jenkins-worker.yml \
+$ ansible-playbook playbooks/jenkins/worker/create.yml \
```
gibfahn: 👍

ansible/README.md
jbergstroem: I thought it was somewhat self-explanatory seeing how the playbook only adds those hosts from the inventory.

santigimeno: @jbergstroem would you prefer I drop this change?

gibfahn: @jbergstroem any reason not to be explicit? I find self-explanatory things tend to be quite subjective.

jbergstroem: @gibfahn not at all; I just think its a "level of knowledge" thing. I mean, should we write operating systems everywhere? If it helps someone that its there I don't mind -- just wanted to state that we have similar variations all over the place.

@santigimeno lets keep it in!

jbergstroem: Not perfectly happy with my last response. Let me elaborate for posterity!

So, as I wrote this; I added two hosts to the playbook; both of which have/are "ubuntu". I can understand that if someone is in the process of helping out by testing different playbooks in different environments, running into "hey this broke on !ubuntu" would point to adding similar remarks.

gibfahn:
@jbergstroem does this LGTY?

PR #727, opened 2017-05-18
initial support for openbsd

qbit:
This still needs some love. The rc script isn't right, but I am not familiar enough with jenkins to know why.

Also on OpenBSD, node won't build on partitions that don't have `wxallowed` set. This means we need to have a fairly well configured system up and running before we can even get to the phase of running ansible on it.

OpenBSD has built in support for automated installs using [autoinstall](http://man.openbsd.org/autoinstall), it would be super cool if we could use this, but I am not too hopeful, as it needs a tftp server and a http server to host config files like the following:

```
System hostname = iojs-openbsd
Password for root = *************
Public ssh key for root = ssh-ed25519 .....
Setup a user = iojs
Password for user = *************
Public ssh key for user = ssh-ed25519 ......
Allow root ssh login = prohibit-password
What timezone are you in = US/Mountain
Location of sets = http
HTTP Server = ftp3.usa.openbsd.org
Server directory = pub/OpenBSD/6.1/amd64
URL to autopartitioning template for disklabel = https://serverwithdisklabel.com/disklabel
```

### DIFF
--- a/setup/openbsd/README.md
+++ b/setup/openbsd/README.md
@@ -1,0 +1,31 @@
+# io.js Build OpenBSD Setup
+
+## Setting up
+
+To set up hosts, make sure you add them to your ssh config first:
+```
+Host test-digitalocean-openbsd61-x64-1
+  HostName 162.243.204.248
+  User openbsd
+
+Host test-digitalocean-openbsd61-x64-2
+  HostName 104.236.54.140
+  User openbsd
+```
+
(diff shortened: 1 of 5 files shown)


jbergstroem:
Apologies for the silence. Seeing ho we are migrating over to the "new" [ansible setup](https://github.com/nodejs/build/tree/master/ansible) it would be great if you could focus your efforts there instead.

Issue #752, opened 2017-06-09
requesting access for test-requireio-osx1010-x64-1


Trott:
I would like to be able to access test-requireio-osx1010-x64-1 so I can test some things in the hopes of identifying the problem in https://github.com/nodejs/node/issues/13527.

My public ssh keys can be found at https://github.com/Trott.keys

gibfahn:
SGTM

gibfahn:
@Trott would you mind PRing the update to the inventory if you've got things working?

I understand from IRC that you've got the IP and the keys, you just need the user (which I assume will be `root`).

Trott:
> I understand from IRC that you've got the IP and the keys, you just need the user (which I assume will be root).

`root` was my guess too but alas no...

gibfahn:
>root was my guess too but alas no...(or I have something else set up wrong)

How did you get the IP address?

@jbergstroem in general how do we get the IP of a machine?

rvagg:
@Trott it's not straightforward but this _should_ get you in:

```
Host jump-nodejs-arm
  User jump
  HostName vagg-arm.nodejs.org
  Port 2222
  IdentityFile ~/.ssh/nodejs_build_test
Host test-requireio-osx1010-x64-1
  HostName 192.168.2.210
  ProxyCommand ssh jump-nodejs-arm nc -q0 %h %p
  User iojs
  IdentityFile ~/.ssh/nodejs_build_test
```

rvagg:
sorry, that's for your .ssh/config

jbergstroem:
@gibfahn all machines should live in our inventory. This one doesn't.

gibfahn:
>@gibfahn all machines should live in our inventory. This one doesn't.

Yeah, I meant how do we get the IP to add to the inventory. Using one of these from the [Script Console](https://ci.nodejs.org/computer/test-requireio-osx1010-x64-1/script) does seem to work, but of course it gave `192.168.2.210`, which I assumed was wrong (didn't think of the jump box). 

```groovy
println InetAddress.localHost.canonicalHostName
// or
println InetAddress.localHost.hostAddress
```

gibfahn:
Followon question, at the moment the generated `~/.ssh/config` entries that use the jumpbox are of the form:

```
Host test-requireio_louiscntr-debian7-arm_pi2-1 
  HostName 192.168.2.68
  IdentityFile ~/.ssh/nodejs_build_test
  User pi
[...]

PR #753, opened 2017-06-10
ansible: add requireio-osx1010-x64-1 to inventory

gibfahn:
Fixes: https://github.com/nodejs/build/issues/752

Not really sure why the user is `iojs` rather than root here, is that intended?

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -40,6 +40,7 @@ hosts:
         andineck-debian7-arm_pi1p-1: {ip: 192.168.2.41, user: pi}
         xgene-ubuntu1404-arm64-1: {ip: 192.168.2.3, user: ubuntu, alias: arm-nfshost}
         rvagg-debian7-arm_pi1p-1: {ip: 192.168.2.40, user: pi}
+        osx1010-x64-1: {ip: 192.168.2.211, user: iojs}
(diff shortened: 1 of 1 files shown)


rvagg:
yeah, the root account isn't active on that machine, we could fix it up but for now let's just leave it as is cause we're migrating to MacStadium anyway. There's a user account "rvagg" that has admin privs, if need be I could probably stick the test key in there too. For full admin though you really need remote desktop, which you could do via SSH tunnel but you then need a password to log in and do anything useful. We're going to have to figure out an answer for this problem, it'll probably mean assigning the same password to an admin account for all of the test OSX machines. /cc @jbergstroem @mhdawson 

@gibfahn while you're in here, could you add another entry in there, under `release`, same name, IP 192.168.2.211 please

gibfahn:
>could you add another entry in there, under release, same name, IP 192.168.2.211 please

Done

rvagg:
lgtm

jbergstroem:
It might be worth mentioning that the create playbooks with fail now seeing how we don't have any support for macos builders. Hopefully this will not be the case any longer ni teh coming week or two, seeing how we now have a few machines up and running at macstadium we will fully ansible-fy (if possible).

PR #759, opened 2017-06-15
A bandaid for nodejs/node/issues/13667

refack:
As suggested in https://github.com/nodejs/node/issues/13667#issuecomment-308651232
I think it's worth considering


gibfahn:
I think if @rvagg releases an 8.1.2 today/tomorrow we'll be fine.

rvagg:
as @refack suggested on IRC, it doesn't have to be permanent, my main concern is with our serious redirect bloat but this can be a temporary one (don't know how long but everyone will move on from this pretty quick I reckon). I've implemented it for now.

Issue #723, opened 2017-05-16
update clang on FreeBSD 10


targos:
Ref: https://github.com/nodejs/node-v8/issues/1

Latest V8 cannot be compiled with version 3.4.1. We should try to update to 3.4.2 and see if that fixes the crash.

/cc @jbergstroem

jbergstroem:
Ok, I will open a PR tomorrow.

targos:
ping

jbergstroem:
Truly sorry :( $wrk, etc etc. I will get to it as soon as I can; likely tomorrow morning.

targos:
ping @nodejs/build (no worries @jbergstroem. I totally understand that you can be busy)

Trott:
@gibfahn Did you mean to add the `wg-agenda` label or is `ctc-agenda` correct? If `ctc-agenda`, can you (or someone else) provide a short description of what you hope CTC will do? Thanks!

joaocgreis:
Was `wg-agenda` and was already discussed in https://github.com/nodejs/build/issues/766

kfarnung:
I'm going to be taking a look at it this week.

kfarnung:
@targos I've taken a look at this and it's not too straightforward.  I have opened #772 for the first step, installing the `clang34` package, but there's a bit more subtlety to the builds on FreeBSD.  I'll keep working on this throughout the week.

As a side note, are the builds still failing with 3.4.1?  I don't see any indication of the original build issue in the CI or in my local builds.

targos:
@kfarnung Thank you for looking into this.
The builds are failing with V8 6.0. You can try to reproduce with this branch: https://github.com/targos/node/tree/v8-6.0

The latest failure on CI is here: https://ci.nodejs.org/job/node-test-commit-freebsd/9863/nodes=freebsd10-64/console

kfarnung:
Thanks @targos, that was the key.  I'm now able to repro the crash.  Unfortunately I'm still seeing the crash on 3.4.2, see attached, so I'm going to try 3.5 and 3.8 as well, just to see if any of the newer versions resolve the issue.

[freebsd10-clang34-output.txt](https://github.com/nodejs/build/files/1106655/freebsd10-clang34-output.txt)

kfarnung:
@jbergstroem Is there any objection to upgrading clang to 3.8.1?  That's approximately the same version as FreeBSD 11 (which uses 3.8.0).

jbergstroem:
[...]

PR #772, opened 2017-06-27
ansible: install clang38 on FreeBSD 10.x machines

kfarnung:
FreeBSD 10.x ships by default with clang-3.4.1, which is below the minimum for Node.js builds.  That version (as well as 3.4.2 and 3.5.2) crashes when attempting to build V8 6.0.  Specifically install the `clang38` package on those hosts.

#### Validation steps
- Installed FreeBSD 10 and 11 on VMs
- Added them to the local Ansible inventory
- Added fake `secret` entries for each machine
- Ran the `jenkins/worker/create` playbook against both hosts
- Verified that `clang38` is only installed on the FreeBSD 10 host
- Verified that the custom CC/CXX variables are only applied to the FreeBSD 10 host

### DIFF
--- a/ansible/roles/jenkins-worker/templates/freebsd.initd.j2
+++ b/ansible/roles/jenkins-worker/templates/freebsd.initd.j2
@@ -28,8 +28,8 @@ jenkins_env=" \
   NODE_TEST_DIR=/home/{{ server_user }}/tmp \
   PATH=/usr/local/libexec/ccache:/usr/local/bin:${PATH} \
   JOBS={{ server_jobs|default(ansible_processor_vcpus|default('2')) }} \
-  CC=cc \
-  CXX=c++"
+  CC={{ custom_cc_command|default('cc') }} \
+  CXX={{ custom_cxx_command|default('c++') }}"
 
 jenkins_jar="/home/{{ server_user }}/slave.jar"
 jenkins_log_file="/home/{{ server_user }}/${name}_console.log"
--- a/ansible/roles/jenkins-worker/templates/freebsd.initd.j2
+++ b/ansible/roles/jenkins-worker/templates/freebsd.initd.j2
@@ -28,8 +28,8 @@ jenkins_env=" \
   NODE_TEST_DIR=/home/{{ server_user }}/tmp \
   PATH=/usr/local/libexec/ccache:/usr/local/bin:${PATH} \
   JOBS={{ server_jobs|default(ansible_processor_vcpus|default('2')) }} \
-  CC={{ custom_cc_command|default('cc') }} \
-  CXX={{ custom_cxx_command|default('c++') }}"
+  CC={{ 'clang38' if os == 'freebsd10' else 'cc' }} \
+  CXX={{ 'clang++38' if os == 'freebsd10' else 'c++' }}"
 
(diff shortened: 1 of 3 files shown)


kfarnung:
/cc @nodejs/build 

I used #650 as a template, but it seems like there was some refactoring done since.  Are the playbooks under `setup/` still used?

kfarnung:
Looks like the CC/CXX variables are set in the init.d template: https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/templates/freebsd.initd.j2#L31

gibfahn:
>Are the playbooks under setup/ still used?

We're moving away from them as quickly as possible.

kfarnung:
@gibfahn Do you want me to attempt to update the older scripts?
@jbergstroem I took a stab at adding a config for the compilers and pushed an update.  I verified on my machine that it applies correctly on both FreeBSD 10 and 11 (doesn't change the latter).

As discussed in #723, it looks like `clang34` (3.4.2) and `clang35` (3.5.2) both crash when compiling V8 6.0.  Building with `clang38` (3.8.1) was successful.  Are we opposed to using `clang38` for FreeBSD 10 builds?

gibfahn:
>@gibfahn Do you want me to attempt to update the older scripts?

I think we should be okay just using the newer scripts, if you've tested building/testing node after using them then that's good enough for me (idk what @jbergstroem thinks though).

mhdawson:
I think we need to document somewhere which one should be used for which platforms.  For example the Raspberry Pi ones are not complete and I burned some time using the new one before realizing that.  Given that we can doc that somewhere prominent I'd agree that just have the latest be updated makes sense.

ansible/roles/jenkins-worker/templates/freebsd.initd.j2
```diff
@@ -28,8 +28,8 @@ jenkins_env=" \
   NODE_TEST_DIR=/home/{{ server_user }}/tmp \
   PATH=/usr/local/libexec/ccache:/usr/local/bin:${PATH} \
   JOBS={{ server_jobs|default(ansible_processor_vcpus|default('2')) }} \
-  CC=cc \
-  CXX=c++"
+  CC={{ custom_cc_command|default('cc') }} \
```
kfarnung: Not a problem, I initially started down that path and discovered the configuration option.  Figured it was worth the exploration.

My one concern with hard-coding into the template was that it could cause confusion down the road.  Ideally it would be nice to have the compiler command right along-side the package config that installs it.  An issue for another time, but that was the thinking.

jbergstroem: @kfarnung Thanks for elaborating on your thoughts; perhaps the config file is the better way - my rationale against it is that having it in the config file isn't alone "enough" to install and pass to CC -- if that is the case we would have to change the variables as well.

kfarnung: Yeah, I think a more complex change is necessary to do it "right" and my change wasn't quite there.  I've updated the PR to use your suggestion as it's the most straightforward approach.

kfarnung:
It sounds like this may no longer be relevant, feel free to close if that's the case.

kfarnung:
Closing this out, feel free to resurrect if necessary.

PR #784, opened 2017-07-07
[WIP] aix ansible setup scripts

BethGriggs:
Work in progress - looking at reducing the steps in [manualBootstrap.md](https://github.com/nodejs/build/blob/master/setup/aix61/manualBootstrap.md)

### DIFF
--- a/setup/aix61/aix-bootstrap.yml
+++ b/setup/aix61/aix-bootstrap.yml
@@ -1,0 +1,99 @@
+# Some RHEL machines require that the system is registered to Red Hat Subscription Management service before you can download packages.
+
+- hosts: all
+  remote_user: root
+  become: yes
+
+  tasks:
+  - block:
+
+  # TODO: Install yum if not already
+
+  #- name: "Change filesystem size"
+  #  shell: |
+  #    chfs -a size=+1300000 /opt
+  #    chfs -a size=+250000 /
+
+    - name: yum upgrade all packages
+      yum:
+        name: '*'
+        state: latest
+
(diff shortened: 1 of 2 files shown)


jbergstroem:
@BethGriggs hey, thanks for this! Seeing how we're moving away from the "old" ansible scripts (that lives in setup) it would be great if you could adapt this to the new structure that lives in "ansible".

gibfahn:
@jbergstroem yeah, I asked @BethGriggs to put this up as is so that people could look at it. The plan is to put it into `ansible/`.

jbergstroem:
seems to be an opportunity for reusing what we've done in other setups too, like ccache. I suspect the bootstrap will be the most different path for AIX.

gibfahn:
>seems to be an opportunity for reusing what we've done in other setups too, like ccache. I suspect bootstrap will be the most different path for AIX.

Yeah definitely. TBH it was easiest for testing to just put a single file together, reusing what we already have is the next step 😁 .

gdams:
We need to add the ramdisk setup to the ansible playbook too, shouldn't be too hard to add

maclover7:
ping @BethGriggs @gibfahn

BethGriggs:
I have not had time to progress this for quite a while, I think I'll close this PR. 

//cc @gibfahn or @nodejs/platform-aix who might want to pick up from where I got to

mhdawson:
I believe that @gibfahn was going to be looking at AIX ansible this week, but if so reasonable to open new PR.

PR #785, opened 2017-07-07
Create the NODE_TEST_DIR env variable on Ubuntu 16.04 and 16.10

piccoloaiutante:
This PR tries to adapt the previous https://github.com/nodejs/build/pull/658 to the new ansible structure that we adopted recently. 

Basically is adding an extra step for Ubuntu 16.04 and 16.10 to create the `tmp` directory used in `NODE_TEST_DIR` variable.

cc @jbergstroem @mcollina


jbergstroem:
any reason we don't do this for every setup?

mcollina:
The other setups already have that folder created.

jbergstroem:
> @mcollina said:
> The other setups already have that folder created.

Ah. We should probably move it to the jenkins role instead then.

piccoloaiutante:
@jbergstroem so creating another folder under `ansible/roles/jenkins-worker/tasks/partials` for this task and adding files only for ubuntu. Then adding this task to `main.yml` similarly to what has been done for `tap2junit` ?

piccoloaiutante:
@jbergstroem should be good ok now.

ansible/roles/bootstrap/tasks/partials/ubuntu1610.yml
```diff
@@ -16,4 +16,4 @@
 
 - name: install python and aptitude
   when: has_python.rc == 1 or has_aptitude.rc == 1
-  raw: apt install -y python-minimal aptitude
+  raw: apt install -y python-minimal aptitude
```
MylesBorins: was this removal of new line at end of file intentional?

ansible/roles/bootstrap/tasks/partials/ubuntu1604.yml
```diff
@@ -16,4 +16,4 @@
 
 - name: install python and aptitude
   when: has_python.rc == 1 or has_aptitude.rc == 1
-  raw: apt install -y python-minimal aptitude
+  raw: apt install -y python-minimal aptitude
```
MylesBorins: was this removal of new line at end of file intentional?

piccoloaiutante: No, let me fix it.

ansible/roles/jenkins-worker/tasks/main.yml
```diff
@@ -19,6 +19,9 @@
     state: present
     line: ::1 localhost.localdomain localhost
 
+- name: create NODE_TEST_DIR directory
+  file: path=/home/{{ server_user }}/tmp state=directory
```
gibfahn: Is this directory created by root, and if so does it have the right permissions/owner?

piccoloaiutante: yes is created by root:
```
drwxr-xr-x  2 root root 4.0K May  5 17:48 tmp
```

I guess I should add also write permission as it will be used by jenkin to run test.

gibfahn: I would have thought it should be chowned to `server_user`:`server_group` (assuming those vars exist).

piccoloaiutante: Yeah, you're right @gibfahn. I fixed it adding an extra setup. Now permission dir look like this:
```
drw-r-xr-x  2 iojs iojs 4.0K May  5 17:48 tmp
```

ansible/roles/jenkins-worker/tasks/main.yml
```diff
@@ -19,6 +19,16 @@
...
+
+- name: set NODE_TEST_DIR permission and owner
+  file:
+    path: /home/{{ server_user }}/tmp
+    owner: "{{ server_user }}"
+    group: "{{ server_user }}"
+    mode: 0655
```
gibfahn: I think it should be 0755 so that iojs can cd into it (see [StackOverflow](https://unix.stackexchange.com/questions/150449/what-does-execute-permission-on-a-folder-means#)). Not entirely sure though.

gibfahn:
I think we're good to land this, it's been sitting for ages and it keeps causing issues with releases.

Of course merging this doesn't mean the tmp directories will magically come back unless someone reruns the playbook on the machines.

PR #796, opened 2017-07-13
Ansible scripts for zos

mhdawson:
Opening for initial comments. Will update commit comment on first revision.

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -4,7 +4,7 @@ gathering           = explicit
 host_key_checking   = False
 inventory           = plugins/inventory/nodejs_yaml.py
 library             = plugins/library
-remote_user         = Unix1
+remote_user         = root
 retry_files_enabled = False
 roles_path          = roles
 squash_actions      = apk
@@ -31,6 +31,7 @@ ansible_python_interpreter = /opt/local/bin/python
 ansible_python_interpreter = /usr/local/bin/python
 
 [hosts:marist]
-remote_user         = Unix1
+ansible_become             = false
 ansible_python_interpreter = /NODEJS2/python-2017-04-12-py27/python27/bin/python
-home                = /u
+home                       = /u
+server_jobs: 4
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -4,7 +4,7 @@ gathering           = explicit
 host_key_checking   = False
 inventory           = plugins/inventory/nodejs_yaml.py
 library             = plugins/library
-remote_user         = root
+remote_user         = Unix1
 retry_files_enabled = False
 roles_path          = roles
 squash_actions      = apk
@@ -29,3 +29,8 @@ ansible_python_interpreter = /opt/local/bin/python
 
 [hosts:freebsd]
 ansible_python_interpreter = /usr/local/bin/python
+
+[hosts:marist]
+remote_user         = Unix1
+ansible_python_interpreter = /NODEJS2/python-2017-04-12-py27/python27/bin/python
+home                = /u
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -34,4 +34,3 @@ ansible_python_interpreter = /usr/local/bin/python
 ansible_become             = false
 ansible_python_interpreter = /NODEJS2/python-2017-04-12-py27/python27/bin/python
 home                       = /u
-server_jobs: 4
--- a/ansible/group_vars/test
+++ b/ansible/group_vars/test
@@ -1,3 +1,4 @@
 jenkins_url: "https://ci.nodejs.org"
 jenkins_worker_jar: "{{ jenkins_url }}/jnlpJars/slave.jar"
 server_user: iojs
+home: "/users"
--- a/ansible/group_vars/test
+++ b/ansible/group_vars/test
@@ -1,5 +1,5 @@
 jenkins_url: "https://ci.nodejs.org"
 jenkins_worker_jar: "{{ jenkins_url }}/jnlpJars/slave.jar"
 server_user: iojs
-home: "/users"
+home: "/home"
 remote_env: {}
--- a/ansible/group_vars/test
+++ b/ansible/group_vars/test
@@ -2,3 +2,4 @@ jenkins_url: "https://ci.nodejs.org"
 jenkins_worker_jar: "{{ jenkins_url }}/jnlpJars/slave.jar"
 server_user: iojs
 home: "/users"
+remote_env: {}
--- a/ansible/host_vars/test-marist-zos13-s390x-template
+++ b/ansible/host_vars/test-marist-zos13-s390x-template
@@ -1,0 +1,21 @@
+secret:
+remote_env:
+  FFI_LIB: /NODEJS2/python-2017-04-12-py27/python27/lib/ffi
+  PYTHON_ENV: python27
+  PATH: /NODEJS2/python-2017-04-12-py27/python27/bin:/bin:/NODEJS2/bin:/NODEJS/bin
+  GIT_EXEC_PATH: /NODEJS/libexec/git-core
+  BASH_PREFIX: /NODEJS2/bin/bash
+  GIT_SHELL: /NODEJS/bin/bash
+  CURL_CA_BUNDLE: /NODEJS2/python-2017-04-12-py27/python27/etc/ssl/cacert.pem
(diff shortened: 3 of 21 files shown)


mhdawson:
Once question is who to change the remote_user properly as what is in the commit is obviously not right for other platforms.

mhdawson:
These are now good enough so that we can run libuv jobs on zOS.  See https://ci.nodejs.org/job/libuv-test-commit-zos/92/

@jbergstroem any suggestions on a better way to change the remote_user ?

jbergstroem:
Sorry for the silence. Will review shortly.

mhdawson:
@jbergstroem thanks, I think we are about ready to image additional machines and would be great to land before then.

joaocgreis:
Some notes:
- The `remote_user` issue must be solved before landing, as landing it in its current form will break every other system. Strange that the override in the section does not work.
- Is there a mechanism to restart Jenkins if it crashes (monit/...)? This can be done later, but please consider it as it's quite important to keep the server reliably online.
- Is `home` needed in the `group_vars` when it is defined in `ansible.cfg`?

Other than that rubber stamp LGTM. I didn't test this but nothing looks wrong.

joaocgreis:
@mhdawson does a `user` tag in the inventory also not work? Like this:

https://github.com/nodejs/build/blob/1512c6c1c583e188eb04b1fd880fcc639f8e5d1e/ansible/inventory.yml#L40

mhdawson:
@joaocgreis thanks for the comments will rebase and look to see how I can address them.

mhdawson:
I've rebased

The suggestion from @joaocgreis allowed me to resolve the remote_user issue with one additional tweak.  Thanks!

I also did some other small clean and addressed an issue I believe would have affected other platforms.

ansible/host_vars/test-marist-zos13-s390x-template
```diff
@@ -0,0 +1,21 @@
...
+  RELEASE_NAME: python-2017-04-12
+  _CEE_RUNOPTS: POSIX(ON)
+  STEPLIB: none
+  PERL_PREFIX: /NODEJS2/bin/perl
+  PKG_CONFIG_PATH: /NODEJS2/python-2017-04-12-py27/python27/lib/pkgconfig:/NODEJS2/python-2017-04-12-py27/python27/share/pkg
+  LIBPATH: /lib:/usr/lib:.:/NODEJS2/lib/perl5/5.24.0/os390/CORE.pod:/NODEJS2/python-2017-04-12-py27/python27/lib/
+  RELEASE_DIR: /NODEJS2/python-2017-04-12-py27
+  RELEASE_TYPE: py27
+  GIT_TEMPLATE_DIR: /NODEJS/share/git-core/templates
+server_jobs: 4
```
joaocgreis: For this file to exist, add a line with
```
!host_vars/*-template
```
to https://github.com/nodejs/build/blob/master/ansible/.gitignore

ansible/ansible.cfg
```diff
@@ -28,3 +28,9 @@ ansible_python_interpreter = /opt/local/bin/python
 
 [hosts:freebsd]
 ansible_python_interpreter = /usr/local/bin/python
+
+[hosts:marist]
+ansible_become             = false
+ansible_python_interpreter = /NODEJS2/python-2017-04-12-py27/python27/bin/python
+home                       = /u
+server_jobs: 4
```
joaocgreis: This like is in the wrong format. I see it is in the host_vars template, was it forgotten here and should be removed?

ansible/group_vars/test
```diff
@@ -1,3 +1,5 @@
 jenkins_url: "https://ci.nodejs.org"
 jenkins_worker_jar: "{{ jenkins_url }}/jnlpJars/slave.jar"
 server_user: iojs
+home: "/users"
```
joaocgreis: Can this be moved to the host_vars template? Also from `release`.

ansible/roles/jenkins-worker/tasks/main.yml
```diff
@@ -5,22 +5,25 @@
...
 - name: add ::1 to /etc/hosts for ipv6 compat
+  when: not os|startswith("zos")
   lineinfile:
     dest: /etc/hosts
     state: present
     line: ::1 localhost.localdomain localhost
 
 - name: prepare installing tap2junit
-  when: type != "release"
+  when: type != "release" and not os|startswith("zos")
```
joaocgreis: Perhaps I'm missing something, but there's a partial added for this for zos. So, is it added but not run for some reason or run somewhere else?

ansible/host_vars/test-marist-zos13-s390x-template
gibfahn: Having templates sounds like a great idea in general, so +1 to this.

refack:
@mhdawson if you need `GYP` I keep my mirror fairly updated https://github.com/refack/gyp

ansible/host_vars/test-marist-zos13-s390x-template
mhdawson: added

mhdawson:
@refack, I've updated to pull from the ci master were we have setup a mechanism to store artifacts.

ansible/group_vars/test
mhdawson: If I move it out then every time somebody creates as host_vars for other platforms they would have to add that (we currently only have a template for s390x).  It is the default to be used in cases other than z/OS, and its overridden in the ansible.cfg for the z/OS platform.  

I think its better to keep it here than having people need to remember to add it when creating their host_vars files...

ansible/roles/jenkins-worker/tasks/main.yml
mhdawson: I think I added it hoping it could install tap2junit, but did not manage to get it working.  I've removed the partial for now, and will revisit when we come back to get the parts needed for build Node.js working (right now I've focused on what was needed to build libuv)

mhdawson:
@joaocgreis pushed changes to address your latest comments. Please take a look.

I've used it to configure the 2 z/OS machines connected to the CI and the libuv job is running nightly (still red though as we have 2 test failures we are working on)

ansible/group_vars/test
joaocgreis: I was reading this wrong before, in that case this is the correct palce. But shouldn't this be `/home` instead?

If I'm reading this right, your objective is to replace references to `/home` with `{{ home }}`, which sounds good to me. But the default should be what was there before replacing, that is `/home` not `/users`. At least the `download slave.jar` step will fail on all other platforms with this line (unless I'm still missing something).

mhdawson:
@joaocgreis @gibfahn any chance you can review so that I can get this landed.

joaocgreis:
@mhdawson I left a comment inline above, is the only thing blocking this now for me: https://github.com/nodejs/build/pull/796#discussion_r141757097

ansible/group_vars/test
mhdawson: Thanks for catching that, I'm not sure where I came up with users as it should be home.  Pushed commit to fix.

mhdawson:
@joaocgreis pushed commit to fix your last comment.

ansible/roles/jenkins-worker/tasks/main.yml
gibfahn: Maybe put a `#TODO(mhdawson): get tap2junit working on zOS` for future reference (so we know it's outstanding).

gibfahn:
Given how much zOS specific stuff we now have in `main.yml`, would it be worth spinning that out into a partial? Not sure how easy that is with ansible, and not a blocking request, more a question.

mhdawson:
Point taken about trying to pull the zOS stuff out, but would like that to be a future TODO.  The comment I'll add.  Just going to wait to see if @joaocgreis has any other comments to I can do the final pass together if possible.

joaocgreis:
@gibfahn I started looking into moving Windows to `ansible/` and I suspect we should probably split `baselayout` into partials like `bootstrap` already is, with a difference: it can have a `default`, so that only `windows` and `aix` need to be exceptions. Still not 100% sure of this though, having groups of subtasks or splitting `baselayout` into smaller roles might make some sense. The best way forward is probably for me (or someone else) to come up with a proposal when time permits.

mhdawson:
Landed as 96af6dc55f0baa39db6ef9e8af6e8ae36c56133d

Issue #513, opened 2016-10-11
Add download testing 


mhdawson:
After a recent release and issue was opened identifying that the tar.gz was corrupted. There were suggestions in https://github.com/nodejs/node/issues/8814 that we could have an automated job/process that automatically checks the download binaries

mhdawson:
@gibfahn @richardlau FYI

jbergstroem:
How about adding a step after creating it instead? something like:
```console
# either
$ gunzip -t $foo.tar.gz && echo $?   # xz supports -t as well
# ..or
$ tar -zxf $foo.tar.gz &> /dev/null && echo $?
```

mhdawson:
This looks like a similar issue that we did not notice for smartos as well https://github.com/nodejs/build/issues/585.

Having a regular job that validates we can download/unzip all of the current releases might help.

gibfahn:
Do we generate md5 checksums for the builds? Seems like something that might be worth doing if we don't.

gdams:
@mhdawson I'd be happy to quickly script one of these up. I wrote a similar job to automate checking the md5 checksums of the IBM Node releases

mhdawson:
@gdams that would be great.  Once you have the script we can look at setting up the ci to run it.  

I can think of a couple of different levels we could valid to:

1) Download and validate unzip works
2) Actually run "hello world", this would mean we'd need to download/unzip on the machine where the version can run.

2) would be great, but 1 would also go a long way to making sure things are ok.

gdams:
Okay so to start with I have written a simple bash tool that will scrape the download dir for a certain node version and check all of the sha sums. Where abouts would it make sense to store this script?

gibfahn:
@gdams I'd put it in [`tools/`](https://github.com/nodejs/node/tree/master/tools)

mhdawson:
Ok the script rocks. I've updated the CI job so that it can be used to:

Run tests on all current shipping releases (StreamandVersion = release)
Run tests on all current shipping releases and nightly releases (StreamAndVersion = latest)
Run tests on the lastest of a particular stream (ex v6) either shipping or nightly (StreamAndVersion = vX, were X is the stream), use 2nd selection box to select Nightly or Release
Run tests on a specific level (shipping or nightly) (StreamAndVersion = vY, were Y is the specific version ex v8.4.1), use 2nd selection box to select Nightly or Release

https://ci.nodejs.org/view/All/job/validate-downloads/

Have also configured to run nightly.

[...]

PR #804, opened 2017-07-19
tools: add download testing

gdams:
This script will take an environment variable (NODE_VERSION) and
download all of the files from `rsync://unencrypted.nodejs.org/nodejs/release/v<NODE_VERSION>/`
It will then loop through the `SHASUMS256.txt` file to check if the
SHA256SUMS match up. Initial issue here: https://github.com/nodejs/build/issues/513

### DIFF
--- a/jenkins/download-test.sh
+++ b/jenkins/download-test.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+# Pass `NODE_VERSION` as an environment variable. Can be an exact or vague version.
+# NODE_VERSION=v8 ./download-test.sh --tap ./test.tap
+
+while [ $# -gt 1 ]; do
+
+  case "$1" in
+      -t|--tap)
+      tapFile="$2"
+      shift # Past option argument.
+      ;;
+      *) # unknown option
+      echo "Ignoring unknown option: $1"
+      ;;
+  esac
+  shift # Past option.
+done
+
+if [ "$DOWNLOAD_LOCATION" == "Nightly" ]; then
+  downloadDir="nightly"
+else
+  downloadDir="release"
+fi
+
+if [ -z "$NODE_VERSION" ]; then
+  echo "NODE_VERSION is undefined! Please declare NODE_VERSION"
+  exit 1
+fi
+
+# If the provided NODE_VERSION is inexact, e.g. `v8`, get the most recent matching version.
+LINK=`rsync rsync://unencrypted.nodejs.org/nodejs/$downloadDir/ | grep $NODE_VERSION | sort -t. -k 1,1n -k 2,2n -k 3,3n | tail -1 | awk '{print $5}'`
+
+# Remove old files
+rm -rf v*
+[ "$tapFile" ] && rm -rf "$tapFile" && echo "TAP version 13" > $tapFile
+
+# Scrape the download server
+rsync -av rsync://unencrypted.nodejs.org/nodejs/"$downloadDir"/"$LINK" .
+
+testNumber=0
+cat "$LINK"/SHASUMS256.txt | {
+  while read line; do
+   let testNumber++ # Increment at the beginning
+   sha=$(echo "${line}" | awk '{print $1}')
+   file=$(echo "${line}" | awk '{print $2}')
+   echo "Checking shasum for $file"
+   calculatedSHA=$(shasum -a 256 "$LINK/$file" | awk '{print $1}')
+   if [ "$calculatedSHA" != "$sha" ]; then
+     echo "Error - SHASUMS256 does not match for $file"
+     echo "Expected - $sha"
+     echo "Found - $calculatedSHA"
+     if [ "$tapFile" ]; then
+       echo "not ok $testNumber SHASUMS256 does not match for $file
+         ---
+           Expected - $sha
+           Found - $remoteSHA
+         ...
+       " >> "$tapFile"
+     fi
+   else
+     echo "Pass - SHASUMS256 is correct for $file"
+     [ "$tapFile" ] && echo "ok $testNumber SHASUMS256 for $file is correct" >> "$tapFile"
+   fi
+  done
+[ "$tapFile" ] && echo "1..$testNumber" >> "$tapFile"
+}


tools/download-test.sh
```diff
@@ -0,0 +1,50 @@
...
+if [ -z ${NODE_VERSION} ]; then
+  echo "NODE_VERSION is undefined! Please declare NODE_VERSION"
+  exit 1
+fi
+# Remove old files
+[ ! -z ${TAP} ]; rm -r ${TAP} || true
+#rm -r nodejs.org || true
+
+# Scrape the download server
+#wget -r --no-parent http://nodejs.org/dist/v${NODE_VERSION}/
```
mhdawson: Should this be commented out ?

tools/download-test.sh
```diff
@@ -0,0 +1,50 @@
...
+  shift # past argument or value
+done
+
+if [ -z ${NODE_VERSION} ]; then
+  echo "NODE_VERSION is undefined! Please declare NODE_VERSION"
+  exit 1
+fi
+# Remove old files
+[ ! -z ${TAP} ]; rm -r ${TAP} || true
+#rm -r nodejs.org || true
```
mhdawson: believe this needs to be -rf to avoid failures if the directory does not already exist.

gdams: the || true should prevent that I think

tools/download-test.sh
mhdawson: It gets further if I uncomment that but still fails later on:


iojs@nodejsb1:~/build/workspace/validate-downloads/temp$ sh build/tools/download-test.sh
build/tools/download-test.sh: 3: build/tools/download-test.sh: [[: not found
rm: missing operand
Try 'rm --help' for more information.
--2017-07-19 14:25:26--  http://nodejs.org/dist/v8.1.4/
Resolving nodejs.org (nodejs.org)... 104.20.22.46, 104.20.23.46, 2400:cb00:2048:1::6814:172e, ...
Connecting to nodejs.org (nodejs.org)|104.20.22.46|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: unspecified [text/html]
Saving to: 'nodejs.org/dist/v8.1.4/index.html'

    [ <=>                                                                                                 ] 5,355       --.-K/s   in 0s

2017-07-19 14:25:26 (14.5 MB/s) - 'nodejs.org/dist/v8.1.4/index.html' saved [5355]

Loading robots.txt; please ignore errors.
--2017-07-19 14:25:26--  http://nodejs.org/robots.txt
Reusing existing connection to nodejs.org:80.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://nodejs.org/robots.txt [following]
--2017-07-19 14:25:26--  https://nodejs.org/robots.txt
Connecting to nodejs.org (nodejs.org)|104.20.22.46|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: unspecified [text/plain]
Saving to: 'nodejs.org/robots.txt'

    [ <=>                                                                                                 ] 112         --.-K/s   in 0s

2017-07-19 14:25:26 (30.8 MB/s) - 'nodejs.org/robots.txt' saved [112]

FINISHED --2017-07-19 14:25:26--
Total wall clock time: 0.05s
Downloaded: 2 files, 5.3K in 0s (14.7 MB/s)
cat: nodejs.org/dist/v8.1.4/SHASUMS256.txt: No such file or directory
build/tools/download-test.sh: 48: build/tools/download-test.sh: let: not found
build/tools/download-test.sh: 49: build/tools/download-test.sh: cannot create : Directory nonexistent
iojs@nodejsb1:~/build/workspace/validate-downloads/temp$

gdams: good spot

gdams:
updated PTAL

mhdawson:
Started building the CI job, noted a few issues in earlier comments.

What I think would also be useful is if it could take an option like "latest" which would figure out the latest for a give release line.  That we won't have to update the job as new releases come out.  Generally I'm thinking we'll only need to be testing the latest.

mhdawson:
still getting this
```
cat: nodejs.org/dist/v8.1.4/SHASUMS256.txt: No such file or directory
./build/tools/download-test.sh: line 49: ${TAP}: ambiguous redirect
```

gdams:
@mhdawson should be fixed now if you want to re-run?

gdams:
CI job here for reference: https://ci.nodejs.org/view/All/job/validate-downloads/

gireeshpunathil:
I think the robot.txt is not allowing recursive download. We may need to parse the index.html to get links individually and then download them.
```
bash-4.1$ wget -r -d  http://nodejs.org/dist/v8.1.4/
...
Deciding whether to enqueue "http://nodejs.org/dist/v8.1.4/node-v8.1.4-x86.msi".
Rejecting path dist/v8.1.4/node-v8.1.4-x86.msi because of rule “dist/”.
Not following http://nodejs.org/dist/v8.1.4/node-v8.1.4-x86.msi because robots.txt forbids it.
Decided NOT to load it.
```

```
bash-4.1$ cat nodejs.org/robots.txt 
User-Agent: *
Disallow: /dist/
Disallow: /docs/
Allow: /dist/latest/
Allow: /dist/latest/docs/api/
Allow: /api/
bash-4.1$ 
```

gireeshpunathil:
ok, the latest script line
`wget -r --no-parent --execute="robots = off" http://nodejs.org/dist/v${NODE_VERSION}/`
seem to have addressed the robot block.

tools/download-test.sh
```diff
@@ -0,0 +1,50 @@
...
+if [ -z ${NODE_VERSION} ]; then
+  echo "NODE_VERSION is undefined! Please declare NODE_VERSION"
+  exit 1
+fi
+# Remove old files
+[ ! -z ${TAP} ]; rm -r ${TAP} || true
+rm -r nodejs.org || true
+
+# Scrape the download server
+wget -r --no-parent --execute="robots = off" http://nodejs.org/dist/v${NODE_VERSION}/
```
rvagg: don't use `/dist`, use `/download/release` instead, that's the new form that allows for more flexibility and we've been trying to transition

gdams: The issue with using `/download/release` is that this link enforces https wheras `/dist` seems to work with http. This can be an issue on machines that don't have ssl enabled for wget

rvagg: The plan is to enforce strict https on that server soon anyway, so you might be best to adapt to it now anyway.

Perhaps it might be best to go the rsync route off unencrypted:

```
rsync rsync://unencrypted.nodejs.org
```

gdams: good idea @rvagg.. updated PTAL

mhdawson:
Ok, so seems to run but I am seeing this 

./build/tools/download-test.sh: line 36: **[: !=: unary operator expected**

@gdams I'd also like to be able to be able to specify something like 8.latest and have it figure out the latest version.

rvagg:
Worth noting that unencrypted.nodejs.org is currently on a 15 minute cron to update its binaries from the main server, that may need to be taken into account if this tool is for automating something close to release. The 15 minute number is arbitrary and could be changed if needed fwiw, it's even possible that we could use a flag mechanism to trigger a sync when a release is promoted so it's near instant, although that's a bit more complicated obviously.

gibfahn:
@gdams when I said tools in https://github.com/nodejs/build/issues/513#issuecomment-316216447 I linked to the `tools/` directory in `nodejs/node`. I don't think it makes sense to start keeping scripts in this repo. so could you move this PR over there?

tools/download-test.sh
```diff
@@ -0,0 +1,50 @@
...
+rsync -av rsync://unencrypted.nodejs.org/nodejs/release/v${NODE_VERSION} .
+
+num=1
+cat v${NODE_VERSION}/SHASUMS256.txt | { while read line
+do
+   sha=$(echo "${line}" | awk '{print $1}')
+   file=$(echo "${line}" | awk '{print $2}')
+   echo "Checking shasum for $file"
+   remoteSHA=$(shasum -a 256 "nodejs.org/dist/v${NODE_VERSION}/${file}" | awk '{print $1}')
+   if [ ${remoteSHA} != ${sha} ]; then
```
gibfahn: @mhdawson is seeing the `unary operator` error because you're not quoting your variables, I suggest dropping the `{}` where they're not needed (everywhere in this script) and putting vars in double quotes everywhere.

tools/download-test.sh
```diff
@@ -0,0 +1,50 @@
...
+     echo "Error - SHASUMS256 does not match for ${file}"
+     echo "Expected - ${sha}"
+     echo "Found - ${remoteSHA}"
+     [ ! -z "${TAP}" ] && echo "not ok ${num} SHASUMS256 does not match for ${file}" >> ${TAP}
```
gibfahn: You can remove the `! -z`

gdams:
@gibfahn I was asked by michael to move to over to `nodejs/build`

gibfahn:
@gdams ahh, don't know how I missed that PR. I'll comment over there.

gdams:
@mhdawson @gibfahn updated PTAL. You can now either pass in a full node version (v6.11.2) or just (v6) and it will fetch the latest 6 build. You can also export `DOWNLOAD_LOCATION="nightly"` and fetch the latest nightly builds

mhdawson:
Ok the script rocks.  I've updated the CI job so that it can be used to:

Run tests on all current shipping releases
Run tests on all current shipping releases and nightly releases
Run tests on the lastest of a particular stream (ex v6) either shipping or nightly
Run tests on a specific level (shipping or nightly)
https://ci.nodejs.org/view/All/job/validate-downloads/

mhdawson:
The other thing we should setup is who gets emails on failures, I'm thinking the release team but we probably need to check with them if they are ok/want that.

mhdawson:
Configured to run nightly.

mhdawson:
@rvagg @jbergstroem any input on whether you think this best fits in the build or core node repo ?

rvagg:
Seems like a reasonable case could be made either way. If it's in nodejs/node then others could run it independently to confirm.

@mhdawson can you explain how the CI job runs? I can't quite work it out. It runs on a schedule but does that mean it's grabbing the latest version every hour to confirm? Also, you may want to make it run on `jenkins-workspace` because those machines are perfect for this kind of work.

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+
+LINK=`rsync rsync://unencrypted.nodejs.org/nodejs/$DOWNLOAD_DIR/ | grep $NODE_VERSION | sort -t. -k 1,1n -k 2,2n -k 3,3n | tail -1 | awk '{print $5}'`
+
+if [ -z "$NODE_VERSION" ]; then
+  echo "NODE_VERSION is undefined! Please declare NODE_VERSION"
+  exit 1
+fi
+
+# Remove old files
+[ "$TAP" ] && rm -r "$TAP" || true
```
rvagg: an `-f` might be good here, and on the next line too, otherwise they show up as errors

rvagg:
just saw the description in https://github.com/nodejs/build/issues/513 about how the script works, fine by me.

I've changed restrict nodes from `ubuntu1404-64 ||  benchmark` to `jenkins-workspace` which is serviced by two packet.net boxes with big 'ol disk.

tools/download-test.sh
rvagg: See https://ci.nodejs.org/view/All/job/validate-downloads/46/console, it looks like a bunch of bad errors

the `|| true` could be removed too once `-f` is in, then it's quieter

gdams: updated PTAL

rvagg:
lgtm, I don't mind where it lands, someone else can have a strong opinion cause I don't

rvagg:
Also, well done @gdams, good work! I didn't see the utility of this initially but I _get it_ now.

mhdawson:
@rvagg thanks for the change to where it runs, I was wondering what the best machines would be.

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+while [[ $# -gt 1 ]]
```
gibfahn: nit: `[[` -> `[`, and could you put the `do` on the same line (for consistency with the `if`s elsewhere)

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+do
+  key="$1"
+
+  case "$key" in
```
gibfahn: Maybe just use `$1` here and drop `$key`.

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+# Remove old files
+[ "$TAP" ] && rm -rf "$TAP"
+rm -rf v*
+
+# Scrape the download server
+rsync -av rsync://unencrypted.nodejs.org/nodejs/"$DOWNLOAD_DIR"/"$LINK" .
+
+num=1
+cat "$LINK"/SHASUMS256.txt | { while read line
+do
```
gibfahn: Maybe just me, but I find this layout really confusing, could you do (change indentation):

```bash
testNumber=1
cat "$LINK"/SHASUMS256.txt | {
  while read line; do
    sha=$(echo "${line}" | awk '{print $1}')
    ...
  done
  let testNumber--
  [ "$tapFile" ] && echo "1..$testNumber" >> "$TAP"
}
```

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+while [[ $# -gt 1 ]]
+do
+  key="$1"
+
+  case "$key" in
+      -t|--tap)
+      TAP="$2"
```
gibfahn: `tapFile`

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+  shift # past argument or value
+done
+
+if [ "$DOWNLOAD_LOCATION" == "Nightly" ]; then
+	DOWNLOAD_DIR="nightly"
+else
+	DOWNLOAD_DIR="release"
+fi
+
+LINK=`rsync rsync://unencrypted.nodejs.org/nodejs/$DOWNLOAD_DIR/ | grep $NODE_VERSION | sort -t. -k 1,1n -k 2,2n -k 3,3n | tail -1 | awk '{print $5}'`
```
gibfahn: `LINK` is really `nodeVersion` right? Could you add a comment along the lines of:

```bash
# If the provided NODE_VERSION is inexact, e.g. `v8`, get the most recent matching version.
```

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+
+if [ -z "$NODE_VERSION" ]; then
+  echo "NODE_VERSION is undefined! Please declare NODE_VERSION"
```
gibfahn: Maybe move this check above the rsync line where you use `NODE_VERSION`

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+# Scrape the download server
+rsync -av rsync://unencrypted.nodejs.org/nodejs/"$DOWNLOAD_DIR"/"$LINK" .
+
+num=1
+cat "$LINK"/SHASUMS256.txt | { while read line
+do
+   sha=$(echo "${line}" | awk '{print $1}')
+   file=$(echo "${line}" | awk '{print $2}')
+   echo "Checking shasum for $file"
+   remoteSHA=$(shasum -a 256 "$LINK/$file" | awk '{print $1}')
```
gibfahn: Maybe `calculatedSHA`

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+   if [ "$remoteSHA" != "$sha" ]; then
+     echo "Error - SHASUMS256 does not match for $file"
+     echo "Expected - $sha"
+     echo "Found - $remoteSHA"
+     [ "${TAP}" ] && echo "not ok $num SHASUMS256 does not match for $file" >> "$TAP"
+     return 1
```
gibfahn: Does this work? I'd expect `return: can only 'return' from a function or sourced script`.

Assuming you meant `break`, do we want to bail here? I think we probably want to calculate all the SHAs, otherwise if the first one is wrong it'll bail out, and we won't know how many were wrong.

So maybe just drop this line.

I assume this script will mostly be run through Jenkins, in which case we'll set it to fail on a failed TAP result. If you want to make sure the script returns `1` if a test fails, maybe have a variable:

```bash
  exitStatus=0 # 1 means tests failed.
  while ...; do
    ...
    if [ error ]; then
      exitStatus=1
    fi
  done
}
exit $exitStatus
```
 
If we do this we should make sure the TAP gets parsed in Jenkins even if there are failures though.

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+do
+   sha=$(echo "${line}" | awk '{print $1}')
+   file=$(echo "${line}" | awk '{print $2}')
+   echo "Checking shasum for $file"
+   remoteSHA=$(shasum -a 256 "$LINK/$file" | awk '{print $1}')
+   if [ "$remoteSHA" != "$sha" ]; then
+     echo "Error - SHASUMS256 does not match for $file"
+     echo "Expected - $sha"
+     echo "Found - $remoteSHA"
+     [ "${TAP}" ] && echo "not ok $num SHASUMS256 does not match for $file" >> "$TAP"
```
gibfahn: You probably want to show the expected and found lines here too (in yaml blocks).

```bash
echo "not ok $num SHASUMS256 does not match for $file
  ---
     echo "Expected - $sha"
     echo "Found - $remoteSHA"
  ...
" >> "$TAP"
```

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+     return 1
+   else
+     echo "Pass - SHASUMS256 for $file is correct"
+     [ "$TAP" ] && echo "ok $num SHASUMS256 for $file is correct" >> "$TAP"
```
gibfahn: `SHASUMS256 for $file is correct` -> `SHASUMS256 is correct for $file` to match the `not ok` case.

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+fi
+
+# Remove old files
+[ "$TAP" ] && rm -rf "$TAP"
+rm -rf v*
+
+# Scrape the download server
+rsync -av rsync://unencrypted.nodejs.org/nodejs/"$DOWNLOAD_DIR"/"$LINK" .
+
+num=1
```
gibfahn: Maybe do:

```bash
testNumber=0

while ...; do
let testNumber++ # Increment at the beginning
...
done
# Don't need let testNumber--
[ "$tapFile" ] && echo "1..$testNumber" >> "$TAP" # Unchanged
```

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+      *) # unknown option
+      ;;
+  esac
+  shift # past argument or value
+done
+
+if [ "$DOWNLOAD_LOCATION" == "Nightly" ]; then
+	DOWNLOAD_DIR="nightly"
+else
+	DOWNLOAD_DIR="release"
```
gibfahn: Seems like you're going for `ALL_CAPS` for env vars, and `camelCase` for local vars, so `downloadDir` (also remove tabs).

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
```
gibfahn: Could you add a comment with example usage, e.g.:

```bash
# Pass `NODE_VERSION` as an environment variable. Can be an exact or vague version.
# NODE_VERSION=v8 ./download-test.sh --tap ./test.tap
```

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+LINK=`rsync rsync://unencrypted.nodejs.org/nodejs/$DOWNLOAD_DIR/ | grep $NODE_VERSION | sort -t. -k 1,1n -k 2,2n -k 3,3n | tail -1 | awk '{print $5}'`
+
+if [ -z "$NODE_VERSION" ]; then
+  echo "NODE_VERSION is undefined! Please declare NODE_VERSION"
+  exit 1
+fi
+
+# Remove old files
+[ "$TAP" ] && rm -rf "$TAP"
```
gibfahn: maybe add an `&& echo "TAP version 13" > $tapFile` here, apparently TAP is supposed to start with that these days.

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+while [[ $# -gt 1 ]]
+do
+  key="$1"
+
+  case "$key" in
+      -t|--tap)
+      TAP="$2"
+      shift # past argument
+      ;;
+      *) # unknown option
```
gibfahn: We should at least warn for unknown options:

```bash
*)
echo "Ignoring unknown option: $1"
;;
```

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+while [[ $# -gt 1 ]]
+do
+  key="$1"
+
+  case "$key" in
+      -t|--tap)
+      TAP="$2"
+      shift # past argument
```
gibfahn: `# Past option argument.`

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+      ;;
+      *) # unknown option
+      ;;
+  esac
+  shift # past argument or value
```
gibfahn:
```bash
# Past option.
```

tools/download-test.sh
```diff
@@ -0,0 +1,59 @@
...
+done
+
+if [ "$DOWNLOAD_LOCATION" == "Nightly" ]; then
+	DOWNLOAD_DIR="nightly"
```
gibfahn: I think you somehow have tabs here.

gibfahn:
>Seems like a reasonable case could be made either way. If it's in nodejs/node then others could run it independently to confirm.

Yeah it's a tricky one. Would be good to get opinions from more people in @nodejs/build.

joaocgreis:
My opinion on this is that `nodejs/node` is **not** the place for this. When we checkout a commit from `nodejs/node`, everything there should relate to that exact version, and be depended on by something there.

The way I would have done this would be directly in a Jenkins job script. We have backups of the server, so it would be safe. However, it would not be public and easily reviewable, but that again links to the discussion about reviewing the jobs source. Since this is already done as a script, I don't see why not land it here.

mhdawson:
@gibfahn I thinking from your +1 on @joaocgreis' comment that you are now ok with landing this this repo ? If so I assume we just need @gdams to address your latest review comments and we'll be good to land ?

gibfahn:
@joaocgreis 

>My opinion on this is that nodejs/node is not the place for this. When we checkout a commit from nodejs/node, everything there should relate to that exact version, and be depended on by something there.

I don't think this applies to a lot of things in [nodejs/node:tools/](https://github.com/nodejs/node/tree/master/tools). 

>The way I would have done this would be directly in a Jenkins job script. We have backups of the server, so it would be safe. However, it would not be public and easily reviewable, but that again links to the discussion about reviewing the jobs source.

Yeah I think putting things in a Jenkins script is the worst possible option (we can't review, there may be backups but in my experience they aren't as easy to version control as git, it's hidden by default).

@mhdawson 

I'm -0 on it, yes. If we're going to do this I'd like us to decide where we're going to put these kinds of scripts, and maybe move more out of Jenkins configs and into this repo. IDK if `tools/` is the best name for this folder in that case, but no strong opinions.

Given that we have a meeting on Tuesday, we could probably talk about this there (in particular the general question of where files like this should go). I'll add it to the agenda.

joaocgreis:
> I don't think this applies to a lot of things in [nodejs/node:tools/](https://github.com/nodejs/node/tree/master/tools). 

In my opinion, it should. But I have no strong feelings about this.

Let's discuss at the meeting.

gibfahn:
@gdams consensus was to create a `nodejs/build:/jenkins/` folder to put this in, as we might want to move more scripts in here going forwards.

If you could fix up my suggestions and change the folder, this should be good to go.

gdams:
@gibfahn updated PTAL

jenkins/download-test.sh
```diff
@@ -0,0 +1,66 @@
...
+   sha=$(echo "${line}" | awk '{print $1}')
+   file=$(echo "${line}" | awk '{print $2}')
+   echo "Checking shasum for $file"
+   calculatedSHA=$(shasum -a 256 "$LINK/$file" | awk '{print $1}')
+   if [ "$calculatedSHA" != "$sha" ]; then
+     echo "Error - SHASUMS256 does not match for $file"
+     echo "Expected - $sha"
+     echo "Found - $calculatedSHA"
+     if [ "$tapFile" ]; then
+       echo "not ok $num SHASUMS256 does not match for $file
```
richardlau: Shouldn't `$num` be `$testNumber`?

gdams: good point, updated PTAL

jenkins/download-test.sh
```diff
@@ -0,0 +1,66 @@
...
+  case "$1y" in
+      -t|--tap)
+      tapFile="$2"
+      shift # Past option argument.
+      ;;
+      *) # unknown option
+      echo "Ignoring unknown option: $1"
+      ;;
+  esac
+  shift # Past option.0
```
gibfahn: Typo? (extra 0)

jenkins/download-test.sh
```diff
@@ -0,0 +1,66 @@
...
+  echo "NODE_VERSION is undefined! Please declare NODE_VERSION"
+  exit 1
+fi
+
+# If the provided NODE_VERSION is inexact, e.g. `v8`, get the most recent matching version.
+LINK=`rsync rsync://unencrypted.nodejs.org/nodejs/$downloadDir/ | grep $NODE_VERSION | sort -t. -k 1,1n -k 2,2n -k 3,3n | tail -1 | awk '{print $5}'`
+
+# Remove old files
+[ "$tapFile" ] && rm -rf "$tapFile" && echo "TAP version 13" > $tapFile
+rm -rf v*
```
gibfahn: move this up a line, otherwise if people call their tapfile vsomething they're going to be very confused.

jenkins/download-test.sh
```diff
@@ -0,0 +1,66 @@
...
+   calculatedSHA=$(shasum -a 256 "$LINK/$file" | awk '{print $1}')
+   if [ "$calculatedSHA" != "$sha" ]; then
+     echo "Error - SHASUMS256 does not match for $file"
+     echo "Expected - $sha"
+     echo "Found - $calculatedSHA"
+     if [ "$tapFile" ]; then
+       echo "not ok $testNumber SHASUMS256 does not match for $file
+         ---
+           echo "Expected - $sha"
+           echo "Found - $remoteSHA"
```
gibfahn: My bad, messed up the suggestion in last review :sob: . Obviously you don't need to echo the word echo (and the quotes will mess stuff up).

```diff
-           echo "Expected - $sha"
-           echo "Found - $remoteSHA"
+           Expected - $sha
+           Found - $remoteSHA
```

gdams:
@gibfahn updated PTAL

gibfahn:
@nodejs/build any other issues? If not I'll land this later today.

gdams:
remember that the download testing job needs to be updated with the new path

mhdawson:
About to land and then fix up the testing job as its already broken since it uses the branch :)

mhdawson:
Landed as 365c7852bab0caf580bf60d0108c90645c748bef

Jobs fixed up.  Still need to discuss email notifications but we can do that in the original issue as opposed to this PR so closing.

Issue #798, opened 2017-07-18
CI flakiness


jasnell:
CI appears to be having a number of issues today. Jobs have either been hanging or erroring out in weird ways. I've also attempting to start CI jobs just to have it do absolutely nothing after pressing submit. Can someone give the CI a bit of a kick?

gibfahn:
Previously discussed in https://github.com/nodejs/build/issues/761#issuecomment-315940357

gibfahn:
Previously discussed in https://github.com/nodejs/build/issues/761#issuecomment-315940357

gibfahn:
This seems relevant:

>Your Jenkins data directory "/var/lib/jenkins" (AKA JENKINS_HOME) is almost full. You should act on it before it gets completely full.

rvagg:
Yeah, full disk, working on it, we bumped into not long ago and it turned out to be a couple of very large jobs. We're going to need to go full block device on this machine, perhaps I'll try and do that now eh?

gibfahn:
>We're going to need to go full block device on this machine, perhaps I'll try and do that now eh?

Sounds reasonable, I don't think there was any objection when it was discussed before.

So how does access to the infra machines work? Everyone in build has access to `test`, the release WG (and some other people?) have access to `release`, and some subset of people have access to `infra`?

It'd be great to document this somewhere, maybe the build README.

refack:
I'm assuming it's probably not the same "disk full issue" but goes under "CI flakiness" title, the `node-test-commit-arm-fanned` is stuck
https://ci.nodejs.org/computer/node-msft-cross-compiler-1/
![image](https://user-images.githubusercontent.com/96947/28313850-54f78076-6b86-11e7-8b40-cc21d93d041c.png)

rvagg:
So I think I figured it out, the backup job that also does the cleanup wasn't doing a full cleanup. It uses `find` to do the job and uses the following roots `/var/lib/jenkins/*/builds/` and `/var/lib/jenkins/*/configurations/axis-nodes/*/builds/` _but_ the axis jobs have names other than 'axis-nodes' including 'axis-MACHINE', 'axis-label', 'axis-v8test' and a bunch of others. Basically any multi-axis job has a name corresponding to the axis selected. What's more, the build directories are pretty deep within some of these due to the way they are configured.

[...]

PR #684, opened 2017-04-14
github-bot: add Jenkins $ENVs to kick off builds for CITGM

phillipj:
These $ENVs are needed for @nodejs-github-bot to kick off Jenkins builds based off PR comments. 

P.S. these are already present on the bot server in production.

Refs https://github.com/nodejs/github-bot/pull/128

PR #811, opened 2017-07-24
doc: add a powers.md to document who has access

gibfahn:
Initial stab at covering who has access to what.

## Outstanding questions:
- [x] @bnoordhuis has read access to nodejs/secrets, as he's an org owner I assume this is no longer necessary? _**EDIT:**_ Yep, historical, removed him.
- [x] is release-* the same group as infra-*? I think it is, but looking at [the commit history](https://github.com/nodejs/secrets/commits/master/build/release) I'm not sure. _**EDIT:**_ It's not
- [x] what about website access? Is there any special access that people have that we should document?

Follow-on proposal: we could create sub-teams of @nodejs/build to list the groups of people who have access to things (which saves having lists of names in this document). I think it's just `nodejs/infra-admins` (which have access to `release-*` and `infra-*`) but I'm not sure.

Refs: https://github.com/nodejs/build/issues/798#issuecomment-316045678

### DIFF
--- a/doc/access.md
+++ b/doc/access.md
@@ -1,0 +1,91 @@
+# Access to Node.js Infrastructure
+
+Documents which groups have access to which Infra assets. Note that links to
+`@nodejs/` teams are not visible to people who aren't in the Nodejs
+organisation, so those links may not work for you. The [secrets repo][] is also
+secret...
+
+## Machine Access
+
+For a list of machines, see the [inventory.yml][]. Secrets are stored in the
+[secrets repo][], which [@nodejs/build][] (and [org owners][]) have access to.
+Secrets are individually encrypted, so access to the repo does not itself
+give access to any of the secrets within. For more info see the repo's README.
+
+### Test machines
+
+[@nodejs/build][] have root access to the test CI machines (`test-*`).
+
+### Infra machines
+
+A subsection of build members have access to infra machines
+(`infra-*`). The current list is:
+
+- Johan Bergström [@jbergstroem](https://github.com/jbergstroem)
+- João Reis [@joaocgreis](https://github.com/joaocgreis)
+- Michael Dawson [@mhdawson](https://github.com/mhdawson)
+- Rod Vagg [@rvagg](https://github.com/rvagg)
+
+### Release machines
+
+A subsection of build members have access to infra machines
+(`infra-*`). The current list is:
+
+- Johan Bergström [@jbergstroem](https://github.com/jbergstroem)
+- João Reis [@joaocgreis](https://github.com/joaocgreis)
+- Rod Vagg [@rvagg](https://github.com/rvagg)
+
+## Infra Access
+
+There are a number of other infra assets maintained by the Build WG, accesses
+are as follows.
+
+Note that the machines that our Jenkins instances run on are `infra` machines,
+and thus any task that requires access to the machine requires `infra` access.
+
+### [ci.nodejs.org](ci.nodejs.org)
+
+- [@nodejs/collaborators][] have access to run Node core tests.
+
+- Run and configure access for other jobs is controlled by the teams who own them
+(for example, the [post-mortem jobs][] are run by [@nodejs/post-mortem][], and
+configured by [@nodejs/post-mortem-admins][]. For more info see the [Jenkins
+access doc][].
+
+- [@nodejs/build][] have machine access (the ability to add, remove, and
+configure machines).
+
+- [@nodejs/jenkins-admins][] have admin access.
+
+### [ci-release.nodejs.org](ci-release.nodejs.org)
+
+- [@nodejs/release][] have access to run builds.
+
+- [@nodejs/jenkins-admins][] have admin access.
+
+### [github-bot][]
+
+Those with `github-bot` access have access to the Github Bot's configuration,
+including Github and Jenkins secrets.
+
+The following have access:
+
+- Johan Bergström [@jbergstroem](https://github.com/jbergstroem)
+- João Reis [@joaocgreis](https://github.com/joaocgreis)
+- Rod Vagg [@rvagg](https://github.com/rvagg)
+- Phillip Johnsen [@phillipj](https://github.com/phillipj)
+- Hans Kristian Flaatten [@Starefossen](https://github.com/Starefossen)
+
+
+[@nodejs/build]: https://github.com/orgs/nodejs/teams/build/members
+[@nodejs/collaborators]: https://github.com/orgs/nodejs/teams/collaborators/members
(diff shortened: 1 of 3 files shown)


rvagg:
might be a bit picking and this is just a suggestion, but maybe "powers" isn't the best name here, "access" or similar might be a bit more .. gentle?

also, who do those links to the team members work for? what level of access does someone need to be able to see who is in jenkins-admins for example and do we care for example if non-collaborators can't see that?

mcollina:
👍 on access.

Also, is there a crash course on the infra or some documents I can read. According to this doc, I have access to the infrastructure secrets and keys. However, there is little guidance in the CTC onboarding process on those.

Moreover, is there a process to _revoke_ access? In other terms, how do we rotate keys?

piccoloaiutante:
I like `access` or even `infra-access.md`.

@rvagg it seems to me that in order to see`jenkins-admins` you need to be at least a `member` of the org. I think that having jenkins-admins not seen by non-collaborators isn't a big issue, as if a new comer needs to contact someone in the team can always ask in issues or irc.

@mcollina I got onboarded by @jbergstroem and @joaocgreis. Maybe would be a good idea to arrange one with you.

gibfahn:
>also, who do those links to the team members work for? what level of access does someone need to be able to see who is in jenkins-admins for example and do we care for example if non-collaborators can't see that?

@rvagg They only work for org members. It would be great if they were public, but that's a Github thing, so there's not much we can do. The only question is whether it's better to have them or not. I'd say it's better to have them. I can add a sentence explaining that if it'd be helpful.

>Also, is there a crash course on the infra or some documents I can read. 

This is my first attempt at having build WG documentation. I figured we needed to work out what we actually have before we talk about how to do stuff.

>According to this doc, I have access to the infrastructure secrets and keys. However, there is little guidance in the CTC onboarding process on those.

You have access to the secrets repo, but not to the gpg encrypted secrets within. Access is documented in that repo, which I think is the best place for it. If it's inadequate then we should add to that repo's docs. I'll add a sentence saying that.

FWIW I don't think it's particularly intentional that org owners have access, it's just how Github works. You won't actually be able to decrypt any of the files inside it.

mcollina:
> You have access to the secrets repo, but not to the gpg encrypted secrets within. Access is documented in that repo, which I think is the best place for it. If it's inadequate then we should add to that repo's docs. I'll add a sentence saying that.

Then this [sentence](https://github.com/nodejs/build/pull/811/files#diff-7ebdd28f2e3f67dab948bd69de05e09eR5) is not correct. Not sure what's the proper rewording.
I'm perfectly ok in not having access BTW, but if I have access to servers, I need to be sure to handle that properly.

gibfahn:
>For a list of machines, see the [inventory.yml][]. Access is controlled through the [secrets repo][], which [@nodejs/build][] (and [org owners][]) have access to.

You have access to the _repo_, just not to the individually encrypted secrets within. I've tried to clarify, does it make more sense? If not please suggest some text (I don't want to get too verbose here).

mcollina:
How about

> Access to the secrets repo does not imply access to the machines themselves.

gibfahn:
I changed `powers.md` to `access.md`. I quite liked `powers`, reminded me of superpowers (which is apt, as with other superpowers, it usually means you have to fix stuff at inconvenient hours etc.) But `access` is fine too (just bland).

gibfahn:
@mcollina okay I pretty much added that, PTAL.

mhdawson:
In terms of employer, I don't think we want to add that as its not necessarily tied to why/who has access.  Having is discover-able through the github profile is good though.

gibfahn:
>Generally LGTM. Thought: do we want to add employer or is that perhaps better expressed through a GitHub profile? A few of us have "keys to the kingdom" of sorts.

I agree with @mhdawson that adding the employer gives the wrong impression, people don't have access because of their employer. At the same time I think we should definitely be open about who we work for. Making sure everyone has it clearly documented in their Github profile would be good. 

Github allows you to fill in the company field in https://github.com/settings/profile, maybe we should encourage everyone to fill it in.

image

image

doc/access.md
```diff
@@ -0,0 +1,64 @@
...
+give access to any of the secrets within. For more info see the repo's README.
+
+### Test machines
+
+[@nodejs/build][] have root access to the test CI machines (`test-*`).
+
+### Infra and Release machines
+
+A subsection of build members have access to infra and release machines
+(`infra-*` and `release-*`). The current list is:
```
gibfahn: I need to add a note to clarify that this includes `ci.nodejs.org`, so any CI issues that can't be resolved within Jenkins will have to be resolved by this group.

gibfahn:
@rvagg I'd like to land this, any objections?

doc/access.md
```diff
@@ -0,0 +1,91 @@
...
+### [ci-release.nodejs.org](ci-release.nodejs.org)
+
+- [@nodejs/release][] have access to run builds.
+
+- [@nodejs/jenkins-admins][] have admin access.
+
+### [github-bot][]
+
+Those with `github-bot` access have access to the tokens that the bot uses to
+authenticate against Github.
```
gibfahn: @phillipj @Starefossen I'm pretty much guessing here, is this accurate?

jbergstroem: Yes; but I haven't been involved in at least 6 months so it might have changed.

jbergstroem:
No objections to landing as-is. Let me try and answer other outstanding questions.

> @gibfahn said:
> @bnoordhuis has read access to nodejs/secrets, as he's an org owner I assume this is no longer necessary?

I would agree. He was historically very active in part of the creation of `io.js` and the general structure of all repos. I would guess this is the remnant of that.

> @gibfahn said:
> what about website access? Is there any special access that people have that we should document?

Guessing you're referring to updating the website. This is handled through the website group in this repo: https://github.com/nodejs/nodejs.org as well as work that is done and landed through the build team for building those assets.

doc/access.md
phillipj: Correct, as well as:

- Github webook secret, verifying the incoming webhook request originates from Github
- Server configuration (port number ATM)
- Jenkins tokens used when communicating with Jenkins to trigger CI builds

The latter isn't added yet, but it's my intention to do so when https://github.com/nodejs/build/pull/684 lands..

gibfahn: Thanks @phillipj, added.

FWIW I don't think we need to go into too much detail here, I suggest putting that kind of info in https://github.com/nodejs/secrets/tree/master/build/github-bot (that way we don't have to worry about who can read it).

gibfahn:
Landed in efb11cc08574026dda0d0eeb45007bccf6f36a17 ef1f86315886b115e9536fa8e3578318ed21546f

Issue #831, opened 2017-08-11
Security: git arbitrary execution w/ ssh:// urls


rvagg:
https://marc.info/?l=git&m=150238802328673&w=2

> A malicious third-party can give a crafted "ssh://..." URL to an unsuspecting victim, and an attempt to visit the URL can result in any program that exists on the victim's machine being executed. Such a URL could be placed in the .gitmodules file of a malicious project, and an unsuspecting victim could be tricked into running "git clone --recurse-submodules" to trigger the vulnerability.

@nodejs/collaborators: please note that for the safety of the project we'd love to make sure you are as protected as possible so that you don't become an attack vector given the level of access you have. Please update your git version today to v2.14.1.

@nodejs/build re infra:

My assessment is this: if you can get a git url into one of our test, benchmark or build machines then you it would probably be more simple to just stick something in Makefile (in test-ci for instance) so we're open to arbitrary already. This is why we only allow trusted users to access our infra and we isolate our release infra even further; plus the "certify safe" is there exactly to try and make a mental note for the collaborator that this isn't an automatically safe operation. 

But, this vulnerability does open up some interesting vectors that could allow an attacker to hide their exploit. The submodules example is an instance of that. So, while I don't think this is critical for us, we should be making moves to update all of our infra to v2.14.1+ for precautionary purposes.

gibfahn:
Good point about upgrading, I notice that my Ubuntu 16.04 machine was pulling git from the Ubuntu default repo, and thus had `2.7.4`. Worth adding the [git stable ppa](https://launchpad.net/~git-core/+archive/ubuntu/ppa) and updating if you haven't already:

```bash
sudo add-apt-repository ppa:git-core/ppa
sudo apt update && sudo apt upgrade
```

or for macOS:

```bash
brew update && brew upgrade
```

TimothyGu:
Ubuntu [fixed](https://people.canonical.com/~ubuntu-security/cve/2017/CVE-2017-1000117.html) this in 1:2.7.4-0ubuntu1.2 for 16.04 (and 1:2.11.0-2ubuntu0.2 for 17.04), so you don't actually need 2.14.1 to get the fix.

For Debian, this was fixed a while back in 1:2.1.4-2.1+deb8u4 in jessie, and 1:2.11.0-3+deb9u1 for stretch. See [Debian Security Tracker](https://security-tracker.debian.org/tracker/CVE-2017-1000117).

addaleax:
[...]

PR #836, opened 2017-08-15
win: update Git to 2.14.1

joaocgreis:
Updated Git to 2.14.1 in all Windows machines. This commit reflects that change.

This caused an issue where Git no longer works with ssh-agent (apparently, Windows only). So, `node-compile-windows` was no longer able to upload the binaries and I had to update the job to a different mechanism.

Ref: https://github.com/nodejs/build/issues/831

R?= @kunalspathak @piccoloaiutante @refack

### DIFF
--- a/setup/ansible-inventory
+++ b/setup/ansible-inventory
@@ -175,6 +175,12 @@ node-win2008r2-x64-3.cloudapp.net
 node-win2012r2-x64-1.cloudapp.net
 node-win2012r2-x64-2.cloudapp.net
 node-win2012r2-x64-3.cloudapp.net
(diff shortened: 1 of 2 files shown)


gibfahn:
What's the current status of the `ansible/` folder for Windows?

joaocgreis:
No support at all yet for Windows in `ansible/`. I have it in my queue and would really like to see it done, but it's no small task and will take me a while to get there.

So, for now, Windows is only in `setup/`.

joaocgreis:
Landed in https://github.com/nodejs/build/commit/2c2a3e20a95b6ed3061b48e404da93440f44b42f

PR #840, opened 2017-08-20
Add EPEL to centos7

rvagg:
I just got caught by this setting up a replacement for test-packetnet-centos7-arm64-1 which has been borked a lot recently (seems to be some bad hardware problems because I tried to do a reboot again which has worked in the past but it's locked up and I've had to contact packet.net support about it—I've had more success in their non-CA datacenters so I switched to their NJ one for this).

I don't know what's changed but `ccache` isn't available for plain install on CentOS 7 on packet.net any more. Perhaps it's something new in CentOS 7.3 or perhaps packet.net used to bundle EPEL automatically but now doesn't. Anyway, it's now installing it manually in this PR and is confirmed working on the new test-packetnet-centos7-arm64-1.

btw there's a note in the CentOS 6 setup about adding EPEL, perhaps it's a TODO and should be added. It can be done in the same way as this but until I have another CentOS 6 box to provision I won't be bothering testing it, someone else can though.

### DIFF
--- a/ansible/roles/baselayout/tasks/partials/repo/centos5.yml
+++ b/ansible/roles/baselayout/tasks/partials/repo/centos5.yml
@@ -14,5 +14,5 @@
 
 - name: install epel
   yum:
-    name: "http://dl.fedoraproject.org/pub/epel/5/{{ ansible_architecture }}/epel-release-5-4.noarch.rpm"
+    name: "http://archives.fedoraproject.org/pub/archive/epel/epel-release-latest-5.noarch.rpm"
     state: present
--- a/ansible/roles/baselayout/tasks/partials/repo/centos6.yml
+++ b/ansible/roles/baselayout/tasks/partials/repo/centos6.yml
@@ -16,4 +16,7 @@
     gpgkey: http://ftp.scientificlinux.org/linux/scientific/5x/{{ ansible_architecture }}/RPM-GPG-KEYs/RPM-GPG-KEY-cern
     gpgcheck: yes
 
-# needs epel! https://fedoraproject.org/wiki/EPEL
+- name: install epel
+  yum:
+    name: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-6.noarch.rpm"
+    state: present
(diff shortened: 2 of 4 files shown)


gibfahn:
I notice that the equivalent for CentOS 5 links to the specific release rather than the latest. Is there any reason to think we would want to do that? I'd have thought keeping with the latest would be better, just checking.

https://github.com/nodejs/build/blob/0d149d7fecaeb84cec33a9804d8b7391b5f08ee1/ansible/roles/baselayout/tasks/partials/repo/centos5.yml#L17

Also while you're at it, that CentOS 5 link is broken [because](https://dl.fedoraproject.org/pub/epel/5/README):

>ATTENTION
>======================================
> The contents of this directory have been moved to our archives available at:
> 
> http://archives.fedoraproject.org/pub/archive/epel/
> 
> If you are having troubles finding something there please stop by #epel on irc.freenode.net

So while you're at it, maybe change:

```diff
-    name: "http://dl.fedoraproject.org/pub/epel/5/{{ ansible_architecture }}/epel-release-5-4.noarch.rpm"
+    name: "http://archives.fedoraproject.org/pub/archive/epel/5/{{ ansible_architecture }}/epel-release-5-4.noarch.rpm"
```

rvagg:
I think the centos5 thing is the way it is because I don't believe there was a non-specific URL when it was set up. These new URLs are new I think.

I've gone and fixed up the centos5 URL and added cento6 for good measure (not tested so crossing fingers on that one!). PTAL @gibfahn

ansible/roles/baselayout/tasks/partials/repo/centos5.yml
```diff
@@ -14,5 +14,5 @@
 
 - name: install epel
   yum:
-    name: "http://dl.fedoraproject.org/pub/epel/5/{{ ansible_architecture }}/epel-release-5-4.noarch.rpm"
+    name: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-5.noarch.rpm"
```
gibfahn: I think this needs to be the archive link:

http://archives.fedoraproject.org/pub/archive/epel/epel-release-latest-5.noarch.rpm

rvagg: hah, you're right, I got that URL off the EPEL main page and it's broken there too (their 6 and 7 links work tho), fixed in latest commit

PR #855, opened 2017-08-29
doc: add instructions for adding fw entry

mhdawson:


mhdawson:
@joaocgreis

refack:
If there is nothing secret there I'd suggest adding a copy of the rules file to this repo for version management. Else keep it in `secrets`
So the procedure will be:
1. Local: edit `X/Y/iptable_rules`
2. Local: commit & push
3. local: `scp X/Y/iptable_rules ci:~/`
4. Remote: `iptables-restore ~iptable_rules`

Maybe add a (0) step of `iptables-save > iptable_rules` and local `scp ci:~/iptable_rules X/Y/iptable_rules` to verify rules are in sync with git

joaocgreis:
The exact firewall rules we use should be private. I don't think there's anything sensitive there at the moment, but we should be able to change them quickly and secretly if needed.

We could keep them in the secrets repo but I don't think we need to move forward with this at the moment, there is some work in progress at https://github.com/nodejs/build/blob/master/ansible/playbooks/jenkins/host/iptables.yml to generate the rules from the inventory. The documentation here can be later updated or integrated with the docs in `ansible/`, but for now this is good to have.

gibfahn:
Added @joaocgreis 's links to the commit message, seems good to document the WIP ansible playbook in the commit message.

PR #867, opened 2017-09-06
aix: no longer need subversion

refack:
FYI it used to drag in `httpd`
P.S. highly recommended to remove it from current machines


gibfahn:
cc/ @mhdawson

mhdawson:
Is there any context as to why it was needed and is no longer needed ?

refack:
> Is there any context as to why it was needed and is no longer needed ?

Sometime is the past V8 needed it for some testing tools. They have moved all their tooling to git.
The main issue is that the script installs `httpd` which is an optional feature even if `svn` was required, and is an unnecessary "wide load".
https://github.com/nodejs/build/blob/18cdefb192a91b339873d239d17cb1e6967b03c0/ansible/README.md#L160

refack:
I run this command on both machines:
```bash
rpm -vv -e subversion httpd serf apr apr-util apr-util-ldap openldap openldap-devel neon file-libs pcre file krb5-libs
```

PR #869, opened 2017-09-08
move common `apt` task out of debian specific task

al-k21:
I think this is right, PTAL.

### DIFF
--- a/ansible/roles/package-upgrade/tasks/partials/apt.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/apt.yml
@@ -4,6 +4,9 @@
 # Updates all packages for debian-based distributions
 #
 
+- name: update repositories
+  raw: apt-get update -y
+
 - name: upgrade installed packages
   apt: upgrade=dist update_cache=yes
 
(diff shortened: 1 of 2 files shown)


ansible/roles/package-upgrade/tasks/partials/apt.yml
```diff
@@ -4,6 +4,9 @@
 # Updates all packages for debian-based distributions
 #
 
+- name: update repositories
+  raw: apt-get update -y
```
gergoerdosi: I think `update_cache=yes` does the same below, so this command seems to be unnecessary.

al-k21: Does this mean the whole `debian7.yml` is unnecessary ?

gibfahn: @al-k21 It does seem like this can be removed, unless there's a reason to rerun it later in the setup process.

cc/ @jbergstroem in case there is a reason for it.

maclover7:
Landed in 888030957977834494ea4f7bf8cf7081b7cd9c09

PR #870, opened 2017-09-08
ansible: add RHEL to the supported OS's

al-k21:
First step in supporting RHEL.
I am not sure how much `centos` stuff should also be run on RHEL.

### DIFF
--- a/ansible/roles/package-upgrade/vars/main.yml
+++ b/ansible/roles/package-upgrade/vars/main.yml
@@ -6,7 +6,7 @@
 
   pm: {
     'apk': 'alpine',
-    'yum': 'centos',
+    'yum': ['centos', 'rhel'],
     'apt': ['debian', 'ubuntu'],
     'dnf': 'fedora',
     'pkg': 'freebsd',


gibfahn:
Obviously this isn't complete RHEL support, but seems like a good first step.

chrislea:
There should be no functional difference between supporting CentOS and RHEL at the same version. IE, if you support CentOS 6.x, then you already support RHEL 6.x.

gibfahn:
@chrislea presumably you might have to configure your yum repos to pull from a different server right?

chrislea:
@gibfahn we certainly don't at NodeSource. Our `rpm` packages, repos, and instructions are identical for CentOS 6 / RHEL 6 and also CentOS 7 / RHEL 7.

refack:
Does RHEL6 have a better `devtoolset-6.i686`? And what about RHEL5 is it easier to find than CentOS5?
(If they are easier to maintain can we use them instead of CentOS?)

chrislea:
CentOS 5 / RHEL 5 has reached production EOL and we shouldn't think about that distro.

The `devtoolset-*` packages are identical for CentOS and RHEL, so there's no better or worse.

It's probably worth quickly explaining the relationship between CentOS and RHEL to clear up confusion. The basic zen as I understand it of the GPL is "if you distribute a binary, you also have to make the sources for that binary available". So that makes it harder to add value as a binary distribution vendor. The way Red Hat gets around this is surprisingly simple. When they make a release, they make their source code packages available everywhere, for free. But if you want a complete binary distribution from those sources that Red Hat supports, then you have to pay them to get those binaries.

CentOS is a group of people that grabs all of the freely available RHEL source packages and builds an entire binary distribution from them. So when I say they are functionally equivalent, I'm being pretty literal. They are the same thing modulo the amount of support you can get from Red Hat (IIRC you actually can pay Red Hat to support your CentOS install, but I think maybe it's more limited and some of their online management tools won't work without the full license).

If I'm being picky, CentOS also has to swap out some branding stuff that Red Hat has trademarked, but that's just for legal reasons. From an engineer's standpoint, CentOS and RHEL are the same thing. Same kernels, same `libc++`, same compilers, same ABIs, etc.

sam-github:
Cleaning up stalled PRs, they can be reopened if necessary.

Note that we use and ansibled multiple rhel hosts right now without this.

PR #880, opened 2017-09-13
inventory: add test-digitalocean-ubuntu1604-x86-1

gibfahn:
Was going to push directly to master, but thought I'd raise a PR in case there are any others.

Would be good to have something that polled the Jenkins API and auto-filled the inventory.yml, maybe a nice project for someone.


gibfahn:
I also noticed that [test-softlayer-ubuntu14-x86-1](https://ci.nodejs.org/computer/test-softlayer-ubuntu14-x86-1/) is misnamed (should be `1404` not `14`). I assume I should change the name in Jenkins rather than the inventory, thoughts?

refack:
> I assume I should change the name in Jenkins rather than the inventory, thoughts?

According to the naming convention, I think rename in Jenkins
https://github.com/nodejs/build/blob/89e15bed2131366c1ea4d38370a69ec879b7f298/ansible/README.md#L94

gibfahn:
Yeah, I'll rename tomorrow if no-one corrects me.

joaocgreis:
About renaming: renaming in Jenkins would be the right thing to do, `ubuntu1404` is better than just `ubuntu14`. However, the workers use that name to connect, so you'll have to rename in the workers as well (looks like that worker was deployed before the ansible refactor, so you'll have to figure out where to rename - in this case is `/etc/init/jenkins.conf`). There are many others `ubuntu12/14`, feel free to rename however many, or just leave them.

Issue #896, opened 2017-09-25
Benchmarking data machine rsync issues


mhdawson:
Twice in the last week the benchmarking machine has become non-responsive due to a large number of rsync processes.

 I saw this once before and thought it might have been due to a stale cron job but I removed the job at that point and it continue to re-occur.

All I can think of now is that for some reason the job pulling the charts is doing something strange ? @rvagg ?

I can paste the rsync command with the sender info but wanted to make sure that is not sensitive first.

mhdawson:
Happened again today

mhdawson:
And again today.

@rvagg which machine is it that the website runs on and does the rsync from the benchmarking machine.  Next time this happens I'd like to look on that machine to see if that gives me any more information as to what is going on.

mhdawson:
It seems to be related to something that is running at around 6:25

From one of the syslog files, and saw the same pattern in another one as well.

```
Oct  2 05:17:06 nodejs-data2 CRON[367]: (root) CMD (   cd / && run-parts --report /etc/cron.hourly)
Oct  2 06:17:07 nodejs-data2 CRON[2380]: (root) CMD (   cd / && run-parts --report /etc/cron.hourly)
Oct  2 06:25:03 nodejs-data2 CRON[2606]: (root) CMD (test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.daily ))
Oct  2 06:27:25 nodejs-data2 kernel: [1203087.357360] rsync invoked oom-killer: gfp_mask=0x201da, order=0, oom_score_adj=0
Oct  2 06:27:25 nodejs-data2 kernel: [1203087.357367] rsync cpuset=/ mems_allowed=0
Oct  2 06:27:25 nodejs-data2 kernel: [1203087.357372] CPU: 0 PID: 29303 Comm: rsync Not tainted 3.13.0-65-generic #105-Ubuntu
Oct  2 06:27:25 nodejs-data2 kernel: [1203087.357375]  0000000000000000 ffff880008a07a98 ffffffff81723f80 ffff88000206c800
Oct  2 06:27:25 nodejs-data2 kernel: [1203087.357379]  ffff880008a07b20 ffffffff8171e587 ffffffff81067a16 ffff880008a07af8
Oct  2 06:27:25 nodejs-data2 kernel: [1203087.357381]  ffffffff810c79dc 00000000000000a9 ffff88003fe1ee28 0000000000000000
[...]

PR #904, opened 2017-10-04
use lock when rsyncing benchmarking data

rvagg:
Fixes: https://github.com/nodejs/build/issues/896


rvagg:
not doing this into /ansible/ because we don't have the web server set up in there yet (or do we?), it's pretty special case so I'm not sure it's worth doing anyway.

PR #911, opened 2017-10-10
doc: add packet as an infra provider

mhdawson:


mhdawson:
@rvagg

README.md
```diff
@@ -139,6 +139,10 @@ and speed in our CI system.
 - **[MacStadium][23]**: Managed hosting provider for Mac. Provides Mac
   hardware used for testing in the Node.js project's CI system.
 
+- **[Packet][24]**: Bare metal cloud for developers. Through their
+  [Works on Arm partnership](https://www.worksonarm.com/),
+  Packet provides Armv8 build infrastructure.
```
rvagg: Ugh, this poor capitalisation is killing me but I see we're guilty of inheriting from their wording. My preference would be "Works on ARM" and "ARMv8", and "ARM64" would be even better here since that's becoming the canonical name for it. What do you think @mhdawson?

mhdawson: I think we probably need to stick to the name they chose for "Works on Arm partnership", but I think its ok change the capitalization to "Works on arm partnership" and then later say "Packet provides ARM64 build infrastructure" if you think that is more accurate.

mhdawson:
@rvagg  Pushed changes to address comments

README.md
```diff
@@ -139,6 +139,10 @@ and speed in our CI system.
 - **[MacStadium][23]**: Managed hosting provider for Mac. Provides Mac
   hardware used for testing in the Node.js project's CI system.
 
+- **[Packet][24]**: Bare metal cloud for developers. Through their
+  [works on arm partnership](https://www.worksonarm.com/),
```
rvagg: I think "Works on Arm" is going to be closest to how they are presenting it. Check out https://www.worksonarm.com/blog/woa-issue-23/ and search for "works on arm" and you'll see a couple of instances where it's used mid-sentence and the capitalisation is "Works on Arm". Sorry to be pedantic but other than this LGTM.

refack:
I can update `provider-logos.ai` if everyone's Ok with the current layout.

mhdawson:
Pushed commit to address comments.

mhdawson:
@rvagg you happy with the latest version ?

rvagg:
yep, lgtm

refack:
@mhdawson I added a suggested tweak-commit, feel free to push it out.
1. replaced logo png with svg
2. added to `.ai`
3. exported new layer from .ai

mhdawson:
@refack I'll leave your commit in but we should talk about what tools are needed for the different formats.  If there is no available free version of the tool (like illustrator) I don't think our workflow should depend on it.

refack:
I just checked. The .ai is totally workable in [inkscape](https://inkscape.org/en/) (which is GPL).

mhdawson:
Had landed this a few days ago, should have closed.

landed as 376a889203bded39f0072147bbfb9ee03023d09c

Issue #838, opened 2017-08-18
Migrate Jenkins job configuration to pipeline files


maclover7:
Currently, it looks like all Jenkins jobs are configured through the web interface. It would be advantageous if all of the jobs were managed through pipelines ([a Jenkins 2 feature](https://jenkins.io/doc/book/pipeline/)), which allows for job configuration to be checked into version control. This would allow for more transparency in the configuration, and using a normal code review process to approve changes.

Some key points to consider are...

- where should the pipeline files live? should they be in the project itself (travis-style), or should they live in nodejs/build in a jenkins directory?
- migrating the jobs with no downtime
- since job configuration is being checked into source control, is there a better way to manage admin groups/users, with the additional metadata?

If this all sounds alright, I can try and draft up a more complete project plan and migration strategy :)

gibfahn:
I've been thinking that this would be a good idea for a long time (and it just came up in https://github.com/nodejs/build/pull/804#issuecomment-323406841). It's a fairly massive change though, as pipelines are pretty different from the existing build trigger plugins and methods.

>where should the pipeline files live? should they be in the project itself (travis-style), or should they live in nodejs/build in a jenkins directory?

I think I agree with @joaocgreis 's point in https://github.com/nodejs/build/pull/804#issuecomment-323234788, we want to have one master copy of the build scripts, not one for each version, so it makes more sense to have them in a separate repository. I think a `jenkins/` directory in this repo might make sense.

>since job configuration is being checked into source control, is there a better way to manage admin groups/users, with the additional metadata?

I think the current system of using Github teams to manage access works quite well, but if Jenkins has some alternative method of managing this we should definitely consider it.

>migrating the jobs with no downtime

I think the first thing we need to look at is whether this is worth doing, and how much work it's going to entail. [Node CI](https://ci.nodejs.org/view/All/) includes non-core CI tests (like CitGM, node-report, and libuv). Those would be a good place to start, if we convert some of those jobs, then that would give us a good indication of how difficult/worthwhile it's going to be.

Another issue is that no-one in the build WG is a Pipeline expert. Many of us are familiar with the existing (Jenkins 1) configuration methods, but pipelines are a new way of doing things.

maclover7:
> I think a jenkins/ directory in this repo might make sense.

I agree -- I think having them in nodejs/build alongside other relevant configuration (ansible, etc) would be helpful

[...]

PR #922, opened 2017-10-16
jenkins: add http-parser.jenkinsfile

maclover7:
This is a simple prototype to try out Jenkins pipelines (ref #838) in a low-ish risk way. This commit introduces `jenkins/http-parser.jenkinsfile`, a pipeline to run tests for [nodejs/http-parser](https://github.com/nodejs/http-parser) -- http-parser is not part of ci.nodejs.org as of right now. I setup a Jenkins instance locally and was using it to test drive out the pipeline file. I can confirm it works correctly for PR 388 as well as master (if you leave PR_ID blank).

Before this can go out, there are a few things (and probably more that I didn't think of) that will be need to tested/done:

- [x] finish setting up integration with nodejs/github-bot
  - get the hostname, and replace ngrok.io with that
  - add http-parser to [this list](https://github.com/nodejs/github-bot/blob/master/scripts/jenkins-status.js#L4)
- [ ] target specific platforms
  - right now, http-parser's Travis CI configuration only targets linux, so we should either stick with that or add more platforms via the [`agent` / `label`](https://jenkins.io/doc/book/pipeline/syntax/#agent) config options.
- [ ]  backfill current open PRs
  - I think a good test for this would be to backfill all current open PRs with a CI run, to make sure it fails / succeeds as appropriate

Some pipelines stuff is a little quirky, but I think this is a decent POC -- please let me know your thoughts :)

cc @refack @gibfahn

### DIFF
--- a/jenkins/http-parser.jenkinsfile
+++ b/jenkins/http-parser.jenkinsfile
@@ -7,8 +7,8 @@ pipeline {
 
     parameters {
         booleanParam(name: 'CERTIFY_SAFE', defaultValue: false, description: 'I have reviewed *the latest version of* these changes and I am sure that they don’t contain any code that could compromise the security of the CI infrastructure.')
-        string(defaultValue: "", description: 'The numeric ID of the pull request (from GitHub URL)', name: 'PR_ID')
-        string(defaultValue: 'master', description: 'Git branch to use for testing (ignore this value for PRs)', name: 'BRANCH')
+        string(name: 'PR_ID', defaultValue: "", description: 'The numeric ID of the pull request (from GitHub URL)')
+        string(name: 'BRANCH', defaultValue: 'master', description: 'Git branch to use for testing (ignore this value for PRs)')
     }
 
     stages {
--- a/jenkins/http-parser.jenkinsfile
+++ b/jenkins/http-parser.jenkinsfile
@@ -32,7 +32,7 @@ pipeline {
                 echo "Testing: ${sh(returnStdout: true, script: 'git log -1 --oneline').trim()}"
 
                 sh 'gcc --version'
-                sendBuildStatus("pending", env.PR_ID, env.BUILD_URL)
+                sendBuildStatus("pending", env.PR_ID, env.BUILD_URL, env.GIT_COMMIT)
             }
         }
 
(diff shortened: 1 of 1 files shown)


jenkins/http-parser.jenkinsfile
```diff
@@ -0,0 +1,87 @@
...
+        stage('Setup Git repository') {
+            when {
+                expression {
+                    return !!env.PR_ID
+                }
+            }
+
+            steps {
+                sh "git fetch origin pull/${env.PR_ID}/head:totest"
+                sh 'git checkout totest'
```
rvagg: Do these two need to be a bit more robust; like what if `totest` exists locally already and has a set of commits that doesn't match orign? What Does it need some `--force` or `--reset` or something?

rvagg:
How does the `curl` for build status get authenticated for this? I'm out of the loop on build status stuff but I'm not seeing anything that's stopping arbitrary hosts from updating build status if they have the right URL.

Rest of it looks pretty good, I'm +1 on getting this hooked up into Jenkins. Great work!

jenkins/http-parser.jenkinsfile
```diff
@@ -0,0 +1,87 @@
...
+
+    parameters {
+        booleanParam(name: 'CERTIFY_SAFE', defaultValue: false, description: 'I have reviewed *the latest version of* these changes and I am sure that they don’t contain any code that could compromise the security of the CI infrastructure.')
+        string(defaultValue: "", description: 'The numeric ID of the pull request (from GitHub URL)', name: 'PR_ID')
+    }
+
+    stages {
+        stage('Clone repository') {
+            steps {
+                git 'https://github.com/nodejs/http-parser.git'
```
gibfahn: When I try the pipeline script generator to check out a PR branch of this repo, it suggests:

```groovy
git branch: 'pull/${env.PR_ID}/head', url: 'https://github.com/nodejs/http-parser'
```

Could you use that rather than this and tne next stage?

maclover7: The reason why I made the PR-branch-switching a separate stage was so CI runs could be done for master... though maybe this is unneeded functionality right now? Might be better to start off with just PRs? 😬

gibfahn: Ahh, good point. Then alternately just do:

```groovy
git(branch: '${env.BRANCH', url: 'https://github.com/nodejs/http-parser')
```

And put `BRANCH` as a parameter:

```groovy
         string(name: 'BRANCH', 
                   defaultValue: "refs/pull/XXXX/head", 
                   description: 'Branch to build, e.g. "master" or "refs/pull/XXXX/head" where XXXX is the PR number') 
```

maclover7: Hmm, trying to get this working on my local instance, but here's the part that's failing:

```console
[Pipeline] git
 > git rev-parse --is-inside-work-tree # timeout=10
Fetching changes from the remote Git repository
 > git config remote.origin.url https://github.com/nodejs/http-parser.git # timeout=10
Fetching upstream changes from https://github.com/nodejs/http-parser.git
 > git --version # timeout=10
 > git fetch --tags --progress https://github.com/nodejs/http-parser.git +refs/heads/*:refs/remotes/origin/*
 > git rev-parse refs/remotes/origin/refs/pull/388/head^{commit} # timeout=10
 > git rev-parse refs/remotes/origin/origin/refs/pull/388/head^{commit} # timeout=10
 > git rev-parse origin/refs/pull/388/head^{commit} # timeout=10
```

I have a feeling that `+refs/heads/*:refs/remotes/origin/*` is the issue, but I haven't been able to find a way to change that 😞

gibfahn: @maclover7 I can't see the error there, what exactly doesn't work?

maclover7: Sorry, my bad for just pasting in an error, here's [a gist](https://gist.github.com/maclover7/21693d452114fd51f27002013772321a) with the updated pipeline file and the full job log. Seems like specifying the branch via `git` is a little wonky :(

gibfahn:
@jkrems would appreciate a review from you if you have the time.

jenkins/http-parser.jenkinsfile
jkrems: FYI, I was playing around with this for a `node-inspect` build script and came up with the following:

```groovy
        stage("Checkout") {
            steps {
                checkout(poll: false, scm: [
                    $class: 'GitSCM',
                    branches: [[
                        name: (params.isPR ? "pr/${params.projectRef}/head" : params.projectRef)
                    ]],
                    userRemoteConfigs: [[
                        url: "https://github.com/${params.projectRepo}",
                        refspec: '+refs/pull/*:refs/remotes/origin/pr/* +refs/heads/*:refs/remotes/origin/*',
                    ]],
                ])
            }
        }
```

maclover7:
@jkrems thank you for the neat git clone/checkout snippet!!

@gibfahn updated PR, for right now +1 on keep the shared bits inside this pipeline file, and then as we need to we can extract out to helper/utility files

maclover7:
@refack @gibfahn updated this to take advantage of `post-build-status-update`, is this okay to land? (Someone will also need to create job on Jenkins since I'm not an admin)

maclover7:
@gibfahn updated, would you be able to create the http parser job on Jenkins?

gibfahn:
Job: https://ci.nodejs.org/view/All/job/node-test-http-parser/

Can you move this into `jenkins/pipelines/`?

maclover7:
> Can you move this into jenkins/pipelines/?

Fixed upon landing, landed in d094879b513beebff067617b08dc3f2b06364746.

PR #930, opened 2017-10-22
doc: list users by GitHub ID rather than name

gibfahn:
Makes it more obvious that we sort like that.

Takes inspiration from [the nodejs/node README](https://github.com/nodejs/node/blob/master/README.md#collaborators).

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ missing please open an issue. If you are interested in joining check out
 
 ## Build WG Members
 
+- [@Starefossen](https://github.com/Starefossen) - Hans Kristian Flaatten
 - [@gdams](https://github.com/gdams) - George Adams
 - [@geek](https://github.com/geek) - Wyatt Preul
 - [@gibfahn](https://github.com/gibfahn) - Gibson Fahnestock
@@ -36,7 +37,6 @@ missing please open an issue. If you are interested in joining check out
 - [@piccoloaiutante](https://github.com/piccoloaiutante) - Michele Capra
 - [@refack](https://github.com/refack) - Refael Ackermann
 - [@rvagg](https://github.com/rvagg) - Rod Vagg
-- [@Starefossen](https://github.com/Starefossen) - Hans Kristian Flaatten
 - [@trott](https://github.com/trott) - Rich Trott
 
 ### Infra Admins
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,41 +24,41 @@ missing please open an issue. If you are interested in joining check out
...
-- Kunal Pathak [@kunalspathak](https://github.com/kunalspathak)
-- Michael Dawson [@mhdawson](https://github.com/mhdawson)
-- Myles Borins [@mylesborins](https://github.com/mylesborins)
-- Phillip Johnsen [@phillipj](https://github.com/phillipj)
-- Michele Capra [@piccoloaiutante](https://github.com/piccoloaiutante)
-- Refael Ackermann [@refack](https://github.com/refack)
-- Rod Vagg [@rvagg](https://github.com/rvagg)
-- Hans Kristian Flaatten [@Starefossen](https://github.com/Starefossen)
-- Rich Trott [@trott](https://github.com/trott)
+- [@gdams](https://github.com/gdams) - George Adams
```
refack: Maybe add `<!-- lists are sorted by GitHub handle -->`

gibfahn: I feel like it's fairly obvious once you order it like this, but 🤷‍♂️ .

README.md
```diff
@@ -24,41 +24,41 @@ missing please open an issue. If you are interested in joining check out
...
 
-- Johan Bergström [@jbergstroem](https://github.com/jbergstroem)
-- João Reis [@joaocgreis](https://github.com/joaocgreis)
-- Rod Vagg [@rvagg](https://github.com/rvagg)
-- Phillip Johnsen [@phillipj](https://github.com/phillipj)
-- Hans Kristian Flaatten [@Starefossen](https://github.com/Starefossen)
+- [@jbergstroem](https://github.com/jbergstroem) - Johan Bergström
+- [@joaocgreis](https://github.com/joaocgreis) - João Reis
+- [@rvagg](https://github.com/rvagg) - Rod Vagg
+- [@phillipj](https://github.com/phillipj) - Phillip Johnsen
```
jasnell: nit: alphabetical order perhaps?

gibfahn:
>nit: alphabetical order perhaps?

@jasnell whoops, should probably have checked that.

Okay, it's not ASCII sorted (i.e. what `vim` does when you do `:sort`). I think this is the way to go, but I'm open to:

1. moving `Starefossen` down to `s`
2. changing it to `starefossen`
3. leaving it as is because it's what most tools will do if you ask them to sort the list.

PR #933, opened 2017-10-22
git: add package-lock.json to .gitignore

gibfahn:
Also removes unneeded `**/` (ignore paths check subdirectories unless
they start with `/`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,14 @@
-node_modules
-**/host_vars/iojs-*
-**/host_vars/node-*
-**/host_vars/test-*
-**/host_vars/release-*
+# tools/jenkins-status:
+node_modules/
+package-lock.json
+# setup secrets files, delete with setup/:
+setup/*/host_vars/iojs-*
+setup/*/host_vars/node-*
+setup/*/host_vars/test-*
+setup/*/host_vars/release-*
+
+# ansible/ secrets files (except templates).
+*.pyc
+ansible/host_vars/*
+!ansible/host_vars/README.md
+!ansible/host_vars/*-template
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,5 +1,14 @@
-node_modules
-**/host_vars/iojs-*
-**/host_vars/node-*
-**/host_vars/test-*
-**/host_vars/release-*
+# tools/jenkins-status:
+node_modules/
```
refack: Non-blocking could this be `/node_modules/`? We might one day want to vendor in a packaged tool

.gitignore
```diff
@@ -1,5 +1,14 @@
...
+package-lock.json
+# setup secrets files, delete with setup/:
+setup/*/host_vars/iojs-*
+setup/*/host_vars/node-*
+setup/*/host_vars/test-*
+setup/*/host_vars/release-*
+
+# ansible/ secrets files (except templates).
+*.pyc
+ansible/host_vars/*
```
refack: Just verify that this is what was meant
```
ansible/host_vars/*
```
means ignore all files in `ansible/host_vars/` I'm not sure it applies recursively to subdirectories.
```
/ansible/host_vars/
```
is probably better (I think)

.gitignore
gibfahn: It could be `/tools/jenkins-status/node_modules/`, but I think defaulting to not vendoring node_modules is a better default, then people can change it if they want to include.

.gitignore
gibfahn: I'm not sure there will ever be directories in `host_vars`, but if there are, maybe we'd want to keep them.

I'd rather leave it as is (this is what was there before) for now, and that way someone can edit the `.gitignore` if they ever want to add subfolders.

PR #956, opened 2017-10-30
[WIP] Add pipeline script for node-inspect

jkrems:
Highly WIP but I guess it's better to put it here before I lose my local docker container...

- [ ] Figure out if there's a maintainable way to do the multi-arch builds
- [ ] Support build against nightlies (quick hack: just support ORG_MIRROR via param?)
- [ ] Support setting the commit/PR status (optional)

See:
* https://github.com/nodejs/build/pull/922
* https://github.com/nodejs/build/issues/838


jenkins/node-inspect.jenkinsfile
```diff
@@ -0,0 +1,83 @@
...
+    ]],
+    userRemoteConfigs: [[
+      url: "https://github.com/${params.projectRepo}",
+      refspec: '+refs/pull/*:refs/remotes/origin/pr/* +refs/heads/*:refs/remotes/origin/*',
+    ]],
+  ])
+}
+
+def runWithNodeVersionLinux(archName) {
+  withEnv(["NVM_DIR=${pwd(tmp: true)}/.nvm"]) {
```
jkrems: Using `nvm` here because I was too lazy to look up how it's done officially in our builds.

jenkins/node-inspect.jenkinsfile
```diff
@@ -0,0 +1,83 @@
...
+    sh 'echo \$NVM_DIR && rm -rf test/*.tap'
+    sh 'curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.6/install.sh | bash'
+    sh "bash -c '. $NVM_DIR/nvm.sh && nvm install ${params.nodeVersion} && npm install && npm test -- --output-file=test/${archName}.tap'"
```
jkrems: `--output-file` depends on https://github.com/nodejs/node-inspect/pull/53

jenkins/node-inspect.jenkinsfile
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env groovy
+
+pipeline {
+  agent any
```
gibfahn: If you do `agent('jenkins-workspace')` does it run on machines with that label? I'm pretty sure that's what we're doing in CI right now.

cc/ @rvagg

jkrems: Yes, using `agent { label 'jenkins-workspace' }` will force execution onto agents with the specified label. From what I can tell in [the docs](https://jenkins.io/doc/book/pipeline/syntax/#agent), the `node(label)` shorthand doesn't seem to exist for `agent`.

jkrems:
I'm really tempted to walk back on the declarative syntax here given the amount of repetition once all possible architectures are added. Unless we'd move that logic to an inline `script {}` section and not model it as parallel `stage`s. Would have to see how that would be rendered in the UI.

refack:
This is what came out https://ci.nodejs.org/job/node-inspect/2/flowGraphTable/
(first steps are checking out the pipeline script itself from @jkrems's fork)

jkrems:
Can someone give https://github.com/nodejs/node-inspect/pull/53 a quick look to get the latest `tap` into master?

jkrems:
For reference, here's how an example run renders in blue ocean when I was testing this script locally:

screen shot 2017-10-31 at 1 44 44 pm

That view makes the job output a lot easier to consume. Unfortunately the out-of-the-box experience of looking at parallel stages in Jenkins is pretty bad (by which I mean pretty much unusable). :(

refack:
> That view makes the job output a lot easier to consume. Unfortunately the out-of-the-box experience of looking at parallel stages in Jenkins is pretty bad (by which I mean pretty much unusable). :(

It's not that bad, it's just not trivial to find:
* each step is avalible from the [flowGraphTable](https://ci.nodejs.org/job/node-inspect/2/flowGraphTable/)
  for example the first failing step:
  https://ci.nodejs.org/job/node-inspect/2/execution/node/43/log/
* An overview of the history of the job can be found on the "Full Stage View"
  image
 --
 image

jkrems:
Yes, I know. But with non-trivial workflows (e.g. parallel + anything happening afterwards) the flattened stage view becomes super confusing. This isn't a problem yet here (since it has just two fake "architectures" right now and nothing happening afterwards) but it won't get any better the more complex the workflow gets. Though maybe it'll be fine for node's builds. I might be scared by our internal deployment pipeline builds (with multiple validation stages consisting of parallelized things).

refack:
@jkrems the log visibility issue was brought up by @joaocgreis and AFAIK is the last hindrance for full adoption.
But as I see it we can continue to use MultiJob. With the current setup the fan-out is the last step of each sub-job and any other steps after that are in a new sub-job. So the flattened log is not that bad, just requires learning where to look for the failures.

On the other hand the step tree:
image
Allows for a not log-reading-dependant way to quickly figure out what failed.

---

tl;dr IMHO we should move away from log-based-groking to using parsing plugins (like tap or tap2junit), while also look-for/develop a better log parser.

sam-github:
Stale pipeline PR related to the pipeline experiment.

PR #971, opened 2017-11-03
Initial macOS ansible scripts

gdams:
Okay so here are my initial ansible scripts for macOS. Obviously the inventory isn't complete yet but I will work with @mhdawson to get that side ready. I have confirmed that this works with a blank machine from macstadium (also ran a test build to confirm all tools were in place).

### Things to note:

this requires `NOPASSWD` to be added to the sudoers file to enable elevation

```bash
sudo visudo
%admin          ALL = (ALL) NOPASSWD:ALL
```

### Possible things to add?

do we want to disable the firewall on the test machines to prevent tonnes of popups?

### DIFF
--- a/ansible/roles/baselayout/tasks/main.yml
+++ b/ansible/roles/baselayout/tasks/main.yml
@@ -17,6 +17,7 @@
 
 - name: set hostname
   when: not os|startswith("smartos") and not os|startswith("zos")
+        and not os|startswith("macos")
   hostname: name="{{ safe_hostname }}"
 
 - name: disable joyent smartconnect
@@ -60,7 +61,19 @@
     state: absent
 
 - name: install packages
-  when: not os|startswith("zos")
+  when: not os|startswith("zos") and not os|startswith("macos")
+  package: name="{{ package }}" state=present
+  loop_control:
+    loop_var: package
+  with_items:
+    # ansible doesn't like empty lists
+    - "{{ packages[os]|default('[]') }}"
+    - "{{ packages[os|stripversion]|default('[]') }}"
+    - "{{ common_packages|default('[]') }}"
+
+- name: install packages (macos)
+  when: os|startswith("macos")
+  become_user: administrator
   package: name="{{ package }}" state=present
   loop_control:
     loop_var: package
@@ -70,6 +83,20 @@
     - "{{ packages[os]|default('[]') }}"
     - "{{ packages[os|stripversion]|default('[]') }}"
     - "{{ common_packages|default('[]') }}"
+
+- name: Check whether /etc/paths contains "/usr/localopt/ccache/libexec" (macos)
+  when: os|startswith("macos")
+  command: grep -Fxq "/usr/localopt/ccache/libexec" /etc/paths
+  register: ccache_mac
+  check_mode: no
+  ignore_errors: yes
+  changed_when: no
+
+- name: add ccache to the path (macos)
+  when: ccache_mac.rc == 1
+  lineinfile: dest=/etc/paths
+    insertbefore=BOF
+    line='/usr/localopt/ccache/libexec'
 
 - name: ubuntu1404 | update package alternatives
   when: os == "ubuntu1404"
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -27,7 +27,7 @@ ntp_service: {
 }
 
 common_packages: [
-  'automake,bash,libtool,sudo',
+  'automake,bash,libtool',
 ]
 
 # you can either add os family or os to this list (see smartos)
@@ -35,45 +35,46 @@ common_packages: [
 packages: {
   centos6_x64: ['centos-release-scl'], # only available on x86_64, 32-bit is from https://copr.fedorainfracloud.org/coprs/mlampe
   centos6: [
-    'ccache,git2u,gcc-c++,devtoolset-6', # even need gcc on centos6 so ccache has symlinks
+    'ccache,git2u,gcc-c++,devtoolset-6,sudo', # even need gcc on centos6 so ccache has symlinks
   ],
 
   centos7_arm64: ['git'], # git2u not available for aarch64 (yet)
-  centos7_x64: ['git2u','centos-release-scl'], # centos-release-scl is required to enable SCLo
+  centos7_x64: ['git2u','centos-release-scl',], # centos-release-scl is required to enable SCLo
                                                # but we do it manually in partials/repo/centos7.yml for arm64
   centos7: [
-    'ccache,gcc-c++,devtoolset-6',
+    'ccache,gcc-c++,devtoolset-6,sudo',
   ],
 
   debian7: [
-    'gcc-4.8,g++-4.8',
+    'gcc-4.8,g++-4.8,sudo',
   ],
 
   debian8: [
-    'ccache,git,gcc-4.9,g++-4.9,libfontconfig1,binutils-2.26',
+    'ccache,git,gcc-4.9,g++-4.9,libfontconfig1,binutils-2.26,sudo',
   ],
 
   debian9: [
-    'gcc-6,g++-6,ccache,git,curl,libfontconfig1,apt-transport-https,ca-certificates',
+    'gcc-6,g++-6,ccache,git,curl,libfontconfig1,apt-transport-https,ca-certificates,sudo',
   ],
 
   fedora: [
-    'bzip2,ccache,gcc-c++,git,fontconfig',
+    'bzip2,ccache,gcc-c++,git,fontconfig,sudo',
   ],
 
   freebsd: [
-    'ccache,git,gmake'
+    'ccache,git,gmake,sudo'
   ],
 
   rhel72: [
-    'gcc-c++',
+    'gcc-c++,sudo',
   ],
 
   smartos: [
     'gccmakedep',
     'git',
     'gmake',
-    'xz'
+    'xz',
+    'sudo',
   ],
 
   smartos14: [
@@ -96,7 +97,7 @@ packages: {
   ],
 
   ubuntu: [
-    'ccache,g++,gcc,git,libfontconfig1',
+    'ccache,g++,gcc,git,libfontconfig1,sudo',
   ],
 
   ubuntu1404: [
--- a/ansible/roles/java-base/tasks/main.yml
+++ b/ansible/roles/java-base/tasks/main.yml
@@ -48,12 +48,18 @@
 
 # if this fails you want to check in vars/main.yml and add package name
 # as appropriate -- try to use generic os family if available.
+
+- name: Check if java is already installed
+  raw: java -version
+  register: java
+  ignore_errors: yes
+
 - name: install java
-  when: not os|startswith("zos") and arch != "ppc64" and not inventory_hostname|regex_search('-arm(v6l|v7l|64)_pi')
+  when: java.rc > 0 and not os|startswith("zos") and arch != "ppc64" and not inventory_hostname|regex_search('-arm(v6l|v7l|64)_pi')
   package: name="{{ java_package_name }}" state=present
 
 - name: install webupd8 oracle java 8 extras
-  when: os in ("ubuntu1204", "ubuntu1404") and arch != "ppc64"
+  when: java.rc > 0 and os in ("ubuntu1204", "ubuntu1404") and arch != "ppc64"
   package: name="{{item}}" state=present
   with_items:
     - ca-certificates
--- a/ansible/roles/jenkins-worker/tasks/main.yml
+++ b/ansible/roles/jenkins-worker/tasks/main.yml
@@ -67,11 +67,11 @@
     line: ::1 localhost.localdomain localhost
 
 - name: create NODE_TEST_DIR directory
-  file: path=/home/{{ server_user }}/tmp state=directory
+  file: path="{{ home }}/{{ server_user }}/tmp" state=directory
 
 - name: set NODE_TEST_DIR permission and owner
   file:
-    path: /home/{{ server_user }}/tmp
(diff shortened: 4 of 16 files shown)


ansible/inventory.yml
```diff
@@ -130,6 +130,9 @@ hosts:
         ubuntu1604-arm64_odroid_c2-2: {ip: 70.167.220.148, user: odroid}
         ubuntu1604-arm64_odroid_c2-3: {ip: 70.167.220.149, user: odroid}
 
+    - macstadium:
+        macos10.12-x64-1: {ip: 207.254.58.162, port: 5, user: administrator}
+
```
mhdawson: Is that the right port ? I thought it was going to be something like 10005

ansible/partials/install-homebrew.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+yes | /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
```
mhdawson: I assume this would always install the latest ?  I'm wondering if we should have it install some fixed version ?

ansible/roles/baselayout/vars/main.yml
```diff
@@ -88,7 +98,8 @@ packages: {
     'gccmakedep',
     'git',
     'gmake',
-    'xz'
+    'xz',
+    'sudo',
```
mhdawson: Just wondering why we need this now and we did not before ?

ansible/roles/java-base/tasks/main.yml
```diff
@@ -29,13 +29,26 @@
...
 - name: install webupd8 oracle java 8 extras
-  when: os in ("ubuntu1204", "ubuntu1404")
+  when: java.rc > 0 and os in ("ubuntu1204", "ubuntu1404")
   package: name="{{item}}" state=present
   with_items:
     - ca-certificates
     - oracle-java8-set-default
+
+- name: install java (macos)
+  become_user: administrator
```
mhdawson: I think the become_user can be set in one place for macos as opposed to in individual jobs

ansible/roles/jenkins-worker/tasks/main.yml
```diff
@@ -97,6 +97,14 @@
     dest: "{{ jenkins.dest }}"
     mode: "{{ jenkins.mode|default('0644') }}"
 
+# TODO - Should this run on every machine?
+- name: copy start.sh to {{ home }}/{{ server_user }}/start.sh
```
gibfahn: cc/ @jbergstroem @rvagg @joaocgreis , should we still be putting start.sh on every machine? The new ansible scripts don't so far, but IDK if that's an oversight or a conscious decision.

ansible/roles/jenkins-worker/tasks/main.yml
```diff
@@ -97,6 +97,14 @@
     dest: "{{ jenkins.dest }}"
     mode: "{{ jenkins.mode|default('0644') }}"
 
+# TODO - Should this run on every machine?
```
mhdawson: There is already config eslsewhere that runs/configures what is needed to start the jenkins agent.  Look in roles/jenkins-worker/vars/main.yml and related which sets up which scripts are run to start the jenkins agent.

ansible/roles/baselayout/vars/main.yml
gibfahn: It used to be in the always install section, but you can't brew install it so now it's moved to all the existing ones.

ansible/partials/install-homebrew.sh
gibfahn: I think with brew it's probably better to stay on latest, there's no concept of LTS versions, and bugs get found and fixed quite frequently.

ansible/roles/jenkins-worker/tasks/main.yml
gibfahn: Yep, but we used to always have `~/start.sh` as well, question here is whether we still need start.sh if we have other startup methods.

I'd suggest we do, because it gives us a consistent way to restart machines, without having to remember all the different commands for different architectures.

ansible/inventory.yml
gdams: This is just a test machine that I spun up, ignore the port for now

ansible/partials/install-homebrew.sh
gdams: Yeah I agree with @gib on this one

ansible/roles/jenkins-worker/tasks/main.yml
gdams: It is worth noting that we need start.sh for the mac machines to start automatically. The question is whether we want this on other platforms too

ansible/roles/java-base/tasks/main.yml
gdams: There are certain places where we wan't to run jobs as root on macOS. The main issue is that brew commands cannot be run as root (hence why I dropped the privileges down for all brew commands)

joaocgreis:
I don't think `start.sh` should be in all machines. There should be some kind of mechanism everywhere to restart Jenkins when it crashes, which is (or was) a relatively frequent issue. So, if Jenkins does not start because of the restart mechanism, we should debug the restart mechanism. If Jenkins starts and we want it stopped for some reason, we must always disable the restart mechanism anyway. Having `start.sh` everywhere will be very misleading and may not be kept in sync with the other way Jenkins is launched.

`start.sh` was traditionally used with `monit`. Here, I see it used with `launchctl` and a `KeepAlive=true` config entry. If that does not work, then we should debug it, not start Jenkins manually and risk having two running.

About this PR, shouldn't the two files under `ansible/partials/` be in `ansible/roles/package-upgrade/files/` instead, since they're only required for `package-upgrade`?

This shouldn't land with the current inventory. A partial one is ok, or even no inventory entries. Also, Gibson raised a valid point about the duplicated `sudo` entry. After these are addressed LGTM. Thanks for doing this!

gibfahn:
>I don't think start.sh should be in all  machines. There should be some kind of mechanism everywhere to restart  Jenkins when it crashes, which is (or was) a relatively frequent issue.  So, if Jenkins does not start because of the restart mechanism, we  should debug the restart mechanism. If Jenkins starts and we want it  stopped for some reason, we must always disable the restart mechanism  anyway. Having start.sh everywhere will be very misleading and may not be kept in sync with the other way Jenkins is launched.

I was thinking that maybe `~iojs/start.sh` could just call the relevant service restart command (to avoid people having to remember what it is for each platform). Might be confusing to call it `start.sh` in that case though.

joaocgreis:
> I was thinking that maybe `~iojs/start.sh` could just call the relevant service restart command

That would be quite nice! The `jenkins-worker` role should probably be separated by the boot system into partials, perhaps with a common section. It would be nice to include a `start.sh` (or restart?) script that does the right thing there, different for each start system.

ansible/roles/jenkins-worker/tasks/main.yml
mhdawson: I think we used to have start.sh only because we had not investigated/implemented the more natural way to do the same thing for each os flavor.

mhdawson:
Agreed it would be nice if there was a script that could be used to start/stop which calls the version for the specific platform.  Maybe "agent-control" with start and stop options ?

ansible/roles/baselayout/tasks/main.yml
```diff
@@ -51,7 +51,19 @@
...
+    loop_var: package
+  with_items:
+    # ansible doesn't like empty lists
+    - "{{ packages[os]|default('[]') }}"
+    - "{{ packages[os|stripversion]|default('[]') }}"
+    - "{{ common_packages|default('[]') }}"
+
+- name: install packages (macos)
+  when: os|startswith("macos")
+  become_user: administrator
```
mhdawson: Could parameterize this instead.  (Fine if in follow on PR).

ansible/roles/baselayout/vars/main.yml
mhdawson: got it

ansible/roles/java-base/tasks/main.yml
```diff
@@ -29,13 +29,26 @@
...
 # as appropriate -- try to use generic os family if available.
+
+- name: Check if java is already installed
+  raw: java -version
+  register: java
+  ignore_errors: yes
+
 - name: install java
-  when: not os|startswith("zos")
+  when: java.rc > 0 and not os|startswith("zos") and not os|startswith("macos")
```
mhdawson: Is there a way to validate that it meets our minimum level (java 8 now)

refack:
Can we land this, and iterate later? On just land the inventory?

refack:
BTW the libuv jobs need `GYP` to be in `~/gyp` so meanwhile I added a curl call in the job:
https://ci.nodejs.org/job/libuv-test-commit-osx/jobConfigHistory/showDiffFiles?timestamp1=2017-11-22_22-16-55&timestamp2=2017-12-01_15-30-59

ansible/plugins/library/ssh_config.py
```diff
@@ -47,7 +47,11 @@
   HostName {{ metadata.ansible_host }}
   IdentityFile {{ metadata.ansible_ssh_private_key_file }}
   User {{ metadata.ansible_user or 'root' }}
-{{ '  ' + ssh_arg|multi_replace(args) + '\n' if ssh_arg }}
+{{
```
refack: I added this code to enable rendering the provided port to the config.
@gdams feel free to edit or remove if you don't like this.

mhdawson:
I'm all for landing minus the inventory and then updating once we have additional changes (like adding in support for ccache)

gibfahn:
>I'm all for landing minus the inventory and then updating once we have additional changes (like adding in support for ccache)

SGTM, but it'd be good to add some TODOs to the source code (like for adding ccache) so we've noted the gaps.

gdams:
We should probably add the final bit for ccache and then I think this is good to land. It should just be a case of getting ccache from homebrew and then setting some symlinks

ansible/roles/baselayout/tasks/main.yml
```diff
@@ -73,6 +73,12 @@
     - "{{ packages[os|stripversion]|default('[]') }}"
     - "{{ common_packages|default('[]') }}"
 
+- name: add ccache to the path (macos)
+  when: os|startswith("macos")
+  lineinfile: dest=/etc/paths
+    insertbefore=BOF
+    line='/usr/localopt/ccache/libexe'
```
gibfahn: `libexec`

gdams: updated

gdams:
Okay ccache bits added

mhdawson:
@gdams have you updated the existing machines after the update to add ccache ? If so I'll kick off a run to check the speed and then we should move forward in adding the rest of the machines that don't have test failures.

gibfahn:
@gdams we should enable putting core dumps in the working directory instead of in `/cores`, that way they'll get cleaned up when jobs finish, rather than being owned by root and filling up the machine.

https://apple.stackexchange.com/questions/108861/can-os-x-dump-cores-in-the-working-directory-instead-of-cores

mhdawson:
@gdams after holiday ping to see if I should check the speed and then add machines to builds.

maclover7:
ping @gdams @gibfahn @mhdawson

maclover7:
ping @gdams @gibfahn @mhdawson

gibfahn:
>ping @gdams @gibfahn @mhdawson

Looks like this needs a rebase from George.

mhdawson:
It was just waiting on a fix up for cache. But I'm +1 for landing and then fixing that

gibfahn:
>So we are able to easily install ccache. My only remaining issue is adding ccache to the path of the jenkins node. I added the path to ccache in etc/paths and this works locally but I am unable to make this extra path appear in jenkins. (I have even tried restarting the machine)

@gdams is `/etc/paths` definitely added even when things are launched directly (not through a terminal)?

gdams:
>is `/etc/paths` definitely added even when things are launched directly (not through a terminal)?

@gibfahn I'm not 100% sure. I'll do some investigation.

rvagg:
I needed the macstadium hosts in my ssh config and @Trott couldn't figure it out to go fix some broken hosts so I've pulled out the inventory.yml and plugin fixes here and pushed them to master as 0495bc2. It'd be good if we could get this merged soon, it can be iterated on once merged if it's not fully working yet.

mhdawson:
I'm +1 on merging and then updating if we need to. @gdams can you rebase and then we'll do that?

gibfahn:
>It'd be good if we could get this merged soon, it can be iterated on once merged if it's not fully working yet.

Yep, we an just leave TODOs where things aren't working yet (or need manual setup).

gdams:
updated PTAL

ansible/inventory.yml
```diff
@@ -144,6 +144,10 @@ hosts:
         macos10.12-x64-1: {ip: 207.254.58.162, port: 10001, user: administrator}
         macos10.12-x64-2: {ip: 207.254.58.162, port: 10002, user: administrator}
 
+<<<<<<< HEAD
+=======
+
+>>>>>>> Initial macOS ansible scripts
```
mhdawson: This looks like a merge prroblem

gdams:
@mhdawson I think those merge issues are now cleared up

mhdawson:
Ok, will try to land by the end of week unless there are objections, comments from other @nodejs/build team members.

mhdawson:
@gdams can you squash your commits.   git am is failing with conflicts even with -3.  I suspect that if you squash that will get rid of the issue.

gdams:
@mhdawson rebased and squashed

mhdawson:
Landed as eb68f591fdd5e07fdd705d1e9fa5316f85d87335

PR #924, opened 2017-10-19
Arm backup

mhdawson:
Changes to ARM Ansible scripts in order to help support an off-site ARM backup for ARM 6

PR #974, opened 2017-11-04
Lots of Raspberry Pi improvements

rvagg:
Closed #841 and am starting again here with a whole lot of improvements. Also includes @mhdawson's updates for his backup Pi @ #924. I've included a bunch more things from the manual list so now all you need to do manually is set password, ssh key and enable ssh.

nfs problems are solved with this too, I went with a similar solution to the one in #924 but just put the mount in /etc/rc.local.

Overclocking is now set by this script, it sets Pi2 to 1000Mhz. the test Pi1 B+ to 800Mhz and release Pi1 B+'s to the standard 700Mhz speed.

A bunch of other things in this too, works really well. I've run it across the whole cluster here. I have 3 Pi1's down right now that I'm going to pull out and do some r/w testing on their SD cards and will reprovision. So I'll get to try this out for new machines shortly too.


rvagg:
nfs problems messed up the Pi3's, see #975, narrowed down to a timing problem, things coming up too quick and out of order. I've inserted a delay for mounting and another delay for starting Jenkins in 9731170d85.

rvagg:
More tweaks after reprovisioning 3 of them and running this against them. Happy with this now. I'm going to assume that nobody else wants to review this or has the experience to do so so I'm just going to merge it.

Issue #200, opened 2015-09-23
missing crontab entries from website server setup


rvagg:
A TODO for myself probably but I'm noticing that /etc/crontab on the server doesn't match what the scripts do to set it up, my fault for not synchronising them:

```
*  *    * * *   root    /home/nodejs/cdn-purge.sh
*/5 *   * * *   dist    /home/staging/promote/promote_nightly.sh nodejs
*/5 *   * * *   dist    /home/staging/promote/promote_nightly.sh iojs
*  *    * * *   nodejs  /home/nodejs/check-build-site.sh nodejs
*  *    * * *   nodejs  /home/nodejs/check-build-site.sh iojs
*/5 *   * * *   dist    /home/staging/tools/promote/promote_nightly.sh nodejs
*/5 *   * * *   dist    /home/staging/tools/promote/promote_nightly.sh iojs
```

Only the first line is done, in tasks/site-setup.yaml

rvagg:
also need to add nightly builder to crontab, that's not there yet

maclover7:
ping @rvagg -- does this need to stay open?

rvagg:
addressed in 976

PR #976, opened 2017-11-05
www: sync w/ current nodejs.org config

rvagg:
Addresses #200 and a few other minor things that are on the server but are not reflected in the scripts for it.

Adds the log-collect stuff that's on there collecting logs for us from cloudflare and doing some basic filtering. There's also some stuff in there to convert the json format to nginx format but it's not being used for anything currently. The log-collect work is only partially complete since we're collecting but not using it for metrics generation yet. We still need to (a) confirm that we're successfully collecting everything with this procedure (logshare isn't the most robust tool unfortunately, or maybe it's their API, hence the sanity-check script in here), and (b) either do the json->nginx conversion for metrics, or make metrics scripts able to read the json files as they are. Either way, this PR reflects current state so is good to merge IMO.

I haven't tested this .. a bit of a stab in the dark, I'm not willing to run this against the live server.

### DIFF
--- a/setup/www/resources/scripts/sync-benchmarking.sh
+++ b/setup/www/resources/scripts/sync-benchmarking.sh
@@ -3,7 +3,8 @@
 pidfile=/var/run/nodejs/sync-benchmarking.pid
 cmd="
   rsync -aqz --delete benchmark:charts/ /home/www/benchmarking/charts/ &&
-  rsync -aqz --delete benchmark:coverage-out/out/ /home/www/coverage/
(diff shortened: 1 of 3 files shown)


maclover7:
Hmm, maybe better to pull the `log-collect` stuff out into a different PR, and land everything else first? The rest of the changes don't look too bad.

rvagg:
OK, removed log-collect from here, will push that in a moment. I've added an upgrade to Node 8 in this too, see https://github.com/nodejs/nodejs.org/pull/1459 for website team's agreement on this.

PR #978, opened 2017-11-05
jenkins: send statuses for non-nodejs/node repos

maclover7:
Related #922. At some point, the CITGM builds can be updated to use this as well.

### DIFF
--- a/jenkins/pipelines/post-build-status-update.jenkinsfile
+++ b/jenkins/pipelines/post-build-status-update.jenkinsfile
@@ -1,8 +1,8 @@
 #!/usr/bin/env groovy
 
 // DESCRIPTION:
-// Sends the status of a node-test-commit-* sub build to the Github Bot, which
-// then updates the corresponding PR's CI status on github.com.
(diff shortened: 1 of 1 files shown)


refack:
And `node-gyp` (which I just noticed today don't receive statuses).
We'll need to do some changes in the bot for `libuv` to receive these as well.

gibfahn:
> We'll need to do some changes in the bot for libuv to receive these as well.

If we're going to do libuv, we'd probably need to add an `org` parameter to this PR (or change `REPO` to accept `nodejs/node`. But I assume that'd require changes to `github-bot` as well.

refack:
> But I assume that'd require changes to github-bot as well.

1. Need to verify the bot has permissions to the `libuv` org.
2. We could do a simpler hack in the bot so that `libuv` == `libuv/libuv`

maclover7:
Going to land this for now-- as this job continues to be used by others, we can keep making changes as necessary :)

maclover7:
Landed in 6b050dc70bd4aa4e7914f694927ef01b7b1d48a9

PR #981, opened 2017-11-06
pipeline: check inputs in post-build-status-update

gibfahn:
Make sure PR is a number, and the `POST_STATUS_TO_PR` is set.

Refs: https://github.com/nodejs/build/pull/973#issuecomment-342130374

### DIFF
--- a/jenkins/pipelines/post-build-status-update.jenkinsfile
+++ b/jenkins/pipelines/post-build-status-update.jenkinsfile
@@ -30,10 +30,8 @@ pipeline {
 
 def sendBuildStatus(params) {
     // Fail loudly if upstream job is configured wrong.
-    for (def i in params) {
-        if (i.value == '') {
-            error('All parameter fields are required, ${i.key} was blank.')
-        }
+    if (params.any{ it.value == '' }) {
+        error('All parameter fields are required.')
     }
 
     // Fail quietly if upstream job shouldn't be posting to GitHub.
--- a/jenkins/pipelines/post-build-status-update.jenkinsfile
+++ b/jenkins/pipelines/post-build-status-update.jenkinsfile
@@ -29,6 +29,14 @@ pipeline {
 }
 
 def sendBuildStatus(repo, identifier, status, url, commit, ref) {
+
+    // Fail quietly if upstream job shouldn't be posting to GitHub.
+    if (!PR.isInteger() || POST_STATUS_TO_PR != 'true') {
+        println "Skipping status update (PR.isInteger(): ${PR.isInteger}, " +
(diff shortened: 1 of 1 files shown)


jenkins/pipelines/post-build-status-update.jenkinsfile
```diff
@@ -29,6 +29,14 @@ pipeline {
 }
 
 def sendBuildStatus(repo, identifier, status, url, commit, ref) {
+
+    // Fail quietly if upstream job shouldn't be posting to GitHub.
+    if (!PR.isInteger() || POST_STATUS_TO_PR != 'true') {
```
maclover7: - Should `PR.isInteger()` change to checking if `ref` includes `refs/pull`? We never explicitly send `PR` to `post-build-status-update`

- `POST_STATUS_TO_PR` will need to be a new string parameter and variable passed to this function

gibfahn: My bad on the first.

For the second, are parameters not passed down as environment variables in pipelines? If not that's a pain.

maclover7: So yes, the parameters are passed down as environment variables, but there are two annoying thing at play here -- 1) `sendBuildStatus` is a separate function, so we have to pass all parameters to it explicitly 2) `POST_STATUS_TO_PR` also needs to be defined as a parameter for the pipeline

gibfahn: How about doing this (what I just pushed)?

jenkins/pipelines/post-build-status-update.jenkinsfile
```diff
@@ -10,28 +10,41 @@ pipeline {
...
 
-def sendBuildStatus(repo, identifier, status, url, commit, ref) {
+def sendBuildStatus(params) {
+    // Fail loudly if upstream job is configured wrong.
+    if (params.any{ it.value == '' }) {
+        error('All parameter fields are required.')
+    }
+
+    // Fail quietly if upstream job shouldn't be posting to GitHub.
+    if (params.REF.contains('refs/pull/') || params.POST_STATUS_TO_PR != 'true') {
```
refack: `!`:
`!params.REF.contains('refs/pull/')`

refack:
Sidebar: why do we keep the `POST_STATUS_TO_PR` arg? Why not try every time?

maclover7:
ping @gibfahn

maclover7:
ping @gibfahn

sam-github:
Stale pipeline PR related to the pipeline experiment.

PR #985, opened 2017-11-07
github-bot: add IPs of Jenkins workers

phillipj:
Providing `$JENKINS_WORKER_IPS` environment variable at startup will activate the whitelist implemented in the bot, validating who's allowed to pushed Jenkins job updates as inline PR statuses on github.com.

Refs https://github.com/nodejs/github-bot/pull/142

/cc @nodejs/github-bot

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -197,6 +197,8 @@ Unsorted stuff of things we need to do/think about
 - [ ] add explicit ARCH and DESTCPU for release machines (RV: I'm adding
       "arm64" manually for both to force the right thing, from memory I've
       needed to do this on x86 and x64, best be explicit to be sure)
+- [ ] github-bot: automate list of whitelisted Jenkins worker IPs with
+      python
 
 [callback]: plugins/inventory/nodejs_yaml.py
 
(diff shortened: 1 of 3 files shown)


phillipj:
@maclover7 thanks! Pushed a commit adding a comment with a friendly reminder.

maclover7:
@phillipj This should be ready to go, right? Would you be able to merge this, and check and see if it helps at all with https://github.com/nodejs/node/pull/17151#issuecomment-346528636?

phillipj:
@maclover7 I'm not able to immediately, but I'll at least have a look tomorrow.

I'm unsure if it would help any issues ATM, if anything probably make it worse? Currently it will happily respond to requests from *anywhere*, having this whitelist activated could if anything, make it worse? Or have I misunderstood the issue in the PR you referenced?

rvagg:
LGTM, although it should be possible to insert some python in plugins/inventory/nodejs_yaml.py to pull out the IP addresses with the `jenkins-workspace` alias and insert them into a new context variable that you can then use in setup/github-bot/resources/environment-file. If you can't figure out how to do this then this change is a good first step. There's a checklist on the README in this ansible directory where you could add a TODO for doing this work.

But as I've mentioned in https://github.com/nodejs/node/pull/17151#issuecomment-346528636 I think the effect of this on non-pipeline jobs is that it's going to stop status reporting completely because the `curl`ing is happening on worker machines, not the workspace ones that are exclusively used by Jenkins.

@phillipj do you have access to update the bot machine with this change?

phillipj:
> @phillipj do you have access to update the bot machine with this change?

Yepp. I'll merge this and update the env variables on the machine out shortly.

phillipj:
@maclover7 this is live in production now.

Issue #108, opened 2015-06-05
Additional builders: shared libraries and debug


jbergstroem:
I'd like to improve testing against shared libraries as well as a debug build. The debug-build is a no-brainer (`./configure --debug`), but the shared build would introduce a bit of flavour.

Since quite a few iojs/nodejs packagers (linux distributions, freebsd, pkgsrc) uses at least a shared zlib and openssl the optimal scenario would be being able to use a package manager to maintain these dependencies instead of building them ahead of every jenkins run. My vision is to – as closely as possible – represent how it's actually used.

My plan is to use archlinux as a base os since it has most of our dependencies both packaged and up to date. If anyone has input on a more suiting distribution, I'm all ears. Since I use Gentoo myself I'm just going to exclude it on the grounds of having enough coverage already (it's a pretty good distribution for working with multiple toolchains and dependencies).

Thoughts?

bnoordhuis:
Just curious, how are you going to deal with floating patches?

jbergstroem:
@bnoordhuis my view is that we should respect abi and not apply floating patches without a high amount of resistance. I for instance PR'ed removing building against a shared c-ares for this reason, but optimally we'd upstream patches and let them flow through there before patching them locally. I wouldn't really see this as a build problem, rather a problem with how the iojs WG/TC handles patches in `deps/`.

bnoordhuis:
Most floating patches aren't ABI breaks, they're bug fixes.  c-ares is an exception because upstream is no longer actively developed.

jbergstroem:
Regardless, the bug fixes should go upstream no? Lets run through them:
 - libuv: you and saghul. I'll assume that whatever we patch we want upstream
 - http_parser: indutny and more. Same thing?
 - openssl: I doubt we'll accept floating patches here that will affect shared vs bundled (for instance the cli patch that we might land to improve windows buildbot status).
 - v8: we might float for a version, but everything we've done so far afaik is upstreamed.

I think we should let whatever bug we fix in tree fail until it lands in upstream. With that said, having it part of the incremental build might not be what we want. We could definitely start by having it part of `iojs+any-pr+multi` until we don't.

bnoordhuis:
I don't think it's that clear cut.  To take that OpenSSL CLI patch as an example, it fixes an issue for us but it's not the kind of patch that upstream will gladly (or ever) take.  Now what?

[...]

PR #992, opened 2017-11-09
docker: add initial docker host playbook and dockerfiles

rvagg:
Moving on from #989 which is trying to use the Infrastructure Containers on Joyent to run Alpine. This sets up a "Docker host" similar to the initial work started in #437 (that PR shows how we're running our Alpine 3.4 test host(s) right now in CI).

Configuring a Docker host means first setting up a machine that can handle the number of containers we need, then using `host_vars` to configure the containers. At the moment I have 2 DIgitalOcean 4 VCPU, 8G machines running and have them set up with this type of `host_vars`:

```yaml
containers:
  - { name: 'test-digitalocean-alpine34_container-x64-1', os: 'alpine34', secret: 'abc123' }
  - { name: 'test-digitalocean-alpine35_container-x64-1', os: 'alpine35', secret: 'abc456' }
  - { name: 'test-digitalocean-alpine36_container-x64-1', os: 'alpine36', secret: 'abc567' }
  - { name: 'test-digitalocean-ubuntu1604_container-x64-1', os: 'ubuntu1604', secret: 'abc890' }
```

So we have 2 x Alpine 3.4, 2 x Alpine 3.5, 2 x Alpine 3.6 and 2 x Ubuntu 16.04. The Alpine machines are labelled in CI like this: `alpine36-container-x64` and the 3 of them are included now in node-test-commit-linux. The Ubuntu containers have the label `jenkins-beta` as @refack suggested we have some floating workers that can be used to test new jobs. They're not assigned to anything so feel free to use them in CI as you like, just don't attach them permanently to anything I suppose.

Dockerfile templates are in `ansible/roles/docker/templates/` and we can add extra types of images in here as required. I could imagine falling back to Docker to run our CentOS5 instances if we lost our current hosts and got desperate for example.

On the host, the containers mount their `/home/iojs/` directories inside the host `/home/iojs/` directory under a subdirectory that has the worker's name. So on test-digitalocean-ubuntu1604-docker-x64-1 we have:

```
# ls /home/iojs/
test-digitalocean-alpine34_container-x64-1  test-digitalocean-alpine36_container-x64-1
test-digitalocean-alpine35_container-x64-1  test-digitalocean-ubuntu1604_container-x64-1
```

Each of those directories looks like a standard `/home/iojs`, complete with `slave.jar`, `.ccache` and `tmp`.

The containers are managed by systemd and have the worker name in them, e.g.:

```
# systemctl list-units | grep jenkins
  jenkins-test-digitalocean-alpine34_container-x64-1.service                                            loaded    active running   Jenkins Slave in Docker for test-digitalocean-alpine34_container-x64-1
  jenkins-test-digitalocean-alpine35_container-x64-1.service                                            loaded    active running   Jenkins Slave in Docker for test-digitalocean-alpine35_container-x64-1
  jenkins-test-digitalocean-alpine36_container-x64-1.service                                            loaded    active running   Jenkins Slave in Docker for test-digitalocean-alpine36_container-x64-1
  jenkins-test-digitalocean-ubuntu1604_container-x64-1.service                                          loaded    active running   Jenkins Slave in Docker for test-digitalocean-ubuntu1604_container-x64-1
```

So you could `systemctl restart jenkins-test-digitalocean-ubuntu1604_container-x64-1` for example.

The two hosts included in the inventory here are the Docker hosts and build/test have access to them to do all of this stuff. Jenkins doesn't know about the hosts, only the workers, but you don't have ssh access to the workers. In Jenkins, I've added descriptions for these to make it clear where the host is if you ever needed to "manage" it. e.g. https://ci.nodejs.org/computer/test-digitalocean-alpine36_container-x64-1/ says "Docker container running on test-digitalocean-ubuntu1604_docker-x64-1 in /home/iojs/test-digitalocean-alpine36_container-x64-1".


rvagg:
As mentioned in https://github.com/nodejs/build/issues/108#issuecomment-343143008 I've started work on a shared-openssl job for 1.1.0g based in a docker container. I've been tinkering with node-test-commit-linux-fips and think we should just roll it into the same set of jobs along with other shared library builds and also a Debug build. Basically we'd use a pool of Ubuntu 16.04 containers that are running on 2 or 3 beefier machines (with `JOBS` relatively high so they can soak up unused capacity where possible) that can run all of these things. node-test-commit-linux-fips currently downloads and compiles the FIPS OpenSSL but I think we should just bake this into the container along with other shared libraries. Then our configs are in Dockerfiles inside our Ansible scripts and we can update everything there and we don't get the extra build overhead for what is identical on each run.

rvagg:
merged for now, can tweak in additional PRs

rvagg:
oh, this includes the "sharedlibs" Ubuntu container that can run OpenSSL 1.1.0 and OpenSSL-FIPS tests too, currently playing with this here: https://ci.nodejs.org/view/All/job/node-test-commit-linux-linked/

PR #993, opened 2017-11-09
ccache: ccache path on fedora

rvagg:
Noticed some of the linux hosts being much slower than they should be, found that fedora hosts aren't properly using ccache so I've fixed up the path to allow for their special location.

It's not only Fedora that's slow, still need to figure out why some of the others are going slow. I've noticed Ubuntu 12.04 and 17.10 going slower than they should. With cache they should rip through these builds.


rvagg:
centos7 arm64 hosts suffering from the same problem, I've run ansible across all of `*centos7*` now. Need to follow-up a bit later to see if it's being used (using a simple `du -s /home/iojs/.ccache` across all the worker hosts to check).

PR #995, opened 2017-11-11
doc: note other accesses for nodejs_build_infra

gibfahn:
I'll add these as I get them :) .

@rvagg I got the invites for Packet and DigitalOcean, should I be expecting others, or are the other ones shared passwords?

### DIFF
--- a/doc/access.md
+++ b/doc/access.md
@@ -23,8 +23,25 @@ A subsection of build members have access to infra machines
 (`infra-*`). The list of members is [here][Infra Admins].
 
 The infra group also have access to:
-- [DigitalOcean Droplets][]
-- [Packet.net][]
+
+#### Servers:
+- [DigitalOcean Droplets][] (individual accounts)
+- [Joyent][]
+- [MacStadium][]
+- [Packet.net][] (individual accounts)
+- [Rackspace][] (individual accounts)
+- [Scaleway][]
+- [SoftLayer][] (individual accounts)
+- [linuxOne][]
+
+#### Certificates
+- [Apple][]
+- [Digicert for Authenticode][]
+- [GoGetSSL for SSL][]
+
+#### Other
+- [Cloudflare][]
+- [Mailgun email][] (uses Rackspace login)
 
 ### Release machines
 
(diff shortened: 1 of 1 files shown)


doc/access.md
```diff
@@ -22,6 +22,10 @@ of members is [here][Build WG Members].
 A subsection of build members have access to infra machines
 (`infra-*`). The list of members is [here][Infra Admins].
 
+The infra group also have access to:
```
maclover7: have --> has

rvagg:
Have a look in admin_logins.md in secrets/build/infra, most of them are in there. I think softlayer might be another one you need an individual account for but I'm not sure how to do that (I haven't looked yet, will try and remember to do so when I'm at the computer next)

doc/access.md
gibfahn: Good nit! I think a group can be singular or plural though ([SO answer](https://english.stackexchange.com/questions/39838/is-group-singular-or-plural)).

rvagg:
Servers:

* MacStadium (shared)
* Digitalocean (individual)
* Rackspace (shared)
* Joyent (shared)
* linuxOne (shared)
* Scaleway (shared)
* packet.net (individual)
* SoftLayer (individual) _(@gibfahn can you talk to @mhdawson about getting a login here? I can't see an option to do so which I think means I don't have the access)._

Certificates:

* Apple (shared)
* DigiCert for Authenticode (I've just added this to admin_logins.md) (shared)
* GoGetSSL for SSL (I've just added this to admin_logins.md) (shared)

Other infra:

* Cloudflare (shared)
* Mailgun email (accessible with Rackspace login)

gibfahn:
@rvagg @mhdawson I've added the accounts you mentioned, with links culled from Google.

PTAL.

doc/access.md
```diff
@@ -22,6 +22,27 @@ of members is [here][Build WG Members].
...
+
+#### Servers:
+- [DigitalOcean Droplets][] (individual accounts)
+- [Joyent][]
+- [MacStadium][]
+- [Packet.net][] (individual accounts)
+- [Rackspace][]
```
rvagg: we should make this individual accounts, this is possible now and I have one

doc/access.md
```diff
@@ -22,6 +22,27 @@ of members is [here][Build WG Members].
...
+- [SoftLayer][] (individual accounts)
+- [linuxOne][]
+
+#### Certificates
+- [Apple][]
+- [Digicert for Authenticode][]
+- [GoGetSSL for SSL][]
+
+#### Other
+- [Cloudflare][]
```
rvagg: Just a note here, this is using my email address but we should change it to an @iojs.org account, it sends 2fa tokens via email when you log in with a new IP and timezone differences make that difficult across the team to coordinate. It's not a trivial change, however, as we're using the CF API in a couple of places and it uses the email address as account key iirc. So ... don't let me forget this is something we need to do.

rvagg:
lgtm other than that single nit

doc/access.md
gibfahn: https://github.com/nodejs/build/issues/1072

doc/access.md
gibfahn: https://github.com/nodejs/build/pull/995#discussion_r159559596

gibfahn:
@rvagg created issues for those two things, do either of them require changes to this PR?

rvagg:
@gibfahn I think the Rackspace one can be done here, just note individual accounts here and we'll use the other issue to make sure each of us have it set up.

Issue #926, opened 2017-10-19
Permission to start culling all pre-v4 infra and build logic


rvagg:
This is something that we should make explicit and get full buy-in from at least the Build team, (if not the TSC itself?). We have a bunch of pre-v4 stuff cluttering up our resources that, if we had explicit clearance to start removing, I think team members would happily start pulling out. The main bits I can think of are in Jenkins, there's a ton of logic around pre-v4 builds in the scripts we have and also for build node labels. There's at least one build resource I can think of (smartos) that is still _only_ required for pre-v4 builds too.

To be clear, this would cover: v0.10, v0.12, and all of io.js. We're essentially precluding the ability to release, or even test off any of these old branches by approving this.

@nodejs/build please 👍 this issue or +1 in some way if you support this and give permission to clean house if anyone stumbles across anything that can be removed; otherwise please explain why you don't support it.

refack:
Only issue I have is `node-gyp@3.x.x`. We are committed to support node > 0.8.0 with that branch, and we also haven't yet cut a `node-gyp@4.0.0` (which drops node < 4).

I'm +1 on cleaning up hacks from Jenkins jobs.

BTW: when was the most recent pre 4 build done / what is the probability we'll have to cut another release?

refack:
Just remembered the `node-gyp` jobs use prebuilt node binaries by definition, so not a real issue.

joaocgreis:
Yes please. I would be very surprised if it all still works. Removing all that logic will make the jobs significantly cleaner and easier to edit. I haven't seen any pre-v4 jobs running in a long time.

This would allow us to remove `NODES_SUBSET` right away, I don't believe the target for docs only is used or usable now (full CI should run for every PR, I don't know why it was there in the first place). Ref: https://github.com/nodejs/build/issues/696#issuecomment-297087910

rvagg:
Not sure if this is exactly the latest for pre-v4 but the last v0.10 was v0.10.28 on Oct 18 2016, the last v0.12 was v0.12.18 on Feb 22 of this year. io.js was last built on Sept 15 2015, that was v3.3.1.

[...]

PR #997, opened 2017-11-13
doc: note that release config is removed after 6m

gibfahn:
Anything stored in GitHub will still be available in the Git history.

Fixes: https://github.com/nodejs/build/issues/926

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,6 @@ while not used actively in the CI infrastructure, is used from time to time for
 R&D purposes by the project.
 
 
-
 ## CI Software
 
 Build and test orchestration is performed by [Jenkins][21].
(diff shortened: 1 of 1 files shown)


gibfahn:
cc/ @nodejs/tsc please review (more info at #926).

README.md
```diff
@@ -214,6 +213,9 @@ Build and test orchestration is performed by [Jenkins][21].
 - A listing of connected servers for testing, building and benchmarking
   can be found at: <https://ci.nodejs.org/computer/>
 
+The Build WG will keep build configuration required for a release line for 6
+months after the release goes End-of-Life, in case further build or test runs
+are required. After that the configuration will be removed.
```
ofrobots: s/will/may, mayhaps?

gibfahn: I'd say will is more accurate. Of course it doesn't specify how soon it will be removed, so there's no rush.

But if anyone reads this, I'd like them to think that this is definitely going away, rather than maybe going away.

rvagg:
love it, thanks @gibfahn, lgtm

rvagg:
Agreement in today's TSC meeting on the proviso that we save current config where possible in git. So we should probably do a checkin of config.xmls here first then we can move ahead.

maclover7:
ping @gibfahn

PR #1004, opened 2017-11-21
jenkins: add node-test-commit-pre

maclover7:
All actions run before pending post-build-status-update job is run. My
thought is that node-test-commit sub jobs will `curl` from master, and
then pipe to bash.

### DIFF
--- a/jenkins/scripts/node-test-commit-pre.sh
+++ b/jenkins/scripts/node-test-commit-pre.sh
@@ -1,0 +1,68 @@
+###
+# node-test-commit-pre
+# All actions run before pending post-build-status-update job is run
+###
+
+# Git
+git --version
+
+# Name and email in git config need to be the same across all jobs
+# so that git rebase --committer-date-is-author-date will produce
+# the same commit SHA1
+git config --replace-all user.name Dummy
+git config --replace-all user.email dummy@dummy.com
+git config user.name
+git config user.email
+echo $GIT_COMMITTER_NAME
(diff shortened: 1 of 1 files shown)


refack:
👍 even if we don't `curl` it in, it's great to have this in a visible place.

mhdawson:
Instead of curl, my first thought would be that we'd clone the build repo at the start of each job and then can run the scripts directly.

Either way seems like a good first step for pulling some logic out of the jenkins config and into git where we can more easily track changes.

maclover7:
Landed in 2e05d83f318a4db5e8f173b4f2dfe7006ccf66ca

refack:
I rolled the change back for `node-test-commit-linux-linked` it was failing to generate `env.properties` so job were failing. https://ci.nodejs.org/job/node-test-commit-linux-linked/295/nodes=ubuntu1604_sharedlibs_debug_x64/console

P.S. I was sure `node-test-commit-linux-linked` was not the the multijob 🤔 yet

rvagg:
that's because I had some custom work in there for node-test-commit-linux-linked to solve a problem related to status posting, the addition and explanation is shown in this comment: https://github.com/nodejs/node/pull/17151#issuecomment-346528636

I could pull that into a separate script and we'd be back to status quo.

Re curl vs git I like the idea of using git cause there's a bunch more things we could pull into it, we could replace most (or all?) of our jenkins scripts with scripts in this repo if we did a clone first.

PR #1005, opened 2017-11-22
Rewrite Foundation resources to https://foundation.nodejs.org

fhemberger:
Ref: https://github.com/nodejs/nodejs.org/issues/1349

I also removed all legacy redirects for foundation member logos, as they are now handled exclusively by the Foundation editorial team.


rvagg:
@fhemberger @gtewallace @hackygolucky this is active now, as is, I've done some brief testing but just in case there's anything wrong please ping me directly, probably via email if it's urgent

PR #1009, opened 2017-11-23
Fix Foundation redirects

fhemberger:
Ref: #1005 

@rvagg Sorry, had to fix an issue and optimized some rules. Please review:

- Fixed `^/en/foundation` overruling subsequent redirects
- Simplified some RegExp (just checking for beginning of a string)
- Added `/?$` to some rules to be more explicit


rvagg:
should be active now

fhemberger:
Thanks!

PR #1017, opened 2017-11-27
jenkins: make curl use ipv4

maclover7:
Having errors with github-bot currently...

16:19:08.251  INFO bot: request finish (req_id=7ada38f2-de42-47b9-8746-e582ad53a907, duration=1, req.query={}, req.remoteAddress=::ffff:147.75.69.113, req.remotePort=44716)

... because curl is trying to use IPv6.

cc @refack @phillipj


maclover7:
Landing...

maclover7:
Landed in 0cf9e6a

PR #1025, opened 2017-12-01
jenkins: add llnode pipeline and shell scripts

gibfahn:
The llnode CI isn't working right now, but hopefully by committing this
we can then iterate in GitHub.

This also makes the node download script runnable manually, and adds docs for it.

@joyeecheung I was thinking if we land this, you can raise a PR to build, and build the jobs directly from your PR. Thoughts?

### DIFF
--- a/jenkins/pipelines/llnode-pipeline.jenkinsfile
+++ b/jenkins/pipelines/llnode-pipeline.jenkinsfile
@@ -5,7 +5,7 @@ properties([
     string(name: 'GIT_REPO', defaultValue: 'nodejs/llnode', description: 'Fetch from this repo (e.g. gibfahn/llnode)'),
     string(name: 'GIT_BRANCH', defaultValue: 'master', description: 'Branch to test (e.g. master, refs/pull/999/head).'),
     string(name: 'NODE_VERSIONS', defaultValue: 'v6 v8 v9 v10 nightly canary', description: 'Space separated list of NODE_VERSIONS to pass to llnode-continuous-integration.'),
-    string(name: 'LLDB_VERSIONS', defaultValue: '3.9.0 4.0.1 5.0.2 6.0.0', description: 'Space separated list of LLDB_VERSIONS to pass to llnode-continuous-integration.'),
+    string(name: 'LLDB_VERSIONS', defaultValue: '3.9 4.0 5.0 6.0', description: 'Space separated list of LLDB_VERSIONS to pass to llnode-continuous-integration.'),
     string(name: 'MACHINES', defaultValue: 'all', description: '''Can be "all" or a space-separated list of nodes, eg:
       debian8-64 fedora22 fedora23 osx1010 ppcle-ubuntu1404 ubuntu1204-64 ubuntu1404-64 ubuntu1604-64 rhel72-s390x aix61-ppc64 ppcbe-ubuntu1404 smartos16-64 smartos15-64 win10 win2012r2'''),
   ]),
--- a/jenkins/pipelines/llnode-pipeline.jenkinsfile
+++ b/jenkins/pipelines/llnode-pipeline.jenkinsfile
@@ -1,0 +1,25 @@
+#!/usr/bin/env groovy
+
+def pr = [] // Mutable parameter set to pass to jobs. I know this is crazy.
+String printParams = '' // String of params to print
+for (param in params) {
+    printParams += param.toString() + '\n'
+    if (param.key != 'NODE_VERSIONs') {
+        pr.push(string(name: param.key, value: param.value))
+    }
+}
+println "\n### BUILD PARAMETERS ###\n${printParams}"
+
+stage('Test llnode') {
+    def versions = params['NODE_VERSIONs'].split()
+    def buildJobs = [:]
+    for (int i = 0; i < versions.length; i++) {
+        int index = i // locally scoped copy.
+        def p = pr.collect() // local copy of params
+        p.push(string(name: 'NODE_VERSION', value: versions[index]))
+        buildJobs["${versions[index]}"] = { build(job: 'llnode-continuous-integration', parameters: p) }
+    }
+
+    println buildJobs
+    parallel(buildJobs)
+}
--- a/jenkins/pipelines/llnode-pipeline.jenkinsfile
+++ b/jenkins/pipelines/llnode-pipeline.jenkinsfile
@@ -4,7 +4,8 @@ properties([
   parameters([
     string(name: 'GIT_REPO', defaultValue: 'nodejs/llnode', description: 'Fetch from this repo (e.g. gibfahn/llnode)'),
     string(name: 'GIT_BRANCH', defaultValue: 'master', description: 'Branch to test (e.g. master, refs/pull/999/head).'),
-    string(name: 'NODE_VERSIONS', defaultValue: 'v6 v8 v9 v10', description: 'Space separated list of NODE_VERSIONS to pass to llnode-continuous-integration.'),
+    string(name: 'NODE_VERSIONS', defaultValue: 'v6 v8 v9 v10 nightly canary', description: 'Space separated list of NODE_VERSIONS to pass to llnode-continuous-integration.'),
+    string(name: 'LLDB_VERSIONS', defaultValue: '3.9.0 4.0.1 5.0.2 6.0.0', description: 'Space separated list of LLDB_VERSIONS to pass to llnode-continuous-integration.'),
     string(name: 'MACHINES', defaultValue: 'all', description: '''Can be "all" or a space-separated list of nodes, eg:
       debian8-64 fedora22 fedora23 osx1010 ppcle-ubuntu1404 ubuntu1204-64 ubuntu1404-64 ubuntu1604-64 rhel72-s390x aix61-ppc64 ppcbe-ubuntu1404 smartos16-64 smartos15-64 win10 win2012r2'''),
   ]),
@@ -23,13 +24,18 @@ for (param in params) {
 println "\n### BUILD PARAMETERS ###\n${printParams}"
 
 stage('Test llnode') {
-    def versions = params['NODE_VERSIONS'].split()
+    def nodeVersions = params['NODE_VERSIONS'].split()
+    def lldbVersions = params['LLDB_VERSIONS'].split()
     def buildJobs = [:]
-    for (int i = 0; i < versions.length; i++) {
-        int index = i // locally scoped copy.
-        def p = pr.collect() // local copy of params
-        p.push(string(name: 'NODE_VERSION', value: versions[index]))
-        buildJobs["${versions[index]}"] = { build(job: 'llnode-continuous-integration', parameters: p) }
+    for (int i = 0; i < nodeVersions.length; i++) {
+        int nodeIndex = i // locally scoped copy.
+        for (int j = 0; j < lldbVersions.length; j++) {
+            int lldbIndex = j // locally scoped copy.
+            def p = pr.collect() // local copy of params
+            p.push(string(name: 'NODE_VERSION', value: nodeVersions[nodeIndex]))
+            p.push(string(name: 'LLDB_VERSION', value: lldbVersions[lldbIndex]))
+            buildJobs["Node.js ${nodeVersions[nodeIndex]} lldb ${lldbVersions[lldbIndex]}"] = { build(job: 'llnode-continuous-integration', parameters: p) }
+        }
     }
 
     println buildJobs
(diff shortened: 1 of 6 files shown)


jenkins/pipelines/llnode-pipeline.jenkinsfile
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env groovy
```
mhdawson: I'm ok with landing so we can iterate but would it be useful to have a comment that indicates "Not yet working" just in case somebody without the context takes a look ?

gibfahn:
>Can the parameters for llnode-pipeline also be added into the pipeline definition?

Good point, added. Also added a check, as the parameters don't always show up until the second time you run the pipeline.

jenkins/scripts/llnode-continuous-integration.sh
```diff
@@ -0,0 +1,34 @@
...
+git clone https://github.com/$GIT_REPO.git llnode
+cd llnode
+if [ $GIT_BRANCH != master ]; then
+	git fetch origin $GIT_BRANCH:testBranch
+	git checkout testBranch
+fi
+
+[ $(uname) = Darwin ] && git clone --depth=1 -b release_39 https://github.com/llvm-mirror/lldb.git lldb
+# Initialize GYP
+git clone https://chromium.googlesource.com/external/gyp.git tools/gyp
```
refack: I found this faster, and it's probably more predictable.
```bash
curl -s http://ci.nodejs.org/downloads/zos/gyp.tar.gz | tar xvf - -C tools/gyp/
```

jenkins/scripts/llnode-continuous-integration.sh
```diff
@@ -0,0 +1,34 @@
...
+export PATH=$PWD:$PATH
+cd $WORKSPACE
+export NPM_CONFIG_USERCONFIG=$WORKSPACE/npmrc
+export NPM_CONFIG_CACHE=$WORKSPACE/npm-cache
+node -v
+npm -v
+export npm_loglevel=error
+npm set progress=false
+ls
+git clone https://github.com/$GIT_REPO.git llnode
```
refack: Why not do a shallow clone here as well?
```
git clone --depth=1 -b $GIT_BRANCH https://github.com/.git llnode
```
or even
```
mkdir llnode
curl -s https://github.com/$GIT_REPO/archive/$GIT_BRANCH.tar.gz | tar xvf - --strip-components=1 -C llnode
```

jenkins/scripts/llnode-continuous-integration.sh
```diff
@@ -0,0 +1,34 @@
...
+fi
+
+[ $(uname) = Darwin ] && git clone --depth=1 -b release_39 https://github.com/llvm-mirror/lldb.git lldb
+# Initialize GYP
+git clone https://chromium.googlesource.com/external/gyp.git tools/gyp
+# Configure
+[ $(uname) = Linux ] && GYP_ARGS="-Dlldb_dir=/usr/lib/llvm-3.8/" || GYP_ARGS=""
+./gyp_llnode $GYP_ARGS
+# Build
+make -C out/ -j9
```
refack: magic number `9`

jenkins/scripts/llnode-continuous-integration.sh
```diff
@@ -0,0 +1,34 @@
...
+	git fetch origin $GIT_BRANCH:testBranch
+	git checkout testBranch
+fi
+
+[ $(uname) = Darwin ] && git clone --depth=1 -b release_39 https://github.com/llvm-mirror/lldb.git lldb
+# Initialize GYP
+git clone https://chromium.googlesource.com/external/gyp.git tools/gyp
+# Configure
+[ $(uname) = Linux ] && GYP_ARGS="-Dlldb_dir=/usr/lib/llvm-3.8/" || GYP_ARGS=""
+./gyp_llnode $GYP_ARGS
```
refack: Maybe add explicit `python` or `python2`

jenkins/scripts/common/getNode.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash -ex
```
refack: So why not just use `nvm` like https://github.com/nodejs/build/blob/1fb7b6046ed7304fef291133a6599decb06243c7/jenkins/pipelines/node-inspect.jenkinsfile#L72-L78

jenkins/scripts/llnode-continuous-integration.sh
gibfahn: Does that get updated automatically?

jenkins/scripts/common/getNode.sh
gibfahn: Does it work on windows?

gibfahn: I don't have an issue with using nvm if it works everywhere and doesn't write to files outside the workspace.

This is what was there already, I just cleaned up the `sort`, which was a bit broken.

jenkins/scripts/llnode-continuous-integration.sh
gibfahn: Fixed.

jenkins/scripts/common/colors.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+# Helper vars for printing in colour. Use with:
```
richardlau: Consistency nit: all the comments in this file use `colour` but the script itself is named `colors`.

jenkins/scripts/llnode-continuous-integration.sh
```diff
@@ -0,0 +1,32 @@
...
+node -v
+npm -v
+export npm_loglevel=error
+npm set progress=false
+ls
+
+git clone --depth=1 -b $GIT_BRANCH https://github.com/$GIT_REPO.git llnode
+cd llnode
+
+[ $(uname) = Darwin ] && git clone --depth=1 -b release_39 https://github.com/llvm-mirror/lldb.git lldb
```
joyeecheung: I think we don't actually need the tools cloning, configuring and build steps now.  `npm install` would do all of that. As long as the machine has a lldb installed, `scripts/configure.js` in llnode would grab the right version of headers, `gyp_llnode` and `make` would be run during `npm install` as well.

joyeecheung: As for the tests, I think on Linux we would need to set `TEST_LLDB_BINARY` to `which lldb-3.9` before running `npm test`. Also it would be helpful to set `TEST_LLNODE_DEBUG=true` and `LLNODE_DEBUG=true` if printing stuff to stderr does not interfere with the job.

jenkins/scripts/common/colors.sh
gibfahn: Good point, done.

jenkins/scripts/llnode-continuous-integration.sh
refack: No. And maybe it's a good thing?

gibfahn: Apparently this is unneeded anyway, so removed.

jenkins/scripts/common/getNode.sh
mmarchini: llnode is not supported on Windows yet (as per https://github.com/nodejs/llnode/issues/17), so I don't think nvm not running on Windows is an issue now. When we get support to Windows, we can use `nvm` for Unix platforms and `nvs` for Windows (or just use `nvs` for everything).

But we can change to `nvm`/`nvs` in the future.

jenkins/scripts/common/helpers.sh
```diff
@@ -0,0 +1,36 @@
+# Helper functions and variables. Script should be sourced.
```
mmarchini: I don't see this file being sourced anywhere. Am I missing something or it will be used in the future?

jenkins/pipelines/llnode-pipeline.jenkinsfile
mmarchini: I think these scripts will work now since tests are working on llnode with Travis.

mmarchini:
Ping @gibfahn

mmarchini:
@gibfahn would you mind if I move forward with this? Looks like Travis on llnode just broke again (our 2 last builds are stuck), so it would be nice to move our tests to Jenkins.

mmarchini:
Commit updating Node.js versions: https://github.com/mmarchini/build/commit/b918f24ca9c1d770dd61d0db14e0e40f7acea40a
(I don't have permission to push to this branch)

I tested with `GIT_REPO="nodejs/llnode" GIT_BRANCH="master" NODE_VERSION="v8" sh jenkins/scripts/llnode-continuous-integration.sh` and it's working. @nodejs/build can we move forward with this and improve in further iterations?

Trott:
>  @nodejs/build can we move forward with this and improve in further iterations?

Works for me, but it would be good to have some input from more Jenkins-y folks like @joaocgreis @maclover7 @rvagg @refack ....

maclover7:
There are a few outstanding review comments that @gibfahn should look at (I, at least personally, am not super familiar with this PR and how it will work, and don't really feel comfortable trying to address those) -- if we don't hear back for a little while longer, we can take another look

refack:
@mmarchini I applied your tweak to the pipeline spec. Since the job runs using the code from @gibfahn's fork it should now be updated.

mmarchini:
@refack I just uploaded a new patch, can you cherry-pick it into `gibfahn:llnode-ci`?
https://github.com/mmarchini/build/commit/417fa906a15b4bf147ec211eb28a8b35f807bcd3

This patch will fix the test runs for Linux and Mac OS by downloading LLDB from http://releases.llvm.org/download.html. It assumes we're only testing on Ubuntu 16.04 and Mac OS for now, which I think is fine for now, since those are the tests we currently have on llnode with Travis.

I also added `nightly` and `canary` as possible parameters for Node.js version. Is it possible to discard the result of those versions on the result of llnode-pipeline, so we can keep the same behavior from our current Travis setup? For reference: 
* https://github.com/nodejs/llnode/blob/master/.travis.yml
* https://travis-ci.org/nodejs/llnode/builds/379496130

image

mmarchini:
Interesting. I tested only with 3.9.0, and lldb was there. Let me see why it's missing from 6.0.0

mmarchini:
Btw, does this build matrix still make sense now that LLDB_VERSIONS is a parameter?

image

mmarchini:
> I tested only with 3.9.0, and lldb was there. Let me see why it's missing from 6.0.0

I actually forgot the global LLDB in my path while testing, my bad. clang+llvm doesn't come with LLDB, and I'm having trouble finding a source to download it from.

refack:
> Btw, does this build matrix still make sense now that LLDB_VERSIONS is a parameter?

Killed the matrix.

joyeecheung:
I have not yet find a way to get reliable prebuilt lldb binaries either.

Here is a gist that I wrote some time ago when I need to test llnode on different linux distros:
https://gist.github.com/joyeecheung/f458f9a668e0e9546d48d899298e7331

mmarchini:
Is Xcode installed in our OSX machines? If so we can use the LLDB that comes with it, and use Docker to test on Linux. What do you think?

refack:
We should have "Xcode Command Line Tools" on the macOS machines.

mmarchini:
New patch, tested on Ubuntu 16.04 (this time without having a global lldb installed):
https://github.com/mmarchini/build/commit/acb541dc66eec2ac94aa60792905023e0b90a3f3

I went with the "download the .deb and extract it" approach.

OSX will still fail, but let's fix one OS at a time :)

refack:
Shaved it some more (`curl` needed `--compressed` and a missing pipe)

```
Reading lldb version...
/home/iojs/build/workspace/llnode-continuous-integration/default/lldb-bin/usr/bin/lldb: error while loading shared libraries: libLLVM-6.0.so.1: cannot open shared object file: No such file or directory
{ Error: Command failed: /home/iojs/build/workspace/llnode-continuous-integration/default/lldb-bin/usr/bin/lldb -v
/home/iojs/build/workspace/llnode-continuous-integration/default/lldb-bin/usr/bin/lldb: error while loading shared libraries: libLLVM-6.0.so.1: cannot open shared object file: No such file or directory
```

~I saw this while trying to to the `deb` dance, there's a long chain of dependencies needed to~ NVM was just missing 1 more; `libllvm$LLDB_VERSION.*$arch.deb` and... success.

https://ci.nodejs.org/job/llnode-continuous-integration/47/default/

mmarchini:
@refack is there anything left to be done here? If so, can I help?

refack:
@mmarchini you should now have write permissions to the underlying branch - ttps://github.com/gibfahn/build/tree/llnode-ci

PR #1041, opened 2017-12-12
add alpine 3.7

rvagg:
active in CI now, first run on master passed just fine, they'll run slow for a few days as the ccache across the 4 docker hosts warms up

### DIFF
--- a/ansible/roles/docker/templates/alpine37.Dockerfile.j2
+++ b/ansible/roles/docker/templates/alpine37.Dockerfile.j2
@@ -1,0 +1,57 @@
+FROM alpine:3.7
+
+ENV LC_ALL C
+ENV USER {{ server_user }}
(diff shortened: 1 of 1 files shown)


mhdawson:
So we just use a long running docker image for our Alpine images.  May be a good template for other OS's we want to support, but don't want to have full machines for.

rvagg:
Indeedy, and we have a few in there already: ansible/roles/docker/templates/ - openssl 1.1.0 and fips  are in the ubuntu1604_sharedlibs.Dockerfile.j2 that's in CI already, I have a branch where I'm messing with zlib and cares too. But aside from that, it's a good fall-back if we find ourselves in a position where we can't get VMs for OSs we want to support, like Ubuntu 12.04, CentOS 5, which we're _just_ hanging on to now because we provisioned those machines before they became unavailable on our IaaS providers.

Issue #416, opened 2016-05-11
Retire Fedora21?


jbergstroem:
Fedora 21 [EOL'ed last year](https://fedoraproject.org/wiki/End_of_life). How about we cut it from the build farm? I know there's no harm in keeping it, but that strategy could have developers spend extra time fixing bugs for platforms that upstream dropped support for (ok, lets just pretend that OS X isn't in this discussion).

bnoordhuis:
Sounds reasonable.  I did wonder why it was still around.

ChALkeR:
+1 for dropping support for anything that doesn't get security updates anymore.

jbergstroem:
Anyone from the build group wants to chip in?

rvagg:
I'm *sure* I fired off a +1, but can't find it, this is the second issue where this has happened in nodejs/build recently so I'm either having mental or technical problems.

+1 from me, we probably need to make an official policy of killing off any Linux distro that has its upstream support dropped so we don't even need to ask these questions.

jbergstroem:
@rvagg fwiw I've had cases where I swear I posted a comment but later finding that it wasn't there.

ChALkeR:
@jbergstroem @rvagg Btw, I remember having similar issues about two weeks ago, and those turned out to be actually GitHub lags — https://twitter.com/githubstatus/status/724933464273903622.

jbergstroem:
@rvagg I believe we need a policy for all platforms/versions. I think as soon as we land in a new OS X environment we should define what we will support (which hopefully equates to what we test and have tested for a while).

I will retire this today and open a new issue about formal support.

jbergstroem:
Retired.

PR #1058, opened 2018-01-01
setup: remove fedora 21

maclover7:
Machines do not appear in https://ci.nodejs.org/computer, or in
`ansible/inventory.yml`.

Refs: https://github.com/nodejs/build/issues/416

---

cc @rvagg, you were the one on the issue retiring fedora 21


rvagg:
👍  the old inventory and setup directory is mostly not used these days, only for a few unique things, so it's pretty safe to yank out these generic worker configs

PR #1066, opened 2018-01-02
setup: remove containers

maclover7:
Machines do not appear in https://ci.nodejs.org/computer, or in
`ansible/inventory.yml`.

Also looks like this approach has been abandoned, and we do not want to
run untrusted builds for the time being.


maclover7:
Would appreciate signoff from @rvagg since he was the one who introduced this, just want to make sure we don't need it anymore

rvagg:
screenshot 2018-01-03 21 32 25

not abandoned, just evolved, see ansible/playbooks/jenkins/docker-host.yaml for all of this now

maclover7:
Landed in 4aee9450c796c9b2a1930950f116b29464555f2e

PR #1069, opened 2018-01-04
ansible: run package-upgrade after adding new sources

maclover7:
Practiced setting up a Debian 8 machine via Vagrant, and
`package-upgrade` needs to be run after adding new sources. After this
update, I think `setup/debian8` can be removed -- I have a Ansible-ized
Vagrant box running a Jenkins worker based on the
`jenkins/worker/create.yml` playbook.


rvagg:
right, good catch lgtm

maclover7:
Landing...

maclover7:
Landed in 008820b

Issue #775, opened 2017-06-28
WARNING: Jenkins update to 2.60.1 requires Java 8


gibfahn:
### Needs doing:

- [x] Create a job to run a `java -version` (and probably a `which java`/`where java`) on each machine, so we can see where we're going to have issues. If the job failed on Java <8 that would be ideal
- [ ] Upgrade ansible scripts for machines which need it.
- [ ] Apply scripts to machines and get us ready to upgrade
- [ ] Upgrade and make sure things still work (some machines may not be using the Java in the path for Jenkins connection)

---

Note for @nodejs/build @nodejs/jenkins-admins considering clicking the shiny red button to update Jenkins

![image](https://user-images.githubusercontent.com/15943089/27628669-ab8204ce-5be7-11e7-8534-13bde397990d.png)

In the [changelog for 2.60.1](https://jenkins.io/changelog-stable/) it notes that it requires Java 8 to be installed on all the hosts.

>2.60.1 is the first Jenkins LTS release that requires Java 8 to run. If you're using the Maven Project type, please note that it needs to use a JDK capable of running Jenkins, i.e. JDK 8 or up. If you configure an older JDK in a Maven Project, Jenkins will attempt to find a newer JDK and use that automatically. If your SSH Slaves fail to start and you have the plugin install the JRE to run them, make sure to update SSH Slaves Plugin to at least version 1.17 (1.20 recommended).

I'm pretty sure we don't have Java 8 on all our hosts, so updating will stop those hosts being able to connect.

gibfahn:
I guess the plan would be to confirm that all our machines are running Java 8 and then update.

jbergstroem:
```console
$ ssh ci java -version              
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)
```

gibfahn:
@jbergstroem sure, but what about the individual hosts? For example I'm pretty sure AIX is using Java 7 (see [here](https://github.com/nodejs/build/blob/master/setup/aix61/manualBootstrap.md#gettext-java-make)).

jbergstroem:
@gibfahn I didn't get that even agents now require java 8. What kind of crazy is this :(

jbergstroem:
So, anyone with any ideas on how we move forward here? We are out of LTS meaning any security bugs (expect them) will hit us pretty hard.

[...]

PR #1075, opened 2018-01-08
WIP: ubuntu 14.04 ppc64le

jBarz:
Enable ubuntu 14.04 ppc64le.
Have both gcc 4.8 and 4.9 on machine available


ansible/roles/java-base/vars/main.yml
```diff
@@ -14,7 +14,7 @@ packages: {
   'freebsd': 'openjdk8-jre',
   'smartos': 'openjdk8',
   'ubuntu': 'openjdk-8-jre-headless',
-  'ubuntu1404': 'oracle-java8-installer',
+  'ubuntu1404': 'openjdk-7-jre-headless',
```
mhdawson: We need Java 8 on all of the jenkins nodes, regardless of the OS level as that is the minimum that works with our levels of Jenkins.

We had to manually update the PPC machines a while back, we should use that template to update these ansible scripts to get the right version on the PPC machines as well.

The templates are here: https://github.com/nodejs/build/issues/775#issuecomment-336245389

gdams: If we wish to fetch the ppcle binaries from AdoptOpenJDK then we should consider using the api https://api.adoptopenjdk.net/README

jBarz: Fixed

mhdawson:
If you have used this to configure ubuntu1404-ppc64_le-3 I think the next step is to connect it up to the so that we can validate the machine works as a test worker.

The one thing we should probably test as part of that is the selection of the 4.9 versus 4.8 compiler.  We should either have a script in the build repo for that or have that script isntalled as part of the ansible setup.  It may make sense to be part of what is installed by the ansible scirpts as we'd need to update when we add support for a new compiler level which would also require ansible script updates.

jBarz:
I have run the ansible script successfuly on ubuntu1404-ppc64_le-3.
To build node using gcc-v4.8, we need to set the following. Equivalent for 4.9.

export CC=/usr/bin/gcc-4.8
export CXX=/usr/bin/g++-4.8
export LINK=/usr/bin/g++-4.8
export LDFLAGS=“-Wl,-rpath,$(dirname `$CC --print-file-name libgcc_s.so`)”

mhdawson:
I think next step is create test CI job as clone of existing job and do run against ubuntu1404-ppc64_le-3.

I would like to check in a script into the build repo called something like set_gcc_level that would looks at the files checked out for Node.js (likely the version info) and then sets the required exports.

The main downside is that we'd need a clone of the build repo but I don't think that would be a big deal.  @nodejs/build any concerns?

maclover7:
Checking scripts into the build repo is fine by me. https://github.com/nodejs/build/tree/master/jenkins/scripts has some prior work, if helpful at all.

maclover7:
ping @jBarz

mhdawson:
@jBarz  The paths don't seem to be quite right based on the ansible setup.  I had to add the path to java 8 to get the jenkins agent to start: /home/iojs/jdk8u144-b01/jre/bin/java in jenkins.conf.

jenkins.conf probably either needs to have the path updated like I did or probably better the path so that it include/home/iojs/jdk8u144-b01/jre/bins

mhdawson:
Testing builds on le-3 here: 
https://ci.nodejs.org/job/node-test-commit-plinux_mdawson/21/
https://ci.nodejs.org/job/node-test-commit-plinux_mdawson/22/
https://ci.nodejs.org/job/node-test-commit-plinux_mdawson/23/

mhdawson:
Tests ran/passed for the 3 combinations.

mhdawson:
PR for script to select compiler level: https://github.com/nodejs/build/pull/1103 that was tested in runs above on be-3.

jBarz:
I modified the ansible script to append "/home/iojs/.profile" with
`export PATH=/home/iojs/jdk8u144-b01:\$PATH`

Is this an acceptable solution?
We are assuming that Jenkins will only run as "iojs".

maclover7:
@jBarz Seems reasonable to me.

@mhdawson Can you run these scripts again on all ppc machines and confirm everything's ok? After that, please land at will.

mhdawson:
@maclover7 can you review/approve https://github.com/nodejs/build/pull/1103 as its related to this work.

maclover7:
@mhdawson Approved 1103. Once this is ready, please feel free to land in master, and then please run on all PPC machines to confirm everything's OK.

mhdawson:
Unfortunately the change to set the path does not quite work as the script that starts the worker sets the path directly. Will discuss with @jBarz to figure out how to fix.

mhdawson:
Looks good with latest update from @jBarz

mhdawson:
Landed as 443f72ec12b3b9c05b4f7a74972680e4a5358ca5

Issue #1074, opened 2018-01-08
SoftLayer instability


rvagg:
I'm currently bogged down getting all 3 SoftLayer CentOS 6 machines back online, release and 2 x test. I've had to reprovision them but am having all sorts of problems. They don't seem to like being rebooted and my guess is that there's some kind of problem with the kernel that gets installed during a `yum upgrade` (a kernel we want for security but doesn't seem to be liked by the VM host is my guess). I also got a host upgrade notice on two VMs that I'd just provisioned so I suppose they're still working through their own upgrade process and that could be related.

So @nodejs/build, don't go messing too much with SoftLayer hosts just now, don't reboot anything, _absolutely do not touch the CentOS 5 machines there!_

maclover7:
ping @rvagg -- did this ever get resolved?

rvagg:
been meaning to test it out, I haven't restarted anything on softlayer yet as I don't have the time to spend fixing up a borked restart if it's still a problem! there was an incident report from them today, critical maintenance or something, my guess is that this churn around Meltdown is causing a lot of internal churn for infra providers that is still to settle down

maclover7:
ping @rvagg -- would like to start working on migrating the `setup` CentOS ansible scripts to `ansible`.

rvagg:
I've been doing plenty of rebooting over the past week an all's fine as far as I can tell

@maclover7 I thought we had centos covered in the ansible scripts already? maybe not CentOS5 but we're waiting for that to die out with Node 4 I think.

maclover7:
@rvagg Not sure -- `setup/centos5` and `setup/centos6` still exist, so was assuming migration was complete.

rvagg:
@maclover7 no, their existence in setup doesn't indicate completeness of migration. It wasn't that well organised! Most of the things in setup/ for standard test machines are redundant. www, raspberry-pi are the two that I still keep updated and have not been migrated. cross-compiler might still be valid for that machine. I don't know about rhel72-linuxonecc or aix61. github-bot I think is still in use. linter I'm not sure about. I _think_ armv7-wheezy might still be the only way to manage those machines. I think the rest are OK to remove, like you've been slowly doing.

maclover7:
@rvagg that's my bad -- meant to say migration was _not_ complete, since those directories still exist :)

> I think the rest are OK to remove, like you've been slowly doing.

[...]

PR #1076, opened 2018-01-09
ansible: new centos6 machines, updated & fixed centos6 config

rvagg:
New set of CentOS 6 machines. Ansible config didn't quite work for it (I think it was written after deployment and hasn't been used to deploy machines from scratch).

Sadly there's something wrong with the SCL devtoolset-2 repo so I've switched to the centos/tru ones which are unsigned but we can get away without errors on that.


ansible/roles/jenkins-worker/templates/centos.initd.j2
```diff
@@ -17,11 +17,12 @@ PIDFILE=/var/run/jenkins/$NAME.pid
...
 JENKINS_SLAVE_LOG="/home/{{ server_user }}/$NAME.log"
-JENKINS_SLAVE_ARGS="-Xmx{{ server_ram|default('128m') }} -jnlpUrl {{ jenkins_url }}/computer/{{ inventory_hostname }}/slave-agent.jnlp -secret {{ secret }}"
+JENKINS_JAVA_ARGS="-Xmx{{ server_ram|default('128m') }}"
+JENKINS_SLAVE_ARGS="-jnlpUrl {{ jenkins_url }}/computer/{{ inventory_hostname }}/slave-agent.jnlp -secret {{ secret }}"
 JENKINS_ENV="JOBS={{ ansible_processor_vcpus }} \
   HOME=/home/{{ server_user }} \
-  DESTCPU={{ ansible_architecture }} \
-  ARCH={{ ansible_architecture }} \
+  DESTCPU={{ server_arch | default(ansible_architecture) }} \
+  ARCH={{ server_arch | default(ansible_architecture) }} \
```
rvagg: centos will do `x86_64` which our config scripts don't recognise so we have to override, I think maybe we should fix configure to recognise it as an alias

rvagg:
opened a PR to allow `x86_64` as an alias, would like to hear opinions on this change https://github.com/nodejs/node/pull/18052

gibfahn:
>opened a PR to allow x86_64 as an alias, would like to hear opinions on this change nodejs/node#18052

Definitely in favour of just doing in in `./configure` rather than duplicating it in every build script.

ansible/roles/jenkins-worker/templates/centos.initd.j2
gibfahn: Does this need to be manually defined every time you run? If so, and if it defaults to `x86_64`, then the chance of us missing this seems high.

mhdawson:
+1 to in ./configure as well.

rvagg:
So, we set `ARCH` / `DESTCPU` explicitly on release machines, we haven't done it as much on non-release but in this case CentOS 6 is used for releases so it ends up being in our test machines too. My preference is to be explicit about architecture on all our machines rather than letting `configure` guess; but it's not a strong preference and I know others have preferred leaving it out. I'm fine as long as we be explicit on our release machines.

mhdawson:
I'm not as worried about having to set it for our test machines, but when people checkout/build it would be good if it just worked whenever possible.

gibfahn:
>My preference is to be explicit about architecture on all our machines rather than letting configure guess; but it's not a strong preference and I know others have preferred leaving it out. 

If we're going to be explicit, as much as possible I'd rather we be explicitly saying `ARCH=$(uname -m)` rather than manually setting it to what we think the ARCH should be. Putting the conversion in `./configure` makes it a lot more visible than having it be an option in a template (see https://github.com/nodejs/build/pull/1076#discussion_r160374859).

>I'm not as worried about having to set it for our test machines, but when people checkout/build it would be good if it just worked whenever possible.

It already worked manually if you didn't specify the ARCH, this is just a question of whether it works if you specifically say `ARCH=$(uname -m)`.

maclover7:
@rvagg is this needed for #1074 to be resolved? Are there any corresponding changes to the secrets repo which need to be made?

rvagg:
No, unrelated to #1074, it's more of a stale-mate of sorts. I'd really like to define ARCH and DESTCPU, mainly for the release machines where being explicit is especially important (IMO), and CentOS 6 is now a release machine type.

nodejs/node#18052 got merged but that's only on master for now so this is still a problem for everything but master and will be for some time even if we backport it all the way (unlikely for 4.x I reckon).

gibfahn:
For now can we just define it in the job (with a comment that says this can be removed when v4.x goes EoL)?

PR #1103, opened 2018-02-06
Add script to select compiler

mhdawson:
Initially used be used on PPC, then zLinux and later on if necessary other linux variants.

### DIFF
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+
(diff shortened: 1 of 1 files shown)


maclover7:
@mhdawson To confirm, this will be called from within the Jenkins job?

mhdawson:
@maclover7 yes.  I clone the repo in the job then call this to set the compiler.

mhdawson:
Landed as 1f63cbd21b0f6c65012f16b0f0c74ca2539e31da

PR #1108, opened 2018-02-08
doc: use ncu-team sync to sync member list

joyeecheung:
This use the new [ncu-team sync](https://github.com/nodejs/node-core-utils/blob/master/docs/ncu-team.md) command in node-core-utils to automatically maintain the list of members in the README with members in the @nodejs/build and @nodejs/github-bot  Github team.

```
npm install -g node-core-utils
ncu-team sync README.md
```

Refs: https://github.com/nodejs/admin/issues/61

### item 1
I could not find corresponding teams for infra admins and release admins. Should we create those teams?

### item 2

@Trott I think you have removed yourself from the README, but have not removed yourself from the build team. Do you want to remove yourself from the team as well?

### item 3

There are also several members added into the README because of their membership in corresponding teams:

@Starefossen in build and github-bot
@Fishrock123 @mscdex @williamkapke in github-bot

Do you want to remain in the team and be added into the README?

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ missing please open an issue. If you are interested in joining check out
 
 ## Build WG Members
 
-- [@Starefossen](https://github.com/Starefossen) - Hans Kristian Flaatten
+<!-- ncu-team-sync.team(nodejs/build) -->
+
 - [@gdams](https://github.com/gdams) - George Adams
 - [@geek](https://github.com/geek) - Wyatt Preul
 - [@gibfahn](https://github.com/gibfahn) - Gibson Fahnestock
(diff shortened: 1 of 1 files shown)


joyeecheung:
Also @rvagg @joaocgreis you are removed from the github bot admins in README because you are not in the @nodejs/github-bot team. Do you want to be added into the team and stay there?

Trott:
@joyeecheung I'll stay on the team. It gives me Jenkins powers I need. Listing me again is fine too. Thanks.

mhdawson:
We've not needed the infra and release teams, but I don't think it would be a bad idea to have them as it would provide a way to @mention people in those specific groups.

gibfahn:
>I could not find corresponding teams for infra admins and release admins. Should we create those teams?

We discussed this before, I don't think there were any objections. They should probably be subteams of build.

joaocgreis:
> Also @rvagg @joaocgreis you are removed from the github bot admins in README because you are not in the @nodejs/github-bot team. Do you want to be added into the team and stay there?

We are in the README list because we have access to the github-bot secrets (I believe the list was made based on the secrets, not teams). We have access because we are on the infra team and might need access to investigate if something comes up. I'd rather not be added to the github-bot team, for the moment I won't be able to contribute much.

The lists in README can reflect either team membership or access rights, but it should probably be clear about which. For me it would make some sense for it to reflect team membership, and perhaps update https://github.com/nodejs/build/blob/master/doc/access.md with any exception or difference in secrets access.

gibfahn:
>For me it would make some sense for it to reflect team membership, and perhaps update /doc/access.md@master with any exception or difference in secrets access.

Agreed, I think it's perfectly reasonable to say that infra admins (for example) have access to github-bot secrets without them being in the github-bot team (or in the github-bot section of the wiki).

joyeecheung:
I am going to merge this for now. We can revisit the other teams later.

PR #1127, opened 2018-02-16
ansible: add support for rhel72-s390x

maclover7:
playbooks/jenkins/worker/create.yml works now.

cc @nodejs/build


gibfahn:
Wow, this is great.

LGTM if we can test it.

mhdawson:
@jbarz can you test and add any changes needed to support gcc 4.8 and 4.9 concurrently?

jBarz:
working on it

gibfahn:
>@jBarz can you test and add any changes needed to support gcc 4.8 and 4.9 concurrently?

Seems like a separate PR, is there a reason for that to block this one?

jBarz:
yea, that would be a separate PR.

maclover7:
@jBarz @mhdawson Agree, I think gcc upgrade would best be kept separate from this. Are you able to review this PR as-is so we can land in current state

maclover7:
Thank you for reviews! Landing

maclover7:
Landed in d2942fe66f6b4bd2dba2b7ee37557ea7c3c8b608

PR #1148, opened 2018-02-25
ansible: added debian9 hosts

rvagg:
Debian 9 (Stretch) was released in July last year but we never added it to CI. Doing so now for completeness although I expect Ubuntu 17.10 has kept us out of trouble here.


ansible/roles/jenkins-worker/templates/systemd.service.j2
```diff
@@ -12,6 +12,7 @@ User={{ server_user }}
 Restart=always
 RestartSec=30
 StartLimitInterval=0
+WorkingDirectory=/home/iojs
```
maclover7: This shouldn't cause problems on other OSes and versions right?

rvagg: I don't believe so, we use the same username and home directory on all .. having said that, we use `{{ server_user }}` everywhere else so I'm going to change it now

PR #1149, opened 2018-02-25
doc: initial supported systems release data doc

rvagg:
First pass at a data sheet of support dates, libc, kernel and gcc versions of our major supported systems. Only RHEL (CentOS), Debian, Ubuntu here. I'm going to skip Fedora and Alpine in the Linuxes since they are short-lived.

Need to add info on SmartOS (I couldn't find a good resource on that), FreeBSD, Windows and the IBM systems.

I'd like to land this with just these initial systems and let others fill in the rest (or I could return to this later). For now I'd like to get discussion on Node 10 supported systems going and these are the key pieces as far as Linux is concerned to start that discussion.

### DIFF
--- a/doc/supported_systems_release_data.md
+++ b/doc/supported_systems_release_data.md
@@ -1,0 +1,19 @@
+
+# Release data for supported operating systems / distributions
+
+## Linux
+
+* [RHEL](https://access.redhat.com/support/policy/updates/errata/) _(versions & dates mirrored by CentOS)_
+* [Debian](https://wiki.debian.org/LTS)
+* [Ubuntu](https://www.ubuntu.com/info/release-end-of-life) _(18.04 versions pending final 18.04 final release)_
+
+| Distro                | glibc | gcc | Linux kernel | Support EOL |
+|-----------------------|-------|-----|--------------|-------------|
+| RHEL 6                | 2.12  | 4.4 | 2.6          | Nov-2020    |
+| RHEL 7                | 2.17  | 4.8 | 3.10         | Jun-2024    |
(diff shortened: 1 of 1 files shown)


gibfahn:
>I'd like to land this with just these initial systems and let others fill in the rest 

Sounds good.

mhdawson:
I'm not sure that data is valid for all platforms in terms of glic versions (ie they might be different for power and or Z).

rvagg:
Who do we have from Joyent these days? @cjihrig is that you? I'd like to get SmartOS into this table and I'd also like to have the SmartOS version locked down for Node 10. It's currently 15 which I think is probably too old so maybe we need to update? Who can tell us what minimum level we should support (and be locked in to supporting for the next 3 years)?

FYI our current SmartOS build rules are encapsulated in this bit of Groovy: https://github.com/nodejs/build/blob/rvagg/VersionSelectorScript.groovy/jenkins/scripts/VersionSelectorScript.groovy#L21-L27

rvagg:
@geek thanks, want to lock one of those in? 17 for Node 10+? I should be able to provision that  and we can get it in to both release and test CI. It'll lock in libc and libstdc++ versions for binaries for the next 3 years.

Rules would end up looking something like:

```groovy
  if (builderLabel.indexOf('smartos14') == 0 && nodeMajorVersion >= 8)
    return false
  if (builderLabel.indexOf('smartos15') == 0 && nodeMajorVersion < 8 || nodeMajorVersion >= 10)
    return false
  if (builderLabel.indexOf('smartos17') == 0 && nodeMajorVersion < 10)
    return false
```

PR #1150, opened 2018-02-25
ansible: install gcc-c++ on centos6 so ccache picks it up properly

rvagg:
otherwise there's no symlinks for `g++` and `c++` in /usr/lib64/ccache and it gets bypassed even when using devtoolset-2 binaries

fixed on the hosts that need it


rvagg:
@maclover7 yep, in fact I replaced the current softlayer centos6 machines just recently with it. CentOS 5 not so much but we haven't tried and we probably shouldn't since it'll die with Node 4.

PR #1155, opened 2018-02-28
Update diagnostics script to handle FreeBSD and macOS

gibfahn:
They can't handle the -p option, and give the following error:

```console
$ netstat -anp
netstat: option requires an argument -- p
Usage:  netstat [-AaLlnW] [-f address_family | -p protocol]
        netstat [-gilns] [-f address_family]
        netstat -i | -I interface [-w wait] [-abdgRtS]
        netstat -s [-s] [-f address_family | -p protocol] [-w wait]
        netstat -i | -I interface -s [-f address_family | -p protocol]
        netstat -m [-m]
        netstat -r [-Aaln] [-f address_family]
        netstat -rs [-s]
```

### DIFF
--- a/jenkins/scripts/node-test-commit-diagnostics.sh
+++ b/jenkins/scripts/node-test-commit-diagnostics.sh
@@ -1,0 +1,42 @@
+#!/bin/bash -ex
+
+# Diagnostics
+
+# $1: Set to before to print "Before building", otherwise prints "After building"
+
+set +x
+DIAGFILE=${HOME}/jenkins_diagnostics.txt
+echo >> ${DIAGFILE}
+echo >> ${DIAGFILE}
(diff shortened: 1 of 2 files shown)


gibfahn:
Any objections to me removing the diagnostics sections from the CI jobs? They're now duplicated as they're already in this script, which is run in all the (non-Unix) test jobs AFAICT.

gibfahn:
>The diagnostics have been useful to me on a couple of occasions, but not much. I don't know if anyone else has used this. Also, became a lot less imported once we started killing node processes. So, I'm not sure having diagnostics at all is useful if the maintenance burden becomes high.

Happy to remove them, but tbh it's probably easier to do that (and get support for it) in a PR to this repo, so might as well add first anyway 😁 .

PR #1157, opened 2018-03-01
build and install gcc 4.9.4 on rhel72

jBarz:
Builds and installs gcc 4.9 which isn't officially supported.
The install could take up to 30 mins.
Must use the following

export PATH=/data/gcc-4.9/bin:$PATH
export LDFLAGS="-Wl,-rpath,$(dirname $(/data/gcc-4.9/bin/gcc --print-file-name libgcc_s.so))"

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -125,6 +125,7 @@ hosts:
     - linuxonecc:
         rhel72-s390x-1: {ip: 148.100.110.63}
         rhel72-s390x-2: {ip: 148.100.110.64}
+        rhel72-s390x-3: {ip: 148.100.110.12}
 
     - mininodes:
         ubuntu1604-arm64_odroid_c2-1: {ip: 70.167.220.147}
--- a/ansible/roles/gcc/tasks/main.yml
+++ b/ansible/roles/gcc/tasks/main.yml
@@ -1,0 +1,292 @@
+---
+
+#
+# installs gcc 4.9 @ /data/gcc-4.9 on rhel7.x
+#
+
+- stat: path=/data/gcc-4.9/bin/gcc-4.9
+  register: build_gcc_49
+  tags: gcc49
+
+- name: Clean path
+  file:
+    state: absent
+    path: /data/tmp/
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: Create working directory
+  file:
+    path: /data/tmp
+    state: directory
+    mode: 0755
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: download gmp 6.1.0
+  get_url:
+    url: http://ftp.gnu.org/gnu/gmp/gmp-6.1.0.tar.xz
+    dest: /data/tmp/
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: unarchive gmp 6.1.0
+  unarchive:
+    src: /data/tmp/gmp-6.1.0.tar.xz
+    remote_src: yes
+    dest: /data/tmp/
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: Create build directory
+  file:
+    path: /data/tmp/build
+    state: directory
+    mode: 0755
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: configure gmp 6.1.0
+  shell: /data/tmp/gmp-6.1.0/configure --prefix=/data/gcc-4.9
+  args:
+    chdir: /data/tmp/build
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: build gmp 6.1.0
+  shell: make -j{{ ansible_processor_cores }} install
+  args:
+    chdir: /data/tmp/build
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: Clean path
+  file:
+    state: absent
+    path: /data/tmp/
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: Create working directory
+  file:
+    path: /data/tmp
+    state: directory
+    mode: 0755
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: download mprf 3.1.4 
+  get_url:
+    url: http://ftp.gnu.org/gnu/mpfr/mpfr-3.1.4.tar.gz
+    dest: /data/tmp/
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: unarchive mprf 3.1.4
+  unarchive:
+    src: /data/tmp/mpfr-3.1.4.tar.gz
+    remote_src: yes
+    dest: /data/tmp/
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: Create build directory
+  file:
+    path: /data/tmp/build
+    state: directory
+    mode: 0755
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: build mprf 3.1.4
+  shell: /data/tmp/mpfr-3.1.4/configure --with-gmp=/data/gcc-4.9 --prefix=/data/gcc-4.9
+  args:
+    chdir: /data/tmp/build
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: build mprf 3.1.4
+  shell: make -j{{ ansible_processor_cores }} install
+  args:
+    chdir: /data/tmp/build
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: Clean path
+  file:
+    state: absent
+    path: /data/tmp/
+  when: build_gcc_49.stat.exists == False
+  tags: gcc49
+
+- name: Create working directory
+  file:
+    path: /data/tmp
+    state: directory
+    mode: 0755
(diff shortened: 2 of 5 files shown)


mhdawson:
Not sure the commit title is quite right as it also adds the parts to install java as well right ?

mhdawson:
or maybe its just the comment that is incorrect at the start ?

jBarz:
Hmm, I think java would get installed before this commit. The only new thing installed is gcc

ansible/roles/gcc/tasks/main.yml
```diff
@@ -0,0 +1,199 @@
+---
+
+#
+# installs java
+#
+
```
mhdawson: I was just commenting on this which says "installs java"

jBarz: will fix

mhdawson:
I assume all of the packages being installed are dependencies required for gcc to be compiled right ?

jBarz:
yea, gcc needs those three dependencies.

ansible/roles/gcc/tasks/main.yml
```diff
@@ -0,0 +1,199 @@
+---
+
+#
+# installs gcc 4.9 @ /data/gcc-4.9
```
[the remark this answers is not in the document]
mhdawson: @jBarz ping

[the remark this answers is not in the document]
jBarz: I made changes to not reinstall gcc49 if already installed previously

ansible/inventory.yml
```diff
@@ -124,6 +124,7 @@ hosts:
     - linuxonecc:
         rhel72-s390x-1: {ip: 148.100.110.63}
         rhel72-s390x-2: {ip: 148.100.110.64}
+        rhel72-s390x-3: {ip: 148.100.110.12, user: root}
```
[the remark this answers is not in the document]
gibfahn: @jBarz Can you remove the `user: root` from this ^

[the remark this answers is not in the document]
jBarz: done

mhdawson:
As part of testing this out we found a pieces that were missed during the port over from the old ansible scripts.  These included setting up the user home directory and ccache.  Just running on existing machines is not really a good enough test as it does not validate if something was missed. 

@gibfahn could you review again as I think it is ready to land.

ansible/roles/jenkins-worker/meta/main.yml
```diff
@@ -6,3 +6,4 @@
 
 dependencies:
 - { 'role': 'java-base' }
+- { 'role': 'gcc', when: os in 'rhel72' }
```
gibfahn: Is this going to break when we update to rhel73? Would it make more sense to run this when arch is `s390x` instead, or would it be run on an x64 rhel7 machine if we had one?

ansible/roles/gcc/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
...
```
gibfahn: If this file is only run on linuxone-rhel72 then it would be good to say that in a comment here, as it looks like the generic "install gcc" task.

ansible/roles/jenkins-worker/meta/main.yml
jBarz: It would need to be run on all platforms because gcc 4.9 is not officially supported on rhel72 and rhel73.

ansible/roles/gcc/tasks/main.yml
jBarz: will add that comment

ansible/roles/jenkins-worker/meta/main.yml
jBarz: AFAIK, devtoolset on RHEL requires a paid subscription.

gibfahn: Okay, if that's true, I guess making it rhel makes sense. But is there a way to make it rhel7 rather than rhel72?

jBarz: yes, will change it

```diff
--- a/ansible/roles/jenkins-worker/meta/main.yml
+++ b/ansible/roles/jenkins-worker/meta/main.yml
@@ -6,4 +6,4 @@

 dependencies:
 - { 'role': 'java-base' }
-- { 'role': 'gcc', when: os in 'rhel72' }
+- { 'role': 'gcc', when: os|startswith("rhel7") }
```

jBarz: done

rvagg:
devtoolset is available from the SCLo from CentOS, I would have assumed that it would work for RHEL as well.

What we're doing for Node 10 on CentOS 6 & 7 is pull in devtoolset-6, this is being done in both test and release now so we're at gcc-6 level. Thankfully because of the Red Hat hackery the binaries are backward compatible with vanilla CentOS 6 without devtoolset-\* installed. I'm about to land those changes here https://github.com/nodejs/build/pull/1223 but they are already active in CI and we are switching between devtoolset-2 for pre-10 (no devtoolset on centos7 for pre-10) and devtoolset-6 for 10+ (for both 6 & 7).

Please be aware that if you're pushing for 4.9.4 you're probably pulling in binutils 2.25 _(I think)_ which isn't going to be enough. The OpenSSL upgrade is introducing some new headaches and we need to have 2.26. See https://github.com/nodejs/build/issues/1229, for EL I've just been assuming that devtoolset-6 gets us out of that tangle, does it not help here?

I currently have a 4.9.4 + binutils 2.28 build for the ARM cross compiler so I know it's possible to munge the two. Just make sure you test `as --version` and get >=2.26.

mhdawson:
@rvagg what do you mean by 'not enough' ?

mhdawson:
In the past I'm sure we looked for a devtoolset equivalent for our platforms but either did not find one or did not find one with support for s390 and Power.

mhdawson:
Testing linuxOne, does look like as is too old

```
user1@user1-lbuntu:~/.ssh$ ssh test-linuxonecc-rhel72-s390x-1
Last login: Fri Apr 13 14:44:44 2018 from 32.97.110.55
[root@test-linuxonecc-rhel72-s390x-1 ~]# export PATH=/data/gcc-4.9/bin:$PATH
[root@test-linuxonecc-rhel72-s390x-1 ~]# as --version
GNU assembler version 2.25.1-32.base.el7_4.2 
```

Testing on power

```
iojs@test-osuosl-ubuntu1404-ppc64-le-1:~/build/workspace/node-test-commit-plinux/nodes/ppcle-ubuntu1404$ source build/jenkins/scripts/select-compiler.sh
Setting compiler for Node version 10 on ppc64le
Compiler set to 4.9
iojs@test-osuosl-ubuntu1404-ppc64-le-1:~/build/workspace/node-test-commit-plinux/nodes/ppcle-ubuntu1404$ as --version
GNU assembler (GNU Binutils for Ubuntu) 2.24
Copyright 2013 Free Software Foundation, Inc.
This program is free software; you may redistribute it under the terms of
the GNU General Public License version 3 or later.
This program has absolutely no warranty.
This assembler was configured for a target of `powerpc64le-linux-gnu'.
iojs@test-osuosl-ubuntu1404-ppc64-le-1:~/build/workspace/node-test-commit-plinux/nodes/ppcle-ubuntu1404$ 
```

mhdawson:
@jBarz can you investigate to see what/how we have to do in order to get a later as onto the machine in  similar way to gcc where it will only be used for newer versions?

jBarz:
Added changes to build and install binutils 2.28.1 in the same location as gcc 4.9.4

rvagg:
sorry for the delay--if you have binutils <2.26 then you won't get asm built into openssl and you won't be able to take advantage of cpu crypto routines, which I assume will be a problem for your platforms? it'll be possible to compile and run with older binutils but you won't get an optimal binary. We're going to have to consider the messaging implications of that .. it's going to be tricky.

@jBarz if dropping in 2.28.1 leads to a :+1: compile then all good! I've taken a similar approach of putting a new binutils with gcc 4.9.4 for the cross-compiler for ARM so we're essentially solving it in the same way.

mhdawson:
@rvagg, if we make it good, otherwise I think we'll need something like 'the performance of the initial linxuOne and Linxux on Power builds for 10.x will not be optimal until we update our  build tooling'.

rvagg:
@mhdawson you technically have until we upgrade to OpenSSL 1.1.1, and there's no urgency to do that once it's released (as far as I understand), folks might be itching for TLS 1.3 but that's about all we get from 1.1.1 over 1.1.0 in the short term. I _think_ you could use an older binutils and then drop in a newer one and the binaries will work on the same distro-level, so you have some runway to figure that out.

mhdawson:
Both the compilers and binutils levels are now upgraded.  AIX we still need to figure out but I think we are ok for 10.X

mhdawson:
I'm going to ahead and land this on Monday.

mhdawson:
Landed as 0683e9985b8df43719f869989b36131a338d5609

nealef:
> In the past I'm we looked for a devtoolset equivalent for our platforms but either did not find one or did not find one with support for s390 and Power. @rvagg have you seen one for those platforms?

You can find the devtoolset for s390x [here](http://download.sinenomine.net/clefos/7/sclo/) - we build and maintain a CentOS clone called ClefOS. Also I think RHEL recently also released a subset of the SCLO for s390x.

PR #1162, opened 2018-03-02
coverage: add gcovr version specific diff

mhdawson:
Add diff specific to the 3.4 version of gcovr.  This will
allow us to avoid breaking coverage in older node versions
when we chose to update the version of gcovr that we are
using.


mhdawson:
Landed as 64c0d261aa79b56a2b0e59b5424a4067867081d6

Issue #419, opened 2016-05-18
Add --without-intl builds to CI matrix


bnoordhuis:
Would it be possible to add `./configure --without-intl` buildbots to the CI matrix?  I volunteer to fix up the test suite if necessary.

See https://github.com/nodejs/node/pull/6820 for background.

jbergstroem:
Flavour irrelevant? Linux?

bnoordhuis:
More is better but Linux + Windows seems like a good start.

jbergstroem:
Slightly off-topic but would you see shared builds in the same light?

bnoordhuis:
I'm not dead set against testing it but it's mainly intended for distro packagers, and they know how to help themselves.

Disabling intl (crypto too, for that matter) is more of an end user thing for people on low-end systems.

jbergstroem:
I'll set this up on our most redundant setup.

bnoordhuis:
`--without-inspector` would be good to have too, see https://github.com/nodejs/node/pull/7258.

bnoordhuis:
Another one: https://github.com/nodejs/node/pull/9041

bnoordhuis:
Would also be good for testing pull requests like https://github.com/nodejs/node/pull/9040.

TimothyGu:
I'm tired of this issue being open and inactive for so long. For my purposes I've created https://github.com/TimothyGu/node-no-icu, which uses Travis CI and a local cronjob to run tests without ICU every 6 hours. Doubtless to say, the [first build](https://travis-ci.org/TimothyGu/node-no-icu/builds/288900245) will certainly fail as soon as it finishes, though the [second build](https://travis-ci.org/TimothyGu/node-no-icu/builds/288902809) with nodejs/node#16251 will probably be much more successful.

gibfahn:
>I'm tired of this issue being open and inactive for so long.

Would you like to set up the Jenkins job? This is just waiting on someone having the time, the only bit that needs special access is the initial job creation. I'm happy to clone the Linux test job and give you access.

gibfahn:
>I think this can now be closed.

Is it actually run in the main flows? We should keep this open till that job is good to go. I also don't think running on that many platforms is a good idea, we have a lot of variations to test, so one or two platforms each is probably best.

TimothyGu:
@gibfahn In that case I'd be happy to be given the permissions to modify the job.

gibfahn:
[...]

PR #1163, opened 2018-03-04
Adopt more agressive removal policy for alpine and fedora

rvagg:
The previous policy we've adopted is that we'd keep test machines around as long as they don't cause too much maintenance overhead and we have the resources.

I'd like to propose we adjust that for the Linuxes without long-term support lifecycles, in particular Alpine and Fedora, and adopt the policy of simply removing them when they expire support. There's a bunch of reasons but the primary is actually maintenance overhead, but not in the particular that they are a problem, but in the general that we have too many hosts to manage and the more we have in our CI mix the more work we have to do and the more chance of false positive failures.

This PR also remove alpine from vanilla ansible setup--using Docker for alpine now.


gibfahn:
>There's a bunch of reasons but the primary is actually maintenance overhead, but not in the particular that they are a problem, but in the general that we have too many hosts to manage and the more we have in our CI mix the more work we have to do and the more chance of false positive failures.

I think this is a reasonable definition of maintenance overhead, and thus completely in keeping with our existing policy.

rvagg:
I did this today. Although I left Alpine 3.6 along with 3.7. I looked at the "fedora-latest-x64" and "fedora-last-latest-x64" strategy and decided I really liked it and applied it to Alpine too. I've gone through the Jenkins jobs and removed direct references to Fedora or Alpine versions and used these labels instead. It just means keeping them updated! I discovered that I didn't roll those labels forward when I introduced Fedora 27 so "fedora-latest-x64" was Fedora 26 and "fedora-last-latest-x64" was Fedora 25. Fixed now and I shut down All of the old Fedora machines.

I have not deleted the old Alpine Docker containers from the Docker hosts (that's not as simple as with IaaS deletion), they'll just become static until the next time I deploy over the top of those machines with a new config. I also just introduced a `--without-intl` builder as per #419 so I need to rebalance those Docker hosts with more containers that can serve node-test-commit-linux-containered instead of leaving space for the unused Alpine.

I'm still considering how best to attack documenting all of this, there are a few issues and PRs open relating to policy and documenting these things and I want to pull them all together somehow without making too much work that won't get done.

PR #1167, opened 2018-03-06
jenkins: update compiler selection arg

mhdawson:
Updated the name of the arg used so that
it does not comflict with arguments used in the
release jobs.


mhdawson:
Landing this as I don't think its controversial and will let me progress the le support for both compilers.

mhdawson:
Landed as cd88b556c075c8e38fbb67f3101075a77105a028

PR #1168, opened 2018-03-06
rename linuxone machines

gdams:
The Linuxone machines were all recently updated from Rhel72 to Rhel74. I imagine that the jenkins nodes will also need to be renamed

CC @nodejs/platform-s390

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -55,7 +55,7 @@ hosts:
         centos6-x64-1: {ip: 50.97.245.10}
 
     - linuxonecc:
-        rhel72-s390x-1: {ip: 148.100.110.65}
+        rhel74-s390x-1: {ip: 148.100.110.65}
 
     - packetnet:
         centos7-arm64-1: {ip: 147.75.104.218}
@@ -122,8 +122,8 @@ hosts:
         zos13-s390x-2: {ip: 148.100.36.134, user: Unix1}
 
     - linuxonecc:
-        rhel72-s390x-1: {ip: 148.100.110.63}
-        rhel72-s390x-2: {ip: 148.100.110.64}
+        rhel74-s390x-1: {ip: 148.100.110.63}
+        rhel74-s390x-2: {ip: 148.100.110.64}
 
     - mininodes:
         ubuntu1604-arm64_odroid_c2-1: {ip: 70.167.220.147}
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -55,7 +55,7 @@ hosts:
         centos6-x64-1: {ip: 50.97.245.10}
 
     - linuxonecc:
-        rhel74-s390x-1: {ip: 148.100.110.65}
+        rhel7-s390x-1: {ip: 148.100.110.65}
 
     - packetnet:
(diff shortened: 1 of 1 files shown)


mhdawson:
If/when we do this I think we should rename to Rhel7 so we don't need to do it again.

maclover7:
Please confirm that `playbooks/jenkins/worker/create.yml` works good after this change, I believe "rhel72" or some variant of that is explicitly defined somewhere in "new Ansible" playbooks/roles.

gibfahn:
>If/when we do this I think we should rename to Rhel7 so we don't need to do it again.

Good idea, who does the updates?

So when the machines were updated the hostname was changed as well?

gibfahn:
Yep, plenty to change:

Files:

```
ansible/roles/java-base/vars/main.yml
ansible/roles/baselayout/vars/main.yml
ansible/roles/jenkins-worker/vars/main.yml
ansible/roles/jenkins-worker/tasks/main.yml
ansible/roles/package-upgrade/vars/main.yml
ansible/playbooks/jenkins/worker/create.yml
ansible/roles/jenkins-worker/tasks/partials/tap2junit/rhel72.yml
```


Hits:

```
ansible/roles/jenkins-worker/vars/main.yml
12:  rhel72: 'rhel72',
40:  rhel72: {

ansible/roles/baselayout/vars/main.yml
89:  rhel72: [

ansible/roles/java-base/vars/main.yml
16:  'rhel72': 'java-1.8.0-openjdk',

ansible/roles/jenkins-worker/tasks/main.yml
15:  when: "'rhel72-s390x' in inventory_hostname"
23:  when: "'rhel72-s390x' in inventory_hostname"
67:  when: "'rhel72-s390x' in inventory_hostname"
74:  when: "'rhel72-s390x' in inventory_hostname"
78:  when: "'rhel72-s390x' in inventory_hostname"

ansible/roles/package-upgrade/vars/main.yml
9:    'yum': ['centos', 'rhel72'],

ansible/playbooks/jenkins/worker/create.yml
26:      when: "'rhel72-s390x' in inventory_hostname"

ansible/roles/jenkins-worker/tasks/partials/tap2junit/rhel72.yml
4:# rhel72: python2.7 is default
```

rvagg:
+1 on changing to "rhel7"

If you change the hostname in Ansible I guess you should do it in Jenkins. And when you do it there you'll need to update the secret since it's tied to the name.

mhdawson:
To be honest I don't remember how we change the machine name, may require doing it both in the UI to match as well as on the machine itself as I don't think the UI will propagate.

You do need to change at least ansible templates when you change it in jenkins otherwise you can have problems when after running ansible.  I already had to figure out why the agent was not connecting when I updated the ppc release machine and I think it was because ansible had not been updated (I did not have time to investigate, I just fixed the name in the file that was starting the agent).

I'd personally prioritize getting the OSX/ccache issue resolved finishing up the OSX machines before this renaming work  (unless its already half done in which case we should do the full job)

sam-github:
Too stale, lets close.

richardlau:
We had replaced the LinuxONE hosts and the new machines just have `rhel7` in their names.

Issue #1147, opened 2018-02-24
Install h2load on benchmarking machine


addaleax:
Hey, I think I remember a similar issue being opened about one of the other benchmarkers, but I can’t find it right now. Anyway, it would be nice to have `h2load` installed on the benchmarking machine so that we can run the HTTP/2 benchmarks there?

Otherwise the result is something like https://ci.nodejs.org/view/Node.js%20benchmark/job/benchmark-node-micro-benchmarks/128/console

maclover7:
I'd be happy to take care of this, but I don't seem to have access to the machine (`ssh root@50.23.85.254 -i /Users/jon/.ssh/nodejs_build_test` isn't working.)

cc @mhdawson @rvagg

rvagg:
It's in the infra group. I've added the keys you have in GitHub onto the host so you can get in (maybe you should remove the one you don't need from there). Try to avoid the work if you see it working on any benchmarking tasks.

mhdawson:
I do have an action to add keys specific to the benchmarking WG onto those machines but I have just never had time to get to it.  I added a comment to https://github.com/nodejs/build/issues/1147 so if we ever get around to the ansible scripts we don't forget to add to the list of ones that should be installed.

There are 3 benchmarking machines now, the original one and the 2 new ones from Intel.  Its the 2 Intel ones that run the microbenchmarks job so we need to install on both of those.

addaleax:
Anything I can help with here? It would be nice to have the benchmarking machine available for https://github.com/nodejs/node/pull/18936. :)

mhdawson:
@maclover7 you still planning to get to this ?

addaleax:
Hm … gentle ping @nodejs/build again :)

mhdawson:
@addaleax have you installed this before.  I was looking to see if it as simple as installing with apt-get but all of the instructions I can find require building along with http2 and doing a full install of http2.

addaleax:
@mhdawson What I did locally was building directly from a https://github.com/nghttp2/nghttp2 checkout and installing it (maybe even manually `cp`’ed the `h2load` binary, not sure), that worked well enough for me

maclover7:
Sorry for not getting back to this. This is already in progress via #1171, just haven't had time to get back to it. Will take a look now

Issue #868, opened 2017-09-07
Ansible scripts needed for Benchmark Machines 


mhdawson:
We need numpy, requests, eventlet installed on the benchmark marching to let us run DC-EIS.

I've given @uttampower an ansible introduction and he is going to start working on scripts for the benchmarking machines.  The first thing they will add to the configuration is these there modules.

In the mean time I've manually added them by running 

```bash
apt-get update
apt-get install python-pip
apt-get install python2.7-dev

pip install numpy
pip install requests (reports as already satisfied)
pip install eventlet
```

rvagg:
I'm OK with pip _if we must_ have it, but often the popular packages are available via apt anyway, in this instance `python-numpy`, `python-requests` and `python-eventlet` all seem to be available in both 2.7 and 3 form.

gibfahn:
>I'm OK with pip if we must have it, but often the popular packages are available via apt anyway, in this instance python-numpy, python-requests and python-eventlet all seem to be available in both 2.7 and 3 form.

We're hopefully moving towards tap2junit at some point, and I don't think that'll be available on apt (and yum etc.).

mhdawson:
@rvagg 

so they can be installed with apt-get ie

apt-get install python-numpy
apt-get install python-request
apt-get install python-eventlet 

?

I had used pip as I believe it was already going to be available due to the need for tap2junit but when we work on the ansible scripts we can install the apt packages instead if they are available.

rvagg:
Ah, well if we're using pip already then I guess it comes down to updating. apt-get is often run manually by myself and maybe others on our test machines but that's all for updating. If there's a pip update process it's unlikely to ever be run. Is that a problem? I don't know! I feel like this is one for the python folks among us to answer.
Sorry for the non-answer, my preference is still for apt-get install (`apt-get install python-numpy python-request python-eventlet -y`) but it's no big deal if we do it with pip _I guess_.

gibfahn:
I'd prefer using `apt` where available too FWIW.

It would make sense to periodically rerun our ansible scripts (once we have them set up) to update everything.

mhdawson:
@uttampower have you had a chance to start working on this yet ?

maclover7:
Related issues: #821, #955

mhdawson:
Related issue https://github.com/nodejs/build/issues/1147

PR #1171, opened 2018-03-07
ansible: add benchmarking

maclover7:
Fixes: https://github.com/nodejs/build/issues/868
Fixes: https://github.com/nodejs/build/issues/1147
    
Creates a new `benchmarking` Ansible role, and runs for all machines
with the `is_benchmark` machine variable set to `true`.
    
Included `infra-softlayer-ubuntu1404-x64-2` since it is also a benchmark
machine, in addition to `test-nearform_intel-ubuntu1604-x64-1` and
`x64-2`.


addaleax:
The YAML file LGTM if it works :)

ansible/host_vars/README.md
```diff
@@ -16,6 +16,8 @@ You will always have to set the following variables to configure a host:
 Variables that _might_ be available for you to change depending on
 what init system your host will be running:
 
+- `is_benchmark`: set to `true`/`false`. If true, will run the
```
mhdawson: I'm not sure this would be the way that @jbergstroem.  Instead I think I would have expected a new role that inherits from jenkins worker but does additional configuration.  I could be wrong though.  @jbergstroem can you comment.

mhdawson: Looking at it in more details I think I better understand the issue, the naming of our machines still uses test_-need something to identify what role.  We might want to rename the benchmarking machines to start with bench- instead  and then use that to select the role.

ansible/roles/benchmarking/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+  become: true
+  become_user: iojs
+
+- name: Add h2load to $PATH
+  lineinfile:
+    path: "/home/{{ server_user }}/.bashrc"
+    regexp: '^export PATH='
+    line: "export PATH=$PATH:/home/{{ server_user }}/nghttp2/src"
+  become: true
+  become_user: iojs
```
mhdawson: I'm not sure if this will achieve what is desired as I'm now 100% sure the .bashrc will run when the jenkins agent is started.

Usually, paths have been added in the file used to start the jenkins agent.  For example: https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/templates/upstart.j2

maclover7:
@mhdawson Related to your first two comments -- the benchmarking role is enabled through a `is_benchmark` host variable, so any host with that will automatically get the role applied to it.

The reason why `infra-test-softlayer-ubuntu1404-x64-2` has to be directly listed is because it is from the `infra` group, not from `test`. I've already run `playbooks/jenkins/worker/create.yml` on the 1 SoftLayer machine, as well as the 2 Intel ones.

mhdawson:
@maclover7 I understand that the roles is enabled through is_benchmark, just not sure that is aligned with how selecting roles has been done so far.  I think what would be consistent would be to rename the machines to start with benchmark (instead of test) and then use that to do the selection.   Not to say that's the right way to do it, only that it seems most consistent with the existing pattern.

maclover7:
@addaleax @mhdawson Confirm that this worked via https://ci.nodejs.org/view/Node.js%20benchmark/job/benchmark-node-micro-benchmarks/204/console.

My suggestion is to merge this is as for now. Then as a followup, one of the Build WG admins can rename the three machines that run benchmarks to start with `bench-` (instead of `infra-`/`test-`), and update the Ansible inventory file. I do not have permissions to make this change.

ansible/roles/jenkins-worker/tasks/main.yml
```diff
@@ -228,7 +228,7 @@
 
 - name: enable jenkins at startup - general
   when: not os|startswith("zos") and not os|startswith("macos")
-  service: name=jenkins state=started enabled=yes
+  service: name=jenkins state=restarted enabled=yes
```
mhdawson: Does this work when the agent is not running?

mhdawson:
I'm ok with landing this but I think ansible/playbooks/jenkins/worker/create.yml needs updates to refer to 
  test-nearform_intel-ubuntu1604-x64-1  and  test-nearform_intel-ubuntu1604-x64-2 as the benchmark machines.  I see the test job you ran, executed on  test-nearform_intel-ubuntu1604-x64-1

ansible/roles/jenkins-worker/tasks/main.yml
maclover7: Just double checked on a machine, and looks to run ok

maclover7:
> I'm ok with landing this but I think ansible/playbooks/jenkins/worker/create.yml needs updates to refer to
test-nearform_intel-ubuntu1604-x64-1 and test-nearform_intel-ubuntu1604-x64-2 as the benchmark machines. I see the test job you ran, executed on test-nearform_intel-ubuntu1604-x64-1

I had run the Ansible script on all three machines -- the way the benchmark machines are picked is via a `is_benchmark` host variable (similar to how we use `secret`, `server_jobs`, etc.), until we make that switchover on renaming the machines

PR #1175, opened 2018-03-08
jenkins: fix ccache use with compiler selectcion

mhdawson:
remove the paths so that we work ok with ccache


mhdawson:
Landed as cf15dd993539b7301b4e27641c5a3ea4b5c1bc72

PR #1181, opened 2018-03-13
Serve security.txt file from the static directory

MarcinHoppe:
The URLs are recommended by the security.txt spec available at https://securitytxt.org.

The security.txt file was added to the static directory in a separate PR: nodejs/nodejs.org#1589.

### DIFF
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -301,11 +301,6 @@ server {
         default_type text/plain;
     }
 
-    location /security.txt {
-        alias /home/www/nodejs/static/security.txt;
-        default_type text/plain;
-    }
-
     location /.well-known/security.txt {
         alias /home/www/nodejs/static/security.txt;
         default_type text/plain;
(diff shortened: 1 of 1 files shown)


setup/www/resources/config/nodejs.org
```diff
@@ -301,6 +301,16 @@ server {
         default_type text/plain;
     }
 
+    location /security.txt {
+        alias /home/www/nodejs/static/security.txt;
```
ChALkeR: Per [spec draft](https://tools.ietf.org/html/draft-foudil-securitytxt-03#section-4), this should be a redirect to `/.well-known/security.txt`.

MarcinHoppe: More specifically:

> A security.txt file located under the top-level path SHOULD either redirect to the security.txt file under the /.well-known/ path or be used as a fall back.

I guess the current implementation above can be considered a fallback, but IANAL :).

I don't have strong feelings about this, happy to configure a redirect.

ChALkeR: The scheme just above that section suggests a redirect. I would prefer this to be a redirect — that would hint that this is a part of some spec, not just a random file.

ChALkeR:
:+1:

Not presisng the «review» button as I don't feel confident enough reviewing changes in this repo — so let's wait for someone from @nodejs/build wg to review this.

gibfahn:
Needs a review from someone who has looked at these files recently, so @rvagg (or anyone else in @nodejs/build who has).

MarcinHoppe:
@nodejs/build is there any way we can move this forward? The `security.txt` is already in the Website content but its usefulness will be little if we don't put it in a well known location.

Let me know if there's anything I can do to move this forward.

gibfahn:
>@nodejs/build is there any way we can move this forward?

Just needs more build members to look at it. I'll ping some people.

rvagg:
fine by me

rvagg:
live now

PR #1189, opened 2018-03-23
update dist.libuv.org config to match what's live now

rvagg:
Prompted by https://github.com/libuv/libuv/issues/1775

This is what's on the server now, strict https. I don't know did this, perhaps it was me! The timestamp on the file is Aug 15 2017.

My guess re https://github.com/libuv/libuv/issues/1775 is that the removal of the ipv6 line is causing problems because it's redirecting to nodejs.org. Unfortunately I'm a little afraid to touch this because of the interactions with the other hosts running off nginx on this particular server. I wish I knew why it was removed!

### DIFF
--- a/setup/www/resources/config/libuv.org
+++ b/setup/www/resources/config/libuv.org
@@ -1,7 +1,18 @@
 server {
     listen *:80;
-    listen [::]:80;
     server_name dist.libuv.org;
+
+    return 301 https://dist.libuv.org$request_uri;
+}
+
+server {
+    listen [::]:443 ssl http2;
+    listen *:443 ssl http2;
+    server_name dist.libuv.org;
+
+    ssl_certificate ssl/libuv_chained.crt;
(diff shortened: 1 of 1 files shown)


setup/www/resources/config/libuv.org
```diff
@@ -1,14 +1,26 @@
 server {
     listen *:80;
-    listen [::]:80;
```
bnoordhuis: Correct me if I'm wrong, but this should be safe to add back in because ultimately the server is selected by the server_name token, right?

Aside: it can probably be generalized to e.g.
```
server_name dist.libuv.org, dist.nodejs.org, ...;
return 301 https://$server_name$request_uri;
```

rvagg: FYI yes it's been added back in in #1402

rvagg:
resolved via #1402

Issue #1193, opened 2018-03-25
Moving CI temporaries repo to our own infra


joaocgreis:
Janea Systems received a request from GitHub to discontinue the usage pattern of `node_binary_tmp`, where we store the temporary binaries used by CI.

I will move forward by creating git repositories in our `jenkins-workspace` machines, that should have enough storage space and be adequate for this usage. For now, I'll set it up so that only one server will be used at a time. I don't see a simple way to implement a load balancing scheme resilient to server failures, and I believe this will be stable enough. If it proves to be a problem, we'll have to investigate a better way or create a dedicated git server.

joaocgreis:
@nodejs/jenkins-admins to give us a way to change the temporaries repo to a different `jenkins-workspace` fast if needed, I added `TEMP_REPO` and `TEMP_REPO_SERVER` as global Jenkins variables, defined in _Manage Jenkins/Configure System_. I removed `TEMP_REPO` as a parameter from everywhere, I believe it was very rarely changed if at all, and the default value would be a problem here.

PR with ansible changes: https://github.com/nodejs/build/pull/1195

I'll leave this issue open until that PR lands.

rvagg:
this sucks, is it a workload thing? they noticed that repo has a _ton_ of churn and it's all bulky?

joaocgreis:
They don't delete the objects after they become unreferenced, so even if we remove the branches the repository grows huge in their servers. They mentioned it was at 140GB, and the servers are not designed for this usage pattern.

PR #1195, opened 2018-03-25
ansible: add jenkins-workspace with git repo

joaocgreis:
Dedicated playbook for the `jenkins-workspace` machines, including a git repo for temporary binaries.

Fixes: https://github.com/nodejs/build/issues/1193

cc @rvagg


ansible/roles/jenkins-workspace/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+- name: Add nodejs-ci to authorized_keys for binary_tmp
+  authorized_key:
+    user: "binary_tmp"
+    key: "{{ lookup('file', '/tmp/nodejs-ci.keys') }}"
+
+# Repository needs to be created in /home/iojs/build because the partition with
+# free space might be mounted in a way that does not include /home/binary_tmp
+- name: Create repository directory
+  file:
+    path: "~iojs/build/binary_tmp.git"
```
rvagg: better be explicit and use `{{ server_user }}` like everywhere else - `/home/{{ server_user }}/build/binary_tmp.git`

ansible/roles/jenkins-workspace/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+    state: link
+    owner: "binary_tmp"
+    group: "binary_tmp"
+    mode: 0755
+
+- name: Initialize Git repository
+  become: true
+  become_user: binary_tmp
+  git:
+    repo: "https://www.github.com/nodejs/node.git"
```
rvagg: `https://github.com/nodejs/node`

rvagg:
mm, I guess this is OK. When you define a new section in `create.yml` will it run these extra commands after running everything else for these hosts, so these are just addons to the normal setup? Hard-wiring so much is a bit of a pain, the hostnames being the biggest annoyance. Perhaps we can do a lookup for the `jenkins-workspace` alias with a `when`. For now, for expedience I guess this is OK.

rvagg:
I think the next step after this quick fix might be to just host the binaries static rather than in git. Give them unique but predictable names, use `scp` to grab them, use crontab to clean out old ones. There's nothing special here that requires git right?

nschonni:
You might look at this too https://developers.googleblog.com/2018/03/artifact-management-for-open-source.html

joaocgreis:
Updated, thanks for the review.

> mm, I guess this is OK. When you define a new section in create.yml will it run these extra commands after running everything else for these hosts, so these are just addons to the normal setup?

Yes. If you're `--limit`ing for hosts that do not match, the whole section is skipped. If not, only hosts that match will run it. I agree it would be better to find a way that let us specify in the inventory which hosts are workspaces.

> I think the next step after this quick fix might be to just host the binaries static rather than in git. Give them unique but predictable names, use scp to grab them, use crontab to clean out old ones. There's nothing special here that requires git right?

I thought about this when I developed the temporaries system, and went with Git as it turned out to be simpler because we could leverage GitHub. Having files out of Git for temporaries is somewhat conceptually tidier, but note that Git solves two problems for us that wouldn't be so straightforward otherwise:
- Matching binaries with Git commit: using Git, it's the parent. Very little chance for mistakes there. Otherwise, we'd have to add at least two parameters to each Jenkins job to specify what commit to test, and the logic to go with it.
- Rebases: `REBASE_ONTO` is usually a commit hash, so jobs that rebase need to fetch a refspec that includes all branches. Not having to do this is an advantage mainly for the Pis, but also to ensure the exact same rebase output is tested everywhere (we had that `POST_REBASE_SHA1_CHECK` parameter to ensure this, but it failed and it was not obvious why at the time).

So, I don't really see not using Git as a step forward, but perhaps I could be persuaded. I quite like the idea of using `git-rebase` as an entry way, and wanted to test using it at `node-test-commit` level but never got to do it.

There is a job to clean old branches: https://ci.nodejs.org/view/All/job/git-clean-tmp-repo/

joaocgreis:
@maclover7 this is already rolled out, I thought I mentioned it in https://github.com/nodejs/build/issues/1193 but I see I was not very clear about it. I made the changes in Jenkins during a quiet period in the weekend, first pointing to the old repo and then monitored closely when I switched it. As for Ansible, `--step` helped a lot (ran only the new parts, avoiding things like upgrading packages).

joaocgreis:
I plan to land this soon if there are no objections.

@rvagg I've addressed your suggestions, let me know if you have more or would like me to hold this longer.

rvagg:
fine by me

Issue #829, opened 2017-08-10
armv7 binaries are compiled for armv6


bnoordhuis:
```
$ strings node-v8.3.0-linux-armv7l/bin/node | grep arm_version
                 'arm_version': '6',
```
It's not new, see https://github.com/nodejs/node/issues/4531#issuecomment-168903077.

gibfahn:
Copying @rvagg 's comment from https://github.com/nodejs/node/issues/4531#issuecomment-168903077

> Thanks for reporting this @igorklopov, I'm a little surprised this went unnoticed for so long. I can confirm your analysis and I've traced it to the fact that we're using Raspbian to get gcc 4.8 on the machine we're using to compile these binaries even though it's a vanilla Debian Wheezy armhf base, and of course Raspbian is armv6 so we get `#define __ARM_ARCH 6` from the compiler...
> 
> Doh!
> 
> Which means we need to go back to finding a Wheezy armhf gcc 4.8(+) source or build one ourselves (which I think I did in the past but that's such a nightmare).
>

gibfahn:
@xadillax are you still working on this (see https://github.com/nodejs/node/issues/4531#issuecomment-309050134)?

rvagg:
:man_facepalming: this is technically stalled and it's my fault. I have a server running on scaleway with custom compiled 4.9 gcc packages that I was going to install on those other boxes. I stalled on trying to figure out _where_ to store these packages so they can be reused when reprovisioning the boxes, I don't really want to check them in to this repo.

piranna:
Don't worry @rvagg, all we have our private lives and sometimes it gets busy :-)

XadillaX:
@gibfahn Yes but slow progress due to my poor PI3.

refack:
ping ya'll

mhdawson:
@rvagg does it make sense to put those on the ci under the download user  like we are for packages needed for other platforms ?

piranna:
> @rvagg does it make sense to put those on the ci under the download user like we are for packages needed for other platforms ?

I think it would be a good idea. Would it be possible to also have downloads of binaries compiled with other standard C libraries beside glibc, like musl?

gibfahn:
>> @rvagg does it make sense to put those on the ci under the download user like we are for packages needed for other platforms ?
>
>I think it would be a good idea. Would it be possible to also have downloads of binaries compiled with other standard C libraries beside glibc, like musl?

[...]

PR #1204, opened 2018-04-02
Refactor ARMv7 Scaleway host config

rvagg:
the armv7-wheezy hosts were never in the ansible/ config but now they are. I've refactored the big Raspberry Pi changes from #1199 so that I could reuse the docker config. Now we have 3 new Scaleway ARMv7 machines running Ubuntu 16.04 with Docker on them and containers for Debian Wheezy, Jessie and Stretch (using Ubuntu for the host because I couldn't get around a strange Docker problem on Stretch). They are in [node-test-commit-arm](https://ci.nodejs.org/job/node-test-commit-arm/) under the labels `debian7-docker-armv7`, `debian8-docker-armv7` and `debian9-docker-armv7`. I'm trying out the [node-version-jenkins-plugin](https://github.com/nodejs/node-version-jenkins-plugin) for this one which uses [VersionSelectorScript.groovy](https://github.com/nodejs/build/blob/rvagg/VersionSelectorScript.groovy/jenkins/scripts/VersionSelectorScript.groovy) as ci-release does to select which builders to use. For Node <10 it'll run tests on debian7 (Wheezy) and debian8 (Jessie) and for Node >=10 it'll run debian8 and debian9 (Stretch) (could run all 3 for <10 but they're not fast so two is more than enough).

As I mentioned in https://github.com/nodejs/build/issues/829, we're still locked in to a Raspbian gcc 4.8 on Wheezy and that's reflected in this PR as well, but thankfully we get a 4.9 in Jessie and a gcc 6 in Stretch so for Node 10 we're now only testing armv7 on gcc 4.9.2 minimum.


refack:
You forgot the libuv job who got starved for `armv7-wheezy`.
I changed labels from `armv7-wheezy` to `debian7-docker-armv7`
https://ci.nodejs.org/job/libuv-test-commit-linux/jobConfigHistory/showDiffFiles?timestamp1=2018-03-06_23-17-14&timestamp2=2018-04-13_15-12-50

PR #1214, opened 2018-04-05
nginx: Redirect old inspector docs

fhemberger:
Ref: https://github.com/nodejs/nodejs.org/pull/1610


rvagg:
whoops, I shouldn't have merged this. The ansible/roles/nginx/templates/nodejs.org.conf.j2 isn't in active use, the refactor for the nodejs.org configs never got fully integrated to the new ansible/ scripts. The one we're using is still @ setup/www/resources/config/nodejs.org. Could you PR to that file instead please @fhemberger?

fhemberger:
@rvagg Sure, here you go: #1215

PR #1216, opened 2018-04-05
build: add compiler selection for s390

mhdawson:


mhdawson:
Please review but do not land, as I need to deploy the compiler to all machines before that happens which I'm going to stage.

mhdawson:
@gibfahn can you take a look

mhdawson:
Although they both use 4.9 there is a fair amount of difference between the config.  If other platforms have more commonality then we can combine.

I probably should have mentioned that this is being used live until it lands.

Plan to update the linxuOne Release machine on Wed after the RC's go out and then land this.

mhdawson:
Landed as cf7370fda47e146ac6e26d2a5baed208c32dd7b4

Issue #1153, opened 2018-02-27
A novel approach to trimming Jenkins nodes by Node.js version


rvagg:
For discussion, although I have a working implementing of the approach I'm detailing below on ci-release atm.

We have 2 different strategies for doing different things in Jenkins according to Node.js version:

1. Detecting `NODE_MAJOR_VERSION` via some Bash scriptery inside a parent job and bubbling it down via environment variables (via env.properties). You can see this in node-test-commit and then it's used in the 3 jobs that are used to compile and test for Windows (node-test-binary-windows, node-compile-windows, node-test-commit-windows-fanned) to select amongst Windows builders.
2. Hacky scripting in Bash and BAT. iojs+release on ci-release is actually pretty entertaining if you look at the many different ways that all of us with access have gone in and added rules that restrict jobs on particular hosts. This is getting more and more complicated.

Our rules are most strict on ci-release where it's critical we get the right builders for a release:

 * CentOS 5 for Node 4, CentOS 6 above that for Linux
 * No PPC BE after Node 8.
 * s390x and AIX only for Node 6 and later
 * SmartOS 14 before Node 8, Smart OS 15 beyond that (we also have SmartOS 13 still in there for <1 ...)
 * MSVC2013 before Node 6. MSVC2015 from 6 up to 10. MSVC2017 from 10 onward

Node 10 is going to make it more complicated. I think for instance we're going to have to introduce new rules for ARMv6 & ARMv7 (and possibly new machines) because we're still using Wheezy, plus I think CentOS 7 might be our new target builder for Node 10+ and probably a new SmartOS too (this discussion is for a new issue).

There are drawbacks to both approaches we use:

1. To get a correct `NODE_MAJOR_VERSION` you need to run a parent job to inspect the source, _or_ you have to manually enter it in the job (it's currently set as default `10` and that needs to be incremented each time we make a major release but, for example, if you want to test Node 8 and you don't select it then you'd get the wrong compiler and you may not even know. We've also kept ci-release pretty simple and only have a single job, with no hierarchy, so we can't use this approach unless we want to start splitting it up (my preference is to KISS but it's getting hard).
2. With our scripting approach in iojs+release, these scripts are run on _all_ machines in the release pool, so they have to check out Node and check the source and exit prematurely if they aren't supposed to run. This can be annoying if you have a flaky machine or a machine offline that you don't even need (CentOS 5 is hardly ever used there but it's also one of the most problematic machines to keep happy).

[...]

PR #1217, opened 2018-04-06
added VersionSelectorScript.groovy to select builders for Node version

rvagg:
Following on from https://github.com/nodejs/build/issues/1153

node-version-jenkins-plugin is now installed on ci.nodejs.org and is being used in node-test-commit-linux-arm, node-test-commit-linux-containered and node-test-commit-linux to select the right builders to run. So we have some ARM and containered specific options in here. The script also has a "buildType"  so that it can do slightly different things for release as for test (and other), e.g. centos6 shouldn't be used for Node 4 in release but it should be used for everything for test.

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -1,10 +1,17 @@
+// Helper closures to make our buildExclusions DSL terse and readable
 def gte = { v -> { nodeVersion -> nodeVersion >= v} }
 def lt = { v -> { nodeVersion -> nodeVersion < v} }
 def releaseType = { buildType -> buildType == 'release' }
 def anyType = { buildType -> true }
 
 def buildExclusions = [
-  // Linux
+  // Given a machine label, build type (release or !release) and a Node.js
+  // major version number, determine which ones to _exclude_ from a build
+
+  // -------------------------------------------------------
+  // Machine Label,                   Build Type,  Node Version
+
+  // Linux -------------------------------------------------
   [ /^centos5/,                       anyType,     gte(8)  ],
   [ /^centos6/,                       releaseType, lt(8)   ],
   [ /centos[67]-(arm)?(64|32)-gcc48/, anyType,     gte(10) ],
@@ -13,38 +20,42 @@ def buildExclusions = [
   [ /^ubuntu1804/,                    anyType,     lt(10)  ], // probably temporary
   [ /^ubuntu1204/,                    anyType,     gte(10) ],
 
-  // Windows
+  // ARM  --------------------------------------------------
+  [ /^debian7-docker-armv7$/,         anyType,     gte(10) ],
+  [ /^debian8-docker-armv7$/,         releaseType, lt(10)  ],
+  [ /^debian9-docker-armv7$/,         anyType,     lt(10)  ],
+
+  // Windows -----------------------------------------------
   [ /^vs2013-/,                       anyType,     gte(6)  ],
   [ /^vs2015-/,                       anyType,     lt(6)   ],
   [ /^vs2015-/,                       anyType,     gte(10) ],
   [ /^vs2017-/,                       anyType,     lt(10)  ],
 
-  // SmartOS
+  // SmartOS -----------------------------------------------
   [ /^smartos14/,                     anyType,     gte(8)  ],
   [ /^smartos15/,                     anyType,     lt(8)   ],
   [ /^smartos15/,                     releaseType, gte(10) ],
   [ /^smartos16/,                     anyType,     lt(8)   ],
   [ /^smartos17/,                     anyType,     lt(10)  ],
 
-  [ /^debian7-docker-armv7$/,         anyType,     gte(10) ],
-  [ /^debian8-docker-armv7$/,         releaseType, lt(10)  ],
-  [ /^debian9-docker-armv7$/,         anyType,     lt(10)  ],
-
-  // PPC BE
+  // PPC BE ------------------------------------------------
   [ /^ppcbe-ubuntu/,                  anyType,     gte(8)  ],
 
-  // s390x
+  // s390x -------------------------------------------------
   [ /s390x/,                          anyType,     lt(6)   ],
 
-  // AIX61
+  // AIX61 -------------------------------------------------
   [ /aix61/,                          anyType,     lt(6)   ],
 
-  // sharedlibs containered
+  // Shared libs docker containers -------------------------
   [ /sharedlibs_openssl111/,          anyType,     lt(9)   ],
   [ /sharedlibs_openssl110/,          anyType,     lt(9)   ],
   [ /sharedlibs_openssl102/,          anyType,     gte(10) ],
   [ /sharedlibs_fips20/,              anyType,     gte(10) ],
   [ /sharedlibs_withoutintl/,         anyType,     lt(9)   ],
+
+  // -------------------------------------------------------
+
 ]
 
 def canBuild = { nodeVersion, builderLabel, buildType ->
@@ -67,8 +78,13 @@ if (parameters['NODEJS_MAJOR_VERSION'])
 println "Node.js major version: $nodeMajorVersion"
 println "Node.js version: ${parameters['NODEJS_VERSION']}"
 
+// NOTE: this assumes that the default "Slaves"->"Name" in the Configuration
+// Matrix is left as "nodes", if it's changed then `it.nodes` below won't work
+// and returning a result with a "nodes" property won't work.
 result['nodes'] = []
 
+// Before running this script, `def buildType = 'release'` or some other value
+// to be able to use the appropriate `buildType` in the exclusions
 def _buildType
 try {
   _buildType = buildType
@@ -78,6 +94,8 @@ try {
 
 combinations.each{
   def builderLabel = it.nodes
+  // Default to running all builders if nodeMajorVersion is still -1
+  // (i.e. the version check failed)
   if (nodeMajorVersion >= 4) {
     if (!canBuild(nodeMajorVersion, builderLabel, _buildType)) {
       println "Skipping $builderLabel for Node.js $nodeMajorVersion"
@@ -86,4 +104,3 @@ combinations.each{
   }
   result['nodes'].add(it)
 }
-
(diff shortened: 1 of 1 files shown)


mhdawson:
Also is this just for release or is it going to be used for the test ci as well?   Was going to look at turning off testing for BE >=8 but this may take care of it if it will apply to the test ci.

rvagg:
@mhdawson re ppc, cause the script is already set up for this I've just edited node-test-commit-plinux and (1) enabled the plugin (under git advanced) and (2) inserted the Groovy Matrix Strategy script to download this script and execute it for filtering and now ppcbe is not running for Node >=8 like on ci-release. If you want it to run later than that then we can modify it to be different for `buildType == 'release'` vs test.

https://ci.nodejs.org/job/node-test-commit-plinux/jobConfigHistory/showDiffFiles?timestamp1=2018-04-04_06-56-16&timestamp2=2018-04-10_04-51-40

Regarding docs, in theory the only thing that should need to be done is edit this script. Unfortunately I'm finding that the Groovy script is caching downloads so it doesn't fetch updates straight away and I'm having to edit the URL inside the jobs to add a unique `?x` on the end to fetch the new version. Not ideal, so I'm going to have to get my Groovy on and figure out a non-caching fetch. Then, if I get that working, it's only editing this script that's needed to change the rules. Of course there are many edge-cases, like the centos gcc one I just implemented that involves an extra layer of hackery, but at least we can remove all version checking into one place.

rvagg:
```groovy
def code = new URL(scriptUrl).getText(useCaches: false, requestProperties: ['Connection': 'close'])
```

these options seem to be enough to force proper single-use fetching 👍

mhdawson:
@rvagg, thanks for the update.  Having the test use the same rules as for release for PPCBE works for now.  Thanks :)

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -0,0 +1,94 @@
+def canBuild(nodeMajorVersion, builderLabel, buildType) {
+
+  // Linux
+  if (builderLabel.indexOf('centos5') == 0 && nodeMajorVersion >= 8)
```
gibfahn: FWIW you can do `.indexOf('x') == 0` -> `startsWith('x')`

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -0,0 +1,94 @@
...
+
+combinations.each{
+  def builderLabel = it.nodes
+  if (nodeMajorVersion >= 4) {
+    if (!canBuild(nodeMajorVersion, builderLabel, _buildType)) {
+      println 'Skipping ' + builderLabel + ' for Node.js ' + nodeMajorVersion
+      return
+    }
+  }
+  result['nodes'] << it
```
gibfahn: does `<<` mean `.add()`? If so maybe just use `.add()`, to keep it simple for people unfamiliar with groovy (i.e. all of us).

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -0,0 +1,94 @@
...
+    return false
+
+  return true
+}
+
+// setup for execution of the above rules
+
+int nodeMajorVersion = -1
+if (parameters['NODEJS_MAJOR_VERSION'])
+  nodeMajorVersion = parameters['NODEJS_MAJOR_VERSION'].toString().toInteger()
```
gibfahn: I don't think this needs the `toString()`

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -0,0 +1,94 @@
...
+    return false
+  if (builderLabel.indexOf('sharedlibs_withoutintl') > -1 && nodeMajorVersion < 9)
+    return false
+
+  return true
+}
+
+// setup for execution of the above rules
+
+int nodeMajorVersion = -1
```
gibfahn: So this will just skip everything if something goes wrong right? If that happens then that means the build will be green right?

It'd be nice to fail the build if we get an unexpected Node version, maybe just ensure it's in a certain range, and then throw if not.

That would also remind us to update the supported platforms list before a new Node version comes out, and give it some urgency.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -0,0 +1,94 @@
...
+  if (buildType == 'release' && builderLabel.indexOf('centos6') == 0 && nodeMajorVersion < 8)
+    return false
+  if (builderLabel =~ /centos[67]-(arm)?(64|32)-gcc48/ && nodeMajorVersion >= 10)
+    return false
+  if (builderLabel =~ /centos[67]-(arm)?(64|32)-gcc6/ && nodeMajorVersion < 10)
+    return false
+  if (builderLabel.indexOf('ubuntu1804') == 0 && nodeMajorVersion < 10) // probably temporary
+    return false
+  if (builderLabel.indexOf('ubuntu1204') == 0 && nodeMajorVersion >= 10)
+    return false
```
gibfahn: I'm not _sure_ about this, so just a suggestion, but it might be easier to parse if we grouped these by Node version as well as arch, i.e.:

```groovy
  if (builderLabel =~ /centos[67]-(arm)?(64|32)-gcc48/ && nodeMajorVersion >= 10)
    return false
  if (builderLabel =~ /centos[67]-(arm)?(64|32)-gcc6/ && nodeMajorVersion < 10)
    return false
  if (builderLabel.indexOf('ubuntu1804') == 0 && nodeMajorVersion < 10) // probably temporary
    return false
  if (builderLabel.indexOf('ubuntu1204') == 0 && nodeMajorVersion >= 10)
    return false
```

to 

```groovy
if (nodeMajorVersion >= 10) {
  if (builderLabel =~ /centos[67]-(arm)?(64|32)-gcc48/)
    return false
  if (builderLabel =~ /centos[67]-(arm)?(64|32)-gcc6/)
    return false
  if (builderLabel.startsWith('ubuntu1804')) // probably temporary
    return false
  if (builderLabel.startsWith('ubuntu1204'))
    return false
}
```

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -0,0 +1,94 @@
...
+  if (builderLabel.indexOf('smartos13') == 0) // Node.js 0.x
+    return false
+  if (builderLabel.indexOf('smartos14') == 0 && nodeMajorVersion >= 8)
+    return false
+  if (builderLabel.indexOf('smartos15') == 0 && nodeMajorVersion < 8)
+    return false
+
+  if (builderLabel.equals('debian7-docker-armv7') && nodeMajorVersion >= 10)
+    return false
+  if (buildType == 'release' && builderLabel.equals('debian8-docker-armv7') && nodeMajorVersion < 10)
```
gibfahn: Is there a difference between `==` and `.equals()`?

rvagg:
@gibfahn I've experimented with two new variations of this that make it a bit more grokkable

1. everything is a regex and names have been simplified: (https://github.com/nodejs/build/blob/88142ce670883a31eaecddb2f7b4bf74ddd9b88e/jenkins/scripts/VersionSelectorScript.groovy)
2. turn it into a DSL, the "build exclusions" are in the form of an array of arrays with 3 elements: regex, build type, version match (current HEAD)

2 is what's being used now, I quite like it cause it reads easily, we end up being able to define rules like:

```groovy
  [ /^centos5/,                       anyType,     gte(8)  ],
  [ /^centos6/,                       releaseType, lt(8)   ],
  [ /centos[67]-(arm)?(64|32)-gcc48/, anyType,     gte(10) ],
  [ /centos[67]-(arm)?(64|32)-gcc6/,  anyType,     lt(10)  ],
  [ /centos6-32-gcc6/,                releaseType, gte(10) ], // 32-bit linux for <10 only
  [ /^ubuntu1804/,                    anyType,     lt(10)  ], // probably temporary
  [ /^ubuntu1204/,                    anyType,     gte(10) ],
```

I've addressed your other comments with these changes I think, two outstanding ones:
 * re the `-1` default, if something messes up then _all_ builders will run, there's a `>=4` in there that makes that happen. Basically if you submit a repo to it that it can't figure out the node version from src/node_version.h then it'll just throw its hands up and guess that you probably didn't intend rules to apply to this one. I could make it fail with a dramatic `throw` or something if you think this is a big enough deal.
 * The `.toString().toInteger()` thing is necessary because it's a `ParameterValue` or somesuch from Jenkins, so the `toString()` extracts the plain value out of it.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -0,0 +1,89 @@
+def gte = { v -> { nodeVersion -> nodeVersion >= v} }
+def lt = { v -> { nodeVersion -> nodeVersion < v} }
+def releaseType = { buildType -> buildType == 'release' }
+def anyType = { buildType -> true }
+
+def buildExclusions = [
+  // Linux
```
gibfahn: Maybe add a nice header like:

```groovy
   // Machine Label                    Release Type  Node Major Version 
   [ /^centos5/,                       anyType,      gte(8)  ], 
```

jenkins/scripts/VersionSelectorScript.groovy
gibfahn: Ahh, I think it's because I normally use `params['NODEJS_MAJOR_VERSION']`, which I think is a proper string.

Nbd though.

maclover7:
@rvagg This should be safe to land, right? It's already been running on public CI for a couple of weeks now

rvagg:
yeah, but we're referencing this branch now in a bunch of places so I need to go find all those references and point them to master!

refack:
moved `node-test-commit-linux` to master

refack:
Please don't delete branch until we're sure Jenkins is branch-use free.

refack:
Just run this in the Script Console and fixed all the jobs
```groovy
import org.jenkinsci.plugins.GroovyScriptMES
import hudson.matrix.MatrixProject

for(item in Jenkins.instance.getAllItems(MatrixProject.class)) {
  e = item.getExecutionStrategy()
  if (e in GroovyScriptMES) {
    println e.secureScript.getScript().split("\n", 2)[0]
  	println item.absoluteUrl + 'configure'
    println '-----------------'
  }
}
```

PR #1218, opened 2018-04-06
cron: reschedule nightlies

refack:
The "nightlies" were triggered at 1,2,3 __PM__ EST (10,11AM 12PM PST).
I suggest:
* moving them to 07:00UTC = 00:00PST = 03:00EST
* Spacing them 2 hour apart

### DIFF
--- a/setup/www/tasks/tools.yaml
+++ b/setup/www/tasks/tools.yaml
@@ -30,9 +30,10 @@
     dest: /etc/crontab
     line: "{{ item }}"
   with_items:
-    - '0 20    * * *   dist    /usr/bin/nodejs-nightly-builder --type v8-canary --ref heads/canary --config /etc/nightly-builder-v8-canary.json'
-    - '0 19    * * *   dist    /usr/bin/nodejs-nightly-builder --type nightly --ref heads/v9.x-staging --config /etc/nightly-builder.json'
-    - '0 18    * * *   dist    /usr/bin/nodejs-nightly-builder --type nightly --ref heads/master --config /etc/nightly-builder.json'
-    - '0 11    * * *   dist    /usr/bin/nodejs-nightly-builder --type nightly --ref heads/v9.x --config /etc/nightly-builder-chakracore.json'
-    - '0 10    * * *   dist    /usr/bin/nodejs-nightly-builder --type nightly --ref heads/master --config /etc/nightly-builder-chakracore.json'
+  # 07:00UTC is 00:00PST and 03:00EST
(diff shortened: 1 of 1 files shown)


rvagg:
So, I don't care much when they are triggered as long as they are less likely to get in the way of release builds. I guess these are likely to fit that criteria? "nightly" was never meant to be literal of course and it's impossible to be literal for everyone around the globe obviously.

refack:
> I guess these are likely to fit that criteria?

I'm assuming the new times would be better suited as to not to minimally interfere with the day's proceedings in the US. But I totally agree this is arbitrary anyway.

The idea to change came to me today when we were debating whether to force push something out, and it was getting close to 18:00UTC so suddenly the content nightlies became a concern.

Maybe ask @nodejs/release

joyeecheung:
I wrote a simple script to find out the frequency of commits in UTC hours (using the latest 10000 commits)

```
'use strict';
const child_process = require('child_process');
const log = child_process.spawnSync('git', ['log', '--format=%cI', '-n', '10000'], {
  cwd: process.cwd()
});
const dates = log.stdout.toString().trim();
const hours = dates.split('\n').map(d => new Date(d).getUTCHours());
const map = new Map();
for (const h of hours) {
  if (map.has(h)) {
    map.set(h, map.get(h) + 1);
  } else {
    map.set(h, 1);
  }
}
const pairs = [...map.entries()].sort((a, b) => a[1] > b[1] ? -1 : 1);
console.log(pairs);
```

Result:

```
[ [ 18, 748 ],
  [ 17, 726 ],
  [ 16, 681 ],
  [ 21, 637 ],
  [ 15, 632 ],
  [ 19, 580 ],
  [ 20, 565 ],
  [ 22, 530 ],
  [ 23, 454 ],
  [ 14, 454 ],
  [ 0, 375 ],
  [ 1, 361 ],
  [ 13, 338 ],
  [ 11, 329 ],
  [ 4, 311 ],
  [ 5, 308 ],
  [ 9, 299 ],
  [ 12, 273 ],
  [ 3, 270 ],
  [ 10, 239 ],
  [ 2, 233 ],
  [ 7, 221 ],
  [ 8, 219 ],
  [ 6, 217 ] ]
```

refack:
> So 6-8 UTC might be the best choice here.

Yay 🕺 
Just an FYI, we have currently have 5 nightly jobs:
Node/V8 10, 9, & Canary-V8 (each taking ~2 hours) and Node/Chakra 10 & 9 (each taking < 1 hour)
Updated PR to moves jobs as follows (at UTC):
06:00 - V8 10
08:00 - V8 9
10:00 - V8 Canary
04:00 - Charka 10
05:00 - Chakra 9
(all times are the below the median commit time)

PR #1219, opened 2018-04-08
nginx: Redirect old inspector docs (node executable)

fhemberger:
Turns out node was linking to nodejs.org/en/docs/inspector.
This fixes the redirect as well.

Ref: nodejs/nodejs.org#1619


refack:
From nodejs/nodejs.org#1619: 
The binary refers to the docs in
https://github.com/nodejs/node/blob/9a6dd07e8d93c7208c6698dfec1296f5caec9948/src/inspector_socket_server.cc#L107-L108
(when running `node --inspect`)

Issue #1034, opened 2017-12-07
Duplicate 'centos7' in the `init` dictionary


seishun:
https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/vars/main.yml#L8

`'centos7'` is present under both `centos` and `systemd`. [This line](https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/vars/main.yml#L61) ends up initializing `init_type` to `systemd` but that might be unintended.

cc @rvagg

Issue #885, opened 2017-09-16
Dropping 32-bit builds


gibfahn:
This came out of discussion of https://github.com/nodejs/build/pull/809 in the last WG meeting (https://github.com/nodejs/build/issues/875, minutes at https://github.com/nodejs/build/pull/881). Discussion paraphrased below.

Basically updating CentOS 6 to gcc 4.9.3 requires `devtoolset-6`, which means that we can no longer do 32-bit xLinux builds.

It's difficult to know how many people are currently using x86_64 32-bit Node. The [metrics](https://nodejs.org/metrics/) show it looking pretty stagnant, but it's difficult to know how many of the downloads are from CI.

@rvagg suggested that the best way to find out would be to drop 32-bit builds in an odd-numbered release and see how many people complain, that way we have the option of adding them back in a minor version. The question was whether 9.x is too soon to do this.

Rod also said that he thought CentOS6 and Ubuntu 14 no longer had a 32-bit version, meaning that anyone relying on a 32-bit OS had already been left behind by Node 8.x. If we could confirm that this is true for most Linux and Windows versions then that would make the decision much easier.

### Todo (if we decide to drop 32-bit)

- Update the jobs to skip 32-bit Intel for Node >8.x
- Update the website to handle the lack of 32-bit builds
- Update the [Supported Platforms](https://github.com/nodejs/node/blob/master/BUILDING.md#supported-platforms-1) list to remove x86.

gibfahn:
More info on the state of play at the OS level would definitely be useful (i.e. when/if major distros stopped supporting 32-bit OSs, i.e. who we'd be breaking).
- Linux: maybe @rvagg or @chrislea would know off the top of their heads
- Windows: @joaocgreis and @refack 
- SmartOS: @nodejs/platform-smartos would be good to know your thoughts, I assume we don't have to drop 32-bit on SmartOS, but if we're doing it elsewhere it might be a good time to do it here too.

refack:
Windows 10 still supports [32bit](https://www.microsoft.com/en-us/windows/windows-10-specifications#system-specifications) and "cross compiling" from a 64bit host to a 32bit target is also fully supported.
Windows has side-by-side technology (WinSxS) that allows 32bit applications to run natively on 64bit hosts, so there should not be an ***external limitation*** for building, testing and running node32 for the foreseeable future.

BTW: AWS/Azure/GCP doesn't rent 32bit VMs of our lowest supported Windows (Server 2008R2)

[...]

PR #1223, opened 2018-04-10
CentOS 6 Update + devtoolset-6 option

rvagg:
Trying again with a new version of my last CentOS 6 PR. This one includes most of those changes but takes a slightly different approach to a few things.

Most notably, it includes devtoolset-2 and devtoolset-6 and doesn't include them in the PATH anywhere.

Instead, what I've done is given the two workers "centos6-64-gcc48" and "centos6-64-gcc6", assigned both labels to node-test-commit-linux and included this in the [VersionSelectorScript](https://github.com/nodejs/build/pull/1217/commits/3bfe64a51f28bcc2cc6ed89b23929bc3ae3df5a2):

```groovy
  if (buildType != 'release' && builderLabel == "centos6-64-gcc48" && nodeMajorVersion >= 10)
    return false
  if (buildType != 'release' && builderLabel == "centos6-64-gcc6" && nodeMajorVersion < 10)
    return false
```

Unfortunately that's not enough so I've also edited Jenkins so it makes the PATH adjustments (note that this is for all Linuxes so it has to be generic too):

```sh
exec_cmd=" \
  NODE_TEST_DIR=${HOME}/node-tmp \
  PYTHON=python \
  FLAKY_TESTS=$FLAKY_TESTS_MODE \
  make run-ci $MAKE_ARGS \
"

if test $nodes = "centos6-64-gcc48"; then
  exec_cmd=". /opt/rh/devtoolset-2/enable; $exec_cmd"
elif test $nodes = "centos6-64-gcc6"; then
  exec_cmd=". /opt/rh/devtoolset-6/enable; $exec_cmd"
fi

sh -c "$exec_cmd"
```

(the previous version just executed the command directly)

And it's working nicely in ci now. Node `master` (10+) compiles with gcc6, everything else with gcc4.8.

ci-release hasn't been done yet, it probably needs exactly the same treatment.

@seishun


rvagg:
Oh yeah, and I also included a way to upgrade git on here, so they're running git 2.16 instead of 1.17 and I'm hoping this will clear up some more of those Jenkins git clone errors that appear sometimes. Debian 8 would be the remaining troublemaker so I'll have to figure out a git upgrade for that next.

rvagg:
so .... https://ci.nodejs.org/job/node-test-commit-linux/17808/nodes=centos7-64/ ... it looks like I need to do the same thing to CentOS7 eh @seishun?

```
$ ssh test-softlayer-centos7-x64-1 gcc -v
...
gcc version 4.8.5 20150623 (Red Hat 4.8.5-16) (GCC)
```

seishun:
A commit that requires gcc 4.9 successfully builds on CentOS 6 again: https://ci.nodejs.org/job/node-test-commit-linux/17808/nodes=centos6-64-gcc6/

>it looks like I need to do the same thing to CentOS7 eh @seishun?

Correct, CentOS 7 requires an upgrade too. I intended to work on that once https://github.com/nodejs/build/issues/1034 is resolved.

rvagg:
OK, I've implemented some of the same changes for CentOS 7

* manual SCLo repo setup for arm64
* setup via centos-release-scl.rpm for x64
* devtoolset-6 from SCLo
* IUS repo setup for x64 to get latest git
* no IUS for arm64 (aarch64) unfortunately so I don't yet have a way to get an updated git so it's stuck on 1.8

I've updated the VersionSelectorScript to choose between `/centos[67]-(arm)?64-gcc(48|6)`/ for <10 & >=10 and updated the hosts with those labels (duplicate for each of them) and included them in the jobs. Then in the Jenkins script for node-test-commit-linux I now have:

```sh
if test $nodes = "centos6-64-gcc48"; then
  exec_cmd=". /opt/rh/devtoolset-2/enable; $exec_cmd"
elif [[ "$nodes" =~ centos[67]-(arm)?64-gcc6 ]]; then
  exec_cmd=". /opt/rh/devtoolset-6/enable; $exec_cmd"
fi
```

i.e. centos6 uses devtoolset-2 to get gcc4.8, centos7 has it natively, but they both use devtoolset-6 to get gcc6

And node-test-commit-arm, which is already doing Debian selection for Node version, has been amended to have the second rule since it only has centos7:

```sh
  if [[ "$nodes" =~ centos[67]-(arm)?64-gcc6 ]]; then
    exec_cmd=". /opt/rh/devtoolset-6/enable; $exec_cmd"
  fi
```

Running master and v8.x on them both now and I have submitted a rerun of @seishun PR with gcc 4.8 incompatible c++ @ https://ci.nodejs.org/job/node-test-commit/17630/

rvagg:
Yeah, it did work, https://ci.nodejs.org/job/node-test-commit-linux/17849/nodes=centos7-64-gcc6/

After that run I realised I didn't remove the centos7-64 label when I added in both centos7-64-gcc48 and centos7-64-gcc6. So the job you're linking to was just a vanilla centos7, same configuration as centos7-64-gcc48 which shouldn't have run. It was removed after that run.

Here's your PR again https://ci.nodejs.org/job/node-test-commit-linux/17861/

rvagg:
Made this work against centos6 release machines, it took a bit of tweaking, however. Unfortunately there's no official devtoolset-6 from SCLo for x86 so I had to use someone's custom build in Copr https://copr.fedorainfracloud.org/coprs/mlampe/devtoolset-6.1/

Latest nightlies were rebuilt with this config, https://nodejs.org/download/nightly/v10.0.0-nightly2018041142d8976dff/ and I've requested testing @ https://github.com/nodejs/node/pull/19091#issuecomment-380425856 but I don't imagine there's going to be many people using testing & nightly builds with older Linuxes.

seishun:
>Made this work against centos6 release machines, it took a bit of tweaking, however. Unfortunately there's no official devtoolset-6 from SCLo for x86 so I had to use someone's custom build in Copr https://copr.fedorainfracloud.org/coprs/mlampe/devtoolset-6.1/

Didn't we decide to drop 32-bit Linux builds? https://github.com/nodejs/build/issues/885

Issue #1222, opened 2018-04-09
infra: spring software updates


refack:
I'm not sure how to coordinate this, but it IMHO we should do some systematic updates to the software on our infra. I'm referring to peripheral software such as Java, `slave.jar` and `git` (not OS or compilers).
Besides minimizing potential bit-rot, and making us feel better in general, I have an intuiting it is already casing failures, and blocking process improvments. For example:
1. https://github.com/nodejs/build/issues/173#issuecomment-379858648 Jenkins failing to communicate with workers &mdash; might be related to stale `slave.jar`:
on failing machines there was an old agent running
![image](https://user-images.githubusercontent.com/96947/38518535-2caeb34a-3c0b-11e8-990a-853761617e77.png)
after restart it bumps to:
![image](https://user-images.githubusercontent.com/96947/38518581-4df299ea-3c0b-11e8-91e5-f3da8497c4cb.png)
"Log" show this warning:
![image](https://user-images.githubusercontent.com/96947/38519217-1134781e-3c0d-11e8-9e90-7fb6278c464b.png)

2. Old git (I mean 1.8 when the latest is 2.17) doesn't handle sparse checkouts, which degrades the overall performance of the cluster:
![image](https://user-images.githubusercontent.com/96947/38518768-df8918ca-3c0b-11e8-808a-582d4d50419b.png)

3. My estimation is that we also have on some platforms outdated `sshd` with potential security issues (also we should disable plain-text password login where possible RE: https://github.com/nodejs/build/issues/866)

---

Since I now have time for such tasks, I'm seeking feedback / pitfalls / warnings. And also ideas on how to coordinate such efforts (RE @gibfahn and the Java8 project).

refack:
A computer similar to my above example just popped (`test-digitalocean-ubuntu1604-x86-1` went into a remoting perma-fail):
![image](https://user-images.githubusercontent.com/96947/38581389-bd787eb8-3cda-11e8-855e-e4f921e8e277.png)
Solved by updating `slave.jar`

gibfahn:
>I'm not sure how to coordinate this, but it IMHO we should do some systematic updates to the software on our infra.

Sounds like a great idea to me.

>Besides minimizing potential bit-rot, and making us feel better in general, I have an intuiting it is already casing failures, and blocking process improvments. For example:
>Since I now have time for such tasks, I'm seeking feedback / pitfalls / warnings. And also ideas on how to coordinate such efforts (RE @gibfahn and the Java8 project).

[...]

PR #1224, opened 2018-04-10
updated git for debian8, ubuntu 14.04 and ubuntu 16.06

rvagg:
they now all have 2.17, including the jenkins-workspace machines. I'm hoping this will clear up some of the git errors we've been seeing in Jenkins.


rvagg:
Lots of infra problems with debian8 still, in the style of #1222, so this hasn't resolved everything. Good to have though.

PR #1228, opened 2018-04-12
ansible: smartos17

rvagg:
I've provisioned 2 test and 1 release smartos 17 machines and this is the required changes for ansible to get it set up. The only iffy bit that I bet there's a better answer for is the gcc7 links into /opt/local/bin, it's strange that there aren't links by default and there must be some kind of equivalent to debian style "alternatives". It works at least.

I've got rules in place to make Node >=10 binaries build on smartos 17, 64-bit only from 10 onward like for linux.

[VersionSelectorScript.groovy](https://github.com/nodejs/build/blob/rvagg/VersionSelectorScript.groovy/jenkins/scripts/VersionSelectorScript.groovy) now has the following:

```groovy
  if (builderLabel.indexOf('smartos14') == 0 && nodeMajorVersion >= 8)
    return false
  if (builderLabel.indexOf('smartos15') == 0 && nodeMajorVersion < 8)
    return false
  if (buildType == 'release' && builderLabel.indexOf('smartos15') == 0 && nodeMajorVersion >= 10)
    return false
  if (builderLabel.indexOf('smartos16') == 0 && nodeMajorVersion < 8)
    return false
  if (builderLabel.indexOf('smartos17') == 0 && nodeMajorVersion < 10)
    return false
```

Roughly translated as:
* releases:
  - Node 4 & 6: smartos14
  - Node 8: smartos15
  - Node 10+: smartos17
* testing:
  - Node 4 & 6: smartos14
  - Node 8: smartos15, smartos16
  - Node 10+: smartos15, smartos16, smartos17

Builds from today onward (nightlies and the Node 10 test builds most notably) are using this configuration. Also including the absence of sunos-x86 binaries.

@geek 

/ref https://github.com/nodejs/build/issues/885
/ref https://github.com/nodejs/build/pull/1149

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -36,6 +36,7 @@ hosts:
         smartos13-x64-1: {ip: 72.2.114.225}
         smartos14-x64-1: {ip: 72.2.113.193}
         smartos15-x64-1: {ip: 165.225.131.14}
+        smartos17-x64-1: {ip: 72.2.114.225}
 
     - requireio:
         rvagg-debian9-armv6l_pi1p-1: {ip: 192.168.2.40, user: pi}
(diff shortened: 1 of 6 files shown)


rvagg:
https://nodejs.org/download/nightly/v10.0.0-nightly201804120aab8ff602/ latest nightly has the new smartos binaries in it, and x64 only

Issue #1229, opened 2018-04-12
Update binutils for OpenSSL asm support


shigeki:
In the discussion of https://github.com/nodejs/node/issues/19944, we have to update binutils version in CI according to the requirements of OpenSSL-1.1.1 asm support if you make opt-in for `openssl_no_asm' option.

The new requirements are
```
### OpenSSL asm support

OpenSSL requires the following asssembler version for use of asm
support.

* gas (GNU assembler) version 2.26 or higher
* xcode version 5.0 or higher
* llvm version 3.3 or higher
* nasm version 2.11.8 or higher in Windows

Otherwise, build with `--openssl-no-asm` in configure.

*Note:* This requirements come from the forthcoming OpenSSL-1.1.1. See
 https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_ia32cap.html for
 details.
```
At the current ,the following CI would be failed due to the older assembler version. Can we make update them?
```
debian8-64: 'gas_version': '2.25',
debian8-x86: 'gas_version': '2.25',
ubuntu1404-32: 'gas_version': '2.24',
ubuntu1404-64: 'gas_version': '2.24',
ppcle-ubuntu1404: 'gas_version': '2.24',
smartos15-64: 'gas_version': '2.25',
rhel72-s390x: 'gas_version': '2.25',
cc-armv6:'gas_version': '2.25',
cc-armv7:'gas_version': '2.25',

aix61-ppc64: defined no_asm? (Not sure it has assembler)
````

rvagg:
OK .. this is going to be a problem .. the cc-armv6 and cc-armv7 machines are doing cross compiling and 2.25 is the highest we have access to for that, at least from the raspbian tools distribution. We may need to go sniffing out a new toolchain for this. I don't know enough about this to know whether we can just drop in a new custom version of `as` or whether we have to get the whole toolchain. Linaro is probably the place to go for a new one I suppose. /cc @joaocgreis 

debian8 and ubuntu1404 might be tricky. I'll have to go hunting down a solution for that.

Unfortunately we're also using debian8 currently to build armv6 and armv7 binaries as well, they are not cross-compiled over there so that's another complication.

smartos15-64 we can retire.

ppcle-ubuntu1404 and rhel72-s390x are in @mhdawson and @gibfahn's hands.

rvagg:
[...]

PR #1235, opened 2018-04-14
install binutils-2.26 on ubuntu

jBarz:
Certain versions of node require binutils version 2.26 (https://github.com/nodejs/build/issues/1229)
This won't affect the default installation of binutils.

To use it, add the following to PATH.
/usr/lib/binutils-2.26/bin/

fyi, @mhdawson

### DIFF
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -100,6 +100,6 @@ packages: {
   ],
 
   ubuntu1404: [
-    'ntp,g++-4.8,gcc-4.8,g++-4.9,gcc-4.9',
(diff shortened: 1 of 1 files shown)


rvagg:
oh, nice! didn't know this was an option. I'll get this sorted. Thanks @jBarz.

mhdawson:
Rolling it out now.  Added in copy of plinux job and tested on le-3.  Looked ok.

Going to  run ansible on the other 2 test machines and add updated selection script to job as next step.

rvagg:
I've run this against the 4 intel ubuntu1404 machines we have in CI. I think @mhdawson when you do the selection script it'll probably be a good idea to switch PATH whenever ubuntu1404 is encountered regardless of arch and then we'll be covered.

PR #1251, opened 2018-04-25
jenkins: remove use of rpath

mhdawson:
rpath will not be useful for customer deployments
use LD_LIBARY_PATH instead on platforms where it is needed

Should fix issue mentioned in https://github.com/nodejs/node-v8/pull/54


mhdawson:
NOTE this is being used live in test now, release later today.

mhdawson:
It is needed on LinuxOne as gcc was compiled and is in a location where it won't naturally be found.  Further confirmed because we had a build failure when @bnoordhuis was brining in some new functionality in node-v8, (torque support) and adding the LD_LIBRARY_PATH resolved it.

There may be a better way to build/install where it might not be required, but what we are currently doing does.  @jBarz can you read up on "prefix directories" as @rvagg has mentioned.

mhdawson:
landed as 497c627a12d53436e7a6e0be92cb04e0ff4541ae

PR #1265, opened 2018-05-10
jenkins: refactor to allow external node version

mhdawson:
Allow NODE_MAJOR_VERSION to be set externally

### DIFF
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -9,18 +9,18 @@ if [ "$DONTSELECT_COMPILER" != "DONT" ]; then
 fi
 
 # get node version
-if [ -z ${NODE_MAJOR_VERSION+x} ]; then
+if [ -z ${NODEJS_MAJOR_VERSION+x} ]; then
   NODE_VERSION="$(python tools/getnodeversion.py)"
-  NODE_MAJOR_VERSION="$(echo "$NODE_VERSION" | cut -d . -f 1)"
+  NODEJS_MAJOR_VERSION="$(echo "$NODE_VERSION" | cut -d . -f 1)"
 fi
 
 if [ "$SELECT_ARCH" = "PPC64LE" ]; then
   # Set default
   export COMPILER_LEVEL="4.8"
 
-  echo "Setting compiler for Node version $NODE_MAJOR_VERSION on ppc64le"
+  echo "Setting compiler for Node version $NODEJS_MAJOR_VERSION on ppc64le"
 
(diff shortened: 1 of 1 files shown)


jenkins/scripts/select-compiler.sh
```diff
@@ -8,13 +8,16 @@ if [ "$DONTSELECT_COMPILER" != "DONT" ]; then
   esac
 fi
 
+# get node version
+if [ "${NODE_MAJOR_VERSION}X" = "X" ]; then
```
refack: Since we know this is going to run in bash, [SO has a very high upvote](https://stackoverflow.com/a/13864829/27955) for using
```bash
if [ -z ${NODE_MAJOR_VERSION+x} ]; then
```

mhdawson:
Seeing failures in test release job with this branch, going to rebase to see if its missing changes.

jenkins/scripts/select-compiler.sh
```diff
@@ -8,13 +8,16 @@ if [ "$DONTSELECT_COMPILER" != "DONT" ]; then
   esac
 fi
 
+# get node version
+if [ -z ${NODE_MAJOR_VERSION+x} ]; then
+  NODE_VERSION="$(python tools/getnodeversion.py)"
+  NODE_MAJOR_VERSION="$(echo "$NODE_VERSION" | cut -d . -f 1)"
```
richardlau: During my work on updating the node-report CI job (https://github.com/nodejs/build/issues/1421#issuecomment-411469565) I had to add a parameter called `NODEJS_MAJOR_VERSION` to use the version selector script. This ends up as an environment variable of a same name (`NODEJS_MAJOR_VERSION`). 

https://github.com/nodejs/build/blob/d121160b7d38cecd98c794a2a9549b89708a4754/jenkins/scripts/VersionSelectorScript.groovy#L87-L88

I can make the job set `NODE_MAJOR_VERSION=NODEJS_MAJOR_VERSION`, but it would be one less line to maintain if I didn't have to (i.e. if the version selector and select compiler script used the same parameter/environment variable name).

mhdawson:
Will look and try to land this On Tuesday oct 2.

refack:
> Will look and try to land this On Tuesday oct 2.

If you could check if it makes sense to use `NODEJS_MAJOR_VERSION` (with the `JS`), since it's rather widely used.

mhdawson:
@refack will do.

mhdawson:
Ok updated to use NODEJS_MAJOR_VERSION and validated in N-API testing and release job testing that it looks ok.  Going to land.

PR #1266, opened 2018-05-10
jenkins: add missing path to LIBPATH for AIX

mhdawson:
We were missing this path which is required with certain compiler options.


mhdawson:
Landing so that we can update across the board.

mhdawson:
Landed as 9334bd05946121e1cc4bf1dd73564c27a897fde7

mhdawson:
Closing

PR #1275, opened 2018-05-17
 Add CORS headers to downloads enabling WebTorrent (closes #678)

### LINKED ISSUE #678: Add CORS headers to downloads enabling WebTorrent
Howdy! Inspired by the recent downtime, I've been working on a way to make a resilient Node downloader using [WebTorrent](https://github.com/feross/webtorrent). Even if the primary download source becomes temporarily unavailable, the files would continue to be available on the bittorrent network. For my proof-of-concept [Download with WebTorrent Button](https://github.com/wmhilton/download-with-webtorrent-button/) to work with nodejs.org, I think all that is needed is to add Cross-Origin Resource Sharing (CORS) headers to the server config so it is possible to download files in `https://nodejs.org/dist/` via AJAX. I know that sounds weird, but how the torrent would work is if there are no other users seeding a torrent it would request the file from the source (via AJAX) to bootstrap the process. This also enables new torrents to be created automatically for files that don't have torrents the first time the button is clicked.

Obviously this isn't a panacea, because a lot of downloads are from 'nvm', WebTorrent only works on modern browsers that support WebRTC, and there are probably a host of other things to consider. But I love WebTorrent and love resilient distributed networking, and I thought adding CORS headers would be preferable to using a proxy service like [cors-anywhere](https://cors-anywhere.herokuapp.com/). You can see the demo of the button [here](https://wmhilton.com/download-with-webtorrent-button)

With that I'll open the floor to questions!

wmhilton:
See the discussion in #678. Basically, we want to be able to download node using AJAX. 😁

### DIFF
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -55,9 +55,24 @@ server {
         autoindex on;
         default_type text/plain;
 
-        location ~ \.json$ {
-            add_header access-control-allow-origin *;
+        # Make it possible for javascript on any website to download the installer files and use the data.
+        # (As of Nginx 1.7.5, add_header supports an "always" parameter which
+        # allows CORS to work if the backend returns 4xx or 5xx status code.)
+        add_header 'Access-Control-Allow-Origin' '*' always;
+        add_header 'Access-Control-Allow-Methods' 'GET, HEAD, OPTIONS' always;
+        add_header 'Access-Control-Allow-Headers' 'Range,Content-Type' always;
+        if ($request_method = 'OPTIONS') {
+            # Tell client that this pre-flight info is valid for 10 minutes
+            add_header 'Access-Control-Max-Age' 600;
+            add_header 'Content-Type' 'text/plain charset=UTF-8';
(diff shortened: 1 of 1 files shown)


setup/www/resources/config/nodejs.org
```diff
@@ -55,9 +55,24 @@ server {
...
+            add_header 'Access-Control-Max-Age' 600;
+            add_header 'Content-Type' 'text/plain charset=UTF-8';
+            add_header 'Content-Length' 0;
+            return 204;
         }
+
+        # JSON files are now covered by the general CORS settings directly above
+        #location ~ \.json$ {
+        #    add_header access-control-allow-origin *;
+        #}
```
feross: Can you delete these commented-out lines?

feross:
cc'ing relevant reviewers: @jbergstroem @maclover7 @rvagg

setup/www/resources/config/nodejs.org
wmhilton: done

sam-github:
Seems like it might be useful, but very stale. Closing unless someone wants to restart this.

PR #1295, opened 2018-05-30
ansible: rpi updates, completed conversions to new format

rvagg:
* migrated the remaining pi's to the new dockerized setup
* mounted /boot/ on all of the nfs-root pi's so any updates get pushed back into the tftp nfs server
* minor fixes

interesting thing happened with this though, one of the Pi's, the last of the Pi 1 B+'s (number 14 according to how I number them, and one of the most recently purchased ones so it's not ancient) test-requireio_davglass-debian9-armv6l_pi1p-1 won't stay running while nfs booted. I've updated firmware, tried using a "next" branch bootloader, all sorts of things but I can't stop it from doing a kernel panic fairly soon after starting up. So I've left it with a classic SD card setup, but it's the only one out of all of them. Pretty strange. Good chance to get these scripts working for both nfs-booted and SD card booted Pi's.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -195,8 +195,8 @@ hosts:
         mhdawson-debian9-armv6l_pi1p-1:         {ip: 192.168.2.49, user: pi}
         securogroup-debian9-armv6l_pi1p-1:      {ip: 192.168.2.50, user: pi}
         securogroup-debian9-armv6l_pi1p-2:      {ip: 192.168.2.51, user: pi}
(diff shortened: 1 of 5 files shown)


maclover7:
@rvagg Excellent work! Does this mean setup/raspberry-pi can be removed now?

rvagg:
yes

maclover7:
ping @rvagg

maclover7:
Landed in 8a70968

PR #1296, opened 2018-05-31
jenkins: add node-linter

maclover7:
Works as a pipeline-ified version node-test-linter. When this lands, will
convert node-test-commit to call this pipeline, instead of node-test-linter.

### DIFF
--- a/jenkins/pipelines/node-linter.jenkinsfile
+++ b/jenkins/pipelines/node-linter.jenkinsfile
@@ -1,0 +1,94 @@
+#!/usr/bin/env groovy
+
+pipeline {
+    agent { label 'linter' }
+    parameters{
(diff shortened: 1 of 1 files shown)


jenkins/pipelines/node-linter.jenkinsfile
```diff
@@ -0,0 +1,94 @@
...
+            steps {
+                checkMake()
+                checkSed()
+                sh """
+                  # this job does not build node, so we symlink the system's node
+                  which node #&& ln -s ${sh(script: "which node", returnStdout: true).trim()}
+                  node --version
+
+                  ${env.MAKE} lint-md-build || true
+                  # If lint-ci fails, print all the interesting lines to the cons::e.
```
richardlau: Should this be `console`?

maclover7: Yes -- this was just a raw copy-paste :)

Issue #1290, opened 2018-05-29
platforms: stop testing on debian8-x86


refack:
RE: https://github.com/nodejs/build/issues/885

Although a usual `node-test-commit` run take less than 15 minutes, we only have 1 worker in this label https://ci.nodejs.org/label/debian8-x86/ so CITGM runs cause a bottleneck.

PR #1298, opened 2018-05-31
jenkins: no debian8-x86 for node 10

refack:
Fixes: https://github.com/nodejs/build/issues/1290


joaocgreis:
I don't see a `debian8-x86` machine in ci-release, so why is this needed?

Perhaps cc @rvagg

rvagg:
yeah, this isn't going to have an impact as there isn't a machine of this type on ci-release (at least there shouldn't be!). Was this intended for `anyType` so it runs on ci?

refack:
> Was this intended for anyType so it runs on ci?

Probably.

refack:
@nodejs/build PTAL.
Since this is in alignment with previous decisions, if there are no objections I'm planning on landing this.

refack:
https://ci.nodejs.org/job/node-test-commit-linux/20645/console:
```
...
16:29:38 Skipping centos5-64 for Node.js 10
16:29:38 Skipping debian8-x86 for Node.js 10
16:29:38 Skipping ubuntu1204-64 for Node.js 10
...
```

Issue #1277, opened 2018-05-21
State of Ansible


maclover7:
As noted in https://github.com/nodejs/build/issues/1272#issuecomment-390502218, would be ideal to get a current sense of a) the transition from `setup` to `ansible`, b) usability of the new scripts and how people feel about running them as well as making changes

maclover7:
Here's what I've been able to gather and summarize from current/past happenings:

**No work:**
`aix61`
`armv7-wheezy`
`centos5`
`centos6`
`debian-wheezy-gcc`
`osx`
`raspberry-pi`
`windows`
`www`

**PR Open:**
`github-bot`
`ubuntu14.04`

**Completed:**
`centos7`
`containers`
`cross-compiler`
`debian8`
`fedora20`
`fedora21`
`fedora22`
`freebsd`
`linter`
`rhel72-linuxonecc`
`smartos`
`ubuntu12.04`
`ubuntu1604`

gdams:
@maclover7 macOS playbooks were created in https://github.com/nodejs/build/pull/971

maclover7:
@gdams Yep, but looks like `setup/osx` wasn't removed -- is that safe to do now?

gdams:
I think so, I've pulled all of the useful stuff out of that playbook already

maclover7:
@gdams Ok -- can you run the new playbook on all OSX machines, confirm everything looks okay, and then remove `setup/osx` if everything works good?

rvagg:
* centos5: I think we can retire it all, Node 4's demise takes out our remaining need for it and it was getting crazy complicated keeping it alive (see non-ansible-configuration-notes.md details on git in centos5 for example)
* debian-wheezy-gcc: can disappear, not used and not useful anymore
* www: too big a job to migrate and I'm not convinced there's utility in migrating it since it's really just a reflection of reality and preparation for emergencies. If it needs to be relocated to clean out "setup" then that's fine by me.
* armv7-wheezy: _"new scripts do not work due to awk error here; can old scripts be removed?"_ what's this referring to @maclover7? it can be removed anyway, the armv7 machines this references are running Ubuntu 16.04 and doing everything in Docker now.
* osx: we have scripts? it's just docs and a few helpers isn't it? they're redundant now with the new osx work.
* windows: I'm pretty sure is out of date but I don't believe we have anything better so it's important for documentation purposes at a minimum

[...]

PR #1300, opened 2018-05-31
ansible: fix and simply ccache version getter

maclover7:
@rvagg This was what I meant in #1277 -- for some reason the awk script doesn't work on my Mac, and this will help to simplify what we're actually trying to do here

### DIFF
--- a/ansible/roles/jenkins-worker/tasks/partials/docker-host.yml
+++ b/ansible/roles/jenkins-worker/tasks/partials/docker-host.yml
@@ -9,13 +9,19 @@
   register: server_user_uid
 
 - name: docker | fetch latest ccache release version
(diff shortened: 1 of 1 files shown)


LucaLanziani:
One issue I can see with this, if I'm reading the code correctly, is that we are assuming that the moment the tag is in GitHub the tar.gz is already in the "samba ftp" endpoint.

refack:
How about splitting this to:
1. A script that fetches whatever version and stores it in a known location (e.g. www/downloads)
2. Ansible can always fetch from known location.

LucaLanziani:
@refack would the script fetch the version on every run?

At the end I've decided to push my version too #1312 ;)

rvagg:
oh, whoops, this must work for me because I have `coreutils` installed on my mac and they are the default executables so it's consistent between my mac and linux machines. +1 on whatever works.

maclover7:
Closing in favor of #1312

PR #1306, opened 2018-06-01
doc: add services.md

maclover7:
A thorough explanation of everything the Build WG is responsible for,
and where configuration lives.

Up next will be a soup-to-nuts guide on how to operate our Jenkins
cluster.

cc @Trott @nodejs/build

### DIFF
--- a/doc/services.md
+++ b/doc/services.md
@@ -1,0 +1,37 @@
+# Build WG Services
+
+This document is intended to outline all services managed by the Build WG,
+and how to configure and operate them.
+
+### Ansible configuration
+
+Ansible is the tool used to managed all machines that the Build WG is responsible for.
+
+Secrets for Ansible scripts live within the `build` directory of the [`nodejs-private/secrets`](https://github.com/nodejs-private/secrets/tree/master/build) repository. Access to credentials is granted upon joining the build team
+
+### `ci.nodejs.org`
+
+`ci.nodejs.org` is the main Jenkins setup used to test projects within
+the Node.js Foundation, the largest being Node itself.
+
+There are several different types of machines that form the test CI
+cluster:
+
+| Type  | Jenkins Agent | Jenkins Workspace | Playbook | Notes |
+|---|---|---|---|---|
+| "Normal"  | On machine | On machine | [`jenkins/worker/create.yml`](https://github.com/nodejs/build/blob/master/ansible/playbooks/jenkins/worker/create.yml) | Run-of-the-mill, most common type of worker |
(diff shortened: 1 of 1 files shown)


doc/services.md
```diff
@@ -0,0 +1,42 @@
...
+There are several different types of machines that form the test CI
+cluster:
+
+| Type  | Jenkins Agent | Jenkins Workspace | Playbook | Notes |
+|---|---|---|---|---|
+| "Normal"  | On machine | On machine | [`jenkins/worker/create.yml`](https://github.com/nodejs/build/blob/master/ansible/playbooks/jenkins/worker/create.yml) | Run-of-the-mill, most common type of worker |
+| "Linter"  | On machine | On machine | [`jenkins/linter.yml`](https://github.com/nodejs/build/blob/master/ansible/playbooks/jenkins/linter.yml) | Dedicated machines for `node-test-linter` |
+| "Half Docker"  | On machine | Docker container | [`jenkins/worker/create.yml`](https://github.com/nodejs/build/blob/master/ansible/playbooks/jenkins/worker/create.yml) |  Raspbery Pi, Scaleway ARM v7 |
+| "Full Docker"  |  Docker container | Docker container  | [`jenkins/docker-host.yaml`](https://github.com/nodejs/build/blob/master/ansible/playbooks/jenkins/docker-host.yaml) | Special case Linux machines |
+
```
Trott: The first and third and entries in the table have the same playbook. Is that intentional?

Trott:
I wonder if this should go directly in the README? (Non-blocking suggestion, just sayin'.)

doc/services.md
maclover7: Yep, they use the same playbook. Did separate entries in the table because they're different "types" of machine

doc/services.md
```diff
@@ -0,0 +1,42 @@
+# Build WG Services
+
+This document is intended to outline all services managed by the Build WG,
+and how to configure and operate them.
+
+### Ansible configuration
+
+Ansible is the tool used to managed all machines that the Build WG is responsible for.
+
+Secrets for Ansible scripts live within the `build` directory of the [`nodejs-private/secrets`](https://github.com/nodejs-private/secrets/tree/master/build) repository. Access to credentials is granted upon joining the build team
```
joyeecheung: Don't people need to open a PR adding their public key there to actually get the access?

maclover7: New members have to send their keys to people that already have access who then run [`dotgpg add`](https://github.com/ConradIrwin/dotgpg#dotgpg-add) to add them

PR #1312, opened 2018-06-03
Simplify the ccache/git latest version fetch

LucaLanziani:
Use an ansible filter to fetch the latest version

Still to verify if the `grep -Eo` works as expected on OSX

### DIFF
--- a/ansible/plugins/filter/filters.py
+++ b/ansible/plugins/filter/filters.py
@@ -21,8 +21,10 @@
 # IN THE SOFTWARE.
 #
 
-from ansible.errors import AnsibleFilterError
 import re
+
+from distutils.version import StrictVersion
+from ansible.errors import AnsibleFilterError
 
 
 def match_key(value, dictionary, raise_error=True, feedback_name='os'):
@@ -42,6 +44,11 @@ def match_key(value, dictionary, raise_error=True, feedback_name='os'):
     return False
 
 
+def latest_version(versions):
+    versions.sort(key=StrictVersion)
+    return versions[-1]
+
+
 def starts_with(value, query):
     return value.startswith(query)
 
@@ -59,5 +66,6 @@ class FilterModule(object):
         return {
             'match_key': match_key,
             'startswith': starts_with,
-            'stripversion': stripversion
+            'stripversion': stripversion,
+            'latest_version': latest_version
         }
--- a/ansible/plugins/filter/filters.py
+++ b/ansible/plugins/filter/filters.py
@@ -23,7 +23,6 @@
 
 import re
 
-from distutils.version import StrictVersion
 from ansible.errors import AnsibleFilterError
 
 
@@ -45,7 +44,7 @@ def match_key(value, dictionary, raise_error=True, feedback_name='os'):
 
 
 def latest_version(versions):
-    versions.sort(key=StrictVersion)
+    versions.sort(key=lambda s: map(int, s.split('.')))
     return versions[-1]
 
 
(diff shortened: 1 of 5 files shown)


LucaLanziani:
If we think it's ok to merge this I'll squash the commits

refack:
> If we think it's ok to merge this I'll squash the commits

Need to make sure `grep -Eo` is available on all platforms 🤕
We have several non GNU (freeBSD, smartOS, AIX) ones.

I'd run `ansible test* -a "cat x | grep -Eo"` but I'm recovering from a complete SSD wipe, and still don't have my WSL (Ubuntu on Windows) up and running.

BTW @LucaLanziani Do you have experience with setting up an Ansible Tower (or [AWX](https://github.com/ansible/awx) or [Semaphore](https://github.com/ansible-semaphore/semaphore)). It might be useful for running playbooks for maintenance and upgrades. ATM we use sometimes use Jenkins jobs for that.

LucaLanziani:
> Need to make sure grep -Eo is available on all platforms face_with_head_bandage
> We have several non GNU (freeBSD, smartOS, AIX) ones.

Yeah, this is exactly my fear.

Unfortunately @refack I have no experience with Ansible Tower or alternatives, never had to use it, I use jenkins too in my projects. 

Btw I saw that in the previous implementation the code was just matching semver version (2 or 3 digits) ignoring the versions with 4 digits.
As for the comment in the second commit I can change the regex to have the same behavior and then we can go back using the `StrictVersion` function in python.

LucaLanziani:
@refack the last commit should solve the `is grep -Eo available on all platforms` problem ;)

ansible/plugins/filter/filters.py
```diff
@@ -21,9 +21,10 @@
 # IN THE SOFTWARE.
 #
 
-from ansible.errors import AnsibleFilterError
 import re
 
+from ansible.errors import AnsibleFilterError
```
maclover7: Does this line need to be moved?

refack: IMHO it does not "need" to be moved, but this is good [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports) style, so I'm +1.

LucaLanziani: It for sure doesn't "need" to be moved but as @refack said it's just about following PEP 8 style guide lines.

ansible/roles/baselayout/tasks/ccache.yml
```diff
@@ -4,13 +4,13 @@
...
-  register: ccache_latest_result
+- name: "ccache : fetch samba ccache HTML page"
+  local_action: shell curl -sL https://www.samba.org/ftp/ccache/
+  register: ccache_html_content
 
-- name: "ccache : extract version string"
+- name: "ccache : extract ccache latest version"
   set_fact:
-    ccache_latest: "{{ ccache_latest_result.stdout }}"
+    ccache_latest: "{{ ccache_html_content.stdout | regex_findall('ccache-[0-9]+.[0-9]+(?:.[0-9]+)*.tar.gz') | map('regex_replace', 'ccache-') | map('regex_replace', '.tar.gz') | list | latest_version }}"
```
LucaLanziani: Some comment went lost between rebases, as part of this I've changed the regex moving from something matching:

`x.y` and `x.y.z`

to something matching versions with 2 or more digits

`x.y`, `x.y.z`, `x.y.z.t` etc

This was mainly done because I've found files like `git-1.8.3.2.tar.gz` in the git web page (https://www.kernel.org/pub/software/scm/git/).

Do you have any opinion on this?
Should we stick to match only semver versions or should we allow for 4+ digits versions?

LucaLanziani:
@maclover7 sure, `setup/*` reverted, left a comment myself about the new regex.

maclover7:
Landed in bb7ac13

PR #1315, opened 2018-06-05
doc: move "join wg" instructions to access.md

maclover7:
Easier to keep in one place, than two separate documents.


richardlau:
> Easier to keep in one place, than two separate documents.

Agreed, but why does this WG not have a GOVERNANCE.md, which is where this information would normally be expected: https://github.com/nodejs/TSC/blob/master/WORKING_GROUPS.md#joining-a-wg

maclover7:
> Agreed, but why does this WG not have a GOVERNANCE.md, which is where this information would normally be expected

@richardlau That's a great question that I honestly don't know the answer to. The Build WG is one of the oldest WGs, and it's possible that no one noticed that there wasn't a GOVERNANCE.md. Would you mind opening up a separate issue as a reminder to add that document?

mhdawson:
I think @richardlau's comment was that instead of moving to access.md we should move to GOVERNANCE.md.  I was thinking moving that moving the info from joining_the_build_wg.md into access.md would make it less obvious where to find that info.

Overall I think it would be better to have the info in the README.md and GOVERNANCE.md files.

Trott:
I think the Build WG pre-dates those guidelines about GOVERNANCE.md. That's an explanation as to why there isn't one, but definitely not a reason to avoid creating one now!

maclover7:
Going to get this in for now, since GOVERNANCE.md seems like a larger project about how the WG functions. Logged issue about creating that document in #1319

Issue #1274, opened 2018-05-16
Install Linux perf on some Linux machines


mmarchini:
Linux perf has been broken on V8 since the Turbofan/Ignition pipeline became the default compiler. Recently on V8 6.7, we got it back to work (through a flag). Since there are some Node.js tools and some huge Node.js deployments relying on Linux perf (and other external profilers), having tests will help to keep these tools more stable.

For those tests to work, we need Linux perf available on at least some Linux machines. We could start on Ubuntu 16.04 machines and if we want we can install it on other machines later. The package can be installed with: `apt install linux-tools-generic`. Is it feasible?

Ref: https://github.com/nodejs/node/pull/20783

mmarchini:
The first one (if the PR lands)

rvagg:
Are we talking about the broken `--perf-basic-prof` stuff here? What would tests look like for this? Is it a simple matter of passing the output through perf and asserting on the results?

I don't _think_ we're going to be able to jam this into the containers unfortunately, even though that's a great place for these kinds of tests. `perf` is tied to the kernel version and that's detached from the OS once you're using containers since you use the host kernel version. Unless we have a match of host OS and container OS. Right now we do have a match, 16.04 on host and in container, but they are not updated at the same time so they get out of sync a bit (I'm not sure minor kernel versions matter much for perf, however). But I'm not sure we want to be restricted to always running the same host OS and container OS for these "sharedlibs" containers.

A better approach might just be installing perf on one of our Ubuntu LTS hosts, maybe 18.04, and let the tests detect its presence and run if its there and skip if it's not. We can lock this in to our ansible scripts so we always get linux-tools-generic installed on some of those hosts.

bnoordhuis:
> perf is tied to the kernel version and that's detached from the OS once you're using containers since you use the host kernel version.

That's the theory but in practice you can just mix them.  For example: I ran stock perf 3.13 on Ubuntu 14.04 for a while against the latest, continually upgraded mainline kernel and it worked just fine.

(Yes, I run mainline kernels.  I even test -rc's.)

joyeecheung:
I usually just clone the Linux master and cd into the perf source to compile it when I need perf on my servers. Does not seem to cause problems when I upgrade kernels from time to time.

joyeecheung:
Can we use the benchmark CI machine for this? It would be nice to have perf there, I am thinking about using `perf stat` in a core benchmark to measure the start up time for our lazy-loading effort.

[...]

PR #1321, opened 2018-06-06
ansible: add role to install Linux perf on Ubuntu

mmarchini:
Ansible role to install Linux perf on Ubuntu by cloning the Linux source
code and building tools/perf to avoid Kernel mismatch errors.

I wasn't sure which playbook should call this role, so I'm submitting only 
the role for now.

Tested on Ubuntu 16.04 from DigitalOcean.

Ref: https://github.com/nodejs/build/issues/1274

### DIFF
--- a/ansible/roles/linux-perf/tasks/main.yml
+++ b/ansible/roles/linux-perf/tasks/main.yml
@@ -24,20 +24,13 @@
         depth: 1
         dest: /data/tmp/linux
 
-    - name: Install build-essential
+    - name: Install packages
       package:
-        name: build-essential
+        name: "{{ package }}"
         state: present
-
-    - name: Install flex
-      package:
-        name: flex
-        state: present
-
-    - name: Install bison
-      package:
-        name: bison
-        state: present
+      loop_control:
+        loop_var: package
+      with_items: "{{ packages }}"
 
     - name: build Linux perf
       shell: make -j{{ ansible_processor_cores | default(1) }}
--- a/ansible/roles/linux-perf/tasks/main.yml
+++ b/ansible/roles/linux-perf/tasks/main.yml
@@ -5,7 +5,7 @@
 #
 
 - stat:
-    path=/usr/local/bin/perf
+    path: /usr/local/bin/perf
   register: linux_perf
   tags: linux_perf
 
@@ -16,13 +16,13 @@
     - name: Clean path
       file:
         state: absent
-        path: /data/tmp/
+        path: "{{ tmp_folder }}"
 
     - name: Download Linux source code
       git:
         repo: https://github.com/torvalds/linux.git
         depth: 1
-        dest: /data/tmp/linux
+        dest: "{{ tmp_folder }}/linux"
 
     - name: Install packages
       package:
@@ -35,12 +35,12 @@
     - name: build Linux perf
       shell: make -j{{ ansible_processor_cores | default(1) }}
       args:
-        chdir: /data/tmp/linux/tools/perf
+        chdir: "{{ tmp_folder }}/linux/tools/perf"
 
     - name: install Linux perf
-      shell: cp /data/tmp/linux/tools/perf/perf /usr/local/bin/perf
+      shell: "cp {{ tmp_folder }}/linux/tools/perf/perf /usr/local/bin/perf"
 
     - name: Clean path
       file:
         state: absent
-        path: /data/tmp/
+        path: "{{ tmp_folder }}"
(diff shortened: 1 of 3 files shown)


ansible/roles/linux-perf/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
+---
+
+#
+# installs Linux perf @ /usr/local/bin/perf
+#
+
+- stat: path=/usr/local/bin/perf
```
maclover7:
```yaml
stat:
  path: /usr/local/bin/perf
```

ansible/roles/linux-perf/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
...
+  register: linux_perf
+  tags: linux_perf
```
maclover7: ✂️ You can remove all of the tags

ansible/roles/linux-perf/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
...
+
+- name: Download Linux source code
+  git:
+    repo: https://github.com/torvalds/linux.git
+    depth: 1
+    dest: /data/tmp/linux
+  when: linux_perf.stat.exists == False
+  tags: linux_perf
+
+- name: Install build-essential
```
maclover7: You can combine all of these `package` calls into a single one, with a loop. An example is in the `benchmarking` role's `tasks/main.yml` file

ansible/roles/linux-perf/tasks/main.yml
refack: Keep the tags. It allows for directed application:
```
ansible-playbook test*gaga --tags "linux_perf"
```

LucaLanziani: if we want to keep the tags we should at least use a block and tag the block instead of adding tag on each task:

```
- name: install perf on linux
  block:
     - task 1
     - task 2
  tags: linux_perf
```

LucaLanziani: Or leave the `stat` out of the block and also add the `when: not (linux_perf.stat.exists | bool)` on the block

```
- stat:
  tags: linux_perf
  register: linux_perf

- name: install perf on linux
  block:
     - task 1
     - task 2
  tags: linux_perf
  when: not (linux_perf.stat.exists | bool)
```

LucaLanziani:
Nice work, just a couple of comments in the PR and a couple of questions for the team to understand how to write those roles in the future (@mmarchini sorry if I'm using your code as example):

1. should we write the roles assuming they are running as privileged user? (eg. setting the `become: true` from the outside)
2. should we check for  presence of perf at the start or should we just run the role every time? Assuming that the run is idempotent (like this seems to be)

mmarchini:
Thanks for the suggestions! I updated the PR with them.

> should we write the roles assuming they are running as privileged user?

I think so, at least looking at `ansible.cfg` and `inventory.yml` this seems to be true for most machines.

> should we check for presence of perf at the start or should we just run the role every time? 

The role will check if `perf` exists in `/usr/local/bin/perf` (i.e., if `perf` was installed by this role).

> Which machine(s) should this be running on?

At least Ubuntu 16.04 machines used by `stress-single-test`. We might want to install perf on other machines later.

mhdawson:
@jBarz since you did a bunch of ansible work for PPC/s390 could you take a look as well.

maclover7:
> should we write the roles assuming they are running as privileged user? (eg. setting the become: true from the outside)

Yes

> should we check for presence of perf at the start or should we just run the role every time? Assuming that the run is idempotent (like this seems to be)

To be honest, probably best to always run the code. That's essentially what the rest of the roles do

ansible/roles/linux-perf/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
+---
+
+#
+# installs Linux perf @ /usr/local/bin/perf
+#
+
+- stat:
+    path=/usr/local/bin/perf
```
maclover7: `path: /usr/local/bin/perf`

ansible/roles/linux-perf/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+  tags: linux_perf
+
+- name: install perf on linux
+  tags: linux_perf
+  when: not (linux_perf.stat.exists | bool)
+  block:
+    - name: Clean path
+      file:
+        state: absent
+        path: /data/tmp/
```
LucaLanziani: Might be worth it to move `/data/tmp` into its own variable considering that it's used in 5 tasks.

mmarchini:
@maclover7 @LucaLanziani done!

maclover7:
Landed in a5d59627dc4338adc6ad416359af66a15c5894a6

PR #1325, opened 2018-06-07
refresh incantations

refack:
I can never remember the exact CLI syntax, so it's nice to have precooked lines documented.

### DIFF
--- a/doc/jenkins-guide.md
+++ b/doc/jenkins-guide.md
@@ -114,10 +114,10 @@ be run on the worker itself, after SSH-ing in and switching to the
 
 ##### Out of memory
 
-First, get statistics on running processes for the machine: `ps aux | grep iojs | grep -v java | grep -v grep`.
+First, get statistics on running processes for the machine: `ps aux | grep node`.
 
 If there are a lot of "hanging" / "abandoned" processes, it's best to
-remove them by running a command like: `ps aux | grep iojs | grep -v java | grep -v grep | awk '{print $2}' | xargs kill`.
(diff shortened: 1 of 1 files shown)


doc/jenkins-guide.md
```diff
@@ -62,10 +62,10 @@ be run on the worker itself, after SSH-ing in and switching to the
 
 ##### Out of memory
 
-First, get statistics on running processes for the machine: `ps aux | grep node | grep -v -egrep -ejava`.
+First, get statistics on running processes for the machine: `ps aux | grep node`.
```
mhdawson: just wondering why the java one does not make sense anymore.  The jenkins agents still runs with Java.

refack: IMHO getting the stats of the `java` proc (or if there is none, or more then 1) is interesting. Also this is a simpler line...

mhdawson: Ok I see now, that its a -v, so its adding not removing the java processes.  But in that case does removing the -egrep make sense?

refack: We should hit those procs with the `node` in `ci.nodejs.org`, e.g.:
```
iojs       839     1  1 647319 174936 0 /usr/bin/java -Xmx128m -jar /home/iojs/slave.jar -jnlpUrl https://ci.nodejs.org
                                                                                                             ^^^^
```

sam-github:
@AshCripps this might be useful to land after rebasing

AshCripps:
@sam-github Ive rebased and correct the typos in the commit titles, want to review before I land?

sam-github:
The commit title is still `refresh incantations`, I think it should be `doc: refresh debugging notes` or some other less arcane title.

Othewise LGTM, and I think a year and a half is long enough for this to land now.

PR #1329, opened 2018-06-10
[WIP] Add node-test-commit-test

maclover7:
Starting point for having all `node-test-commit` sub jobs use a single script. Will be helpful when/if we move to pipelines, since all will be in one place

- [ ] freebsd
- [x] linux
- [x] osx
- [x] plinux
- [x] smartos
- [ ] windows
- [x] linux-containered
- [ ] linux-containered --> ubuntu1604_sharedlibs_debug_x64
- [ ] arm
- [ ] arm-fanned
- [x] linuxone
- [x] aix

### DIFF
--- a/jenkins/scripts/node-test-commit-test.sh
+++ b/jenkins/scripts/node-test-commit-test.sh
@@ -1,0 +1,273 @@
+#/bin/bash
+
+rm -rf build
+git clone https://github.com/nodejs/build.git
+
+. ./build/jenkins/scripts/node-test-commit-pre.sh
+
+if test $IGNORE_FLAKY_TESTS = "true"
+then
+  FLAKY_TESTS_MODE=dontcare
+else
+  FLAKY_TESTS_MODE=run
+fi
+echo FLAKY_TESTS_MODE=$FLAKY_TESTS_MODE
+
+echo $NODE_NAME
+
+if [[ $NODE_NAME = *"aix"* ]]; then
+  # Some of the tests require a file size limit that is greating that
+  # string size limite (see https://github.com/nodejs/node/pull/16273#pullrequestreview-70282286)
+  # Set the limit so the test can run
+  # Disable for now as need more config to make it allowable
+  #ulimit -f 4194304
+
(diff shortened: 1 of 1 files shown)


refack:
Great idea. We also have the [regular job DSL plugin](https://github.com/jenkinsci/job-dsl-plugin) (and there the XML to DSL tool in the main screen).
BTW it's now quite possible to script the windows jobs in bash.

jenkins/scripts/node-test-commit-test.sh
```diff
@@ -0,0 +1,43 @@
...
+else
+  FLAKY_TESTS_MODE=run
+fi
+echo FLAKY_TESTS_MODE=$FLAKY_TESTS_MODE
+
+echo $NODE_NAME
+
+. ./build/jenkins/scripts/select-compiler.sh
+
+if [ -x "$(command -v make)" ]
```
richardlau: Does this work on AIX? I don't know how the community AIX systems are set up but (on the internal IBM ones I have access to) [`make` exists](https://www.ibm.com/support/knowledgecenter/en/ssw_aix_61/com.ibm.aix.cmds3/make.htm) but is not `GNU make` that is required to build Node.js. Maybe test for `gmake` and default to `make` if not found?

mhdawson:
I'm +1 assuming we test across all of the platforms.

refack:
A few comments:
1. The script needs to be run with `bash -xe`
2. IMHO this should be a new sub-job and platform `labels` added to is as other subjobs are retired. (I'm starting this now)
    1. This will get us a simpler outcome
    2. This allows for only a single point where the branch URI is used
    3. It's easier to switch back in case of bug
3. Testing needs to be done with 3 positive controls
    1. Known green
    2. Known red
    3. known yellow

maclover7:
Going to close this for now, and roll out in smaller increments

refack:
You deleted the branch 😢 so now
https://raw.githubusercontent.com/maclover7/build/jm-node-test/jenkins/scripts/node-test-commit-test.sh is 404

PR #1335, opened 2018-06-12
Add basic testing on linux-perf role using kitchen

LucaLanziani:
Will execute the role in a docker container and test the presence of perf

### DIFF
--- a/ansible/roles/linux-perf/.kitchen.yml
+++ b/ansible/roles/linux-perf/.kitchen.yml
@@ -1,0 +1,28 @@
+---
+
+driver:
+  name: docker
+  use_sudo: false
+  cap_add:
+    - SYS_ADMIN
+
+platforms:
+  - name: ubuntu-14.04
+#  - name: ubuntu-16.04
+
+provisioner:
+  hosts: test-kitchen
+  name: ansible_playbook
+  playbook: default_kitchen.yml
+  roles_path: ../
+  filter_plugins_path: ../../plugins/filter
+  extra_vars: 
+    os: ubuntu1404
+  require_chef_for_busser: true
+  require_ruby_for_busser: false
+  ansible_verbosity: 1
+  ansible_verbose: true
+  sudo: true
+
+suites:
+  - name: default
--- a/ansible/roles/linux-perf/test/integration/default/bats/verified_installed.bats
+++ b/ansible/roles/linux-perf/test/integration/default/bats/verified_installed.bats
@@ -1,0 +1,3 @@
+@test "tmux is installed and in the path" {
+  which perf
+}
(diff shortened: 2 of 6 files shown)


ansible/roles/linux-perf/test/integration/default/bats/verified_installed.bats
```diff
@@ -0,0 +1,3 @@
+@test "tmux is installed and in the path" {
```
mmarchini: `s/tmux/perf/`

ansible/roles/linux-perf/.kitchen.yml
```diff
@@ -0,0 +1,28 @@
...
+
+driver:
+  name: docker
+  use_sudo: false
+  cap_add:
+    - SYS_ADMIN
+
+platforms:
+  - name: ubuntu-14.04
+#  - name: ubuntu-16.04
```
mmarchini: Can we test on ubuntu-16.04 instead of ubuntu-14.04? `linux-perf` role is executed only on ubuntu-16.04 machines right now (see https://github.com/nodejs/build/blob/master/ansible/playbooks/jenkins/worker/create.yml#L63-L66)

ansible/roles/linux-perf/test/integration/default/bats/verified_installed.bats
LucaLanziani: Thanks @mmarchini I'll fix this

ansible/roles/linux-perf/.kitchen.yml
LucaLanziani: Sure we can ;)

LucaLanziani:
Let's wait for @maclover7 to have a look at this and once we know if think we are going to merge this or not I'll fix the above issues ;)

maclover7:
@LucaLanziani Very interesting! Is there a way to run these tests inside of a container? If so, we can make a CI test job to run on PRs, so we don't have to install more dependencies on the test machines.

LucaLanziani:
@maclover7 those tests are already running in a container:

https://github.com/nodejs/build/pull/1335/files#diff-651ce8be44824032669549a986abcc42R6

The platforms are containers

maclover7:
@LucaLanziani Sorry, I had meant as far as running `bundle install` and `kitchen test` -- is it possible to setup a Dockerfile with stock Ruby and bundler so we don't have to add any additional dependencies

LucaLanziani:
Ah yeah @maclover7 it can be done but if you want to run kitchen with the current config we will have to run docker inside docker.
The container will have to include `ruby` and `docker` and if I'm not wrong will have to run with some privileged.

For now this can be an simple way to test roles on your local machine during development

sam-github:
Closing as inactive.

Issue #1342, opened 2018-06-14
run `make test-v8-updates` as a sub-job of `node-test-commit-v8-linux`


mmarchini:
As discussed in https://github.com/nodejs/build/issues/1274 and https://github.com/nodejs/node/pull/20783, we should add a sub-job to `node-test-commit-v8-linux` which runs `make test-v8-updates` on a Ubuntu 16.04 machine, otherwise tests for missing postmortem metadata and for Linux perf won't run during V8 updates.

maclover7:
@mmarchini Can you please prepare the full bash script for the job, and then I can drop that into a new job and setup the configuration for you

mmarchini:
Sure. The script should look like https://github.com/nodejs/build/pull/1329/files, right?

mmarchini:
`v8-updates` are running for `node-test-commit-v8-linux`, so this can be closed.

Ref: https://github.com/nodejs/build/pull/1344 and https://github.com/nodejs/build/pull/1344#issuecomment-398144436

PR #1344, opened 2018-06-14
jenkins: add node-test-commit-v8-linux script

mmarchini:
Ref: https://github.com/nodejs/build/issues/1342
Ref: https://github.com/nodejs/node/pull/20783

### DIFF
--- a/jenkins/scripts/node-test-v8-updates.sh
+++ b/jenkins/scripts/node-test-v8-updates.sh
@@ -6,9 +6,19 @@
 # are inteded to be run in Ubuntu 16.04 machines.
 ###
 
-rm -rf build
-git clone https://github.com/nodejs/build.git
+FLAKY_TESTS_MODE=run
+if test $IGNORE_FLAKY_TESTS = "true"; then
+  FLAKY_TESTS_MODE=dontcare
+fi
 
-. ./build/jenkins/scripts/node-test-commit-pre.sh
+echo FLAKY_TESTS_MODE=$FLAKY_TESTS_MODE
 
-make run-ci CI_JS_SUITES="v8-updates" CI_NATIVE_SUITES="" CI_DOC=""
+export JOBS=$(getconf _NPROCESSORS_ONLN)
+
+MAKE_ARGS="--output-sync=target -j $JOBS CI_JS_SUITES='v8-updates' CI_NATIVE_SUITES='' CI_DOC=''"
+
+NODE_TEST_DIR=${HOME}/node-tmp \
+PYTHON=python \
+FLAKY_TESTS=$FLAKY_TESTS_MODE \
+V=1 \
+make run-ci $MAKE_ARGS
--- a/jenkins/scripts/node-test-v8-updates.sh
+++ b/jenkins/scripts/node-test-v8-updates.sh
@@ -11,7 +11,4 @@ git clone https://github.com/nodejs/build.git
 
 . ./build/jenkins/scripts/node-test-commit-pre.sh
 
-./configure;
-make -j $(getconf NPROCESSORS_ONLN);
-
-python tools/test.py -p tap --logfile test.tap --mode=release v8-updates
+make run-ci CI_JS_SUITES="v8-updates" CI_NATIVE_SUITES="" CI_DOC=""
--- a/jenkins/scripts/node-test-v8-updates.sh
+++ b/jenkins/scripts/node-test-v8-updates.sh
@@ -1,0 +1,17 @@
+#/bin/bash -ex
+
+###
+# node-test-v8-updates
+# Run specific Node.js tests for V8 updates (from test/v8-updates). These tests
+# are inteded to be run in Ubuntu 16.04 machines.
+###
+
+rm -rf build
+git clone https://github.com/nodejs/build.git
+
+. ./build/jenkins/scripts/node-test-commit-pre.sh
+
+./configure;
+make -j $(getconf NPROCESSORS_ONLN);
+
+python tools/test.py -p tap --logfile test.tap --mode=release v8-updates


mmarchini:
Should be enough to run `v8-updates` tests on Ubuntu 16.04. @maclover7 PTAL

refack:
@mmarchini the build phase of the job should be similar to a regular `node-test-commit`?

mmarchini:
Yes

refack:
So this is our build script
```bash
#!/bin/bash -xe

FLAKY_TESTS_MODE=run
if test $IGNORE_FLAKY_TESTS = "true"; then
  FLAKY_TESTS_MODE=dontcare
fi

echo FLAKY_TESTS_MODE=$FLAKY_TESTS_MODE

if [ -z ${JOBS+x} ]; then
  export JOBS=$(getconf _NPROCESSORS_ONLN)
fi

MAKE_ARGS="-j $JOBS"

if [ $(make -v | grep 'GNU Make 4' -c) -ne 0 ]; then
  MAKE_ARGS="$MAKE_ARGS --output-sync=target"
fi

exec_cmd=" \
  NODE_TEST_DIR=${HOME}/node-tmp \
  PYTHON=python \
  FLAKY_TESTS=$FLAKY_TESTS_MODE \
  make run-ci $MAKE_ARGS \
"

if test $nodes = "centos6-64-gcc48"; then
  exec_cmd=". /opt/rh/devtoolset-2/enable; $exec_cmd"
elif [[ "$nodes" =~ centos[67]-(arm)?64-gcc6 ]]; then
  exec_cmd=". /opt/rh/devtoolset-6/enable; $exec_cmd"
fi

echo $exec_cmd
V=1

$SHELL -xec "$exec_cmd"
```

I'm sure there is a ton of cross platform unneeded code there, but the tl;dr is we run `make run-ci`

refack:
@mmarchini I created a new job and given you edit perms: https://ci.nodejs.org/job/node-test-commit-linux-v8-updates/configure
Let's see if it works

mmarchini:
Worked 😄 

````

**Comment by maclover7** [comment]
@mmarchini Ahh, didn't fully understand what was needed here. In theory, you should just be able to utilize `node-test-commit-custom-suites`, I started a sample of what this could look like at https://ci.nodejs.org/job/node-test-commit-custom-suites/11

**Comment by mmarchini** [comment]
@maclover7 thanks for pointing this out, but `node-test-commit-custom-suites` will run on `rhel72-s390x` and `test/v8-updates` should be run on Ubuntu 16.04 (otherwise the Linux perf test will be skipped).

**Comment by refack** [comment]
I added `node-test-commit-linux-v8-updates` as a build step of `node-test-commit-v8-linux. I don't think we used this feature before, let's see how it behaves.
https://ci.nodejs.org/job/node-test-commit-v8-linux/1427/

**Comment by maclover7** [comment]
> test/v8-updates should be run on Ubuntu 16.04

@mmarchini Ah true -- let me update that job configuration, and then we should be okay to use that (would rather try and reuse existing job pieces if possible than introduce new stuff, to avoid more complexity in Jenkins)

**Comment by refack** [comment]
Well that gives weird results, every platform job triggers a `linux-v8-updates` jobs
https://ci.nodejs.org/job/node-test-commit-v8-linux/1427/nodes=benchmark,v8test=v8test/

Let me do some key mashing... 🤔

**Comment by refack** [comment]
@mklebrasseur do you have a quick answer for this?

tl;dr:
we have an exsisting multi-configuration project (https://ci.nodejs.org/job/node-test-commit-v8-linux/), we want to add a parameterized sub job (https://ci.nodejs.org/job/node-test-commit-custom-suites) to it with minimal refactoring

**Comment by mklebrasseur** [comment]
@refack sure I can have a look. Do I need any additional access for those jobs to view the build steps. I am rather new to this project and have not got my grounds yet.

**Comment by refack** [comment]
> Do I need any additional access for those jobs to view the build steps.

Unfortunately, Yes... If you know how to enable "read only" view, that's also interesting.

Best I can do is a screenshot and the XML
https://gist.github.com/refack/05e9afb15c722ce2fece27bab620a1f2
[node-test-commit-v8-linux Config [Jenkins].pdf](https://github.com/nodejs/build/files/2107346/node-test-commit-v8-linux.Config.Jenkins.pdf)

ATM I added the sub job as the last post-build-action.

**Comment by mklebrasseur** [comment]
Yea, my mistake I was looking at the repo before with the use of some pipelines. I've never enjoyed having to use the UI to build out any builds. Shared libraries I found to be much easier and allow for granular access using git (as well as read only). We recently converted all our legacy builds to pipelines this month. It's very fresh in my head.

Conditional will work fine, also the use of a shared library would work too. 

Apologies for the delayed response.

**Comment by refack** [comment]
@mklebrasseur and @mmarchini I gave you read access to https://ci.nodejs.org/job/node-test-commit-v8-linux/configure
Please say you can see that. This is a test of a new plugin that allows read only access to joob configs.

**Comment by mklebrasseur** [comment]
@refack yes I can confirm that can see the configuration. No save options appear (y)

**Comment by mmarchini** [comment]
Doesn't seems to be working here :/

screenshot_20180616-113640

**Comment by mmarchini** [comment]
`v8-updates` tests are still not running for `node-test-commit-v8-linux`, even though settings seems to be correct:

image

image

Relevant Jenkins log:

```
05:53:51 Join notifier cannot find upstream project: node-test-commit-v8-linux/nodes=benchmark,v8test=v8test
```

https://ci.nodejs.org/job/node-test-commit-custom-suites/17/

mklebrasseur:
Wonder if you should use Regexp instead of Strings

refack:
> v8-updates tests are still not running for node-test-commit-v8-linux, even though settings seems to be

I should have verified that `node-test-commit-custom-suites` actually understands the platform label parameter (and doesn't just run on empty).

https://ci.nodejs.org/job/node-test-commit-custom-suites/20/ seems to be working.
Let's see - https://ci.nodejs.org/job/node-test-commit-v8-linux/1439/

mmarchini:
Worked this time 😄

It's not very intuitive to see the test result (node-test-commit-v8-linux#1439 > nodes=benchmark > node-test-commit-custom-suites#21), but we can improve that in the future if we think it's necessary.

PR #1347, opened 2018-06-15
add ccache to jenkins_path for macOS

gdams:
This fix means that ccache is now in the path on new mac. I have tested tested this change on `test-macstadium-macos10.10-x64-1` and the jenkins_path now shows to be `/usr/localopt/ccache/libexec:/usr/bin:/bin:/usr/sbin:/sbin`

### DIFF
--- a/ansible/roles/jenkins-worker/templates/start.j2
+++ b/ansible/roles/jenkins-worker/templates/start.j2
@@ -6,7 +6,8 @@ export JOBS="{{ jobs_env }}"
 export OSTYPE=osx
 export ARCH=x64
 export DESTCPU=x64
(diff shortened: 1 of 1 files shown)


maclover7:
Rolled this out across all Mac machines

gdams:
@maclover7 did you restart the jenkins agent as well?

maclover7:
@gdams Yep, I'm working on a PR to add this to Ansible

refack:
We might want to roll this back. It makes command line testing harder since the environments for the jenkins agent and a vanilla ssh login are different. AFAICT other jenkins daemonizers are also kept minimal...

gdams:
ccache is also added to the vanilla ssh login path so there should be no issue?

PR #1352, opened 2018-06-16
setup: remove raspberry-pi

maclover7:
Tested new Ansible "half docker" setup on all Pi machines, and seems to
be working well. @rvagg, would you be able to take a look at the docs in
`doc/non-ansible-configuration-notes.md`, and see what can be
:scissors:?


rvagg:
OK, the two items in doc/non-ansible-configuration-notes.md are the first line talking about stand-alone ansible that you're now deleting and the use of Wheezy, it needs to be Stretch (or whatever the latest is) so probably needs instructions for how to get that. I guess I'll have to make this a TODO for myself.

+1 on deletion of setup/ files tho.

maclover7:
@rvagg Opened up https://github.com/nodejs/build/issues/1355 so we don't forget to update the docs, going to merge this in

PR #1356, opened 2018-06-18
setup: remove osx

maclover7:
cc @gdams


maclover7:
@gdams See #1357

maclover7:
Actually, wait, yeah that probably won't cut it. Let me add to `doc/jenkins-guide.md`...

gdams:
yeah we should add it somewhere convinent for people to easily look it up

maclover7:
@gdams updated

mhdawson:
I'm wondering if the non-MacStadium OSX machines (those I think that are in @rvagg's site) still need some of the parts being removed.  @rvagg can you comment.

rvagg:
No, I think we're good to go with this. I believe I should be able to re-use the new Ansible scripts to re-setup the existing VMs if I dared blow away the existing state.

lgtm

maclover7:
Landed in b804f76

PR #1365, opened 2018-06-19
doc: move NPM management to access document

maclover7:
Try to centralize documentation that deals with "access"


maclover7:
Landed in 66729dd

PR #1368, opened 2018-06-19
doc: update members list

maclover7:
One member would like to be pinged on SmartOS only things, and the other
has been removed due to inactivity. Both already are gone from
the `nodejs/build` and `nodejs-private/build` GitHub teams


richardlau:
Any merit in a list of WG emeriti?

maclover7:
> Any merit in a list of WG emeriti?

@richardlau Maybe? Looking at what we've done in the past, we usually just remove them from the README with no further mention

PR #1376, opened 2018-06-20
www: nodejs.org/github-webhook.log

rvagg:
Expose the website build log, I don't believe there's anything in here that needs to be protected, it's all run inside a Docker container too so you can't even manipulate the website repo to expose secrets through this file.

Need a sanity check from @nodejs/website: is there anything in the build output that we wouldn't want to expose publicly? `npm install --cache-min 1440 --production && npm run deploy` is what's run on the server.

Would also appreciate sanity check from any build folks who might think of a security concern that I haven't.

### DIFF
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -306,6 +306,11 @@ server {
         default_type text/plain;
     }
 
+    location /github-webhook.log {
+        alias /home/nodejs/github-webhook.log;
+        default_type text/plain;
+    }
+
     location /metrics {
         alias /home/dist/metrics/;
         autoindex on;


maclover7:
Seems reasonable to me. Maybe run the playbook (and re-send one or two "push" webhooks) to test this out, before announcing it's actually live, in case anything does leak out.

---

On a separate note, I'm honestly more worried about this playbook being run than the actual change itself... Are all of the NGINX files up-to-date in `setup/www`, as well as `dist_users`/`libuv_users`? Concerned that `www` may end up in a weird state after using the playbook. (If I end up getting access to www, the first item on my list is to make sure the Ansible playbook matches the current server setup...)

fhemberger:
@rvagg There's nothing special in the build log on our side, so LGTM.

rvagg:
Yes, the config files are up to date with current server state, changes get made in both places. Unfortunately there's far too much complex state on the server for me to be comfortable running it fresh over the server (although I think the last time I did that, by accident, there was nothing severe).

rvagg:
This is live now, @nodejs/website if you ever want to know why the site isn't updating you can just view https://nodejs.org/github-webhook.log. It bypasses Cloudflare so should be a live feed of what's happening. It's pretty large, there's no logrotate on this one, perhaps that'll be the next PR.

PR #1363, opened 2018-06-18
doc: update ONBOARDING.md

maclover7:
Adds more concrete steps to the onboarding guide for the WG. This is
updated with the information I used to onboard @mmarchini.

PR #1382, opened 2018-06-22
github-bot: restart service without sudo

maclover7:
`iojs` on the github-bot server is not a sudo user, and we don't need
sudo anyways. This was preventing the server from relaunching.


maclover7:
cc @phillipj, would like to expedite this if possible

maclover7:
Landed in 55714b8, thank you for the quick reviews!

joaocgreis:
@maclover7 considering what I wrote in https://github.com/nodejs/build/pull/1363#discussion_r197549666 , in cases like this if something is not working, feel free to fix it by running Ansible from your branch or editing in the server directly. The PR here would give us a chance of reviewing the fix, if something changes it can always be run again. Anyway, thanks for fixing stuff!

PR #708, opened 2017-05-04
[WIP] win: add a "VS2017 Build Tools" playbook

refack:
A new playbook to set-up "VS2017 Build Tools".

PR #1383, opened 2018-06-22
Refactor the Windows world

joaocgreis:
This moves the Windows Ansible scripts to `ansible/`. Since some refactoring was necessary to make them fit, I took the opportunity to change most packages to install from Chocolatey.

This can be used to deploy new workers, but I'm not sure it can be safely used to refresh the current ones. Hence, this does not include changes in the inventory yet. I will keep working on this. When that happens, I will merge the `create-windows.yml` into `create.yml`.

cc @maclover7


ansible/README.md
```diff
@@ -16,16 +17,31 @@
...
+running one (or multiple) of below playbooks.
+
+See the [manual setup instructions](../doc/non-ansible-configuration-notes.md)
+for how to prepare both the control and target machines to run the commands
+below. To ensure that the secrets are in place and test the connection to a
+host use:
+
+```console
+$ ansible test-digitalocean-debian8-x64-1 -m ping -vvvv
+Or, for Windows hosts:
```
maclover7: Maybe add a `#` so this appears as a comment

setup/ansible-inventory
```diff
@@ -45,40 +45,3 @@ iojs-build-centos5
 
 [iojs-aix61]
 test-ibm-aix61-ppc64-1:18822
-
-[node-windows]
-node-win10-x64-1.cloudapp.net
```
maclover7: Just to confirm, all of these inventory entries exist in `ansible/inventory.yml` (ignoring whether or not the Ansible playbook has been run on them) Don't want these hosts to get lost somehow

maclover7:
Will this cover what's added in https://github.com/nodejs/build/pull/708?

ansible/README.md
joaocgreis: This highlights as `console`, meaning the prompt starts a command line. If I only add a `#` it looks bad, so I added two. Could also have added the prompt `$ #`, let me know if you prefer this. Any of these looks better than dividing this into two blocks I think.

joaocgreis:
@nodejs/build I plan to land this soon. Please keep in mind that this is only good for creating new machines, I don't know if this will run well on existing ones. I will keep working on this, I plan to review our current inventory and update `inventory.yml` as I do so.

joaocgreis:
Landed in https://github.com/nodejs/build/commit/054bb4eea6fd07e951ee9c0ad82e3a252f8eb724. Thanks for your review @maclover7!

PR #1384, opened 2018-06-22
github-bot: fix privilege for iojs user

maclover7:
Turns out #1382 was not the right fix, the issue was that the `iojs`
user did not have proper privileges to run the command, not the command
itself.


maclover7:
Landed in 9dcaaf5

PR #1389, opened 2018-07-01
Add virtualenv to ansible directory

LucaLanziani:
In my Linux distribution the default python is python3, thus
I was then having issues with the python scripts in the repo
where the shabang is `#!/usr/bin/env python`.

First idea was to change the shabang to `... python2` but I
realized that a virtual env can also help to pin the versions of
ansible and libraries eg. Jinja2 and make everyone env consistent.

### DIFF
--- a/ansible/Makefile
+++ b/ansible/Makefile
@@ -7,7 +7,7 @@ ENV: Makefile
 
 CHECK_ENV:
 ifndef VIRTUAL_ENV
-	$(error PLEASE ENTER THE VIRTUALENV BEFORE FUN THE COMMAND)
+	$(error PLEASE ENTER THE VIRTUALENV BEFORE RUNNING THE COMMAND)
 endif
 
 UPDATE_ENV: requirements.txt
@@ -17,4 +17,3 @@ UPDATE_ENV: requirements.txt
 
 FLAKE8: CHECK_ENV
 	flake8
-
--- a/ansible/Makefile
+++ b/ansible/Makefile
@@ -1,0 +1,20 @@
+.PHONY: ENV
+
+virtualenv_path = ./virtualenv
+
+ENV: Makefile
+	virtualenv2 --distribute $(virtualenv_path)
+
+CHECK_ENV:
+ifndef VIRTUAL_ENV
+	$(error PLEASE ENTER THE VIRTUALENV BEFORE FUN THE COMMAND)
+endif
+
+UPDATE_ENV: requirements.txt
+	$(virtualenv_path)/bin/pip install --upgrade setuptools
+	$(virtualenv_path)/bin/pip install --upgrade pip
+	$(virtualenv_path)/bin/pip install -r requirements.txt
+
+FLAKE8: CHECK_ENV
+	flake8
+
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -8,15 +8,15 @@
 1. Read this document.
 2. For SSH access, see the [SSH guide](../doc/ssh.md).
 
-### Install ansible system wide
+### Install ansible systemwide
 
 Follow the [instructions to install the latest version of Ansible][ansible-install].
-* In most cases, using pip: `pip install ansible`.
+* In most cases, use pip: `pip install ansible`.
 * If you use brew, then `brew install python2 ansible`, and then run
 `export PYTHONPATH=$(pip2 show pyyaml | grep Location | awk '{print $2}') `
 before you use `ansible-playbook`.
 
-### Install ansible in virtualenv
+### Install ansible via virtualenv
 
 From the `ansible` directory:
 
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -5,14 +5,27 @@
 
 ## Getting started
 
-1. Follow the [instructions to install the latest version of Ansible][ansible-install].
-   * In most cases, using pip: `pip install ansible`.
-   * If you use brew, then `brew install python2 ansible`, and then run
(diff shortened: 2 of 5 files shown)


ansible/Makefile
```diff
@@ -0,0 +1,20 @@
...
+endif
+
+UPDATE_ENV: requirements.txt
+	$(virtualenv_path)/bin/pip install --upgrade setuptools
+	$(virtualenv_path)/bin/pip install --upgrade pip
+	$(virtualenv_path)/bin/pip install -r requirements.txt
+
+FLAKE8: CHECK_ENV
+	flake8
+
```
maclover7: ✂️

ansible/Makefile
```diff
@@ -0,0 +1,20 @@
+.PHONY: ENV
+
+virtualenv_path = ./virtualenv
+
+ENV: Makefile
+	virtualenv2 --distribute $(virtualenv_path)
+
+CHECK_ENV:
+ifndef VIRTUAL_ENV
+	$(error PLEASE ENTER THE VIRTUALENV BEFORE FUN THE COMMAND)
```
maclover7: `FUN`?

ansible/README.md
```diff
@@ -5,14 +5,27 @@
...
-   * In most cases, using pip: `pip install ansible`.
-   * If you use brew, then `brew install python2 ansible`, and then run
-   `export PYTHONPATH=$(pip2 show pyyaml | grep Location | awk '{print $2}') `
-   before you use `ansible-playbook`.
-2. Read this document.
-3. For SSH access, see the [SSH guide](../doc/ssh.md).
+1. Read this document.
+2. For SSH access, see the [SSH guide](../doc/ssh.md).
 
+### Install ansible system wide
```
maclover7: `Install Ansible systemwide`

ansible/README.md
```diff
@@ -5,14 +5,27 @@
...
+
+Follow the [instructions to install the latest version of Ansible][ansible-install].
+* In most cases, using pip: `pip install ansible`.
+* If you use brew, then `brew install python2 ansible`, and then run
+`export PYTHONPATH=$(pip2 show pyyaml | grep Location | awk '{print $2}') `
+before you use `ansible-playbook`.
+
+### Install ansible in virtualenv
```
maclover7: Install Ansible via `virtualenv`

ansible/README.md
```diff
@@ -5,14 +5,27 @@
...
-   before you use `ansible-playbook`.
-2. Read this document.
-3. For SSH access, see the [SSH guide](../doc/ssh.md).
+1. Read this document.
+2. For SSH access, see the [SSH guide](../doc/ssh.md).
 
+### Install ansible system wide
+
+Follow the [instructions to install the latest version of Ansible][ansible-install].
+* In most cases, using pip: `pip install ansible`.
```
maclover7: using --> use

ansible/Makefile
LucaLanziani: That was `RUNNING`, no idea why it become `FUN`, I'll fix it

LucaLanziani:
@maclover7 comments addressed

mhdawson:
Sorry if this is a dumb questions, but I assume that virtualenv support is consistent across platforms and available on all linux distros?

refack:
> Sorry if this is a dumb questions, but I assume that virtualenv support is consistent across platforms and available on all linux distros?

Yes, and so is `pipenv`, both are pure python & a little bit of POSIX shell (they are kind of the equivalent to `npm`).
But since this is only required for the `ansible` host, it anyway should not affect the way we manage the "workers".

maclover7:
@refack What are the +/- of pipenv versus virtualenv (personally I'm not familiar with it) Would you be willing to file a pull request that implements pipenv, and then we can compare both options side by side?

refack:
> What are the +/- of pipenv versus virtualenv (personally I'm not familiar with it) Would you be willing to file a pull request that implements pipenv, and then we can compare both options side by side?

Will do.

tl;dr:
* `pip` is probably the most popular package manager for python (but defaults to global scope)
* `virtualenv` is the conventional way to enable local scoped dependencies (a la  `node_modules`)
* `pipenv` is kind of new (~1 yo) and is the latest "blessed" incarnation of a wrapper for the last two. It's the most similar to `npm` in that by default it installs dependencies in local scope.

refack:
See seggestions in https://github.com/nodejs/build/pull/1399

LucaLanziani:
Sorry I've should have asked before proceeding with the implementation. 

The only reason I didn't go with pipenv is that imho it adds extra features like:

- dev dependencies (do we have a dev/prod env definition? genuine question)
- lock file (do we need this on releases? genuine question)
- install packages and update requirements for you (for this I can see a use)

over something that works already well enough on simple cases like the one we have here (I might be underestimating the complexity of the repo).

That said, more than happy to use pipenv if you guys thinks it's better/we need this functionalities.

refack:
> Sorry I've should have asked before proceeding with the implementation.

No worries. I mostly like it because it's newer and is more `npm` like. Only two features it has that make a difference are newer [PEP508](https://www.python.org/dev/peps/pep-0508/) dependency URIs, and it captures the required python engined needed.

What I really wanted to do is get the code to be 2/3 compatible (https://github.com/nodejs/build/pull/1399/commits/2b5e0639855db44d5d1fbf96d7e244dc1930f4e0).
With a formal python2 EOL date, we need to start moving...

sam-github:
stale, and not what we are doing ATM.

PR #1390, opened 2018-07-01
WIP: ansible: add ansible tower playbook

gdams:
This playbook will create the basic AWX tower machine. There are a few things that we will probably still want to add:

1. Create a github app and auto configure the github auth CC @rvagg (we'll need one to be created).
2. Auto create the job templates so that we don't have to manually create each job template from a playbook.
3. Add a DNS entry for the machine and enable SSL using Cloudflare.


gdams:
@maclover7 I'm happy to move it over if others from @nodejs/build agree?

mhdawson:
Moving to ansible-tower role sounds good to me.

gdams:
@maclover7 I've now moved this to an ansible-tower role

ansible/playbooks/create-ansible-tower.yml
```diff
@@ -0,0 +1,22 @@
+---
+
+#
+# sets up the host that runs ansible.nodejs.org
+#
+
+- hosts: infra-softlayer-ubuntu1804-x64-2
```
joaocgreis: To be safe, can you explicitly exclude this host in https://github.com/nodejs/build/blob/master/ansible/playbooks/jenkins/worker/create.yml ?

rvagg:
DNS entry done. Although I've had to change SSL to "flexible" to allow the backend to be http rather than https. Apparently in "full", as it was set, we could even expose https on the server with a self-signed certificate and it would be cool with that. Perhaps we should do that so the Cloudflare<->host comms are encrypted as well? We have a wildcard certificate but that should go away soon if we switch to letsencrypt or we might just switch to Cloudflare's own certificates. Or, maybe we could do letsencrypt on this ansible host for comms with Cloudflare? Can we expose /.well-known/ on it to verify with letsencrypt?

So next up, authenticating with GitHub. There is an OAuth app, like our Jenkins servers, and there's GitHub apps. It looks like we might need the latter so it can access GitHub resources? I'm not really sure how that works and what it can access and as who, but this is the config screen:

screenshot 2018-07-31 15 55 22

Are there docs somewhere about this we can see for Ansible Tower to understand what the implications are? We'll probably need to give the TSC a heads-up.

rvagg:
I had to switch back to "full" ssl on Cloudflare because it screwed up nodejs.org! I don't know how or why but it made it stick clients into a redirect loop.

So for now, we're going to need an ssl cert of some kind on this new server for it to work via https.

gdams:
@rvagg I'm happy to use letsencrypt! Looks like there is already a nice ansible module that we could implement... https://docs.ansible.com/ansible/2.5/modules/letsencrypt_module.html

sam-github:
This appears to be abandoned, but should it be picked up again? Its not clear.

@mhdawson @rvagg, close as not-in-progress, label as needs-help? Other?

PR #1395, opened 2018-07-08
ansible: move linter tasks to worker/create.yml playbook

maclover7:
Also fixes an issue with the FreeBSD `jenkins-worker` template


maclover7:
Going to land this, fixes a bug that prevents playbooks from being run on linter (or any other FreeBSD) machines (I have to keep switching to this branch to get the fix)

PR #1399, opened 2018-07-12
ansible: add Pipfile and make python3 compatible

refack:
1. Add `Pipfile` to describe dependencies
2. add needed git ignore rules
3. make code `PEP8` compliant and add `tox.ini`
4. make code python3 (3.6 specifically) compatible


refack:
Ref: https://github.com/nodejs/build/pull/1389

maclover7:
@refack Would you be able to update `ansible/README.md` and the other docs with the new `pipenv` commands?

rvagg:
I don't understand any of this, I just know ansible is broken for me on my mac

does it have to be so heavy-weight? switching to `items()` in filter.py fixes my immediate issue. Why do we need these dependencies and what is PEP8 and why do we need it? Is there a minimal version of this?

cclauss:
Rebase?

refack:
Code has been reworked and rebased.
Some stuff that "work for me":
```
ansible -m ping test*

ansible-playbook -vvv playbooks/upgrade-packages.yml -l test-rackspace-ubuntu1604-x64-2

ansible-playbook -vvvv playbooks/jenkins/worker/create.yml -l test-rackspace-ubuntu1604-x64-2 --check
```

ansible/plugins/inventory/nodejs_yaml.py
```diff
@@ -23,35 +23,33 @@
...
-  # valid roles - add as necessary
-  'type': ('infra', 'release', 'test'),
-
-  # providers - validated for consistency
-  'provider': ('azure', 'digitalocean', 'joyent', 'ibm', 'linuxonecc',
-               'macstadium', 'marist', 'mininodes', 'msft', 'osuosl',
-               'rackspace', 'requireio', 'scaleway', 'softlayer', 'voxer',
-               'packetnet', 'nearform')
+    # taken from nodejs/node.git: ./configure
+    'arch': ('armv6l', 'armv7l', 'arm64', 'ia32', 'mips', 'mipsel', 'ppc', 'ppc64', 'x32', 'x64', 'x86', 's390', 's390x'),
```
rvagg: if you're still wrapping `provider` then you should probably wrap this too

setup/www/host_vars/.gitignore
```diff
@@ -1 +1,2 @@
-infra-*
\ No newline at end of file
+infra-*
+!node-www.tmpl
```
rvagg: what's this for?

rvagg:
So, with this we'd need to execute ansible from inside `pipenv` and not bother with having a global ansible package?

Can you explain the logic of this for me and what the dependencies are for?

rvagg:
And: does this crate necessary dependencies? Will running the playbooks directly with `ansible-playbook` even on a machine with Python 3 still work or must we use the steps above? If so, maybe we could script it? `./run-playbook.sh .....`

setup/www/host_vars/.gitignore
refack: We're tracking this file https://github.com/nodejs/build/blob/master/setup/www/host_vars/node-www.tmpl
(keeping it ignored, make my IDE angry)

refack:
> So, with this we'd need to execute ansible from inside `pipenv` and not bother with having a global ansible package?
 
> Can you explain the logic of this for me and what the dependencies are for?

It's an alternative approach.
tl;dr python out-of-the-box works with global dependencies.
Then came virtualenv that wraps python with some monkey-patches so it first looks for dependencies in a local dependency store.
Then came `pipenv` and the `Pipefile` that allows for a workflow similar to `npm` (i.e.  dependencies stored locally by default, and `pip install` installs locally and updates the `Pipfile`).

> And: does this crate necessary dependencies? Will running the playbooks directly with `ansible-playbook` even on a machine with Python 3 still work or must we use the steps above? If so, maybe we could script it? `./run-playbook.sh .....`

It's still possible to use global dependencies just as before. Or we can use `npm run` style recipes that allow doing things like:
```
pipenv run playbook [what [ever [args]]]
```
example: 
https://github.com/refack/GYP/blob/86ab71caad4640dcc3ec572c2118a1ce3f24150f/Pipfile#L12-L14
used as:
https://github.com/refack/GYP/blob/master/.travis.yml#L35

ansible/plugins/inventory/nodejs_yaml.py
refack: ack

rvagg:
OK, as long as it's documented I could live with it. I'll try it out soon on some of my known-breaking Python 3 Ansible tasks.

rvagg:
this worked on a freebsd and ubuntu 14.04 reprovision but the problem I have with ccache jinja2 scripting persists:

```
fatal: [test-requireio_kahwee-debian9-arm64_pi3-1]: FAILED! => {"msg": "Unexpected templating type error occurred on ({{ ccache_html_content.stdout | regex_findall('ccache-[0-9]+.[0-9]+(?:.[0-9]+)*.tar.gz') | map('regex_replace', 'ccache-') | map('regex_replace', '.tar.gz') | list | latest_version }}): '<' not supported between instances of 'map' and 'map'"}
```

cclauss:
You could try putting __| list |__ in between the two calls to map.  Python 3 often creates generators instead of lists and perhaps forcing the results into a list will placate Jinja2.

refack:
>> `'<' not supported between instances of 'map' and 'map'"`

> forcing the results of the first **map()** call into a list will placate Jinja2.

👍 make sense.

anyway.
I'll try to move that super-filter out of the template and into code

ansible/plugins/inventory/nodejs_yaml.py
```diff
@@ -23,35 +23,36 @@
...
-
-  # valid roles - add as necessary
-  'type': ('infra', 'release', 'test'),
-
-  # providers - validated for consistency
-  'provider': ('azure', 'digitalocean', 'joyent', 'ibm', 'linuxonecc',
-               'macstadium', 'marist', 'mininodes', 'msft', 'osuosl',
-               'rackspace', 'requireio', 'scaleway', 'softlayer', 'voxer',
-               'packetnet', 'nearform')
+    # taken from nodejs/node.git: ./configure
```
thefourtheye: I don't think they are in sync anymore. https://github.com/nodejs/node/blob/ed2c6965d2f901f3c786f9d24bcd57b2cd523611/configure.py#L48-L49

```python
valid_arch = ('arm', 'arm64', 'ia32', 'ppc',
              'ppc64', 'x32','x64', 'x86', 'x86_64', 's390', 's390x')
```

ansible/plugins/inventory/nodejs_yaml.py
```diff
@@ -296,9 +300,8 @@ def has_metadata(info):
 
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser()
-    parser.add_argument('--list', action='store_true')
-    parser.add_argument('--host', action='store')
-    args = parser.parse_args()
-
+    # parser = argparse.ArgumentParser()
```
thefourtheye: Why are we checking in commented code?

jenkins/scripts/coverage/generate-index-html.py
```diff
@@ -6,9 +6,9 @@
 with open('out/index.csv') as index:
   index_csv = filter(lambda line: line, index.read().split('\n'))
 
+# noqa
```
thefourtheye: Is this comment incomplete?

cclauss: LGTM: This flake8 linter directive will silence the linter as discussed at 
http://flake8.pycqa.org/en/3.7.7/user/violations.html#in-line-ignoring-errors

cclauss:
@refack Can you please go through the comments (esp. https://github.com/nodejs/build/pull/1399#discussion_r266725917) and make any required updates.  It would be a big help to have [__pipenv__](https://docs.pipenv.org) in place.

Trott:
Rebase to eliminate conflicts and force-pushed. @cclauss @refack

Trott:
Rebased to eliminate conflicts and force-pushed. @cclauss @refack

cclauss:
My sense is that this can be closed unless @refack has objections.  I like the idea of automating the creation of virtualenvs for our builds to take place in and pipenv would deliver that.  However @sam-github and I agreed that there might be compatibility issues on non-mainstream Win/Mac/Linux so we should delay that experimentation until after we have Python 3 working.

Trott:
> My sense is that this can be closed unless @refack has objections. I like the idea of automating the creation of virtualenvs for our builds to take place in and pipenv would deliver that. However @sam-github and I agreed that there might be compatibility issues on non-mainstream Win/Mac/Linux so we should delay that experimentation until after we have Python 3 working.

In that case, I"m going to close this, but @refack or anyone else should feel absolutely free to re-open it if they think that's the wrong move.

PR #1401, opened 2018-07-12
WIP: load secrets on the fly

LucaLanziani:
This will only work with ansible >= 2.4

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -4,6 +4,7 @@ gathering           = explicit
 host_key_checking   = False
 inventory           = plugins/inventory/nodejs_yaml.py
 library             = plugins/library
+vars_plugins        = plugins/vars
 remote_user         = root
 retry_files_enabled = False
 roles_path          = roles
--- a/ansible/plugins/vars/load_secrets.py
+++ b/ansible/plugins/vars/load_secrets.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import os
+import tempfile
+import errno
+import subprocess
+
+# from ansible import __version__
+from ansible.errors import AnsibleParserError
+from ansible.module_utils._text import to_bytes, to_native
+from ansible.inventory.host import Host
+from ansible.inventory.group import Group
+from ansible.utils.vars import combine_vars
+
+try:
+    from ansible.plugins.vars import BaseVarsPlugin
+except ImportError:
+    class BaseVarsPlugin(object):
+        def __init__(self, args):
+            try:
+                from __main__ import display
+            except ImportError:
+                from ansible.utils.display import Display
+                display = Display()
+
+            display.warning("%s will not work if ansible < 2.4" % __name__)
+
+DECRYPT_TOOL = "gpg"
+FOUND = {}
+
+
+class VarsModule(BaseVarsPlugin):
+
+    def _can_i_run(self):
+        try:
+            subprocess.Popen([DECRYPT_TOOL, "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE).communicate()
+        except OSError as e:
+            if e.errno == errno.ENOENT:
+                self._display.warning("load_secret plugin cannot find %s executable" % DECRYPT_TOOL)
+            return False
+        else:
+            return True
+
+    def _decrypt_file(self, file_path, temp_plain_file):
+        self._display.display("Trying to decrypt %s" % file_path)
+        command = [DECRYPT_TOOL, "-q", "--decrypt", file_path]
+        process = subprocess.Popen(command, stdout=temp_plain_file, stderr=subprocess.PIPE)
+        _, stderr = process.communicate()
+        if process.returncode > 0:
+            self._display.warning("Decryption failed with error %s for file %s" % (process.returncode, file_path))
+            self._display.warning("Got following error while running '%s':" % " ".join(command[0:-1]))
+            self._display.warning(stderr)
+        return (process.returncode == 0)
+
+
+    def get_vars(self, loader, path, entities, cache=True):
+
+        data = {}
+        if not self._can_i_run():
+            return data
+
+        ''' parses the inventory file '''
+        if not isinstance(entities, list):
+            entities = [entities]
+
+        super(VarsModule, self).get_vars(loader, path, entities)
+        for entity in entities:
+            if isinstance(entity, Host):
+                secrets_dir = os.path.realpath(os.environ.get('NODE_BUILD_SECRETS', os.path.join(os.getcwd(), '../../secrets/build/')))
+                if not os.path.isdir(secrets_dir):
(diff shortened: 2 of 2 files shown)


ansible/ansible.cfg
```diff
@@ -4,6 +4,7 @@ gathering           = explicit
 host_key_checking   = False
 inventory           = plugins/inventory/nodejs_yaml.py
 library             = plugins/library
+vars_plugins        = vars_plugins
```
maclover7: Can you move `load_secrets.py` to inside `plugins/vars`

ansible/vars_plugins/load_secrets.py
```diff
@@ -0,0 +1,108 @@
...
+
+from ansible.inventory.host import Host
+from ansible.inventory.group import Group
+from ansible.utils.vars import combine_vars
+
+FOUND = {}
+
+
+class VarsModule(BaseVarsPlugin):
+
```
maclover7: ✂️

ansible/vars_plugins/load_secrets.py
```diff
@@ -0,0 +1,108 @@
...
+else:
+    BaseVarsPlugin = object
+
+from ansible.inventory.host import Host
+from ansible.inventory.group import Group
+from ansible.utils.vars import combine_vars
+
+FOUND = {}
+
+
```
maclover7: ✂️

ansible/vars_plugins/load_secrets.py
```diff
@@ -0,0 +1,108 @@
...
+        self._display.display("Trying to decrypt %s" % file_path)
+        temp_plain_file=tempfile.NamedTemporaryFile(dir="./host_vars", prefix="%s-" % entity.name, suffix='-tmp-secrets')
+        exit_code = subprocess.call([DECRYPT_TOOL, "cat", file_path], stdout=temp_plain_file)
+        if exit_code > 0:
+            self._display.warning("Decryption failed with error %s for file %s" % (exit_code, file_path))
+        return (exit_code, temp_plain_file)
+
+
+    def get_vars(self, loader, path, entities, cache=True):
+
```
maclover7: ✂️

ansible/vars_plugins/load_secrets.py
```diff
@@ -0,0 +1,108 @@
...
+            return data
+
+        ''' parses the inventory file '''
+        if not isinstance(entities, list):
+            entities = [entities]
+
+        super(VarsModule, self).get_vars(loader, path, entities)
+        for entity in entities:
+            if isinstance(entity, Host):
+                secrets_dir = os.path.realpath(os.path.join(os.getcwd(), '../../secret/build/'))
```
maclover7: Is it possible to make this configurable? For example, I store my Git repos locally as `~/code/nodejs-private/secrets` and `~/code/nodejs/build` 😬

LucaLanziani: @maclover7 mmm not sure, I can have a look, where would you store the configuration though?

ansible/ansible.cfg
LucaLanziani: yeah, much better place for the plugin, I'll do it

ansible/vars_plugins/load_secrets.py
maclover7: Maybe an environment variable? Not sure. More just thinking out loud, since we can't guarantee that everyone will have the same folder layout on their machines

joaocgreis:
<details>
<summary>My dotgpg is running very slow (looks like it is scanning all keys before deciding to use mine, not sure exactly what is wrong with it, I'll have to investigate). So I replaced it with gpg, leaving the diff here for reference.</summary>

```diff
diff --git a/ansible/plugins/vars/load_secrets.py b/ansible/plugins/vars/load_secrets.py
index 32ffe57..03995dd 100755
--- a/ansible/plugins/vars/load_secrets.py
+++ b/ansible/plugins/vars/load_secrets.py
@@ -29,7 +29,7 @@ except ImportError:
 
             display.warning("%s will not work if ansible < 2.4" % __name__)
 
-DECRYPT_TOOL = "dotgpg"
+DECRYPT_TOOL = "gpg"
 FOUND = {}
 
 
@@ -37,7 +37,7 @@ class VarsModule(BaseVarsPlugin):
 
     def _can_i_run(self):
         try:
-            subprocess.call([DECRYPT_TOOL], stdout=subprocess.PIPE)
+            subprocess.call([DECRYPT_TOOL, "--version"], stdout=subprocess.PIPE)
         except OSError as e:
             if e.errno == errno.ENOENT:
                 self._display.warning("load_secret plugin cannot find %s executable" % DECRYPT_TOOL)
@@ -48,7 +48,7 @@ class VarsModule(BaseVarsPlugin):
     def _decrypt_file(self, file_path, entity):
         self._display.display("Trying to decrypt %s" % file_path)
         temp_plain_file = tempfile.NamedTemporaryFile(dir="./host_vars", prefix="%s-" % entity.name, suffix='-tmp-secrets')
-        exit_code = subprocess.call([DECRYPT_TOOL, "cat", file_path], stdout=temp_plain_file)
+        exit_code = subprocess.call([DECRYPT_TOOL, "-q", "--decrypt", file_path], stdout=temp_plain_file)
         if exit_code > 0:
             self._display.warning("Decryption failed with error %s for file %s" % (exit_code, file_path))
         return (exit_code, temp_plain_file)
```

</details><br>

joaocgreis:
@LucaLanziani it would be good to adapt the README and add a note in `host_vars/README` that the files now live in secrets. Can you include this in this PR?

cc @nodejs/build - this makes running Ansible a lot easier

LucaLanziani:
@joaocgreis I'll add a section on the readme after work

@gibfahn @joaocgreis if everyone is happy with it I'm ok with replacing `dotgpg` with `gpg`, it should be more than fine for decrypting files.

gibfahn:
>@gibfahn @joaocgreis if everyone is happy with it I'm ok with replacing dotgpg with gpg, it should be more than fine for decrypting files.

Maybe do that as a separate commit so you can roll it back if others have concerns (hopefully it should just be a matter of applying @joaocgreis 's diff).

This looks really impressive by the way, thanks for doing it!

joaocgreis:
> Maybe do that as a separate commit so you can roll it back if others have concerns (hopefully it should just be a matter of applying @joaocgreis 's diff).

Sounds good to me. @LucaLanziani each `gpg` run sends a few lines to stderr, if you can redirect them and print only if `gpg` fails it would be great. If not, it is good enough as is, I don't think the lines will be a problem (and are legitimate warnings, even if not very relevant).

gibfahn:
> If not, it is good enough as is, I don't think the lines will be a problem (and are legitimate warnings, even if not very relevant).

Agree with @joaocgreis 100% here (as usual), less noisy output in the passing case is a nice-to-have, but don't spend time on it if it's non-trivial, that can easily be done in a follow-up PR.

LucaLanziani:
@joaocgreis @gibfahn you might like this better ;)

joaocgreis:
Landed in https://github.com/nodejs/build/commit/6046c1c03d411f2b1b4b0590227296580be7f534

Thanks @LucaLanziani!

PR #1402, opened 2018-07-18
www: letsencrypt config for dist.libuv.org

rvagg:
Existing `libuv.org` nginx config file isn't accurate for what's on the server at the moment, it looks like we (me probably) forgot to sync the two when we went https a while back. This update fixes that but there is a minor adjustment to what was on the server: port 80 doesn't just do an everything-redirect, there's a fall-through for `/.well-known/` because letsencrypt apparently likes to go straight for port 80 to verify (there may be a way around this but I couldn't find anything other than doing DNS validation). I don't think this is a big deal, we're serving `Strict-Transport-Security` anyway so browsers at least will always just jump to https.

For letsencrypt, I've opted to include instructions in doc form rather than ansible, it just gets too complicated, particularly since we have two servers this ansible setup gets applied to but only one needs the Certbot stuff. This is probably something to rectify when (if) we re-do the www ansible scripts.

So, https://dist.libuv.org/ is now serving a certificate generated from letsencrypt, its Comodo certificate expires next month so we'll let it go. We have another year on the nodejs.org certificate I think but since we front it with Cloudflare we need to come up with something more creative for that, or possibly just rely on Cloudflare's own issued certificates.

### DIFF
--- a/doc/non-ansible-configuration-notes.md
+++ b/doc/non-ansible-configuration-notes.md
@@ -2,6 +2,21 @@
 
 There are a number of infrastructure setup tasks that are not currently automated by Ansible, either for technical reasons or due to lack of time available by the individuals involved in these processes. This document is intended to collect that information, to serve as a task list for additional Ansible work and as a central place to note special tasks.
 
+## nodejs.org / web host
+
+For dist.libuv.org we use letsencrypt.org for HTTPS certificate. Use Certbot to register and generate a certificate on the main nodejs.org server; only the single server serves dist.libuv.org so this configuration is simple. Certbot sets up auto-renewal for the certificate in /etc/cron.d/certbot.
+
+```sh
+sudo apt-get install software-properties-common
+sudo add-apt-repository ppa:certbot/certbot
+sudo apt-get update
+sudo apt-get install python-certbot-nginx
+mkdir -p /home/libuv/www/.well-known/acme-challenge/
+certbot --manual certonly -d dist.libuv.org -m buil@iojs.org
+# take the CODE and the FILE and insert into the following command:
+echo "CODE" > /home/libuv/www/.well-known/acme-challenge/FILE
+```
+
 ## Windows (Azure/Rackspace)
 
 In order to get Windows machines to a state where Ansible can be run against them, some manual steps need to be taken so that Ansible can connect.
--- a/setup/www/resources/config/libuv.org
+++ b/setup/www/resources/config/libuv.org
@@ -3,12 +3,40 @@ server {
     listen [::]:80;
     server_name dist.libuv.org;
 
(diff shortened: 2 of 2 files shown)


doc/non-ansible-configuration-notes.md
```diff
@@ -2,6 +2,21 @@
...
+
+For dist.libuv.org we use letsencrypt.org for HTTPS certificate. Use Certbot to register and generate a certificate on the main nodejs.org server; only the single server serves dist.libuv.org so this configuration is simple. Certbot sets up auto-renewal for the certificate in /etc/cron.d/certbot.
+
+```sh
+sudo apt-get install software-properties-common
+sudo add-apt-repository ppa:certbot/certbot
+sudo apt-get update
+sudo apt-get install python-certbot-nginx
+mkdir -p /home/libuv/www/.well-known/acme-challenge/
+certbot --manual certonly -d dist.libuv.org -m buil@iojs.org
```
maclover7: `buil@iojs.org` --> `build@iojs.org`

cjihrig:
Is this related to https://github.com/nodejs/build/pull/1189?

rvagg:
thanks @cjihrig, yes you're right, I've closed that one and Ben's comment there is taken care of here already

rvagg:
merged 2cf8811

glensc:
i can confirm problem reported in https://github.com/libuv/libuv/issues/1775 is resolved (ipv6 urls did not work)

```
➔ wget http://dist.libuv.org/dist/v1.8.0/libuv-v1.8.0.tar.gz --no-hsts -6
--2018-08-01 10:40:05--  http://dist.libuv.org/dist/v1.8.0/libuv-v1.8.0.tar.gz
Resolving dist.libuv.org (dist.libuv.org)... 2604:a880:400:d0::b2c:a001
Connecting to dist.libuv.org (dist.libuv.org)|2604:a880:400:d0::b2c:a001|:80... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://dist.libuv.org/dist/v1.8.0/libuv-v1.8.0.tar.gz [following]
--2018-08-01 10:40:05--  https://dist.libuv.org/dist/v1.8.0/libuv-v1.8.0.tar.gz
Connecting to dist.libuv.org (dist.libuv.org)|2604:a880:400:d0::b2c:a001|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1042536 (1018K) [application/gzip]
Saving to: ‘libuv-v1.8.0.tar.gz’

libuv-v1.8.0.tar.gz                                         100%[==========================================================================================================================================>]   1018K  1,30MB/s    in 0,8s

2018-08-01 10:40:07 (1,30 MB/s) - ‘libuv-v1.8.0.tar.gz’ saved [1042536/1042536]
```

PR #1407, opened 2018-07-18
doc: Windows and Visual Studio supported versions

joaocgreis:
This is a summary of supported versions for Windows and Visual Studio. This matches what is in the main Node repo for each version, but given the complexity of the matrix it's good to have this summarized in some place.

This is somewhat similar to https://github.com/nodejs/build/pull/1149, but more focused on what we supported in the past. It may be good to merge both in the future, but for now it's probably better to let them mature separately.

cc @nodejs/build

### DIFF
--- a/doc/windows-visualstudio-supported-versions.md
+++ b/doc/windows-visualstudio-supported-versions.md
@@ -1,5 +1,9 @@
 # Windows and Visual Studio versions supported for Node.js
 
+**Notes:**
+- The exact Windows builds and Visual Studio releases are not tracked. Assume the latest or a recent version at the time of each commit.
+- Only 64 bit machines are available in the CI system. Thus, all 32 bit binaries are cross-compiled and tested on WoW64.
+
 ## For running Node.js
 
 Supported versions for running the Node.js installer and executable as released.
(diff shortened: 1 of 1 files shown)


refack:
A few questions:
1. Do we need to explicitly mention `VCBT2015`? (I don't remember if at the time  any differences were found)? Maybe move it to a footnote (`VS2015 & vs2017 support includes using the C++ Build Tools package`
2. RE https://github.com/nodejs/node/pull/21443 we only test on 64 bit hosts, so maybe add a comment to that effect.
3. VS2017 has a much more active release cycle, and there have been many public releases, so maybe add which version we use? Or suggest to always use latest version...

joaocgreis:
@refack I'd like to keep VCBT2015 separate there. Support was added separately from VS2015 and we use dedicated machines to test it. The VS2017 Build Tools have been almost issue-free (as other VS2017 editions), but VCBT2015 was more different from VS2015. I believe it's good to have it listed explicitly.

Updated about 2 & 3, PTAL.

joaocgreis:
Landed in https://github.com/nodejs/build/commit/639b55d6e7efb911898c2ada97d0b4099ee86d4f

Thanks!

PR #1409, opened 2018-07-22
WIP: jenkins: add selection for building on osx1011

mhdawson:


mhdawson:
Selector script does not seem to be getting the node version properly on the test machines, even though it seemed ok on the release machines: https://ci.nodejs.org/view/All/job/node-test-commit-osx-mdawson/1/console

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -61,8 +61,11 @@ def buildExclusions = [
   [ /sharedlibs_withoutssl/,          anyType,     lt(10)  ],
   [ /sharedlibs_shared/,              anyType,     lt(9)   ],
 
-  // -------------------------------------------------------
+  // OSX ---------------------------------------------------
+  [ /^osx1010/,                       anyType,     gte(11) ],
```
targos: Can it be `gte(10)`? There seems to be consensus in https://github.com/nodejs/node/pull/21883 that dropping support for 10.10 is not semver-major.

refack: Although it's not semver major, we tend not to "pull the rug under the feet" of current users without an actual need.
I'm +1 on landing as is, and re-evaluating if the need arises.

refack: (this is assuming the bump to V8 6.8 is not planned for v10)

refack: I'm now testing backward compatibility of binaries built with 10.11 on 10.10. If the tests pass, I'll be 100% with excluding 10.10 for node v10 as well.

refack:
For posterity: when enabling `VersionSelectorScript.groovy` also need to enable the `NodeVersionExtension` SCM plugin with `exportNodejsVersion` checked.

mhdawson:
Hmm, from what I can tell landing this may have caused us to no longer have releases on osx.

I was half way through testing and had not yet added the 1011 release machine to the infra.  Once this landed that would mean we would not build at all on osx for 11 and greater.

I'll add the machine in now.

mhdawson:
@refack the problem was that in the release ci the nightlies are built and are version 11.  Since the script landed and excluded building 11 on the existing release machines (which were only osx 1010) we've had no nightly builds.

I'm just in the process of adding the 1011 release machine, which I was going to do once I'd finished testing things out.

Issue #1253, opened 2018-04-27
Remove prebuilt node executables from test machines


joaocgreis:
@nodejs/build 

For all the machines that are used to run the sub-jobs of `node-test-commit`, we should not have any node executable in the path. If we do, we open up the possibility of testing the wrong executable. While this looks rather unlikely, the consequences can be severe and we would probably only catch this too late, after releasing.

The machines used for linting have node in the path, but are exclusive for linting and run no other jobs. Some jobs (like npm or node-gyp tests) require node, but only download it to the local workspace.

Currently some machines have node installed and are also used in `node-test-commit`:

- [ ] test-joyent-freebsd10-x64-1
- [ ] test-requireio_kahwee-debian8-arm64_pi3-1
- [ ] test-joyent-smartos15-x64-2
- [ ] test-joyent-smartos15-x64-1
- [ ] test-joyent-smartos16-x64-1
- [ ] test-joyent-smartos16-x64-2
- [ ] test-joyent-smartos14-x86-2
- [ ] test-joyent-smartos14-x86-1
- [ ] test-joyent-smartos17-x64-1
- [ ] test-joyent-smartos17-x64-2
- [ ] test-joyent-smartos14-x64-2
- [ ] test-joyent-smartos14-x64-1

Are there any legitimate use cases for having node in these machines? If so, I can change the Jenkins script to download a local node, or we can find a dedicated machine if needed. I'll start removing node from the machines above next week if there are no objections.

(Credit for @Trott who mentioned this in https://github.com/nodejs/node/pull/20322#issuecomment-384806639 . Thanks!)

Trott:
```console
root@test-digitalocean-ubuntu1604-x86-1:~# which node
/snap/bin/node
root@test-digitalocean-ubuntu1604-x86-1:~# /snap/bin/node -v
v9.11.1
root@test-digitalocean-ubuntu1604-x86-1:~#
```

According to @maclover7, that one might be a side effect of node-inspect CI runs.

joaocgreis:
Removed node from all the `joyent` machines but `test-joyent-smartos17-x64-1`, which does not seem to have the test key installed and I can't connect. Not sure who created it, @rvagg perhaps?

The ARM cluster is down at the moment, will remove from `test-requireio_kahwee-debian8-arm64_pi3-1` when it comes back.

joaocgreis:
[...]

PR #1416, opened 2018-07-25
ansible: ensure node is not in the path

joaocgreis:
Ensure node is not installed anywhere but the linter servers.

Ref: https://github.com/nodejs/build/issues/1253


joaocgreis:
Landed in https://github.com/nodejs/build/commit/29a8f29a03503f1e1d0d5830a667d2527780aa14

Thanks!

PR #1434, opened 2018-08-09
ansible: add fix playbook

maclover7:
Does some common cleanup tasks on a Jenkins worker machine (clean disk space, clean old processes, restarts the local Jenkins agent).

cc @gdams, let me know what I can do to get this setup in the Ansible Tower UI, this might be an ideal thing to try out with some non-Build WG nodejs/node collaborators

### DIFF
--- a/ansible/playbooks/jenkins/worker/fix.yml
+++ b/ansible/playbooks/jenkins/worker/fix.yml
@@ -1,0 +1,52 @@
+---
+
+#
+# Fix common problems with Jenkins workers
+#
+
+- hosts:
+  - test
+  - release
+  # Benchmarking machine
+  - infra-softlayer-ubuntu1404-x64-2
+
+  tasks:
+    # Clean disk space
+    - name: Get disk space
+      command: "df -h"
+      register: disk_usage
+    - set_fact:
+        disk_usage: "{{ disk_usage.stdout_lines }}"
+    - debug:
+        var: "{{ disk_usage }}"
+
+    - name: Get directories to delete
+      find:
+        paths: /home/iojs/build/workspace
+        patterns: "node-test-binary-arm,node-test-commit-*,libuv-test-commit-*"
+        recurse: no
+        file_type: directory
+      register: directories
+
+    - name: Clean disk space
+      file:
+        path: "{{ item.path }}"
+        state: absent
+      with_items: "{{ directories.files }}"
+
+    - name: Get disk space
+      command: "df -h"
+      register: disk_usage
+    - set_fact:
+        disk_usage: "{{ disk_usage.stdout_lines }}"
+    - debug:
+        var: "{{ disk_usage }}"
+
+    # Clean stale processes
+    - name: Clean stale processes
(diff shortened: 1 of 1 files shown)


maclover7:
Also cc @Trott, not sure if there are other commands you use to fixup machines that should be added here

gdams:
Does this need to be ran as root? It would be great if we could run this as the iojs user to take some of the risk away

mhdawson:
Looks good to me assuming we can run as iojs as suggested by @gdams

mhdawson:
On other suggestion which could be in a follow on PR is that we look to see if we can include cleanup of citgm directories as well. I've seen those beeing problems of out of space issues as well.

maclover7:
updated @mhdawson @gdams

ansible/playbooks/jenkins/worker/fix.yml
```diff
@@ -0,0 +1,60 @@
+---
+
+#
+# Fix common problems with Jenkins workers
+#
```
joaocgreis: Can you add a note here making it explicit that this cannot be run while Jenkins is active? Something like:
```
Please make sure no jobs are running on this worker before starting this
script. The worker must be marked as offline in Jenkins, or by any other
means it must not be possible to start a new job on this worker while this
script is running.
```

ansible/playbooks/jenkins/worker/fix.yml
```diff
@@ -0,0 +1,60 @@
...
+          register: disk_usage
+        - set_fact:
+            disk_usage: "{{ disk_usage.stdout_lines }}"
+        - debug:
+            var: "{{ disk_usage }}"
+
+        - name: Get directories to delete
+          find:
+            paths: /home/iojs/build/workspace
+            patterns: "node-test-binary-arm,node-test-commit-*,libuv-test-commit-*,citgm-*"
```
joaocgreis: Just delete everything inside the workspace. Perhaps we should make this explicit in the docs somewhere: there should be nothing of value anywhere in the workers, we should be able to delete and recreate workers at any time.

ansible/playbooks/jenkins/worker/fix.yml
```diff
@@ -0,0 +1,60 @@
...
+            recurse: no
+            file_type: directory
+          register: directories
+
+        - name: Remove directories
+          file:
+            path: "{{ item.path }}"
+            state: absent
+          with_items: "{{ directories.files }}"
```
joaocgreis: The best way to delete would be to move to a temporary directory in the same filesystem (inside the workspace should be safe), and then delete that directory. I've seen issues in the Pis where a file can't be deleted because of filesystem errors. Moving first at least ensures the original directories can be created again clean. I understand this requires some more work, so I'm just mentioning this here and completely ok to leave this to another PR sometime. Please don't block this PR on this.

refack: > I've seen issues in the Pis where a file can't be deleted because of filesystem errors.

That usually happens if there are hanging `node` processes (live or undead).
I support the move->delete strategy anyway.

sam-github:
Stale, closing, though if someone wants to pick it up again, it looks kindof useful.

PR #1436, opened 2018-08-09
ci-public: replace multijob with freestyle for `custom-suites`

refack:
I've configure a new job https://ci.nodejs.org/job/node-test-commit-custom-suites-freestyle/ (job XML attached).
This should replace https://ci.nodejs.org/job/node-test-commit-custom-suites/ in node-test-commit, as the current one is based on `multijob` which ignores the `PLATFORM_LABEL` parameter, and is unneccecerly combersome in this scenario.
Because of its limitations the current job is recursive (the flyweight job runs on a `jenkins-workspace` node and the actual build also runs on a `jenkins-workspace` node) so when the systems under load it can lead to unexpected behavior.

This is the diff from the current job https://www.diffchecker.com/MKqrFehH

### DIFF
--- a/jenkins/xml/node-test-commit-custom-suites-freestyle.xml
+++ b/jenkins/xml/node-test-commit-custom-suites-freestyle.xml
@@ -1,0 +1,208 @@
+<?xml version='1.1' encoding='UTF-8'?>
+<project>
+  <actions/>
+  <description></description>
+  <keepDependencies>false</keepDependencies>
+  <properties>
+    <com.coravy.hudson.plugins.github.GithubProjectProperty plugin="github@1.29.2">
+      <projectUrl>https://github.com/nodejs/node/</projectUrl>
+      <displayName></displayName>
+    </com.coravy.hudson.plugins.github.GithubProjectProperty>
+    <com.sonyericsson.rebuild.RebuildSettings plugin="rebuild@1.28">
+      <autoRebuild>false</autoRebuild>
+      <rebuildDisabled>false</rebuildDisabled>
+    </com.sonyericsson.rebuild.RebuildSettings>
+    <hudson.model.ParametersDefinitionProperty>
+      <parameterDefinitions>
+        <hudson.model.StringParameterDefinition>
+          <name>GITHUB_ORG</name>
+          <description>The user/org of the GitHub repo</description>
+          <defaultValue>nodejs</defaultValue>
+          <trim>true</trim>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>REPO_NAME</name>
+          <description>The name of the repo</description>
+          <defaultValue>node</defaultValue>
+          <trim>true</trim>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>GIT_REMOTE_REF</name>
+          <description>The remote portion of the Git refspec to fetch and test</description>
+          <defaultValue>refs/heads/master</defaultValue>
+          <trim>true</trim>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>REBASE_ONTO</name>
+          <description>Optionally, rebase onto the given ref before testing. Leave blank to skip rebasing.</description>
+          <defaultValue></defaultValue>
+          <trim>true</trim>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>POST_REBASE_SHA1_CHECK</name>
+          <description></description>
+          <defaultValue>After rebasing, check that the resulting commit sha1 matches the given one. If left blank, no check is performed.</defaultValue>
+          <trim>true</trim>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.BooleanParameterDefinition>
+          <name>IGNORE_FLAKY_TESTS</name>
+          <description>Mark the build as unstable instead of failure if only flaky tests fail</description>
+          <defaultValue>false</defaultValue>
+        </hudson.model.BooleanParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>CONFIG_FLAGS</name>
+          <description>Add arguments to ./configure</description>
+          <defaultValue></defaultValue>
+          <trim>false</trim>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>CI_JS_SUITES</name>
+          <description>Suites to run tools/test.py for, separated by spaces (e.g. internet)</description>
+          <defaultValue></defaultValue>
+          <trim>false</trim>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>TEST_ARGS</name>
+          <description>Additional arguments to pass to `python tools/test.py` (e.g. --worker)</description>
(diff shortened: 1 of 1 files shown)


jenkins/xml/node-test-commit-custom-suites-freestyle.xml
```diff
@@ -0,0 +1,208 @@
...
+        <hudson.model.StringParameterDefinition>
+          <name>REBASE_ONTO</name>
+          <description>Optionally, rebase onto the given ref before testing. Leave blank to skip rebasing.</description>
+          <defaultValue></defaultValue>
+          <trim>true</trim>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>POST_REBASE_SHA1_CHECK</name>
+          <description></description>
+          <defaultValue>After rebasing, check that the resulting commit sha1 matches the given one. If left blank, no check is performed.</defaultValue>
```
maclover7: Should this be the description?

jenkins/xml/node-test-commit-custom-suites-freestyle.xml
```diff
@@ -0,0 +1,208 @@
...
+  echo &quot;ok 1 # skip $SKIP_MESSAGE&quot; &gt;&gt; test.tap
+  echo &quot;  ---&quot; &gt;&gt; test.tap
+  echo &quot;  duration_ms: 4.925&quot; &gt;&gt; test.tap
+  echo &quot;  ...&quot; &gt;&gt; test.tap
+fi</command>
+    </hudson.tasks.Shell>
+    <hudson.tasks.Shell>
+      <command>curl https://raw.githubusercontent.com/nodejs/build/master/jenkins/scripts/node-test-commit-diagnostics.sh | bash -ex -s after</command>
+    </hudson.tasks.Shell>
+    <hudson.tasks.Shell>
```
maclover7: I guess it's okay here, but isn't this usually part of the "post build actions" section in the UI (just for consistency)?

jenkins/xml/node-test-commit-custom-suites-freestyle.xml
```diff
@@ -0,0 +1,208 @@
...
+          <trim>false</trim>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>TEST_ARGS</name>
+          <description>Additional arguments to pass to `python tools/test.py` (e.g. --worker)</description>
+          <defaultValue></defaultValue>
+          <trim>false</trim>
+        </hudson.model.StringParameterDefinition>
+        <org.jvnet.jenkins.plugins.nodelabelparameter.LabelParameterDefinition plugin="nodelabelparameter@1.7.2">
+          <name>PLATFORM_LABEL</name>
```
maclover7: Could be missing it, but where is this parameter used to actually filter the machines?

jenkins/xml/node-test-commit-custom-suites-freestyle.xml
refack: That's a copy&paste of the current desc so ¯\_(ツ)_/¯
Probably should be /s/check that/checks that/

jenkins/xml/node-test-commit-custom-suites-freestyle.xml
refack: I figured recently that it wasn't optimal to put it as a post-build action. I want to start moving them all to a last build stage.
But maybe it should be before the diags-after...
Also it could wait for a future iteration.

richardlau:
If this is for use in node-test-commit, does it need to be custom-suites with all the customisation, or should it be a node-test-worker (since currently the chained job is for running with `--worker`)?

refack:
> If this is for use in node-test-commit, does it need to be custom-suites with all the customisation, or should it be a node-test-worker (since currently the chained job is for running with --worker)?

It was designed for running customizable tests. We are already using it in three configurations:
1. to run the default suits with `--worker` - [example](https://ci.nodejs.org/job/node-test-commit-custom-suites/618/)
2. as part of `daily-master` to run the `internet` suite - [example](https://ci.nodejs.org/job/node-test-commit-custom-suites/611/)
3. as part of the `test-v8` job to test the `v8-update` suite - [example]( https://ci.nodejs.org/job/node-test-commit-custom-suites/614/)

So IMHO the conscumizations have proven very (re)useful.

refack:
Due to an [apparent degradation](http://logs.libuv.org/node-build/2018-08-10#10:17:34.019) of stability, [I've implemented the switch](https://ci.nodejs.org/job/node-test-commit/jobConfigHistory/showDiffFiles?timestamp1=2018-08-08_22-46-25&timestamp2=2018-08-10_09-54-22), and am monitoring the situation.

richardlau:
@refack I agree that customisations are very useful, especially for one of runs (e.g. like the stress job). My query is whether we want to be reusing the same job in this way as part of larger jobs/flows/pipelines. For example, it's harder to see the history of `--worker` runs because the job history will be a mixture of the three scenarios you have outlined (plus any manual runs).

refack:
> My query is whether we want to be reusing the same job in this way as part of larger jobs/flows/pipelines?

It's a balance of useability/reusability... This case is actually a good example, as if we had 3 seperate jobs it would have been harder to identify the issue and correct the configuration.
So IMHO for the time being using it for `--worker` is a good solution. But we should be exploring better code reuse solutions, such as project inheritance, managed scripts and shared groovy steps.

jenkins/xml/node-test-commit-custom-suites-freestyle.xml
maclover7: Ah, meant should this be in the description field, rather than defaultValue

refack: 🤦‍♂️

maclover7:
> My query is whether we want to be reusing the same job in this way as part of larger jobs/flows/pipelines? For example, it's harder to see the history of --worker runs because the job history will be a mixture of the three scenarios you have outlined (plus any manual runs).

To view the history of `--worker` builds, you can check out the node-test-commit-custom-suites part from https://nodejs-ci-health.mmarchini.me/#/detailed-jobs

richardlau:
@maclover7 but are all of those are `--worker` builds (e.g. https://ci.nodejs.org/job/node-test-commit-custom-suites/410/ isn't a `--worker` build)?

refack:
FTR the issue that the jobs were running on `jenkins-workspace` nodes, was because I did not RTFM and missed that there's a specific way to trigger such jobs. (That is most probably why we have been seeing very frequent infra related failure these last couple of days.
image

sam-github:
@richardlau is this still relevant? in-progress? already done some other way? Maybe its stale and should be closed, but its not immediately obvious, and it looks like you've refed it in an open issue.

richardlau:
https://ci.nodejs.org/job/node-test-commit-custom-suites/ is disabled in Jenkins and has a description saying to instead use https://ci.nodejs.org/job/node-test-commit-custom-suites-freestyle/ (incidentally I've just PR'ed a change to the COLLABORATOR_GUIDE (https://github.com/nodejs/node/pull/31557) in core to update the job link).

I think we can just close this out. I don't know what policy (or if we have one written down) covers what is backed up to https://github.com/nodejs/build/tree/master/jenkins/xml but it's hard to tell from a glance if the XML in this PR is current (i.e. has anyone changed the job configuration since). In any case backing up Jenkins changes from test and release CI is being handled by the proposed scripts @rvagg is working on in https://github.com/nodejs/build/issues/2133.

PR #1451, opened 2018-08-15
ansible: upgrade to OpenSSL 1.1.0i & 1.0.2p in sharedlibs

rvagg:
running upgrades now, it'll require restart of those containers but since CI is locked down that should be easy


rvagg:
pulled in OpenSSL 1.1.1-pre8 too, we're getting close to 1.1.1 final so it's time to start getting serious I reckon

refack:
So this makes for an interesting situation. ATM nothing in our repo passes CI (e.g. [v10.8](https://ci.nodejs.org/job/node-test-commit-linux-containered/6300/nodes=ubuntu1604_sharedlibs_openssl110_x64/), or master).
* We can land https://github.com/nodejs/node/pull/22318 and live with this.
* We can write sophisticated matrix filter.

/CC @Trott @joyeecheung @nodejs/testing

Trott:
Since nodejs/node#22318 is going out in a release Real Soon Now, any reason not to fast-track it and land it on master? It's already landed on `v10.9.0-proposal` branch.

refack:
FTR the expected failures are:
- ubuntu1604_sharedlibs_openssl110_x64
  - `parallel/test-crypto-scrypt`
  - `parallel/test-tls-passphrase`

PR #1476, opened 2018-08-27
add yamllint

gdams:
The Ansible playbooks are long overdue having some linting run on them, this PR adds yamllint (will require travis to be enabled on this repo).

Key changes:

- Enforce the use of `true`/`false` rather than `yes`/`no`
- Enforce spaces between braces (`{ blah }` rather than `{blah}`)
- Enforce `---` at the top of all yaml files


gdams:
@nodejs/build could I get a few more reviews before landing this please?

gdams:
@phillipj good spot! updated PTAL

ansible/roles/visual-studio/tasks/partials/vs2015.yml
```diff
@@ -8,7 +8,7 @@
   win_get_url:
     url: 'https://go.microsoft.com/fwlink/?LinkId=532606'
     dest: 'C:\TEMP\vs2015_community.exe'
-    force: no
+    force: false
```
joaocgreis: Indentation updates and `---` at the top sound great to me, but text should not change. Take this line for example, the Ansible documentation at https://docs.ansible.com/ansible/latest/modules/win_get_url_module.html clearly shows that the choices are `yes` and `no`.

gdams: so `yes` and `no` are always interchangeable with `True` and `False` and ansible is pushing for defaulting to `True`/`False` in all newer versions. I'm open to either but I feel that we should agree on a format to use

yamllint.yml
```diff
@@ -0,0 +1,6 @@
+---
+extends: default
+rules:
+  line-length: disable
+  braces: { min-spaces-inside: 1, max-spaces-inside: 1 }
+  brackets: { min-spaces-inside: 1, max-spaces-inside: 1 }
```
smpeters: If there are still some questions on true/false vs. yes/no, then maybe ignoring these issues would be the best for now.

```suggestion
  brackets: { min-spaces-inside: 1, max-spaces-inside: 1 }
  truthy: disable
```

sam-github:
Maybe worth resurrecting this?

richardlau:
> 
> 
> Maybe worth resurrecting this?

I'll resurrect this using GitHub Actions (assigned myself for now but I'll probably close and open a new PR).

sam-github:
Tried it out, mostly cosmetic or confusing comments from yamllint. Closing (though if anyone wants to take another shot, go for it).

PR #1505, opened 2018-09-26
setup: archive old setup scripts

refack:
Move `/setup/` to ~`/backup/archived-machine-setup/`~ `doc\reference-setup-scripts`

Fixes: https://github.com/nodejs/build/issues/1277


richardlau:
Why not branch/tag the repository and then delete the scripts instead of moving them if they are no longer used?

refack:
> Even if unusable, I try to keep `www` updated when changes are made, and I see at least @rvagg does as well. It's not great but it's the best place we have to document changes, so I'm reluctant to change it to a folder named `archive` that doesn't encourage changes.

I get what you are saying, and I agree. That was also my understanding of the situation, and indeed it should be reflected in folder names. I'll try to come up with alternative suggestions.

joaocgreis:
Having it in `doc` seems strange to me. I don't see a need to remove `setup/`, but I won't block this if you do.

It would be good to have explicit approval from @rvagg since he also uses this folder.

rvagg:
I might try and move `www` into somewhere in `ansible` because it is still relevant, I can't speak for aix and some other bits in there but we can probably just remove it all, git is our "archive"

sam-github:
It looks from conversation that there was some objections, and now it has no champion. Closing, but it can be reopened if someone wants to take it up again.

PR #1510, opened 2018-09-27
jenkins: refactor select-compiler.sh

refack:
Test job - https://ci.nodejs.org/job/test-selector/1/nodes=rhel72-s390x/console

Current script always used `$(python tools/getnodeversion.py)`.
I added an escape hatch by setting `NODEJS_MAJOR_VERSION`, there is no need for having an actual node src tree present (for example `citgm-nobuild`).

### DIFF
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -6,15 +6,20 @@ if [ "$DONTSELECT_COMPILER" != "DONT" ]; then
     *s390x* ) SELECT_ARCH=S390X ;;
     *aix* ) SELECT_ARCH=AIXPPC ;;
   esac
+  
+  # Get node version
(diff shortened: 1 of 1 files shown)


richardlau:
See https://github.com/nodejs/build/pull/1265 (although this PR addresses https://github.com/nodejs/build/pull/1265#discussion_r219523989).

refack:
> See #1265 (although this PR addresses [#1265 (comment)](https://github.com/nodejs/build/pull/1265#discussion_r219523989)).

😄 I wasn't around so much in May...

mhdawson:
I've not landed #1265 yet as I want to be really really sure it works properly as its used in release as soon as we land.  An earlier version would have silently build on the wrong compiler levels...

I'll try to schedule block of 2 hours next week to update 1265 and land.

mhdawson:
Closing since #1265 landed.

PR #1511, opened 2018-09-27
ansible: tweak macOS PATH

refack:
* Specify PATH explicitly so that it matches the environment
  when SSHing and running `sudo su iojs`


refack:
Otherwise there's a mess with where `python` is and how to pip install stuff to it.

gdams:
As a future note, we might be able to make use of the built in pathhelper tool `/usr/libexec/path_helper -s` to generate the PATH for us rather than having to hard code it.

Issue #1496, opened 2018-09-15
Keeping old binaries to allow resuming fanned jobs


joaocgreis:
Currently, we delete the temporary branch at the end of the fanned jobs. This means that when a fanned job is resumed, if the compile job succeeded only the test job will run again, and will not be able to find the temporary branch from last run.

I'd like to stop deleting temporary branches at the end of the runs, but this change is not urgent so it should be discussed with @nodejs/build first. Also, it would be good to do this when we have people available in the next few days to put out any fire.

At first, this sounds like the size of the binary repo could become a problem. However, git keeps references around long after the branches are deleted, so this might have a very small impact. Note that we have a job that runs daily and deletes branches for which the Jenkins job is no longer accessible, so branches will still get deleted after about one week.

<details>
<summary>Emergency cleanup of the binary repo</summary>

```
ssh jenkins-workspace-1
su binary_tmp -s /bin/bash
cd ~/binary_tmp.git
du -sh .
git prune
du -sh .
```

</details><br>

Trott:
> this change is not urgent 

If we can get this working before Code + Learn, that could be immensely helpful.

joaocgreis:
I've changed the jobs, will monitor the temporary repo.

For now the changes were minimal (changed the delete parameter to false). If all goes well, in a couple of months will remove the parameter and the build step.

joyeecheung:
https://ci.nodejs.org/job/node-test-commit-windows-fanned/20827/

https://ci.nodejs.org/job/node-test-binary-windows/COMPILED_BY=vs2017-x86,RUNNER=win2012r2,RUN_SUBSET=1/20072/console

```

10:25:29 c:\workspace\node-test-binary-windows>git checkout -f refs/remotes/jenkins_tmp/jenkins-node-test-commit-windows-fanned-20827-bin-win-vs2017-x86 
10:25:29 error: pathspec 'refs/remotes/jenkins_tmp/jenkins-node-test-commit-windows-fanned-20827-bin-win-vs2017-x86' did not match any file(s) known to git.
```

This job started 6hrs ago.  Looks like it did not resume successfully

refack:
> I've changed in to `${JOB_NAME}-${GIT_COMMIT}`

I reverted that because now `node-test-commit-windows-fanned` is not a git job...

joaocgreis:
[...]

PR #1515, opened 2018-10-03
Updates for jenkins-workspace binary temporaries repository

joaocgreis:
This does two things:
- Increase MaxSartups in `sshd_config`. This is necessary when many workers running tests try to access the temp repo to download binaries simultaneously. Not sure if there is an optimal value for this. Since the previous value seemed to just barely not be enough, I increased it 10x considering it is possible that more than one job might be started at the same time.
- Weekly clean-up of the repository. This deletes branches where the last commit is more than one month old. This is not ideal, but since https://github.com/nodejs/build/issues/1496 the branches no longer have the number of the job that started them, so looking at Jenkins jobs is no longer possible. This replaces `git-clean-temp-repo`, which was broken not only because of this but also because unauthenticated access to Jenkins is disabled.

### DIFF
--- a/ansible/roles/jenkins-workspace/files/clean_binary_tmp.sh
+++ b/ansible/roles/jenkins-workspace/files/clean_binary_tmp.sh
@@ -2,12 +2,13 @@
 
 cd ~binary_tmp/binary_tmp.git
 (echo; date) >> ~binary_tmp/clean_binary_tmp.log
+du -sh ~binary_tmp/binary_tmp.git/ >> ~binary_tmp/clean_binary_tmp.log
 
 git fetch origin +master:master
 
 for b in $(git branch | sed /\*/d); do
-  if [ -z "$(git log -1 --since='4 weeks ago' -s $b)" ]; then
-    (git branch -D $b || true) |& tee -a ~binary_tmp/clean_binary_tmp.log
+  if [ -z "$(git log -1 --since='7 days ago' -s $b)" ]; then
+    (git branch -D $b |& tee -a ~binary_tmp/clean_binary_tmp.log) || true
   fi
 done
 
--- a/ansible/roles/jenkins-workspace/files/clean_binary_tmp.sh
+++ b/ansible/roles/jenkins-workspace/files/clean_binary_tmp.sh
@@ -1,0 +1,16 @@
+#!/bin/bash -ex
+
+cd ~binary_tmp/binary_tmp.git
+(echo; date) >> ~binary_tmp/clean_binary_tmp.log
+
+git fetch origin +master:master
+
+for b in $(git branch | sed /\*/d); do
+  if [ -z "$(git log -1 --since='4 weeks ago' -s $b)" ]; then
+    (git branch -D $b || true) |& tee -a ~binary_tmp/clean_binary_tmp.log
+  fi
+done
+
+git prune
+
+du -sh ~binary_tmp/binary_tmp.git/ >> ~binary_tmp/clean_binary_tmp.log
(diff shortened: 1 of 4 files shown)


refack:
Question: any reason not to use Jenkins' built-in cron (just so that we don't have configuration in too many places)?
image

joaocgreis:
@refack was there before, I removed it because the job is broken. Deleting the job and archiving as XML here is the next step there.

ansible/roles/jenkins-workspace/files/clean_binary_tmp.sh
```diff
@@ -0,0 +1,16 @@
...
+
+git fetch origin +master:master
+
+for b in $(git branch | sed /\*/d); do
+  if [ -z "$(git log -1 --since='4 weeks ago' -s $b)" ]; then
+    (git branch -D $b || true) |& tee -a ~binary_tmp/clean_binary_tmp.log
+  fi
+done
+
+git prune
```
refack: maybe add a ```nohup git gc >& ~binary_tmp/git_gc_`date +"%s"`.log ```?
That way we won't get an auto `gc` in the middle of a future `fetch`

ansible/roles/jenkins-workspace/files/clean_binary_tmp.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -ex
+
+cd ~binary_tmp/binary_tmp.git
+(echo; date) >> ~binary_tmp/clean_binary_tmp.log
+
+git fetch origin +master:master
+
+for b in $(git branch | sed /\*/d); do
+  if [ -z "$(git log -1 --since='4 weeks ago' -s $b)" ]; then
```
refack:
```suggested
  if [ -z "$(git log -1 --since='5 days ago' -s $b)" ]; then
```
IMO 5 days is enough. After five days we should run a rebuild not a resume (since `master` probably progressed).

rvagg: agreed, 4 weeks is a long time, any good reason for that @joaocgreis?

mhdawson: I find that after a week the jobs are most often no longer in the history anyway.  Maybe 7 days?

ansible/roles/jenkins-workspace/files/clean_binary_tmp.sh
joaocgreis: Please see below about gc.

ansible/roles/jenkins-workspace/files/clean_binary_tmp.sh
joaocgreis: 7 days sounds reasonable, updated.

joaocgreis:
Updated (9b6f4b03aa2ff8f00c2c52a4056170a73b36a761), PTAL.

- Changed the expiry time to 7 days.
- Changed the script to run daily.
- Disabled git auto gc, since it can interfere with fetching from test jobs. I don't think there's a need to run the full gc in the script, since repacking does not make sense for this usage and we already do the pruning explicitly. The full `git gc` takes hours to complete, so I've let it out, but we can add it to the script if there's a need for it.
- Changed logging to not include bash output and added another `du` in the beginning.

joaocgreis:
Landed in https://github.com/nodejs/build/compare/a1957ef2c75ad576f5932adc2ed32852b4e58df3...5961359b794190575ebc53b304e93d38ef25c552

PR #1547, opened 2018-10-25
ansible: add atspi-2 to s390x and ppc64le

gdams:
This is required for V8 7.1 https://github.com/nodejs/node/pull/23423

### DIFF
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -70,6 +70,10 @@ packages: {
     'gcc-c++,sudo',
   ],
 
+  rhel72_s390x: [
+    'at-spi2-core-devel',
+  ],
+
   smartos: [
     'gccmakedep',
     'git',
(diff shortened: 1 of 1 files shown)


ansible/roles/baselayout/vars/main.yml
```diff
@@ -70,6 +70,10 @@ packages: {
     'gcc-c++,sudo',
   ],
 
+  rhel72_s390x: [
+    'at-spi2-core-devel',
+  ],
```
richardlau: Are these additive (i.e. on `rhel72_s390x` both `rhel72` and `rhel72_s390x` apply) or does one take preference over the other?

Asking because the comment on L34 says:
> ```
> # you can either add os family or os to this list (see smartos)
> # but the playbook chooses os over family - not both
> ```

gdams: I just copied the example on https://github.com/nodejs/build/blob/master/ansible/roles/baselayout/vars/main.yml#L43

sam-github:
I lack context on this. It hasn't landed, so it doesn't seem to be required, but I could misunderstand.

@miladfarca @mhdawson @richardlau  do any of you see this as useful to get landed, or is it stale and should be closed?

miladfarca:
Doesn't seem like it's needed anymore, `atspi2`  as a requirement was modified/moved in this CL: https://chromium-review.googlesource.com/c/chromium/src/+/1356979

PR #1552, opened 2018-10-30
www: letsencrypt --manual doesn't allow auto-renew, use --nginx

rvagg:
found this in my commit queue, discovered a few weeks back when the dist.libuv.org cert didn't renew. It turns out that `--manual` doesn't allow for auto-renewal with certbot, you have to use `--nginx`, which is unfortunate, but it works and is apparently safe for our config.


rvagg:
Made some changes here, switched from `certonly` to `run`, it works quite well and I haven't had any troubles lately so it's safe(ish).

Also added `--no-redirect` to avoid a prompt.

Also added commands for the iojs.org domains that are now using letsencrypt (some discussion @ https://github.com/nodejs/build/pull/1557).

PR #1557, opened 2018-11-01
Redirect iojs.org to nodejs.org

fhemberger:
Using full domain redirects, because the little amount of traffic still pointing
to iojs.org probably doesn't justify full url mapping redirects.

Ref: https://github.com/nodejs/iojs.org/issues/432

Open issues:

- [ ] Remove DNS entry for roadmap.iojs.org  
  (linking to https://github.com/nodejs/roadmap/tree/gh-pages)
- [ ] `*.iojs.org` also resolves to GitHub pages
- [ ] What should happen with `https://iojs.org/dist/`?
Drop it entirely? Integrate it somehow into `https://nodejs.org/dist/`?
- [ ] What about tools like `nvm` or `n`, which still point to `dist`?

### DIFF
--- a/setup/www/resources/config/iojs.org
+++ b/setup/www/resources/config/iojs.org
@@ -2,97 +2,17 @@ server {
     listen *:80;
     server_name iojs.org www.iojs.org;
 
-    return 301 https://iojs.org$request_uri;
+    return 301 https://nodejs.org;
 }
 
 server {
     listen [::]:443 ssl http2;
     listen *:443 ssl http2;
-    server_name www.iojs.org;
+    server_name iojs.org www.iojs.org;
 
     ssl_certificate ssl/iojs_chained.crt;
     ssl_certificate_key ssl/iojs.key;
     ssl_trusted_certificate ssl/iojs_chained.crt;
 
-    return 301 https://iojs.org$request_uri;
-}
-
-server {
-    listen [::]:443 ssl http2;
-    listen *:443 ssl http2;
-    server_name iojs.org;
-
-    ssl_certificate ssl/iojs_chained.crt;
-    ssl_certificate_key ssl/iojs.key;
-    ssl_trusted_certificate ssl/iojs_chained.crt;
-
-    keepalive_timeout 60;
-    server_tokens off;
-
-    resolver 8.8.4.4 8.8.8.8 valid=300s;
-    resolver_timeout 10s;
-
-    add_header Strict-Transport-Security max-age=63072000;
-    add_header X-Frame-Options DENY;
-    add_header X-Content-Type-Options nosniff;
-
-    access_log /var/log/nginx/iojs/iojs.org-access.log nodejs;
-    error_log /var/log/nginx/iojs/iojs.org-error.log;
-
-    gzip on;
-    gzip_static on;
-    gzip_disable "MSIE [1-6]\.";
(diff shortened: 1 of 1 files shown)


fhemberger:
What should happen with `https://iojs.org/dist/`?
Drop it entirely? Integrate it somehow into `https://nodejs.org/dist/`?

fhemberger:
@Trott Any thoughts about the open questions? Should we just ditch all iojs downloads completely (while we still keep stuff like Node.js 0.1.10 around)?
How will this affect 3rd-party-tools?

Trott:
> @Trott Any thoughts about the open questions? Should we just ditch all iojs downloads completely (while we still keep stuff like Node.js 0.1.10 around)?
How will this affect 3rd-party-tools?

/ping @nodejs/version-management

refack:
> For nvm to work properly, it needs nodejs.org/dist to _not_ contain io.js versions, and iojs.org/dist to _only_ contain io.js versions.

What I meant by integrating, is that we do just that, i.e. keep:
```yaml
    location /dist {	
        alias /home/dist/iojs/release/;	
        autoindex on;	
        default_type text/plain;	
         location ~ \.json$ {	
            add_header access-control-allow-origin *;	
        }	
    }	
```

ljharb:
As long as all the same versions are accessible, and show up in distinct lists depending on the domain requested, everything will continue to work.

fhemberger:
Updated the PR to keep all `location` directives in place. PTAL.

rvagg:
I'm not sure the config matches the discussion here. Can we be more precise with what we're trying to achieve given the above discussion?

As far as the website is concerned, it's very simple and only has localisation directories plus an /images. Nothing deeper than that.

We also have locations that point to distributables, they are: /dist, /download, /api, and /docs.

My suggestion would be to redirect everything but those distributables, leaving them intact. Each of them lets you access specific versions and there's probably links out there on the web to them and people using weird versions that it might not be worth disrupting.

Use of assets in /images might also be a concern. We've tried really hard to not break historical links on nodejs.org since we started the Foundation, maybe that should extend here too?

Aside from the question of /images. If the redirect goes only in `location /`, the rest could stay intact.

FYI here's the DNS records for iojs.org, and also we actively use @iojs.org for emails.

iojs_dns

fhemberger:
@rvagg Ok, so basically leave everything as it is (redirect http+www -> https, etc.) and only redirect `/` (as more specific location paths still have a higher rank).

I updated the PR accordingly, PTAL.

Removing the `*` CNAME record should also be fine to remove `roadmap.iojs.org` as well (added redirect in nginx config).

rvagg:
I think this looks good. Thankfully iojs.org isn't critical in the same way as nodejs.org so I'll be able to deploy and test it out. Our SSL cert expires this week so I'll need to sort that out too, either cloudflare or letsencrypt.

rvagg:
switched to letsencrypt on iojs.org

rvagg:
All done, landed in 7a7e700c, I reworked some of the config to make nginx work and also activate ipv6 on some of the `server` stanzas (letsencrypt verifies via AAAA records when they exist).

Some letsencrypt config details for this added to https://github.com/nodejs/build/pull/1552

rvagg:
and I removed the `*` -> gh-pages DNS record and put `roadmap.iojs.org` as a CNAME for `iojs.org` so the redirect works there too.

Issue #1540, opened 2018-10-19
request: a server for Gzemnid


ChALkeR:
[Gzemnid](https://github.com/nodejs/Gzemnid) is currently hosted on my personal VPS, which is limited in resources and unoptimal — and that's what currently limits improving it further. Also, virtually no one uses it from there (even though I tried setting up access once).

Now, with lz4 compression fully supported (since recently), I can outline the required specs for the server to move that to a foundation-owned server.

Performing a move would both remove the need in me hosting that and provide the room for future improvements (e.g. translate that to a close to real-time follower with rebuilds and a proper web interface for contributors once I or someone else would have time to implement that).

# CPU

A 4-core CPU is preferred if we would do both the UI and the scheduled rebuilds on the server.

No hard speed requirements, faster CPU would mean faster rebuilds and queues.
Storage uses compression, and a significant part of operations are CPU-bound.

# Memory

Something around 4 GiB should be fine, more could be used for caches.

<details>
Rebuild consumes around 2 GiB for a single task, and we still want the UI.
I am not sure if more would be needed for complex tasks, though, but I assume that could be solved later, if  needed.
</details>

# Fast storage (SSD)

Around 120 GiB should be fine here, but that could grow.
Faster is better here.

<details>
Fast storage is the data that is constantly needed for queues and rebuilds.

| Title | Purpose | Size |
| --- | --- | --- |
| slim.code | code search | 12 GiB |
| slim.topcode | partial code search | 1.2 GiB |
| deps | dependency stuff | 0.3 GiB |
| files | file listings | 0.2 GiB |
| byField | [system] packages listing from npm | 0.2 GiB |
| meta | [system] meta info about packages | 4 GiB |
| partials | [system] slim.code partials | 60 GiB |
| OS | operating system | 2 GiB |

Total: ~80 GiB.

As the ecosystem grows exponentially, I expect an under ~1.5 times growth in a year or so, so we should probably plan storage resources according to that.
</details>

# Slow storage (possibly network)

* ~530 GiB — if we don't have fast access to an npm mirror, or
* ~100 GiB — if we have fast access to an npm mirror.

Could be merged with «fast storage», but also could be stored separately, even on a remote network drive.
50-100 Mbps would be enough.

<details>

| Title | Purpose | Size |
| --- | --- | --- |
| current | current versions of npm packages | 285 GiB |
| historic | historic versions of slim.code/deps — to check how the ecosystem evolves | 45 GiB |

Total: ~330 GiB.

[...]

PR #1565, opened 2018-11-06
added infra-digitalocean-ubuntu1804-x64-1 gzemnid

rvagg:
as per #1540


refack:
I see it (at least) has nginx. Should we have a playbook for it?

PR #1571, opened 2018-11-14
jenkins: remove deprecated NOP target lint-md-build

thefourtheye:
I noticed this for the first time in https://ci.nodejs.org/job/node-linter/1598/console


refack:
> On second thought I could add `|| true` to the CI script.

I wasn't paying attention, I thought you're eliminating the target from the Makefile.
In the pipeline file it's still required.

thefourtheye:
> We still need that in CI since we also test old versions 🤷‍♂️

Makes sense, I thought it was left during a cleanup.

PR #1579, opened 2018-11-20
groovy: re-include centos5-32 for release builds

refack:
Refs: http://logs.libuv.org/node-build/2018-11-20#18:24:29.343

/CC @MylesBorins


refack:
Assumptions were wrong

PR #1598, opened 2018-11-23
Create VersionSelectorScript.md

tiagorvmartins:
First draft version taking into account the exclusion rules of VersionSelectorScript.groovy


tiagorvmartins:
I still have some changes to do

Issue #1522, opened 2018-10-10
chore: list compiler versions used to build each node version on each platform


refack:
(for new contributors, this one might require some GitHub archaeology)

tiagorvmartins:
Hi @refack do you think I can help out in this issue? Thanks :)

refack:
Hello @tiagorvmartins, you sure could give it a try 👍 

I'll give you some leads.
1. We started the based for this kind of list at https://github.com/nodejs/build/blob/master/doc/windows-visualstudio-supported-versions.md
2. You can see the output of our testing CI runs, for example https://ci.nodejs.org/job/node-daily-master/1361/, more specifically https://ci.nodejs.org/job/node-test-commit-osx/22932/nodes=osx1011/consoleText
   Most of those have a call to print the compiler version (or `xcode_version`).
3. The platforms in CI are selected according to this filtering script https://github.com/nodejs/build/blob/master/jenkins/scripts/VersionSelectorScript.groovy (notice it's an exclusion rules)
4. We setup our platforms with `Ansible` and the scripts are in https://github.com/nodejs/build/tree/master/ansible

I think if you can do this for node 11, and 12 (which is the current `master`) that would be amazing.

tiagorvmartins:
Thanks for the info @refack , seems quite hard for me for a starting point (first contribution on github here xD), but after a quick investigation can you just clarify me that I only need to update the VersionSelectorScript.groovy with the new hosts from the first list you noted? Or does it involve more work beside it? Thanks!

refack:
FTR:

Some more references:

* https://github.com/nodejs/build/commits/master/setup (this is a history of our old machine setup scripts)
* https://github.com/nodejs/build/commits/master/ansible (this is a history of our current setup scripts)
* https://github.com/nodejs/build/issues/259
* https://github.com/nodejs/build/pull/809
* https://github.com/nodejs/build/issues/1580 (wrong assumption, but has some history)
* https://github.com/nodejs/node/pull/16969
* https://github.com/nodejs/node/pull/14004
* https://github.com/nodejs/node/issues/8172
* https://github.com/nodejs/node/commits/master/BUILDING.md
* https://github.com/nodejs/build/commits/master/ansible/roles/package-upgrade/tasks/partials/brew.yml
* https://github.com/nodejs/build/pull/971


More references about the minimal supported platforms, but we sometimes test & release with newer compilers.

[...]

PR #1599, opened 2018-11-23
Create VersionSelectorScript.md

tiagorvmartins:
First version of the documentation showing the list of compiler versions used to build each node version on each platform

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.md
+++ b/jenkins/scripts/VersionSelectorScript.md
@@ -1,0 +1,97 @@
+### The current file list the compiler versions used to build each node version on each platform:
+
+#### Linux
+
+node 6,7 => built on **centos5**
+
+node 8,9,10,11 => built on **centos6**
+
+node 6,7,8,9 => built on **centos[67]-(arm)?(64|32)-gcc48**
+
+node 10,11 => built on **centos[67]-(arm)?(64|32)-gcc6**
+
+node 6,7,8,9 => built on **centos6-32-gcc6**
+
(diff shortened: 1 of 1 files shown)


rvagg:
@tiagorvmartins I wonder if you could use my script in https://github.com/nodejs/build/issues/1522#issuecomment-441331423 to make this automatically? I don't really like that this is manual and will need to be edited in parallel with VersionSelectorScript.groovy and it'd be nice to have an easy way for this to be generated and updated—automatic would be best but I don't have any bright ideas for achieving that yet.

I have the same problem with a `NODE_MODULE_VERSION` registry I'm trying to make @ https://github.com/nodejs/node/pull/24114. Where parallel documentation that isn't integrated into the single .json file is going to get out of sync really easily. For that case I'm going to put descriptive metadata straight into the .json file so you're forced to see and edit it if you're touching the file.

refack:
> I don't really like that this is manual and will need to be edited in parallel with VersionSelectorScript.groovy and it'd be nice to have an easy way for this to be generated and updated—automatic would be best but I don't have any bright ideas for achieving that yet.

The problem I asked @tiagorvmartins to help solve, is to have *some sort* of human readable documentations.
We could refactor `VersionSelectorScript.groovy` to include all this info, but there is still several gaps, like which compiler version each platforms uses, and what it used for test and what is used for release.

sam-github:
Cleaning up stalled PRs, they can be reopened if necessary.

Also, I don't see this as maintainable. It duplicates the actual config in .groovy, and the painstakingly maintained BUILDING.md in nodejs/node.

PR #1608, opened 2018-11-28
jenkins: add an llhttp builder to the mix for 11+

rvagg:
I've added this tag already, just not build config, need this first


rvagg:
I think we're now default llhttp so maybe it doesn't matter anymore

PR #1618, opened 2018-11-30
jenkins: add deleted git-clean-tmp-repo job

joaocgreis:
This job no longer works because Jenkins requires a token. Archiving here since this can still be useful in the future.

Replaced by: https://github.com/nodejs/build/pull/1515


joaocgreis:
Landed in https://github.com/nodejs/build/commit/bcc589f90e2ff722508e78e28f0b3e7c64d1af73

PR #1624, opened 2018-12-06
Change how secrets are loaded to only one file per level

joaocgreis:
Instead of using `host_files`, this PR adds support to have one public and three private inventory files (one for each access level). This centralizes the secret information in the minimum possible amount of files, making it easier to manage.

Having the information in this format will also make it easier to manage Windows secrets. I will move the Windows entries to the new format, and this will allow to extract RDP connection information and remove the current duplication (it's not currently committed, but usernames/passwords/ports/etc have to be copied to several places when working with Windows hosts).

This goes together with https://github.com/nodejs-private/secrets/pull/35.

I've left the `wip` commits to make this easier to review. Will squash when landing.

@LucaLanziani this removes the file you committed, but uses parts of it and the work done there made this change much easier. Would be great to have your review here if you have the time!

cc @nodejs/build


rvagg:
so this essentially makes host_vars redundant?

and is this happy on both python 2 and 3?

joaocgreis:
Rebased. Kept the commit structure, will squash into two commits when landing.

joaocgreis:
Thanks for the review @mhdawson, landed in https://github.com/nodejs/build/compare/b70ff37441e3967a6705f2e220d1a22ede55203f...5f16061ec7a633b17cd7009d7c70773ebf4dcb52

@rvagg I assume it's ok to move this forward, but please let me know if you find any issue later.

PR #1627, opened 2018-12-09
initial ansible support for aix71

gdams:
TODO:
- [x] Install Java if it doesn't already exist

### DIFF
--- a/ansible/roles/jenkins-worker/vars/main.yml
+++ b/ansible/roles/jenkins-worker/vars/main.yml
@@ -5,6 +5,7 @@
 #
 
 init: {
+  aix: 'aix71',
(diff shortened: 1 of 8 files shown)


gdams:
@nodejs/build please review

ansible/roles/jenkins-worker/vars/main.yml
```diff
@@ -97,6 +104,7 @@ java_path: {
 
 # same for bash. will default to /bin/bash if you don't set it
 bash_path: {
+  aix: '/usr/bin/ksh',
```
richardlau: Would things be simpler if this actually pointed to `bash` instead of `ksh`?

PR #1630, opened 2018-12-10
Fix VersionSelectorScript.groovy on non-release CI

richardlau:
Windows labels on https://ci.nodejs.org/ are of the form `*-vs201?`, 

e.g. `win10-vs2017`.
image

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -34,10 +34,10 @@ def buildExclusions = [
   [ /^cross-compiler-armv[67]-gcc-4.9/, anyType,   lt(10)  ],
 
   // Windows -----------------------------------------------
-  [ /^vs2013-/,                       anyType,     gte(6)  ],
-  [ /^vs2015-/,                       anyType,     lt(6)   ],
-  [ /^vs2015-/,                       anyType,     gte(10) ],
-  [ /^vs2017-/,                       anyType,     lt(10)  ],
+  [ /vs2013/,                       anyType,     gte(6)  ],
+  [ /vs2015/,                       anyType,     lt(6)   ],
+  [ /vs2015/,                       anyType,     gte(10) ],
+  [ /vs2017/,                       anyType,     lt(10)  ],
 
   // SmartOS -----------------------------------------------
   [ /^smartos14/,                     anyType,     gte(8)  ],
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -34,10 +34,10 @@ def buildExclusions = [
   [ /^cross-compiler-armv[67]-gcc-4.9/, anyType,   lt(10)  ],
 
   // Windows -----------------------------------------------
-  [ /vs2013/,                       anyType,     gte(6)  ],
-  [ /vs2015/,                       anyType,     lt(6)   ],
-  [ /vs2015/,                       anyType,     gte(10) ],
-  [ /vs2017/,                       anyType,     lt(10)  ],
+  [ /vs2013/,                         anyType,     gte(6)  ],
+  [ /vs2015/,                         anyType,     lt(6)   ],
+  [ /vs2015/,                         anyType,     gte(10) ],
+  [ /vs2017/,                         anyType,     lt(10)  ],
 
   // SmartOS -----------------------------------------------
   [ /^smartos14/,                     anyType,     gte(8)  ],


richardlau:
I have no idea how to go about testing this (short of creating a temporary job, which I don't think I can do by myself). I also have no visibility of the release CI (which I assume this script is working for and therefore assume uses labels which are the other way around (VS version first)).

richardlau:
I noticed this with 
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/1677/console
which didn't exclude `win-vs2015` from Node.js 11.

targos:
Would it help if I give you a screenshot of the labels present in the release ci?

richardlau:
> 
> 
> Would it help if I give you a screenshot of the labels present in the release ci?

Possibly? I mean it still won't help test the proposed change, but might be good for auditing (unless there's some reason that even the labels in use on the release CI should remain private). 

I guess what we're looking for is that the removed characters from the regular expression will still match the labels they were matching before and not falsely match any additional labels afterwards.

richardlau:
(As an aside I'm sure that I read somewhere a comment/issue about the Windows labels in the release and non-release CI's not matching but I can't find it.)

targos:
Labels in ci-release are of the form `vs201?-*`. For example `vs2015-x64`.

joaocgreis:
@richardlau tested in https://ci.nodejs.org/view/All/job/reis-citgm-smoker/1/console, looks good, you have write access to that job if you want to try anything. I see you don't have commit access to this repo, I'll land this after 2 days if there are no objections. I don't think this will cause any issues in release-ci, I'll start some test builds after landing to make sure.

Actually, using the VersionSelectorScript for citgm does not give it full coverage, but since it's the simplest we have and won't run on unsupported systems, it's a good way to do this for now. The VersionSelectorScript is not to filter out unsupported platforms, but to select a single platform to build the official release. That is, it follows the last table ("Official Releases") in https://github.com/nodejs/build/blob/master/doc/windows-visualstudio-supported-versions.md, when for citgm it would be better to follow the one above ("For building Node.js Addons").

richardlau:
> 
> Actually, using the VersionSelectorScript for citgm does not give it full coverage, but since it's the simplest we have and won't run on unsupported systems, it's a good way to do this for now. The VersionSelectorScript is not to filter out unsupported platforms, but to select a single platform to build the official release. That is, it follows the last table ("Official Releases") in https://github.com/nodejs/build/blob/master/doc/windows-visualstudio-supported-versions.md, when for citgm it would be better to follow the one above ("For building Node.js Addons").

Ah but the `citgm-smoker` job also builds Node.js (i.e. this job is used to run citgm against Node.js pull requests) so probably should run on the same platforms as the official release? 

The `citgm-smoker-nobuild` job uses prebuilt Node.js binaries and should probably be using the other list as you point out. Perhaps we can add an `addonsType` build type in a follow up PR?

joaocgreis:
To be clear, this PR is good as it is.

An `addonsType` sounds quite good for a future PR. Ideally, the citgm job would cross-compile, build with all supported versions for building and test with all supported Windows and VS versions. But that would also use way to much resources, so I think we have a good balance as it is for `citgm-smoker`. It would be good to use this script in `node-compile-windows` and `node-test-binary-windows`, so I also have this in my queue to take a better look at some time.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -34,10 +34,10 @@ def buildExclusions = [
   [ /^cross-compiler-armv[67]-gcc-4.9/, anyType,   lt(10)  ],
 
   // Windows -----------------------------------------------
-  [ /^vs2013-/,                       anyType,     gte(6)  ],
-  [ /^vs2015-/,                       anyType,     lt(6)   ],
-  [ /^vs2015-/,                       anyType,     gte(10) ],
-  [ /^vs2017-/,                       anyType,     lt(10)  ],
+  [ /vs2013/,                       anyType,     gte(6)  ],
```
rvagg: Tiny nit, 2 extra spaces before `anyType` to make them line up. The alignment is intended to make it easier to visually parse

joaocgreis:
Landed in https://github.com/nodejs/build/commit/8eab9cb2bba9055124518b334fb0f6518a48491c

Will confirm that there are no issues with release Jenkins.

joaocgreis:
`master`: https://ci-release.nodejs.org/job/iojs+release/4057/console
`v6.x`: https://ci-release.nodejs.org/job/iojs+release/4058/console

Looks as expected.

PR #1648, opened 2018-12-19
Use intermediate variables to remain DRY

cclauss:
Easier to read and more difficult to mess up.


rvagg:
this is fine, except that it doesn't even work anymore since anonymous access to jenkins was turned off, so we either need to delete this or make it work again with github login or token

PR #1649, opened 2018-12-20
Fix/nodejs.org language fallback

fhemberger:
@rvagg Sorry, missed the second part for the content redirect, first commit only serves English 404-fallback.


fhemberger:
Can this be merged?

Also, someone from build might want to look at https://github.com/nodejs/nodejs.org/pull/1940 (which would make this PR superfluous, but will have an impact on website build time).

PR #1652, opened 2018-12-27
Generate Remmina configuration for Windows RDP access

joaocgreis:
Depends on https://github.com/nodejs/build/pull/1624 and includes a squashed version of that PR. Please review only the commits after it, for now https://github.com/nodejs/build/commit/a10086f6a3a4c7538fef9c0568c90dee3c238e73.

This generates configuration files to be used by [Remmina](https://remmina.org/) to connect to our Windows machines using RDP. This works with both Python 2 and 3. I will follow up with documentation and an inventory update when possible, for now this already works when `rdp_port` is set for a machine. This will allow us to remove the Windows specific files from secrets.

Another option would be to generate native Windows RDP files, but password management is not so straightforward and doing everything from a single Unix machine is probably better for most Build WG members. Anyway, nothing prevents us from having both in the future.


joaocgreis:
Thanks, landed in https://github.com/nodejs/build/commit/0f2b367325ee6f4383b64ffb109a388796ac5589

PR #1661, opened 2019-01-15
WIP ansible: add role to install ci-health-dashboard on github-bot server

mmarchini:
I haven't tested it yet, but this should work.

I still have to update https://github.com/mmarchini/nodejs-ci-health-dashboard to accept credentials in the format provided by `envs.jenkins_api_credentials`.

Also, we should probably use nginx as a proxy instead of exposing the port directly. There's a nginx role already, but I think it is specific to the website. Should I try to reuse the existing `nginx` role, or is it ok to install nginx as a dependency in this new `ci-health-dashboard` role?

cc @maclover7 @rvagg

### DIFF
--- a/ansible/playbooks/create-github-bot.yml
+++ b/ansible/playbooks/create-github-bot.yml
@@ -13,6 +13,7 @@
     - bootstrap
     - package-upgrade
     - github-bot
+    - ci-health-dashboard
   pre_tasks:
     - name: check if secrets are properly set
       fail:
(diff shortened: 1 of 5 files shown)


rvagg:
@mmarchini whoops! I just noticed this, sorry. How are we going with this? Re nginx I think you should probably do nginx from scratch in here, the one for the website is very particular for the website. You can copy config files, letsencrypt etc. if you want though but I think you'll end up with something very different for this purpose.

sam-github:
Seems pretty stale, so I'm closing, but if its still active, or should be, anyone is welcome to reopen or ask me to reopen it.

PR #1671, opened 2019-01-25
jenkins: add patch to work around 24966

mhdawson:
Patch required to work around problem affecting
coverage builds.  24966 is not related to coverage
but a more general issue that manifests in the
coverage builds.

We need this in CI to be able to run coverage on more than one machine which we are trying to do as part of adding a coverage check to the main PR regression job.


mhdawson:
@gdams could you review/approve so I can land this?

sam-github:
@mhdawson rubber stamped it... do you still want to land, or should it close?

mhdawson:
The job https://ci.nodejs.org/view/Node.js%20Daily/job/node-test-commit-linux-coverage-daily

still has

```
# work around ISSUE #24996
git apply /home/iojs/cov-patch-24966
```

Not sure if its still required or not though.  Given that there is work towards a different coverage implementation I'd say we could just close.  It may not longer be required and if we are going to EOL that job in the nearish future no point investing time to figure that out.

Issue #1674, opened 2019-01-30
Add Python 3 to more build machines


cclauss:
Continue the process started in #1644 of putting both Python 2 and Python 3 on various build machines so that we can perform tests and experimental builds on the Python 3. 
https://github.com/nodejs/node/pull/25759#issuecomment-458146087

/R = @rvagg

cclauss:
On those build machines where Python 3 is already installed, we need to do __python3 -m pip install --upgrade pipenv".  See test failures on nodejs/node#25878

refack:
Need to add to ansible recipes (https://github.com/nodejs/build/search?p=3&q=python&type=&utf8=%E2%9C%93)
Might be a good chance to refactor common prerequisites.

cclauss:
Can we please get this resolves and closed? 303 daze left...

rvagg:
some help getting ansible working properly with python3 would be great, I can't run it on my mac anymore for machines that need ccache and have to resort to linux where it uses python2. The line causing problems is https://github.com/nodejs/build/blob/master/ansible/roles/baselayout/tasks/ccache.yml#L13 and there's a similar one in ansible/roles/jenkins-worker/tasks/partials/docker-host.yml.
Something about jinja2 processing of `"{{ ccache_html_content.stdout | regex_findall('ccache-[0-9]+.[0-9]+(?:.[0-9]+)*.tar.gz') | map('regex_replace', 'ccache-') | map('regex_replace', '.tar.gz') | list | latest_version }}"` the chaining of the maps seems to be the problem and I don't have the error at hand. If needed I could probably generate the error again.

refack:
@rvagg does https://github.com/nodejs/build/pull/1399 help?

(BTW this has the new "npm" like pipenv work flow, `git fetch`, `pipenv install`, `pipenv shell`)

cclauss:
Should we close this one or is this still an issue?

refack:
We still need Python3 on _more_ machines.

sam-github:
Is https://github.com/nodejs/build/issues/1674#issuecomment-459875569 needed on the AIX machines? If so, after updating them all to have python3 available, https://github.com/nodejs/build/issues/1775#issuecomment-492803274, I could do it.

sam-github:
- `test-linuxonecc-rhel72-s390x-*`: needs python3, and it isn't in the default repo.
[...]

PR #1679, opened 2019-02-02
build: add python3-pip to ubuntu.Dockerfiles

cclauss:
https://pypi.org/project/tap2junit says that __tap2junit__ is compatible with legacy Python but says nothing about Python 3 compatibility.  Related to #1674

### DIFF
--- a/ansible/roles/docker/templates/ubuntu1604.Dockerfile.j2
+++ b/ansible/roles/docker/templates/ubuntu1604.Dockerfile.j2
@@ -21,9 +21,12 @@ RUN apt-get update && apt-get dist-upgrade -y && apt-get install -y \
     openjdk-8-jre-headless \
     curl \
     python-pip \
+    python3-pip \
     libfontconfig1
 
 RUN pip install tap2junit
+
+RUN pip3 install tap2junit
 
 RUN addgroup --gid {{ server_user_gid.stdout_lines[0] }} {{ server_user }}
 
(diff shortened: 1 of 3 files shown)


thefourtheye:
cc @nodejs/build

cclauss:
@refack @rvagg @richardlau Is this PR going in the right direction or should I close it?

rvagg:
I have no idea @cclauss, you tell me, what does this even do? What does it mean to have both `python-pip` and `python3-pip` and to install `tap2junit` with both? What will that do when it's invoked by different versions of Node? I can merge this and rebuild the images but I'd be doing it completely blind to the consequences and I don't like having to scramble to fix something that's broken because I didn't understand it.

cclauss:
__Python 2__ and __Python 3__ have two _separate_ copies of __pip__.  These are usually called __pip2__ and __pip3__ but not always so a safe approach is to do __python 2 -m pip__ and __python3 -m pip__ to ensure that you are getting a hold of the pip that corresponds to the Python you are using.

TL;DR They maintain _separate_ "site-packages" directories so that a module like __tap2junit__ needs to be installed into both if we want that module to be accessible to both Py2 and Py3.

On my machine I get:
```
$ python2 -m pip --version
pip 19.1.1 from /usr/local/lib/python2.7/site-packages/pip (python 2.7)
$ python3 -m pip --version
pip 19.1.1 from /usr/local/lib/python3.7/site-packages/pip (python 3.7)
```
We will need to have a similar set up on all boxes where we want to test both Pythons.

rvagg:
should be deployed across all the containers now

PR #1694, opened 2019-02-19
jenkins: backup benchmark-node-micro-benchmarks-compare

refack:
AFAICT it never worked.
https://ci.nodejs.org/job/benchmark-node-micro-benchmarks/ is used instead.

### DIFF
--- a/jenkins/xml/benchmark-node-micro-benchmarks-compare.xml
+++ b/jenkins/xml/benchmark-node-micro-benchmarks-compare.xml
@@ -1,0 +1,206 @@
+<?xml version='1.1' encoding='UTF-8'?>
+<project>
+  <actions/>
+  <description>We want to compare two branches, tags or commits.</description>
+  <keepDependencies>false</keepDependencies>
+  <properties>
+    <hudson.security.AuthorizationMatrixProperty>
+      <inheritanceStrategy class="org.jenkinsci.plugins.matrixauth.inheritance.InheritParentStrategy"/>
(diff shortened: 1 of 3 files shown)


mhdawson:
@refack can you fill in any context.  As far as I know the job was working fine for people (unless there is a different one that lets people run the microbenchmarks) and there was even requests to extend it.

There seem to be a few successful runs as well>

refack:
> unless there is a different one that lets people run the microbenchmarks

That what I think is going on.
This job was called `benchmark-node-micro-benchmarks-compare` and it was never used, and AFAICT never worked.
The one that is used and works Ok is the one I linked in the OP - https://ci.nodejs.org/job/benchmark-node-micro-benchmarks/ - which I assume is the one you are looking at.

mhdawson:
@refack ok that makes sense to me.  My understanding based on the last discussion around  https://ci.nodejs.org/job/benchmark-node-micro-benchmarks/ was that instead of a new job we could just add options to the existing one.  We can always create a new job if needed as well.

refack:
Added two other unused jobs.

P.S. @mhdawson is this https://ci.nodejs.org/view/Node.js%20Daily/job/node-test-commit-linux-coverage-new/jobConfigHistory/ still needed?

mhdawson:
@refack I think we can get rid of s https://ci.nodejs.org/view/Node.js%20Daily/job/node-test-commit-linux-coverage-new/jobConfigHistory/

sam-github:
stale, and @AshCripps recently cleaned out unused jobs.

PR #1720, opened 2019-03-12
docker: upgrade to openssl-1.1.1b for sharedlibs

richardlau:
Refs: https://github.com/nodejs/node/pull/26209#issuecomment-472047076

cc @sam-github

### DIFF
--- a/ansible/roles/docker/templates/ubuntu1604_sharedlibs.Dockerfile.j2
+++ b/ansible/roles/docker/templates/ubuntu1604_sharedlibs.Dockerfile.j2
@@ -31,15 +31,15 @@ RUN addgroup --gid {{ server_user_gid.stdout_lines[0] }} {{ server_user }}
 
 RUN adduser --gid {{ server_user_gid.stdout_lines[0] }} --uid {{ server_user_uid.stdout_lines[0] }} --disabled-password --gecos {{ server_user }} {{ server_user }}
 
-ENV OPENSSL102DIR /opt/openssl-1.0.2q
+ENV OPENSSL102DIR /opt/openssl-1.0.2r
 
-RUN mkdir -p /tmp/openssl_1.0.2q && \
-    cd /tmp/openssl_1.0.2q && \
-    curl -sL https://www.openssl.org/source/openssl-1.0.2q.tar.gz | tar zxv --strip=1 && \
+RUN mkdir -p /tmp/openssl_1.0.2r && \
+    cd /tmp/openssl_1.0.2r && \
+    curl -sL https://www.openssl.org/source/openssl-1.0.2r.tar.gz | tar zxv --strip=1 && \
     ./Configure shared linux-x86_64 --prefix=$OPENSSL102DIR -fPIC && \
     make -j 6 && \
     make install && \
-    rm -rf /tmp/openssl_1.0.2q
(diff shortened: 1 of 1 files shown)


rvagg:
LGTM but @richardlau would you mind doing a bump for 1.0.2r while you're in there?

FYI this requires stopping, deleting and rebuilding docker images on all 4 docker hosts. It needs to be automated better because it involves `systemctl stop` for all of the jenkins jobs running containers, then `docker` cleanup commands and then a rerun of the ansible scripts. It's so tedious in fact that a few times I've had to do this I've just reprovisioned or re-imaged the hosts and started from scratch!

I guess docs would help too eh?

richardlau:
> 
> 
> LGTM but @richardlau would you mind doing a bump for 1.0.2r while you're in there?

Done!

rvagg:
I'm just going to take them all offline by disabling node-test-commit-linux-containered and using parallel-ssh to do the cleanup in batch then ansible them back again

rvagg:
here's what I've done after taking them offline (/tmp/docker-hosts has the 4 docker hosts listed in it):

```
parallel-ssh -i -h /tmp/docker-hosts 'systemctl list-units -t service --plain --all jenkins* | grep jenkins-test | awk '\''{print $1}'\'' | xargs -l sudo systemctl stop'
parallel-ssh -i -h /tmp/docker-hosts 'systemctl list-units -t service --plain --all jenkins* | grep jenkins-test | awk '\''{print $1}'\'' | xargs -l sudo systemctl disable'
parallel-ssh -i -h /tmp/docker-hosts 'sudo rm /lib/systemd/system/jenkins-test-*'
parallel-ssh -i -h /tmp/docker-hosts 'sudo systemctl daemon-reload'
parallel-ssh -i -h /tmp/docker-hosts 'sudo systemctl reset-failed'
parallel-ssh -i -h /tmp/docker-hosts 'sudo docker system prune -fa'
```

Then I did some upgrades and reboots and in the process test-softlayer-ubuntu1604-docker-x64-1 got bricked and now it's sitting in limbo. I requested a re-image but even that's stuck. I tried to request a new one in a different DC but that seems to be in limbo too.. But in the meantime I also:

* brought test-joyent-ubuntu1604-docker-x64-1 and fixed up the inventory (PR incoming) because it was wrong
* removed Alpine 3.7 and added 3.9 and shuffled alpine-last and alpine-last-latest (coming in same PR)

Then I've ansibled the 4 machines, one not present but the other 3 are all set up again and running. The workspaces and ccache still exists, I haven't blown that away, but the OpenSSL updates and Alpine 3.9 will require some warm-up time to get back to normal speed across the board.

Will keep an eye on the SoftLayer portal for movement on those machines but for now it's out of action. If/when it comes back it'll need a reprovision.

refack:
wait... if `test-softlayer-ubuntu1604-docker-x64-1` is down, how is it possible that [test-softlayer-ubuntu1604_sharedlibs_container-x64-2](https://ci.nodejs.org/computer/test-softlayer-ubuntu1604_sharedlibs_container-x64-2/) is working 😕

rvagg:
@refack because it came back up after I posted that and I ran ansible on my computer and walked away. FYI the old server came back up and I've deleted it now so we have a fresh one there running. PR incoming.

PR #1722, opened 2019-03-13
[WIP]ansible: low hanging GCC-8 fruit

refack:
* CentOS7 - `devtoolset-8`
* Ubuntu - `gcc-8`

### DIFF
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -41,9 +41,9 @@ packages: {
 
   centos7_arm64: ['git'], # git2u not available for aarch64 (yet)
   centos7_x64: ['git2u','centos-release-scl',], # centos-release-scl is required to enable SCLo
-                                               # but we do it manually in partials/repo/centos7.yml for arm64
+                                                # but we do it manually in partials/repo/centos7.yml for arm64
   centos7: [
-    'ccache,gcc-c++,devtoolset-6,sudo',
+    'ccache,gcc-c++,devtoolset-6,sudo,devtoolset-8',
   ],
 
   aix: [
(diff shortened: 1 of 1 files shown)


rvagg:
with CentOS7 that'll have to be an opt-in at the Jenkins level, for Ubuntu will that work across all our infra? Do we have the toolchain ppa setup by default? And then, when enabled it'll switch to GCC 8 by default and that'll begin removing testing of older GCC, are we OK with that?

refack:
> with CentOS7 that'll have to be an opt-in at the Jenkins level, for Ubuntu will that work across all our infra? Do we have the toolchain ppa setup by default? And then, when enabled it'll switch to GCC 8 by default and that'll begin removing testing of older GCC, are we OK with that?

Optimally they should live SxS, so I'd like to set up select-compiler-per-node-version for Ubuntu as well.

(for CentOS we already have an if-elif for devtoolset-6)
image

rvagg:
ok, so this is blocked until we have something in place to do side-by-side? and `apt-get install gcc gcc-8` is that going to install gcc-8 only or both gcc-8 and the default gcc? I have a feeling that since gcc-8 "provides" gcc that we'll only get 8, so it won't be side-by side unless we specify the version for each ubuntu.

refack:
> ok, so this is blocked until we have something in place to do side-by-side?

Ack.

> and `apt-get install gcc gcc-8` is that going to install gcc-8 only or both gcc-8 and the default gcc?

As far as I've tested (16.04 and 18.04), both, since `gcc` is hard-aliased to a specific version:
https://packages.ubuntu.com/xenial/gcc
https://packages.ubuntu.com/disco/gcc

sam-github:
Stale.

Issue #1543, opened 2018-10-21
compilers: install GCC 8 /CLANG 7 (as V8 does)


refack:
[Chromium and V8 are actually build](https://chromium.googlesource.com/chromium/src/tools/clang/+/a245b955fe9cd620081ed267fae303c88d033fef/scripts/update.py#71) with a very recent revision off the LLVM dev branch (svn revision 344066 which is ver 8 that is in progress)

GCC 8.1 has been available since May 2018 (http://gnu.mirrors.pair.com/gcc/gcc-8.1.0/) with 8.2 the latest.
CLang 7 has been available since Sep 2018 (http://releases.llvm.org/).
We are already using a very advanced MSVC compiler.

Using the latest compiler includes the latest optimizers, which is free performance gain. It also enables new C++ language fatures which can make the code simpler, and safer.

As I see it we need to make sure we can do this, then recommend an update to core.

/CC @nodejs/build @nodejs/release

refack:
> Is this for 11? Otherwise what is the impact for native add-on building if introduced in 10

This will probably be for 12.

hashseed:
Whether it's Node 11 or Node 12 is not particularly relevant for V8, since we take a snapshot of the master branch from time to time to use for our CI.

mhdawson:
@gdams FYI, need to figure out plan for newer compiler on IBM platforms.

rvagg:
I think we should probably start making moves in this direction for `master` for Node 12. #1583 is about release builds, I think devtoolset-7 might make sense on CentOS 7 for that. But we'll need to assess what we can do across our Linux test cluster, I'm sure there are going to be awkward edges.

For instance, Debian Stable (stretch) doesn't come with gcc-8, it's queued in Testing & Unstable. Thankfully Ubuntu 18.04 has it though.

refack:
> For instance, Debian Stable (stretch) doesn't come with gcc-8, it's queued in Testing & Unstable. Thankfully Ubuntu 18.04 has it though.

I'm more concerned with the IBM machines. It might be worth considering switching to clang for those, as it seems LLVM is easier to bootstrap.

rvagg:
clang might be easier in lots of places, https://apt.llvm.org/ is pretty helpful. We'd need to balance that against needing to have gcc compilability tested though unless we're prepared to say gcc isn't a Tier 1 supported compiler.

hashseed:
We still have a bot that tests building with GCC. We would love to get rid of it though :)

[...]

PR #1723, opened 2019-03-14
ansible: update minimum gcc on linux-ppc64le to gcc-6

sam-github:


ansible/roles/baselayout/vars/main.yml
```diff
@@ -106,6 +106,6 @@ packages: {
   ],
 
   ubuntu1404: [
-    'ntp,g++-4.8,gcc-4.8,g++-4.9,gcc-4.9,binutils-2.26',
+    'ntp,g++-4.8,gcc-4.8,g++-4.9,gcc-4.9,binutils-2.26,gcc-6,g++-6',
```
refack: AFAICT GCC-8 is available, so that's preferable, as per https://github.com/nodejs/build/issues/1543, and latest discussion.

jenkins/scripts/select-compiler.sh
```diff
@@ -20,7 +20,12 @@ if [ "$SELECT_ARCH" = "PPC64LE" ]; then
 
   echo "Setting compiler for Node version $NODEJS_MAJOR_VERSION on ppc64le"
 
-  if [ "$NODEJS_MAJOR_VERSION" -gt "9" ]; then
+  if [ "$NODEJS_MAJOR_VERSION" -gt "11" ]; then
+    # XXX(@sam-github) Why? I didn't need binutils in path to build/link/test
+    # node, the system default of 2.24 worked fine for me.
+    export PATH=/usr/lib/binutils-2.26/bin/:$PATH
```
refack: Maybe it's required only for a subset of platforms (also IIUC GCC > 5 is better at backwards compatibility)? Let's try without it.

ansible/roles/baselayout/vars/main.yml
sam-github: Latest build WG discussion was 6.x is target, or so @mhdawson said. If there isn't consensus on that @refack, could you comment in #1543, to keep the discussion centralized?

jenkins/scripts/select-compiler.sh
mhdawson: I'd prefer to figure that out as a separate step. We needed to do it separately when we introduced the 4.9.4 compiler and I don't want to break anything. @refack if you want we can make it conditional on PPC if you disagree?

ansible/roles/baselayout/vars/main.yml
mhdawson: See my comment on #1543, Tegardless of whether GCC-8 is available we still need to test with GCC-6 to make sure we are supporting the minimum.  In the case of PPC we don't have enough existing machines to double up on the testing (which as discussed in the build wg meeting we already do for x86 so it is feasible there) to test GCC-6 and GCC-8 initially.  Therefore, for the 12.x release of PPC we will quite likely test and release on GCC-6.  There is some chance we can do better depending on how quickly we can get our centos machines up and running but I don't want to block or depend on that for 12.X.

Having said that, @sam-github if it's just a matter of adding gcc-8, and g++-8 to the list it might make sense to do that so that we have them installed for when we look to do more?

mhdawson: The other thing I'll add is that we don't have the developer toolset for PPC for the existing machines and without that I'm not confident a binary built with 8 would run on a system that only has 6.x which is another thing we'd have to convince ourselves is safe.  With the developer tools (as we have on x86) I'm more convinced it will just work.  We may figure out the developer toolset on the centos machines for PPC  but again I'm not sure we will get that time for 12.x

jenkins/scripts/select-compiler.sh
refack: Ack.

ansible/roles/baselayout/vars/main.yml
refack: Ok. So lets land this as is, and I'll test compatibility of GCC-8 built binaries on a machine that does not have it.

refack:
Actually let's wait with this as little bit. I want to test GCC-8 binaries on machines with only 4.9.

mhdawson:
Pushed a commit. When testing on a new PPC machine the dependencies were not quite right between roles. The new commit fixes that.

jenkins/scripts/select-compiler.sh
```diff
@@ -20,7 +20,12 @@ if [ "$SELECT_ARCH" = "PPC64LE" ]; then
 
   echo "Setting compiler for Node version $NODEJS_MAJOR_VERSION on ppc64le"
 
-  if [ "$NODEJS_MAJOR_VERSION" -gt "9" ]; then
+  if [ "$NODEJS_MAJOR_VERSION" -gt "11" ]; then
```
mhdawson: We've been building with binutils-2.26 for 9+ so I think we should continue to do that, we might adjust for 11+ but I'm not sure it's worth it.

We should remove the Why? comment through or have an issue at least to take an action if we believe we need to.  My understanding from Rod is that it was needed to ensure we get OpenSSL assembly acceleration.

mhdawson:
Create a new machine (test-osuosl-ubuntu1404-ppc64_le-4), configure it with this script and test job to ensure builds ok without changes to select-compiler.

https://ci.nodejs.org/job/node-test-commit-plinux-mdawson-test/nodes=test-osuosl-ubuntu1404-ppc64_le-4/1/console

mhdawson:
@refack in terms of testing GCC 8 binaries on machine with only 4.9.4 one thing to keep in mind is that it depends on what symbols are used from the libraries that come with the compiler. Even it things run today (I think we've seen problems in the past) there is nothing stopping the new of new symbols in the future which might break the backwards compatibility.  The great thing about the RHEL developer toolset is that it understands the deltas introduced in newer gcc versions and statically instead of dynamically links (based on my understanding) avoiding that problem.  From the IBM perspective I understand that our binaries might not be quite as fast having been built with an older gcc, but I value the stability/compatibility of using the older compiler more until we can get the developer toolset in place.

refack:
> depends on what symbols are used from the libraries that come with the compiler.

A minimal sanity test is checking that the libraries have a minimal version. The error we're seeing now is:
```
./node
/lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by ./node)
```
while the library does export it's minimal supported version:
```
strings /lib64/libstdc++.so.6 | grep LIBCXX
GLIBCXX_3.4
GLIBCXX_3.4.1
GLIBCXX_3.4.2
GLIBCXX_3.4.3
GLIBCXX_3.4.4
GLIBCXX_3.4.5
GLIBCXX_3.4.6
GLIBCXX_3.4.7
GLIBCXX_3.4.8
GLIBCXX_3.4.9
GLIBCXX_3.4.10
GLIBCXX_3.4.11
GLIBCXX_3.4.12
GLIBCXX_3.4.13
GLIBCXX_3.4.14
GLIBCXX_3.4.15
GLIBCXX_3.4.16
GLIBCXX_3.4.17
GLIBCXX_3.4.18
GLIBCXX_3.4.19
GLIBCXX_LDBL_3.4
GLIBCXX_LDBL_3.4.7
GLIBCXX_LDBL_3.4.10
GLIBCXX_DEBUG_MESSAGE_LENGTH
```

refack:
Another simple option is to simply statically link `libstdc++`, and in that case IIUC we are only dependent on the system's `libc`

mhdawson:
@refack back when we ramped up support for PPC we explored the available options including statically linking. Unless we do that for all platforms (x86) included we did not feel it was a good option.   I think the developer toolset has progressed and has better platform support (as does our options for OS's at OSU) but as mentioned I don't think we can count on figuring that out before 12.X goes out.

mhdawson:
@sam-github noticed one other problem in that after setup on a new machine the wrong gcc is the default.  We need to figure out how to update the ansible script to fix that

mhdawson:
I think the select compiler script will still do the right thing if the default is wrong, but it would be better if the default was set properly for cases were we have not properly configured the jobs.

refack:
> @sam-github noticed one other problem in that after setup on a new machine the wrong gcc is the default. We need to figure out how to update the ansible script to fix that

@rvagg noticed that too in https://github.com/nodejs/build/pull/1722#issuecomment-472631934 I think this will require us to use a compiler selector for all platforms.

(I'm now thinking that we should start storing that information in the node repo and read it for each checkout)

mhdawson:
Build with PR being used to select the compiler as gcc-6 : https://ci.nodejs.org/job/node-test-commit-plinux-mdawson-test/nodes=test-osuosl-ubuntu1404-ppc64_le-4/2/

sam-github:
re: https://github.com/nodejs/build/pull/1723#issuecomment-473063761

I don't see why the default matters, we explicitly set the compiler version with select-compiler.sh. If someone forgets to use select-compiler.sh, they will get the wrong compiler for some cases (it will be wrong for 9.x+, or wrong for 11.x+, but can't be correct for both).

I've no idea how to change the default. I'll check, but I suspect it depends on the order in which packages are installed, which depends on current machine state, so is indeterminant.

ansible/roles/baselayout/vars/main.yml
```diff
@@ -106,6 +106,6 @@ packages: {
   ],
 
   ubuntu1404: [
-    'ntp,g++-4.8,gcc-4.8,g++-4.9,gcc-4.9,binutils-2.26',
+    'ntp,gcc-6,g++-6,g++-4.8,gcc-4.8,g++-4.9,gcc-4.9,binutils-2.26',
```
sam-github: switching order might make the last installed the default, but if gcc-4.9 is already installed, YMMV

jenkins/scripts/select-compiler.sh
sam-github: replaced comment with a ref to information above

refack:
> I've no idea how to change the default. I'll check, but I suspect it depends on the order in which packages are installed, which depends on current machine state, so is indeterminant.

https://askubuntu.com/questions/26498/how-to-choose-the-default-gcc-and-g-version

But you are right it doesn't matter.

refack:
P.S. maybe we should run:
```
sudo update-alternatives --remove-all gcc 
sudo update-alternatives --remove-all g++
```
So there will be no default... But that's for a separate PR.

rvagg:
> P.S. maybe we should run:
> ... update-alternatives

yeah, I like that a lot: always explicit and fail when unspecified

sam-github:
My ansible skills are barely up to copy-n-paste, they are not up to adding a new shell command like https://github.com/nodejs/build/pull/1723#issuecomment-473493615. I don't know where to put it, and I don't know how to make sure it only runs after the packages are installed, and ideally, run after *if-and-only-if* packages are installed. I can test it if I'm given some pointers to what kind of config to put in what place.

refack:
@sam-github IMHO this PR looks good as is. The default discussion just gravitated here. I'll pick it up in a separate PR.

sam-github:
@mhdawson PTAL, to make sure the changes you pushed are back

mhdawson:
@sam-github I think it looks good.

mhdawson:
So given that we agree this can move forward we need to FIRST configure all of the machines to have gcc6 otherwise it will break test/build.

richardlau:
> 
> 
> So given that we agree this can move forward we need to FIRST configure all of the machines to have gcc6 otherwise it will break test/build.

Perhaps we could we break this up into two parts:
1. The ansible changes, which could land first (and then be used to put gcc6 on the machines)
1. The `select-compiler.sh` changes which should only land when all the machines have gcc6

sam-github:
Breaking this into two commits is trivial, if that's the right way,  I can do it.

Note that @mhdawson is already in process of using the PR branch to update the machines, so the whole PR can be merged.

refack:
Another off-shoot idea. Store `select-compiler.sh` in the node repo, not here. That way we can keep it coupled to the version it lives in.
(Jenkins job could check if it's there. If not, use the old script).

mhdawson:
I'm fine either way (1 PR or 2). 2 is probably better in case we need to revert the compiler selection for some reason.

in test le-2, le-4 are updated I'll plan to do le-1 later today and therelease machine some time next week (or possibly the week after since I'm on holiday next week even though I'm not going anywhere.)

mhdawson:
Deployed across the 3 test machines and the release machine.  Should be good to land.  After landing please what both test/release jobs for a few days to make sure things look ok.

One extra todo is that we need to add 

ubuntu1404-ppc64_le-4: {ip: 140.211.168.140}

to inventory.yml.  

I created that as a replacement for ubuntu1404-ppc64_le-3 which seems to be completely broken

sam-github:
@mhdawson https://github.com/nodejs/build/pull/1739

PR #1734, opened 2019-03-19
Put a list between the two maps for Python 3

cclauss:
As discussed at https://github.com/nodejs/build/pull/1399#issuecomment-473137730 and https://github.com/nodejs/build/issues/1674#issuecomment-472631285

> You could try putting | list | in between the two calls to map(). Python 3 often creates generators instead of lists so perhaps forcing the results of the first map() call into a list will placate Jinja2.

### DIFF
--- a/ansible/roles/baselayout/tasks/ccache.yml
+++ b/ansible/roles/baselayout/tasks/ccache.yml
@@ -10,7 +10,7 @@
 
 - name: "ccache : extract ccache latest version"
   set_fact:
-    ccache_latest: "{{ ccache_html_content.stdout | regex_findall('ccache-[0-9]+.[0-9]+(?:.[0-9]+)*.tar.gz') | map('regex_replace', 'ccache-') | map('regex_replace', '.tar.gz') | list | latest_version }}"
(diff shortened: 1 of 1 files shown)


rvagg:
yeah, I tried it:

```
fatal: [test-requireio_kahwee-debian9-arm64_pi3-1]: FAILED! => {"msg": "Unexpected templating type error occurred on ({{ ccache_html_content.stdout | regex_findall('ccache-[0-9]+.[0-9]+(?:.[0-9]+)*.tar.gz') | map('regex_replace', 'ccache-') | list | map('regex_replace', '.tar.gz') | list | latest_version }}): '<' not supported between instances of 'map' and 'map'"}
```

tried adding another too:

```
fatal: [test-requireio_kahwee-debian9-arm64_pi3-1]: FAILED! => {"msg": "Unexpected templating type error occurred on ({{ ccache_html_content.stdout | regex_findall('ccache-[0-9]+.[0-9]+(?:.[0-9]+)*.tar.gz') | list | map('regex_replace', 'ccache-') | list | map('regex_replace', '.tar.gz') | list | latest_version }}): '<' not supported between instances of 'map' and 'map'"}
```

btw I'm getting this from roles/jenkins-worker/tasks/partials/docker-host.yml where it's repeated.

cclauss:
Closing because (of course) @refack was correct when [he wrote](https://github.com/nodejs/build/pull/1399#issuecomment-473666420):
> I'll try to move that super-filter out of the template and into code

PR #1735, opened 2019-03-19
Machines updated: Update linux s390 gcc

sam-github:
@mhdawson

### DIFF
--- a/ansible/roles/gcc/tasks/main.yml
+++ b/ansible/roles/gcc/tasks/main.yml
@@ -10,14 +10,14 @@
 
 - stat: path=/data/gmp-6.1/include/gmp.h
   register: build_gmp
-  tags: gcc49
+  tags: gcc
 
 - name: Clean path
   file:
     state: absent
     path: /data/tmp/
   when: build_gmp.stat.exists == False
-  tags: gcc49
+  tags: gcc
 
 
 - name: Create working directory
@@ -26,14 +26,14 @@
     state: directory
     mode: 0755
   when: build_gmp.stat.exists == False
-  tags: gcc49
+  tags: gcc
 
 - name: download gmp 6.1.0
   get_url:
     url: http://ftp.gnu.org/gnu/gmp/gmp-6.1.0.tar.xz
     dest: /data/tmp/
   when: build_gmp.stat.exists == False
-  tags: gcc49
+  tags: gcc
 
 - name: unarchive gmp 6.1.0
   unarchive:
@@ -41,7 +41,7 @@
     remote_src: yes
     dest: /data/tmp/
   when: build_gmp.stat.exists == False
-  tags: gcc49
+  tags: gcc
 
 - name: Create build directory
   file:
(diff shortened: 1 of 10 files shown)


ansible/roles/gcc/tasks/main.yml
```diff
@@ -290,3 +323,79 @@
...
+
+- name: Create working directory
+  file:
+    path: /data/tmp
+    state: directory
+    mode: 0755
+  when: build_gcc_63.stat.exists == False
+  tags: gcc63
+
+- name: download gcc 6.3.4
```
richardlau: .4?

ansible/roles/gcc/tasks/main.yml
```diff
@@ -2,225 +2,258 @@
...
+- stat: path=/data/gmp-6.1/include/gmp.h
+  register: build_gmp
   tags: gcc49
 
 - name: Clean path
   file:
     state: absent
     path: /data/tmp/
-  when: build_gcc_49.stat.exists == False
+  when: build_gmp.stat.exists == False
```
mhdawson: Should the tags: gcc49 lines which are not specific to 49 like this one be updated to something more generic like `tags:gcc`

sam-github: No idea. The tags have no purpose I could discern.

refack:
Seems like the Fedora project provides GCC 5,7,8 (but not 6) precompiled RPMs for s390x:
https://rpmfind.net/linux/RPM/fedora/devel/rawhide/s390x/g/gcc-9.0.1-0.10.fc30.s390x.html
https://rpmfind.net/linux/RPM/fedora/29/s390x/g/gcc-8.2.1-2.fc29.s390x.html
https://rpmfind.net/linux/rpm2html/search.php?query=gcc

ansible/roles/gcc/tasks/main.yml
sam-github: fixed

mhdawson:
@refack not sure how Fedora relates here.  We use RHEL 7.X to build for s390....

ansible/roles/gcc/tasks/main.yml
mhdawson: I think they are just a tag to show related items.  In that context I think we could just change them from gcc49 to gcc as those steps are now not specific to the later gcc49 build steps.

sam-github:
I updated -3 last week, I updated -2 yesterday, and restarted jenkins today.

I'm updating -1 today.

mhdawson:
@sam-github just remember I also have to update the release machine first.  If things look good on -1 today I can do that tomorrow.

mhdawson:
Took a while as we had to fix package dependencies on the release machine before ansible ran, but its complete now.  Kicked off builds against 12.x, 10.x, 8.x and 6.x and we'll do a sniff test on them due to the dependency fixup we had to do.  

We should be able to land this soon.

sam-github:
We've updated all test and release machines to have gcc6, so this can land (and switch to using gcc6 for  `> 11.x`)

mhdawson:
Landed as 2b17715d34817026980e45aba5eee110663f7f6e

PR #1739, opened 2019-03-22
ansible: add ubuntu1404-ppc64_le-4 to inventory

sam-github:
See: https://github.com/nodejs/build/pull/1723#issuecomment-475717268


sam-github:
@mhdawson Should -3 be removed? It sounds like its broken or gone?

mhdawson:
We should remove -3  as well. Landing this and you can do that as a follow on PR.

PR #1741, opened 2019-03-24
github-bot: update Node.js from 6 -> 10

phillipj:
It's been a while since we setup the github-bot initially and a few Node.js major releases has been cut since then..

Thought we better settle on current LTS for now.

/cc @nodejs/github-bot


rvagg:
updated and restarted node services

PR #1751, opened 2019-04-04
ansible: remove infra "collector"

rvagg:
I've turned this off in DO (not destroyed) to see if anyone/anything yelps. It's an InfluxDB + Grafana server that @jbergstroem set up some time ago but I don't believe we ever hooked it up to collect stats for us. I've also turned off an old www server that I believe is redundant now.

Please yelp if you have a reason either of these shouldn't be destroyed.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -10,7 +10,6 @@ hosts:
   - infra:
 
     - digitalocean:
-        debian8-x64-1: {ip: 198.199.72.83, alias: collector}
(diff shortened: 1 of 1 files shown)


rvagg:
turned off another old www server that's being kept around for just-in-case reasons but this one is much older, with these 3 removals we have nothing else running in DO that's not in inventory.yml so we're 👌

rvagg:
removed 3 servers from DO

PR #1754, opened 2019-04-05
Use devtoolset-6 for s390x on node gt 11.x

sam-github:
Depends on https://github.com/nodejs/build/pull/1755

### DIFF
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
 if [ "$DONTSELECT_COMPILER" != "DONT" ]; then
+  NODE_NAME=${NODE_NAME:-$HOSTNAME}
+  echo "Selecting compiler based on $NODE_NAME"
   case $NODE_NAME in
     *ppc64_le* ) SELECT_ARCH=PPC64LE ;;
     *s390x* ) SELECT_ARCH=S390X ;;
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -46,22 +46,26 @@ elif [ "$SELECT_ARCH" = "S390X" ]; then
 
   echo "Setting compiler for Node version $NODEJS_MAJOR_VERSION on s390x"
 
-  if [ "$NODEJS_MAJOR_VERSION" -gt "11" ]; then
-    export PATH="/data/gcc-6.3/bin:/data/binutils-2.28/bin:$PATH"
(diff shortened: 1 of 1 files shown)


sam-github:
devtoolset-6 is authorized for use on IBM systems (and the s390x boxes are IBM systems) for the purposes Node.js would use it for, but installing it requires downloading RPMs through an authenticated portal. I don't see how this can be ansibled, so I created /data/devtoolset-6-s390x-rpms/ on the 3 s390x boxes. I installed it on only -3, using `yum install /data/devtoolset-6-s390x-rpms/*.rpm`. I haven't done this on the other 2 yet. Or added docs to the README on this manual setup task.

/to @mhdawson

sam-github:
It appears to me that devtoolset does indeed build more portable binaries, even with _no additional configuration_. See:
```
[root@test-linuxonecc-rhel72-s390x-3 data]# docker run --rm -it -v /home/iojs/sam-github:/sam-github rhel:7.6 /sam-github/node-gcc -p process.versions
/sam-github/node-gcc: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /sam-github/node-gcc)
[root@test-linuxonecc-rhel72-s390x-3 data]# docker run --rm -it -v /home/iojs/sam-github:/sam-github rhel:7.5 /sam-github/node-devtools -p process.versions
{ node: '12.0.0-pre',
  v8: '7.4.288.18-node.16',
  uv: '1.27.0',
  zlib: '1.2.11',
  brotli: '1.0.7',
  ares: '1.15.0',
  modules: '72',
  nghttp2: '1.37.0',
  napi: '4',
  llhttp: '1.1.1',
  http_parser: '2.8.0',
  openssl: '1.1.1b',
  cldr: '34.0',
  icu: '63.1',
  tz: '2018e',
  unicode: '11.0' }
[root@test-linuxonecc-rhel72-s390x-3 data]# cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.6 (Maipo)
```

mhdawson:
We might want to leave this PR with the selection being gcc6, until we get the ansible changes in to help install the dev toolset as well?

mhdawson:
We've discussed this approach for the dev toolset.  Store the rpms on one of the s390 machines and then have ansible pull/install from there when it is run.  This is because we believe the licence allows them to be on IBM machines (which the ones in the community are) but not otherwise redistributed.  This approach is not perfect but a step better than simply manually installing on all machines.

mhdawson:
Landed as cf02bffb85f18ed0be2073cb3f30aa0ec97c698d

PR #1758, opened 2019-04-11
ansible: remove non-existent ubuntu1404-ppc64_le-3

sam-github:
See: https://github.com/nodejs/build/pull/1739#issuecomment-482115248


refack:
If it's not there, it's not there 🤷‍♂️

Issue #1705, opened 2019-02-26
Add centos7 PPCle machines


mhdawson:
If possible we'd like to build on Centos76  for PPC now that centos is available at OSU.  Starting with using them for 12.x and above and then seeing if we can use them for older releases as well.

- [x]  Add 2 test centos machines (test-osuosl-centos7-ppc64-le-1, test-osuosl-centos7-ppc64-le-2)
- [x]   Configure test machines with existing Ansible scripts if possible
- [x]  Validate that compiler is installed properly: devtoolset-6 installed
- [x] Update ansible scripts if necessary
- [x]  Update regression tests so that tests also run on centos76 for 12.x+, validate all passes
- [x]  Add 1 release centos machine (release-osuosl-centos7-ppc64-le-1)
- [X] update select-compiler.sh: select devtoolset-6 ONLY for centos7-ppc64le (if it's ubuntu, continue ot use gcc-6)
- [x]  Configure release machine with Ansible
- [x]  Add release machine to release CI.
- [x]  ... add additional required keys to release machine etc.
- [x] Do test builds with cloned release job where we build releases on centos7 for 13+ instead of unbuntu
- [x] Land PR for groovy selector to start doing releases on centos7 for 13+ instead of ubuntu
- [ ] Audit community test jobs to make sure we have added centos7 to all test jobs
- [ ] Identify next release, and plan switch over so that we get X days of nightly build/testing before next
      real release.
- [ ]  Land PR for groovy selector to start doing releases on centos7 for 12+ instead of ubuntu

refack:
Is the plan to (if all goes well) use there and retire the `ubuntu14-ppc64-le` workers?

mhdawson:
@refack assuming we convince ourselves that older binaries still run on all target OS's (of which I'm hopeful) then yes but I don't want to rush that so my plan would be to do that slowly.

sam-github:
@mhdawson Tried to tick off some of the items above, but only build-wg members can do that.

What are the "regression tests"?

sam-github:
@refack You probably already know this, but for anyone else: ubuntu 14.04 is aging out of LTS soon, so yes, we want to move to centos7 because it's support horizon will last long enough for 12.x.

To replace the ubuntu14.04 with centos7 for < 12.x, I need to find gcc 4.8, and 4.9 (at least) for them, and I haven't had any luck with that yet. Worst case is we'll have to build them ourselves via ansible, which would be frustrating.

[...]

PR #1763, opened 2019-04-15
Add support for Centos7/ppc64

sam-github:
- Add the repository for devtoolset-6.
- Do not use the "ius" repository, it doesn't support ppc64, and was not
  necessary.

Fixes: https://github.com/nodejs/build/issues/1705

### DIFF
--- a/ansible/roles/baselayout/tasks/partials/repo/centos7.yml
+++ b/ansible/roles/baselayout/tasks/partials/repo/centos7.yml
@@ -6,16 +6,25 @@
     state: present
 
 - name: centos7 | install ius
-  when: "arch != 'arm64'"
+  when: "arch != 'arm64' and arch != 'ppc64'"
   yum:
     name: "https://centos{{ ansible_distribution_major_version }}.iuscommunity.org/ius-release.rpm"
     state: present
+
 
 - name: centos7 | install scl for devtoolset-6
   when: "arch != 'arm64' and arch != 'ppc64'"
   yum:
     name: centos-release-scl
     state: present
+
+- name: centos7 | ppc64 | install sclo7 devtoolset-6
+  when: "arch == 'ppc64'"
+  yum_repository:
+    name: sclo7-devtoolset-6
+    description: cbs.centos.org sclo7-devtoolset-6-rh-release
+    baseurl: https://cbs.centos.org/repos/sclo7-devtoolset-6-rh-release/$basearch/os/
+    gpgcheck: no
 
 - name: centos7 | aarch64 | install sclo repo
   when: "arch == 'arm64'"
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -44,7 +44,7 @@ packages: {
   centos7_x64: ['git2u','centos-release-scl',], # centos-release-scl is required to enable SCLo
                                                # but we do it manually in partials/repo/centos7.yml for arm64
   centos7: [
-    'ccache,gcc-c++,devtoolset-6,sudo,git',
+    'ccache,gcc-c++,devtoolset-6,sudo,git,devtoolset-6-libatomic-devel',
   ],
 
   aix: [
--- a/ansible/roles/java-base/tasks/main.yml
+++ b/ansible/roles/java-base/tasks/main.yml
@@ -32,19 +32,34 @@
   include_role:
     name: package-upgrade
 
-- name: download java
-  when: os in "ubuntu1404" and arch == "ppc64"
+- name: download java ppc64
+  when: os in ("ubuntu1404", "centos7") and arch == "ppc64"
   get_url:
     url: https://github.com/AdoptOpenJDK/openjdk8-releases/releases/download/jdk8u172-b11/OpenJDK8_ppc64le_Linux_jdk8u172-b11.tar.gz
     dest: /tmp/
 
-- name: unarchive java
+- name: unarchive java ubuntu1404 ppc64
   when: os in "ubuntu1404" and arch == "ppc64"
   unarchive:
     src: /tmp/OpenJDK8_ppc64le_Linux_jdk8u172-b11.tar.gz
     remote_src: yes
     dest: /home/iojs
   tags: java
+
+- name: unarchive java centos7 ppc64
+  when: os in "centos7" and arch == "ppc64"
+  unarchive:
+    src: /tmp/OpenJDK8_ppc64le_Linux_jdk8u172-b11.tar.gz
+    remote_src: yes
+    dest: /opt
+  tags: java
+
+- name: symlink java centos7 ppc64
+  when: os in "centos7" and arch == "ppc64"
+  file:
+    src: "/opt/jdk8u172-b11/bin/java"
+    dest: "/usr/bin/java"
+    state: link
 
 # if this fails you want to check in vars/main.yml and add package name
 # as appropriate -- try to use generic os family if available.
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -19,6 +19,7 @@ def buildExclusions = [
   [ /centos[67]-(arm)?(64|32)-gcc6/,  anyType,     lt(10)  ],
   [ /centos6-32-gcc6/,                releaseType, gte(10) ], // 32-bit linux for <10 only
   [ /^centos7-64/,                    releaseType, lt(12)  ],
+  [ /^centos7-ppcle/,                 anyType,     lt(12)  ],
   [ /debian8-x86/,                    anyType,     gte(10) ], // 32-bit linux for <10 only
   [ /^ubuntu1804/,                    anyType,     lt(10)  ], // probably temporary
   [ /^ubuntu1204/,                    anyType,     gte(10) ],
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -4,7 +4,7 @@ if [ "$DONTSELECT_COMPILER" != "DONT" ]; then
   NODE_NAME=${NODE_NAME:-$HOSTNAME}
   echo "Selecting compiler based on $NODE_NAME"
   case $NODE_NAME in
-    *ppc64_le* ) SELECT_ARCH=PPC64LE ;;
+    *ppc64*le* ) SELECT_ARCH=PPC64LE ;;
     *s390x* ) SELECT_ARCH=S390X ;;
     *aix* ) SELECT_ARCH=AIXPPC ;;
   esac
@@ -21,6 +21,17 @@ if [ "$SELECT_ARCH" = "PPC64LE" ]; then
   export COMPILER_LEVEL="4.8"
 
   echo "Setting compiler for Node version $NODEJS_MAJOR_VERSION on ppc64le"
+
+  case $NODE_NAME in
+    *centos7* )
+      if [ "$NODEJS_MAJOR_VERSION" -gt "11" ]; then
+        # Setup devtoolset-6, sets LD_LIBRARY_PATH, PATH, etc.
+        . /opt/rh/devtoolset-6/enable
+        echo "Compiler set to devtoolset-6"
+	return
+      fi
+      ;;
+   esac
 
   if [ "$NODEJS_MAJOR_VERSION" -gt "11" ]; then
     # See: https://github.com/nodejs/build/pull/1723#discussion_r265740122
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -50,6 +50,7 @@ hosts:
         aix61-ppc64_be-1: {ip: 140.211.9.99}
         ubuntu1404-ppc64_be-1: {ip: 140.211.168.76}
         ubuntu1404-ppc64_le-1: {ip: 140.211.168.66}
+        centos7-ppc64_le-1: {ip: 140.211.168.61, user: centos}
 
     - scaleway:
         ubuntu1604-armv7l-1: {ip: 212.47.245.242}
@@ -150,6 +151,8 @@ hosts:
         ubuntu1404-ppc64_le-1: {ip: 140.211.168.106}
         ubuntu1404-ppc64_le-2: {ip: 140.211.168.94}
         ubuntu1404-ppc64_le-4: {ip: 140.211.168.140}
+        centos7-ppc64_le-1: {ip: 140.211.168.193, user: centos}
+        centos7-ppc64_le-2: {ip: 140.211.168.244, user: centos}
 
     - rackspace:
         centos7-x64-1: {ip: 119.9.27.82}


sam-github:
Hm, gcc-6 and gcc-4.9 aren't installed. For ubuntu, they appear to have been official packages (gcc-4.8, etc.). For 11.x, I assume it doesn't matter, we'll use devtoolset-6 for 11.x+, and only use the centos machines to build them.

sam-github:
Hm, link failed on `-latomic`, so I patched v8, and node linked:

```
[centos@test-osuosl-centos7-ppc64--le-1 node]$ git diff
diff --git a/tools/v8_gypfiles/v8.gyp b/tools/v8_gypfiles/v8.gyp
index 72a8dfe..9835491 100644
--- a/tools/v8_gypfiles/v8.gyp
+++ b/tools/v8_gypfiles/v8.gyp
@@ -391,7 +391,7 @@
         }],
         # Platforms that don't have Compare-And-Swap support need to link atomic
         # library to implement atomic memory access
-        [ 'v8_current_cpu in ["mips", "mipsel", "mips64", "mips64el", "ppc", "ppc64",
+        [ 'v8_current_cpu in ["mips", "mipsel", "mips64", "mips64el", "ppc", "s390", "
             'link_settings': {
               'libraries': [ '-latomic', ],
             },
```

@nodejs/v8-update any ideas on this?

sam-github:
Hm, went the other way, installed `devtoolset-6-libatomic-devel`

refack:
Does this need changes in
https://github.com/nodejs/build/blob/cf02bffb85f18ed0be2073cb3f30aa0ec97c698d/jenkins/scripts/select-compiler.sh#L19
, doing something like
https://github.com/nodejs/build/blob/cf02bffb85f18ed0be2073cb3f30aa0ec97c698d/jenkins/scripts/select-compiler.sh#L57
?

sam-github:
Yes... and no. Generally we land the ansiblization first, then the select-compiler.sh second in another PR. That said, I have the changes locally to select-compiler, and this machine isn't currently in use, and I'm very  unfamiliar with the process of bringin online  a set of machines that will (initially) only be used to build master.12.x..

@mhdawson the select-compiler update, want it in this PR, or another?

refack:
> I'm very unfamiliar with the process of bringin online a set of machines

Well... We don't do it very often... We'd also need to patch somewhere near
https://github.com/nodejs/build/blob/cf02bffb85f18ed0be2073cb3f30aa0ec97c698d/jenkins/scripts/VersionSelectorScript.groovy#L51-L53
...

IMO in this specific case the order or PRs is not very critical since the change will not go live until we actually add the new machines to the job matrix (`VersionSelectorScript.groovy` is an exclusion filter on top of that).

ansible/roles/baselayout/vars/main.yml
```diff
@@ -43,7 +43,7 @@ packages: {
   centos7_x64: ['git2u','centos-release-scl',], # centos-release-scl is required to enable SCLo
                                                # but we do it manually in partials/repo/centos7.yml for arm64
   centos7: [
-    'ccache,gcc-c++,devtoolset-6,sudo',
+    'ccache,gcc-c++,devtoolset-6,sudo,git',
```
refack: You mentioned `devtoolset-6-libatomic-devel`

refack: While you're here could you add `python3` (or equivalent)?

mhdawson: makes sense that both python3 and libatomic should be here.

ansible/roles/baselayout/tasks/partials/repo/centos7.yml
```diff
@@ -6,11 +6,18 @@
     state: present
 
 - name: centos7 | install ius
-  when: "arch != 'arm64'"
+  when: "arch != 'arm64' and arch != 'ppc64'"
   yum:
     name: "https://centos{{ ansible_distribution_major_version }}.iuscommunity.org/ius-release.rpm"
     state: present
 
+- name: centos7 | install sclo7 devtoolset-6
```
rvagg: I have #1765 which is for x64 centos7 and it's going to conflict with this, one of us probably needs to change, or we need to add arch exclusions. Mine is just:

```
 - name: centos7 | install scl for devtoolset-6
  when: "arch != 'arm64'"
  yum:
    name: centos-release-scl
    state: present
```

and I'm not sure about the arm64 exclusion there, we might be able to get rid of the sclo aarch64 bits further down this file but it'll need some testing (which I don't have time for right now!)

@sam-github might a simple `yum install centos-release-scl` work for ppc64 too?

refack:
Closes: https://github.com/nodejs/build/issues/1705

sam-github:
Added groovy exclusion, added more packages to centos7, still looking around to compare to Rod's changes, and for his comments.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -51,6 +51,7 @@ def buildExclusions = [
 
   // PPC BE ------------------------------------------------
   [ /^ppcbe-ubuntu/,                  anyType,     gte(8)  ],
+  [ /^ppcbe-ubuntu/,                  anyType,     gte(12) ],
```
refack:
```suggestion
  [ /^ppcle-ubuntu/,                  anyType,     gte(12) ],
```

ansible/roles/baselayout/tasks/partials/repo/centos7.yml
sam-github: I'm not so sure what these incantations do, but I replaced mine with your example above, and am trying it on a fresh host. I'll see if it works.

sam-github: Unfortunately, no, I don't think that repo works for ppc64le:
```
TASK [baselayout : install packages] **************************************************
failed: [release-osuosl-centos7-ppc64_le-1] (item=ccache,gcc-c++,devtoolset-6,sudo,git,devtoolset-6-libatomic-devel,python3) => {"changed": false, "msg": "Failure talking to yum: Cannot find a valid baseurl for repo: centos-sclo-sclo/ppc64le", "package": "ccache,gcc-c++,devtoolset-6,sudo,git,devtoolset-6-libatomic-devel,python3"}                
failed: [release-osuosl-centos7-ppc64_le-1] (item=automake,bash,libtool) => {"changed": false, "msg": "Failure talking to yum: Cannot find a valid baseurl for repo: centos-sclo-sclo/ppc64le", "package": "automake,bash,libtool"}
```

sam-github: I'm reverting back to what I did, but with `when: "arch == 'ppc64'"`.

sam-github:
@refack Sorry, I've had no success finding python 3.

jenkins/scripts/VersionSelectorScript.groovy
sam-github: I don't understand this change. I'm adding support for ppc64le, why should I change anything about ppcbe or ppcle?

sam-github: Ah, sorry. I see what you mean, I added the wrong exclusion. :eyeroll: I fixed this.

sam-github:
I've ansibled the 3 ppc64_le machines added in inventory.yml, several times (just doing it again now, against all 3 simultaneously, to make sure they are in same state).

What's next, do we land this, or add them as ci machines to make sure the config is sufficient?

ansible/roles/jenkins-worker/vars/main.yml
```diff
@@ -91,6 +91,7 @@ needs_monit: [
 
 # some os'es needs different paths to java. add them here.
 java_path: {
+  'centos7': '/home/iojs/jdk8u172-b11/bin/java',
```
sam-github: This java path is actually specific to centos-ppc64le, but I'm not sure if any other centos7 platforms use systemd.

ansible/roles/jenkins-worker/templates/systemd.service.j2
```diff
@@ -23,7 +23,7 @@ Environment="NODE_COMMON_PIPE=/home/{{ server_user }}/test.pipe"
 Environment="NODE_TEST_DIR=/home/{{ server_user }}/tmp"
 Environment="OSTYPE=linux-gnu"
 
-ExecStart=/usr/bin/java -Xmx{{ server_ram|default('128m') }} \
+ExecStart={{ java_path[os] }} -Xmx{{ server_ram|default('128m') }} \
```
sam-github: This causes some trouble. ExecStart *must* be an absolute path, or systemd refuses to use it. The path is usually /usr/bin, but on centos-ppc64le, its in /hom/iojs/.../bin/java (see comment below). So, I use the `java_path`, but its hard to know if that will work for everyone.

Is this the right way?

sam-github:
I re-ansibled with the correct jenkins secrets (@mhdawson set the machines up in jenkins), and realized that there was no java on the machine. Please see last commit, it works, but I'm not sure if its the right way, or what other options there are.

ansible/roles/jenkins-worker/templates/systemd.service.j2
mhdawson: I think it may be the right way but will likely break other platforms as I'm not sure that java_path is fully populated for all of the existing platforms.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -51,6 +51,7 @@ def buildExclusions = [
 
   // PPC BE ------------------------------------------------
   [ /^ppcbe-ubuntu/,                  anyType,     gte(8)  ],
+  [ /^ppcle-ubuntu/,                  anyType,     gte(12) ],
```
mhdawson: This should only go in once we are ready to transition from unbuntu to centos for ppce le.  To start it should instead exclude running centos on versions lower than 12.x.  The result should be builds on both ubuntu and centos for 12.x and above.

jenkins/scripts/select-compiler.sh
```diff
@@ -4,39 +4,41 @@ if [ "$DONTSELECT_COMPILER" != "DONT" ]; then
...
 
   if [ "$NODEJS_MAJOR_VERSION" -gt "11" ]; then
-    # See: https://github.com/nodejs/build/pull/1723#discussion_r265740122
-    export PATH=/usr/lib/binutils-2.26/bin/:$PATH
-    export COMPILER_LEVEL="6"
-  elif [ "$NODEJS_MAJOR_VERSION" -gt "9" ]; then
-    export PATH=/usr/lib/binutils-2.26/bin/:$PATH
-    export COMPILER_LEVEL="4.9"
-  fi
+    # Setup devtoolset-6, sets LD_LIBRARY_PATH, PATH, etc.
```
mhdawson: We need to support both ubuntu and centos so this will need to check that and then ensure the right compiler is selected based on the os

ansible/roles/jenkins-worker/templates/systemd.service.j2
sam-github: Maybe `{{ java_path[os] || '/usr/bin/java' }}` ?

jenkins/scripts/VersionSelectorScript.groovy
sam-github: `[ /^centos7/,                       releaseType, lt(12)  ],` is already in the file, so I should just change that to `anyType`, and remove the line I added, and we'll be good?

jenkins/scripts/select-compiler.sh
```diff
@@ -22,6 +22,17 @@ if [ "$SELECT_ARCH" = "PPC64LE" ]; then
 
   echo "Setting compiler for Node version $NODEJS_MAJOR_VERSION on ppc64le"
 
+  case $NODE_NAME in
```
sam-github: So, now it uses devtoolset-6 only on centos7, and only for node > 11.

jenkins/scripts/VersionSelectorScript.groovy
rvagg: that was jsut changed to centos7-64, it was catching arm64 which it shouldn't, so you should be specific

rvagg:
so we're seeing the fruits of removing `java-base` as a dependency here again, can someone explain why it was removed, especially since it seems to break everything else--because everything needs java

sam-github:
I'm not sure what `java-base` is, so I searched git history, and found https://github.com/nodejs/build/pull/1723/files

I'm slightly horrified to see I changed how the `java-base` role is reached, I don't recall doing that on purpose, or  even understand what the changes are, I suspect they made it into my PR via rebasing, but I can't prove it.

Anyhow, I'm not sure if this is the `java-base` dependency you are talking about. If it is, the java-base role is (and was) executed for centos7-ppc64le, it just so happens that because of the `when: ` clauses, no stanza to install java matched so the role ended up being a no-op.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -18,7 +18,7 @@ def buildExclusions = [
   [ /centos[67]-(arm)?(64|32)-gcc48/, anyType,     gte(10) ],
   [ /centos[67]-(arm)?(64|32)-gcc6/,  anyType,     lt(10)  ],
   [ /centos6-32-gcc6/,                releaseType, gte(10) ], // 32-bit linux for <10 only
-  [ /^centos7-64/,                    releaseType, lt(12)  ],
+  [ /^centos7-64/,                    anyType,     lt(12)  ],
```
mhdawson: Can you explain the need for this a bit more.  Wondering if maybe the addition should be ppc specific

jenkins/scripts/select-compiler.sh
```diff
@@ -4,14 +4,14 @@ if [ "$DONTSELECT_COMPILER" != "DONT" ]; then
...
-    *ppc64_le* ) SELECT_ARCH=PPC64LE ;;
+    *ppc64*le* ) SELECT_ARCH=PPC64LE ;;
     *s390x* ) SELECT_ARCH=S390X ;;
     *aix* ) SELECT_ARCH=AIXPPC ;;
   esac
 fi
 
 # get node version
-if [ -z ${NODEJS_MAJOR_VERSION+x} ]; then
+if [ -z "$NODEJS_MAJOR_VERSION" ]; then
```
mhdawson: Wondering about this change.

jenkins/scripts/VersionSelectorScript.groovy
mhdawson: In specific I'm wondering why we would not have already had the exclude for x64 in test.  Do we not have centos in the builds, that seems odd to me.

jenkins/scripts/select-compiler.sh
```diff
@@ -4,14 +4,14 @@ if [ "$DONTSELECT_COMPILER" != "DONT" ]; then
   NODE_NAME=${NODE_NAME:-$HOSTNAME}
   echo "Selecting compiler based on $NODE_NAME"
   case $NODE_NAME in
-    *ppc64_le* ) SELECT_ARCH=PPC64LE ;;
+    *ppc64*le* ) SELECT_ARCH=PPC64LE ;;
```
mhdawson: It this because the node names are inconsistent?

sam-github: because of https://github.com/nodejs/build/blob/6d12c28b01dbdaaf71fc810dcd11e5ffc3095d5d/ansible/roles/baselayout/tasks/main.yml#L16, it means that `HOSTNAME := NODE_NAME.replace(_, --)`, so select compiler's attempt to use the HOSTNAME as a default, which usually works, fails here.

jenkins/scripts/select-compiler.sh
sam-github: The previous code wasn't working for me, this does.  Possibly a shell difference? I am not at all sure why there was such a convoluted way of checking of the env var is set, this seems to work better:
```
core/build (centos7-ppc-support % u=) % if [ -z ${ZZZ+x} ]; then echo T; fi
T
core/build (centos7-ppc-support % u=) % if [ -z "$ZZZ" ]; then echo T; fi
T
core/build (centos7-ppc-support % u=) % env | grep ZZZ
```

Tomorrow I can revert it and post the error caused by the original line.

jenkins/scripts/VersionSelectorScript.groovy
sam-github: I backed this out.

jenkins/scripts/VersionSelectorScript.groovy
sam-github: https://github.com/nodejs/build/pull/1763#discussion_r276013923

> To start it should instead exclude running centos on versions lower than 12.x. The result should be builds on both ubuntu and centos for 12.x and above.

I am misunderstanding what you meant by that. What did you mean?

ansible/roles/jenkins-worker/templates/systemd.service.j2
sam-github: I backed this out, an instead I untar java to /opt, and symlink /usr/bin/java (the expeted path) to /opt/BLAH/bin/java. It worked with -1, according to `systemctl status jenkins`, so now I'm reansibling -1 and -2.

ansible/roles/java-base/tasks/main.yml
```diff
@@ -32,20 +32,35 @@
...
-- name: unarchive java
+- name: unarchive java ubuntu1404 ppc64
   when: os in "ubuntu1404" and arch == "ppc64"
   unarchive:
     src: /tmp/OpenJDK8_ppc64le_Linux_jdk8u172-b11.tar.gz
     remote_src: yes
     dest: /home/iojs
   tags: java
 
+- name: unarchive java centos7 ppc64
```
sam-github: these two stanzas have the effect of putting java in /usr/bin/java, where the systemd jenkins unit file (and anyone else) would expect it to be.

jenkins/scripts/VersionSelectorScript.groovy
rvagg: maybe that applies only to centos7-ppc64? that version rule is wrong, we want the centos7-64 exclusion to only exist for `releaseType` because that's how we say that centos6 should be used for making binaries prior to 12. `anyType` is going to stop it even testing.

jenkins/scripts/select-compiler.sh
rvagg: This change will match both an empty string and an unset environment variable. I think this change is OK because if it's empty then it's still useless and should be re-generated, but it's a bit weird that it's necessary. In what context is `NODEJS_MAJOR_VERSION` not set? Are there Jenkins jobs that aren't using the versioning plugin?

jenkins/scripts/select-compiler.sh
rvagg: The change should be OK, but again it's an oddity because `NODE_NAME` should come from Jenkins and the only situation where it's not set is when this is run independently, I guess if this is run as an independent test?

jenkins/scripts/select-compiler.sh
richardlau: > 
> 
> This change will match both an empty string and an unset environment variable. I think this change is OK because if it's empty then it's still useless and should be re-generated, but it's a bit weird that it's necessary. In what context is `NODEJS_MAJOR_VERSION` not set? Are there Jenkins jobs that aren't using the versioning plugin?

Yes. Anything that isn't building Node.js from source but still needs to do compilation (i.e. natve addons), e.g. [citgm-smoker-nobuild](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-nobuild/configure). [node-test-node-addon-api](https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api/configure).

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -19,6 +19,7 @@ def buildExclusions = [
   [ /centos[67]-(arm)?(64|32)-gcc6/,  anyType,     lt(10)  ],
   [ /centos6-32-gcc6/,                releaseType, gte(10) ], // 32-bit linux for <10 only
   [ /^centos7-64/,                    releaseType, lt(12)  ],
+  [ /^centos7-ppcle/,                 anyType,     lt(12)  ],
```
sam-github: OK, the new machines are labelled `centos7-ppcle`, this excludes them when node lt 12.

jenkins/scripts/select-compiler.sh
sam-github: Yes, anyone attempting to trouble shoot build issues or bootstrap a ci machine needs to read select-compiler by hand, after reading it to figure out what they should set NODE_NAME to. I added HOSTNAME as the default. If we don't want that, I can back it out. Given that ansible sets the HOSTNAME to the NODE_NAME (modulo the `-` and `_`), I'm not sure why this script depends on ci env vars. If this change is out of scope, I'll back it out. Its more of a public service for future users.

jenkins/scripts/VersionSelectorScript.groovy
mhdawson: My comment was because the tag is centos7-64 and I expected changes to only be affecting ppcle in this PR.

mhdawson:
So Java base was not removed it was moved in https://github.com/nodejs/build/pull/1723/files

 ansible/roles/jenkins-worker/tasks/main.yml now has

```
- include_role:
    name: java-base
```

instead. That was necessary because java base installs java to a location only created earlier on in main.yml (at least for ppc) so it could not work properly on a new machine.  

So it was moved (my fault as I added it to Sam's PR to get it working).  It worked fine for both PPC and s390 but maybe there is something about it that does not work on other platforms?  It may be because of the associated list of packages in build/ansible/roles/java-base/vars/main.yml.  

@rvagg we can revert if its causing problems on other platforms.  @sam-github can you see if you can figure out another solution which works for ppc, s390 and other platforms?  It might be something like moving the package: {} definition but I really don't know.

rvagg:
I think the java-base thing should be sorted in the merged smartos18 PR that pulled its inclusion up before the `service` calls.

ansible/roles/baselayout/vars/main.yml
```diff
@@ -43,7 +43,7 @@ packages: {
   centos7_x64: ['git2u','centos-release-scl',], # centos-release-scl is required to enable SCLo
                                                # but we do it manually in partials/repo/centos7.yml for arm64
   centos7: [
-    'ccache,gcc-c++,devtoolset-6,sudo',
+    'ccache,gcc-c++,devtoolset-6,sudo,git,devtoolset-6-libatomic-devel',
```
rvagg: I've added `git` in #1765, if it gets merged first then you'll just have to resolve that conflict

jenkins/scripts/select-compiler.sh
mhdawson: Ok, understood now  and makes sense.  The result is that you can call select-compiler.sh on the host by hand and get the right behavior without having to export NODE_NAME.

mhdawson: +1

mhdawson:
Landing as I think this is ready to go based on the discussion.

PR #1769, opened 2019-04-16
ansible,jenkins: smartos18 in ci and ci-release

rvagg:
Got the machines provisioned, won't be active until we get the versionselectorscript changes in place though.

Picked up a couple of strange removals from #1723, probably just typos or debugging changes /cc @sam-github.

### DIFF
--- a/ansible/roles/baselayout/tasks/ccache.yml
+++ b/ansible/roles/baselayout/tasks/ccache.yml
@@ -8,7 +8,7 @@
   # This depends on ansible being able to run curl locally, YMMV, if it doesn't
   # work, try changing the local_action to a remote one:
   #  raw: curl -sL https://www.samba.org/ftp/ccache/
-  local_action: command shell -sL https://www.samba.org/ftp/ccache/
+  local_action: command shell curl -sL https://www.samba.org/ftp/ccache/
   register: ccache_html_content
 
 - name: "ccache : extract ccache latest version"
--- a/ansible/roles/jenkins-worker/meta/main.yml
+++ b/ansible/roles/jenkins-worker/meta/main.yml
@@ -5,5 +5,4 @@
 #
 
 dependencies:
-- { 'role': 'java-base' }
 - { 'role': 'gcc', when: os|startswith("rhel7") }
--- a/ansible/roles/jenkins-worker/meta/main.yml
+++ b/ansible/roles/jenkins-worker/meta/main.yml
@@ -5,4 +5,5 @@
 #
 
 dependencies:
+- { 'role': 'java-base' }
 - { 'role': 'gcc', when: os|startswith("rhel7") }
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -37,6 +37,7 @@ hosts:
         smartos14-x64-1: {ip: 72.2.113.193}
         smartos15-x64-1: {ip: 72.2.113.195}
         smartos17-x64-1: {ip: 72.2.114.225}
+        smartos18-x64-1: {ip: 72.2.119.47}
         ubuntu1604_arm_cross-x64-1: {ip: 72.2.114.100, user: ubuntu}
 
     - requireio:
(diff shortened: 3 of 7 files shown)


ansible/roles/baselayout/tasks/ccache.yml
```diff
@@ -8,7 +8,7 @@
   # This depends on ansible being able to run curl locally, YMMV, if it doesn't
   # work, try changing the local_action to a remote one:
   #  raw: curl -sL https://www.samba.org/ftp/ccache/
-  local_action: command shell -sL https://www.samba.org/ftp/ccache/
+  local_action: command shell curl -sL https://www.samba.org/ftp/ccache/
```
rvagg: removed in #1723, hopefully by accident, also I discovered that smartos is shipping with a recent enough ccache now so we don't even take this path for smartos18

ansible/roles/jenkins-worker/meta/main.yml
```diff
@@ -5,4 +5,5 @@
 #
 
 dependencies:
+- { 'role': 'java-base' }
```
rvagg: removed in #1723, just for debugging or speedup during dev hopefully

refack:
The old machines are running a little heavy, could we provision the new ones a little beefier (maybe 4 cores)?
https://ci.nodejs.org/job/node-test-commit-smartos/buildTimeTrend
image

ansible/roles/jenkins-worker/meta/main.yml
mhdawson: It was removed because the dependencies were wrong.  The task that depended on it needed to create some directory locations before it would be able to be installed.

mhdawson: Instead, it was integrated into the task itself in the right place to satisfy ordering.

ansible/roles/baselayout/tasks/ccache.yml
sam-github: Was in #1735, and was unintentional, sorry, I was flailing around trying to get the wget to work.

rvagg: ohhh, yeah, been there with `wget`, sorry for your pain

ansible/roles/jenkins-worker/meta/main.yml
rvagg: So this is broken for the smartos install at least, it doesn't get java installed without it. How are we supposed to handle this now?

rvagg:
@refack in my versionselectorscript I've tightened the rules a little more, basically only smartos18 is going to be run for master+, maybe this will help matters? Post merge that graph will change to incorporate these new machines but they're still only 2-core ones, might be worth waiting to see though?

ansible/roles/jenkins-worker/meta/main.yml
sam-github: https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/tasks/main.yml#L298-L299

sam-github: ^--- that stanza should be hit, does it not get triggered on smartos?

rvagg: ah, here's the problem: that java-base inclusion is probably OK but it's too late because it comes after the service call `service: name=jenkins state=restarted enabled=yes` - so that's going to try and start jenkins but if there's no Java it'll fail, which is exactly where I was failing.

In my latest commit here, 436b080, I've removed it again from meta/main.yml and instead have moved up the `include_role` to just before the first startup (`service`) call. I don't have a bare host to test on but re-ran it on one of the smartos18 machines and watched the output and it appeared to do it in a workable order.

This isn't ideal IMO because an `include_role` in the middle of a set of tasks is going to get confusing for future debugging. It's probably technical debt we have to accept for now but some more refactoring is probably in order for jenkins-worker.

PTAL at latest.

sam-github: Your explanation makes sense to me, I ran into the issue on a fresh install, so didn't hit the restart problem.

refack:
> might be worth waiting to see though?

Sure, no rush. Was just a "while you're there"... What I did notice is that for SmartOS 17 there seem to be a substantial difference between worker 1 and 2:
image

ansible/roles/jenkins-worker/meta/main.yml
mhdawson: @rvagg thanks for figuring that out. My bad as I made the move in Sam's PR.  Not sure why we did not see the failure when running ansible on the machine I had to make the change for but glad you have a single solution at least for now.

rvagg:
smartos17-x64-1 and -2 are the same type, g4-highcpu-4G, but it doesn't really surprise me that there are differences. I'll try and remember to reprovision -2 soon, just not right now.

PR #1770, opened 2019-04-16
we need devtoolset-2 for node8

refack:


rvagg:
I think you might have branches mixed up here, you've included gcc-8 stuff too

also, where is centos6 getting devtoolset-2 from? do we have the scl or epel or whatever logic in place to make this happen? we had the same problem with centos7 which wanted devtoolset-6 in this same way but doesn't have a place to install it.

PR #1772, opened 2019-04-16
jenkins: more specific centos7 exclude

refack:
We still need `centos7-arm64-gcc48` for node 8


rvagg:
thanks for rescuing 8.x and fixing this @refack

Issue #1730, opened 2019-03-17
Node.js 12 TODO: Switch to Linux for generating source & headers tarballs


rvagg:
This is currently done on macOS along with docs. Depending on the workflow we might want to switch docs to Linux too. This has been a persistent problem (e.g. #1658) and is to do with the difference between the macOS / BSD tar and GNU tar. I believe GNU tar provides us with maximum compatibility and should make errors go away.

For safety I think we should make this change only happen for Node >= 12. It's going to require some work in the VersionSelectorScript and/or ci-release/iojs+release scripts to switch logic.

rvagg:
solved in #1777, nightlies as of today use centos7 to generate all non-binary assets.

PR #1777, opened 2019-04-17
jenkins: new "release-sources" labels for source, headers, docs

rvagg:
Ref #1730

We need to switch to Linux for creation of our generic tarballs and 12 is the time to get it done. Here's my strategy and I'd like some sanity checking before I put it into effect for nightlies.

* Current strategy piggy-backs the `osx1010-release-tar` and `osx1011-release-tar` labels, these are also used for creating the `darwin` binary tarballs so we'll leave them as they are
* I've added labels `osx1010-release-sources` and `osx1011-release-sources` to the existing osx release machines, I've also added `centos7-release-sources` to the new centos7-64 release machine.
* I'll add all 3 of these new labels to iojs+release and VersionSelectorScript here should take care of selecting the right one.
* The iojs+release job has a dedicated block for doing this work, it comes under a label match for `osx10\d\d-release-tar` and runs 3 separate items for `tar-upload`, `doc-upload` and `tar-headers-upload`. I'll change that to `.*-release-sources` and leave the rest alone, I don't see anything that's macOS specific.

That should leave <12 working as it does now and switch to CentOS 7 x64 for making these assets for Node >= 12.

Can I get some sanity checking please?

I'm also noticing as I do this that osx1011 can be removed entirely from ci-release when we drop Node 11, assuming we get 10.13 in place in time.

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -75,6 +75,12 @@ def buildExclusions = [
   // FreeBSD -----------------------------------------------
   [ /^freebsd10/,                     anyType,     gte(11) ],
 
+  // Source / headers / docs -------------------------------
+  [ /^osx1010-release-sources$/,      releaseType, gte(11) ],
+  [ /^osx1011-release-sources$/,      releaseType, lt(11)  ],
+  [ /^osx1011-release-sources$/,      releaseType, gte(12) ],
+  [ /^centos7-release-sources$/,      releaseType, lt(12)  ],
+
   // -------------------------------------------------------
 ]
 


richardlau:
SGTM.

rvagg:
this turned out to be more complicated: by uncoupling this from the osx-release-tar label we end up with non-deterministic run order and with a clean workspace. The `tar-upload`, `doc-upload` and `tar-headers-upload` all depend on `$(NODE_EXE)` but don't go through the motions of configuring and building like `binary` does. Previously we could rely on the osx-release-tar builder to do a `binary-upload` before going through these other stages so everything would be ready to go. But now the same machine is re-used but as an entirely separate run.

So, in the release-sources jobs I've had to do a `make binary` before anything else and rely on ccache to do the heavy lifting, but it means lots more copypasta which is unfortunate. On top of this we need devtoolset-6 enabled for centos7 for _all_ of the `make` calls, so more copypasta and conditionals. Then, I also discovered that ccache isn't even activated for the devtoolset jobs (centos6/devtoolset-2 as well) because we're relying on `$PATH` to do that work on these machines. So even more copypasta from the nice block of Bash in node-test-commit-linux that sets `CC` and `CXX` prepended with `ccache`.

Anyway, it appears to be working now. I re-ran a 10.x nightly and a master nightly and they used the right builders, so I think we can call this solved for now. iojs+release (maybe the whole ci-release) really needs an overhaul though, it's getting out of hand but it's such a critical piece that it's difficult to do major surgery on.

PR #1783, opened 2019-04-18
jenkins: exclude debian8 from armv7 tests for node 12+

rvagg:
This one will remove Debian 8 from the ARMv7 mix for Node 12+. This is for node-test-commit-arm on the Scaleway ARMv7 machines, not the Raspberry Pi machines.


rvagg:
Debian 8 uses GCC 4.9, Debian 9 has GCC 6 which we're using: https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/templates/armv7_stretch.Dockerfile.j2#L19-L20

Issue #1780, opened 2019-04-18
Add Windows ARM64 temporary test machines


joaocgreis:
I will add two ARM64 Windows test machines to Jenkins if there are no objections. For now, these will be used to have some initial CI support and start configuring Jenkins. I plan to add the test job to the daily jobs, but not yet to node-test-commit.

These are actually two laptops sitting in my office. They won't be very fast, but enough for now.

joaocgreis:
https://github.com/nodejs/build/pull/1795 landed with Ansible support, and 3 Jenkins jobs have been created:
- https://ci.nodejs.org/view/All/job/build-release-windows-arm64/ (https://github.com/nodejs/unofficial-builds/issues/2)
- https://ci.nodejs.org/view/All/job/node-test-commit-windows-arm64/
- https://ci.nodejs.org/view/All/job/nodegyp-test-commit-arm64/

I didn't yet add the test-commit job to the node-daily job. Since the node-gyp in npm does not yet have support for ARM64, all addons tests fail. When a node-gyp release with support lands in node, I will add the job.

PR #1795, opened 2019-04-30
ansible: add ARM64 Windows machines

joaocgreis:
Refs: https://github.com/nodejs/build/issues/1780

This adds the public part of the inventory for two ARM64 Windows machines (secret parts already updated). This also adds two entries to the Windows configuration that previously were in `host_vars` files, changes the `vs` variable to be a string everywhere and adds ARM support to the Visual Studio 2017 installation (running on existing hosts won't add the new components to VS, `cuninst -y visualstudio2017-workload-nativedesktop` needs to be run manually first).

The machines are already online and connected to Jenkins.

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -34,3 +34,7 @@ ansible_python_interpreter = /usr/local/bin/python
 ansible_become             = false
 ansible_python_interpreter = /NODEJS2/python-2017-04-12-py27/python27/bin/python
 home                       = /u
+
+[hosts:win]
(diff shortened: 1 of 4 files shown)


rvagg:
if it works then rubber-stamp lgtm because I don't quite understand how this actually works against windows hosts

joaocgreis:
Thanks, landed in https://github.com/nodejs/build/commit/be6d6dec9c1c7bb169a0e72a00a7a1024b1085dc

PR #1806, opened 2019-05-22
jenkins: ppc platform ranges

refack:
added `  [ /^ppcle-ubuntu/,                  anyType,     gte(13) ],` (and moved centos rule).

why now? for https://ci.nodejs.org/job/node-test-commit-v8-linux

@mhdawson @sam-github PTAL.

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -53,7 +53,7 @@ def buildExclusions = [
   [ /^smartos17/,                     anyType,     gte(12) ],
   [ /^smartos18/,                     anyType,     lt(12)  ],
 
-  // PPC BE ------------------------------------------------
+  // PPC ---------------------------------------------------
   [ /^ppcbe-ubuntu/,                  anyType,     gte(8)  ],
   [ /^ppcle-ubuntu/,                  anyType,     gte(13) ],
   [ /^centos7-ppcle/,                 anyType,     lt(12)  ],
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -19,7 +19,6 @@ def buildExclusions = [
   [ /centos[67]-(arm)?(64|32)-gcc6/,  anyType,     lt(10)  ],
   [ /centos6-32-gcc6/,                releaseType, gte(10) ], // 32-bit linux for <10 only
   [ /^centos7-64/,                    releaseType, lt(12)  ],
-  [ /^centos7-ppcle/,                 anyType,     lt(12)  ],
   [ /debian8-x86/,                    anyType,     gte(10) ], // 32-bit linux for <10 only
   [ /^ubuntu1804/,                    anyType,     lt(10)  ], // probably temporary
   [ /^ubuntu1204/,                    anyType,     gte(10) ],
@@ -56,6 +55,8 @@ def buildExclusions = [
 
   // PPC BE ------------------------------------------------
   [ /^ppcbe-ubuntu/,                  anyType,     gte(8)  ],
+  [ /^ppcle-ubuntu/,                  anyType,     gte(13) ],
+  [ /^centos7-ppcle/,                 anyType,     lt(12)  ],
 
   // s390x -------------------------------------------------
   [ /s390x/,                          anyType,     lt(6)   ],


jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -56,6 +55,8 @@ def buildExclusions = [
 
   // PPC BE ------------------------------------------------
   [ /^ppcbe-ubuntu/,                  anyType,     gte(8)  ],
+  [ /^ppcle-ubuntu/,                  anyType,     gte(13) ],
+  [ /^centos7-ppcle/,                 anyType,     lt(12)  ],
```
richardlau: These are le, not be (as the comment implies).

refack: Ack. I generalized the comment since IMO it more convenient to have these grouped.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -54,8 +53,10 @@ def buildExclusions = [
   [ /^smartos17/,                     anyType,     gte(12) ],
   [ /^smartos18/,                     anyType,     lt(12)  ],
 
-  // PPC BE ------------------------------------------------
+  // PPC ---------------------------------------------------
   [ /^ppcbe-ubuntu/,                  anyType,     gte(8)  ],
+  [ /^ppcle-ubuntu/,                  anyType,     gte(13) ],
+  [ /^centos7-ppcle/,                 anyType,     lt(12)  ],
```
sam-github: I don't understand the categorization, or have strong feelings about it. Some of the categories are OS, some are machine architecture, and its not clear which organization to prefer. Note that before this change, all the `centos*` were under Linux (arm, intel, and ppc). After, the centos-ppc moved to "PPC", but lthe centos-arm is still in "Linux", instead of in "ARM". Seems like it gains and loses consistency in equal amounts.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -54,8 +53,10 @@ def buildExclusions = [
   [ /^smartos17/,                     anyType,     gte(12) ],
   [ /^smartos18/,                     anyType,     lt(12)  ],
 
-  // PPC BE ------------------------------------------------
+  // PPC ---------------------------------------------------
   [ /^ppcbe-ubuntu/,                  anyType,     gte(8)  ],
+  [ /^ppcle-ubuntu/,                  anyType,     gte(13) ],
```
sam-github: This line is new, not just moved, but I can't tell from https://ci.nodejs.org/job/node-test-commit-v8-linux/ what its for. A bit more info in the commit message would be helpful.

Is the version of V8 in master not passing on ppcle-ubuntu? If so, is that an issue we have to look at, or do we not care because we expect 13.x to build on centos7?

refack: the tl;dr is we want to test V8 updates on PPC, but one variant is enough. I think the assumption is that for node 13 we'll drop ubuntu14 and prefer CentOS and devtoolset for testing and release.
I'll document this.

mhdawson: I don't completely agree with one variant is enough.  Testing on ubuntu as well is good.  Since we'll have to keep it around for earlier releases anyway might be good to keep using 14 until we think about if we can have other machines.  Either way I don't see a rush to change.

sam-github: I am not seeing why we should be testing on different distros unless we have some credible reason to think there are distro (or toolset?) specific problems. V8 shouldn't depend on the distro. Also, isn't this V8 testing for new V8 releases? V8s that aren't going to get ported back to old node.js majors? Its not possible to prove that coverage isn't needed, but there is also a cost in stability, maintainability, and usability as the number of test systems grow. Btw, this is a bit of an aside, but is it impossible to start building the `< 12.x` node.js releases on centos7, even though they used to build on ubuntu1404? I thought centos7+devtoolset gave use more portable binaries, so would be an improvement, and we would decomission the ubuntu1404 machines?

mhdawson:
I don't get the comment that this is related to testing for v8, it will change the testing for all jobs?

mhdawson:
I'll also note that until we get the centos release machine ready and being used we continue to release on ubuntu for ppc so this should not land regardless of other discussions until then.  Going to mark as changes requested until then.

refack:
Just FTR the V8 job was a motivation to do it sooner rather then later. I agree with @mhdawson this should block until we have a CentOS7/PPC release machine.

sam-github:
Stale.

Issue #1804, opened 2019-05-21
Make sure fedora-latest and fedora-last-latest are up to date


targos:
I don't know how to see the versions of Fedora installed in CI but according to @refack there's at least one host with Fedora 27.

The latest Fedora is 30 so we should update the CI to have fedora-latest=30 and fedora-last-latest=29.

rvagg:
We have a bit of a resourcing problem with this one. Rackspace is only offering 28 and we've split these in the past between Rackspace and DigitalOcean. I think we might have to trim last-latest and just go a latest with both on DO.

rvagg:
aaaand ansible doesn't properly support fedora 30, at least the ansible I can get on my machine (2.8.0). It appears to be fixed here https://github.com/ansible/ansible/pull/49402 but I'm not sure if that's in a release or even if it works properly. `python2-dnf` isn't available on fedora 30, even though python2 is. ansible hard-wires python2-dnf as a dependency for doing any package management.

At the moment I've switched the fedora27 machines to be both last-latest and latest and remove latest from node-test-commit-linux. I have the DigitalOcean Fedora 26 machine now running on a Fedora 30 image for testing but this is blocking me for now.

targos:
@rvagg I found this issue: https://github.com/ansible/ansible/issues/54855

They say we need to set `ansible_python_interpreter=/usr/bin/python3` to make it work.

PR #1812, opened 2019-05-28
ansible: add fedora30, remove fedora26

rvagg:
Fixes: #1804

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -30,6 +30,9 @@ ansible_python_interpreter = /opt/local/bin/python
 [hosts:freebsd]
 ansible_python_interpreter = /usr/local/bin/python
 
+[hosts:fedora30]
+ansible_python_interpreter = /usr/bin/python3
+
 [hosts:marist]
 ansible_become             = false
 ansible_python_interpreter = /NODEJS2/python-2017-04-12-py27/python27/bin/python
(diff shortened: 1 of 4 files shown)


ansible/ansible.cfg
```diff
@@ -30,6 +30,9 @@ ansible_python_interpreter = /opt/local/bin/python
 [hosts:freebsd]
 ansible_python_interpreter = /usr/local/bin/python
 
+[hosts:fedora30]
```
rvagg: this will become `hosts:fedora` on the next update, from now on we need python3, there is no python2-dnf from Fedora 30+ which makes Ansible bork as per https://github.com/ansible/ansible/issues/54855

rvagg:
fedora27 is last-latest, fedora30 is latest. We could remove fedora27 whenever we want but I figure that it's easier to just leave it in place for now.
Note that Rackspace doesn't offer 30 so both of these are on DO. I've tested them with 8.x, 10.x and master and they're green.

ansible/roles/baselayout/vars/main.yml
```diff
@@ -64,7 +64,7 @@ packages: {
   ],
 
   fedora: [
-    'bzip2,ccache,gcc-c++,git,fontconfig,sudo',
+    'bzip2,ccache,gcc-c++,git,fontconfig,sudo,make',
```
refack: could we add `ninja-build` as well?
```suggestion
    'bzip2,ccache,gcc-c++,git,fontconfig,sudo,make,ninja-build',
```

rvagg: we could, but why? are we testing that now? and do you know for sure that's available on fedora30?

PR #1822, opened 2019-06-05
aix: set jenkins TERM to a consistent value

sam-github:
Currently, it is inherited from the user that started Jenkins, which can
be `TERM=dumb` in some ssh environments. That value of TERM breaks some
of the terminal related tests, so ensure a consistent TERM value.

Fixes: https://github.com/nodejs/node/issues/28064

### DIFF
--- a/setup/aix61/resources/S20jenkins
+++ b/setup/aix61/resources/S20jenkins
@@ -8,7 +8,7 @@
 
 case "$1" in
 start )
-        su - {{user}} -c 'ulimit -c 0; \
+        TERM=ansi su - {{user}} -c 'ulimit -c 0; \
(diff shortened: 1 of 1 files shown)


sam-github:
Fix manually applied to:
- test-osuosl-aix61-ppc64_be-1
- test-osuosl-aix61-ppc64_be-2
- test-osuosl-aix61-ppc64_be-3

@mhdawson I do not have credentials to update release-osuosl-aix61-ppc64_be-1, you should look at it (assuming it runs the node test suite during the release proces).

mhdawson:
@sam-github we don't run the test suite on the release machines.

Issue #1695, opened 2019-02-19
provision mac minis at nearform as osx machines in jenkins CI


mhdawson:
We've realized we dont' have an issue covering this as email was used to do the initial arrangements with the Foundation.

We have 2 mac minis at NearForm (bought by the Foundation) to be added to the CI.

We have VMware so we should be able to have 3 machines on each (1 bare metal and 2 VMs).

Next steps include:

- [x] Nearform to configure with ssh and vnc access
- [X] Installation of vmware (not sure if build team members can do this through ssh and vnc or not)
- [X] Creation of the virtual machines, key question here is what osx versions can we get/install
- [x] Assign the IP given by nearform to a VM
- [x] Verify we can reach the VM using this IP
- [ ] Run and tweak the ansible playbooks on the VM
- [ ] Node build/test passes
- [ ] Re-check Compatability tests - (build binaries on latest, run tests on 10.10)
- [X] Decide which OS level and how many we want (ATM all 10.15 - 1 release, 3 test. Can revaluate later if needed)
- [ ] Set up and ansible the remaining VMs (reinstall any 10.14 as 10.15)
- [ ] Add the VMs to jenkins
- [ ] Add the 3 test machines to node-test-commit-osx, and 1 release machine to iojs+release
- [ ] Notarization: should be possible on these release builds, confirm that it works (or possibly complete the notarization process ongoing in nodejs/node#29216)
- [ ] Check the nightlys do work on all platforms
- [ ] Check that all other jobs using osx can run on 10.15/nearform (v8, etc.)
- [ ] Update all jobs to use these machines (Which jobs use OSX?)

mhdawson:
@BethGriggs,  @jasnell  I'm hoping we only need Nearform to do the first step on the list above.

jasnell:
We have it most of the way. I need to follow up this week with our infrastructure guy

BethGriggs:
@jasnell, any updates?

rvagg:
I'm not sure running bare is going to be a good idea, isolating via vm is nice for security and the ease of clean-up & replace as needed

mhdawson:
Using just the 2 VMs makes sense to me.  I thought somebody had mentioned using bare metal as the third, but avoiding that seems safer to me as well.  Don't think that really changes any of the steps above except we won't do it for the bare metal machine.

sam-github:
@ashcripps and I have started looking at this. ssh and vnc was already setup, its working.

The next steps are VMWare setup.

[...]

PR #1838, opened 2019-06-14
ansible: macOS tweak - disable Spotlight

refack:
It seems like a resource hog, and IIUC used main in GUI, so we don't need it.


mhdawson:
@refack did you test this out?

sam-github:
Seems stale... but then again, maybe once we do #1695 it will be relevant again, so leaving open for a bit.

sam-github:
Closing, the presence of spotlight (if its there on Orka) has not been causing issues.

cc: @AshCripps

Issue #1798, opened 2019-05-14
Problems with CITGM job on AIX


targos:
Run: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/1848/nodes=aix61-ppc64/console

Issue 1:

```
Receiving objects:  58% (293725/506422), 311.83 MiB | 15.06 MiB/s   
fatal: write error: There is not enough space in the file system.
```

Issue 2: 

```
01:02:52 + ps awwx
01:02:52 + grep citgm
01:02:52 + grep -v grep
01:02:52 + awk '{print $1}'
10:02:52 + xargs -rl kill -9
10:02:52 xargs: The -r flag is not valid.
10:02:52 Usage: xargs [-ptx] [-e[EndOfFileString]] [-E EndOfFileString]
10:02:52         [-i[ReplacementString]] [-I ReplacementString | -L Number |-n Number ]
10:02:52        [-l[Number]] [-s Size] [Command [Argument ...]]
```

sam-github:
machine 3 has the least disk usage of any of the three test machines... howerver, it also has the smallest ramdisk of any of the three machines. I recommend not running citgm on that machine.

```
ansible/host_vars (master % u=) % parallel-ssh -ih hosts.test-osuosl-aix61-ppc64_be du -sk /home/iojs/build/workspace                                                        
[1] 12:26:23 [SUCCESS] test-osuosl-aix61-ppc64_be-3
2032864 /home/iojs/build/workspace
[3] 12:26:30 [SUCCESS] test-osuosl-aix61-ppc64_be-2
4736800 /home/iojs/build/workspace
[2] 12:26:26 [SUCCESS] test-osuosl-aix61-ppc64_be-1
2696936 /home/iojs/build/workspace
ansible/host_vars (master % u=) % parallel-ssh -ih hosts.test-osuosl-aix61-ppc64_be "df|grep iojs"                                                                           
[1] 12:27:05 [SUCCESS] test-osuosl-aix61-ppc64_be-3
/dev/ramdisk1   15000000    584008   97%   162889    18% /home/iojs/build
[2] 12:27:05 [SUCCESS] test-osuosl-aix61-ppc64_be-2
/dev/ramdisk0   23000000   9157912   61%   470210    21% /home/iojs/build
[3] 12:27:05 [SUCCESS] test-osuosl-aix61-ppc64_be-1
```

sam-github:
[...]

PR #1850, opened 2019-06-21
Add note about not restarting AIX machines

sam-github:
see https://github.com/nodejs/build/issues/1798#issuecomment-504577167


doc/jenkins-guide.md
```diff
@@ -193,6 +194,9 @@ shutdown -r now
 reboot
 ```
 
+On the advice of the system adminstrators managing the AIX machines, please do
+not restart them without good reason, it will make things worse.
```
Trott:
```suggestion
not restart them without good reason; it will make things worse.
```

PR #1858, opened 2019-06-28
doc: updates to provider listings

rvagg:
* update scaleway logo
* properly remove NodeSource and Voxer from list (they were already removed from the images but the references remained)
* minor tweak to image description at top of doc

For future reference for anyone wanting to edit the static-assets/provider-logos.svg file: it's been edited with both Illustrator and Inkscape. When Inkscape saves it, it leaves `inkscape:connector-curvature="0"` attributes in some of the `<path>` elements but the `inkscape` namespace isn't declared so it won't open in a standard SVG viewer or editor so that has to be manually removed.

/cc @remyleone I think we need to have a better one-sentence description of Scaleway, the old one seems out of date now that Scaleway has expanded its offerings, can you help us with a better tagline?

> a "BareMetal" SSD cloud server provider


rvagg:
Added updates for the text for Joyent (expanded description of what they provide) and Rackspace (removed reference to object storage).

remyleone:
@rvagg I think you can put: Scalable cloud platform designed for developers & growing companies

rvagg:
updated

rvagg:
landed in b35df97

Issue #1864, opened 2019-07-18
Removing node-test-commit-custom-suites-freestyle from node-test-commit


rvagg:
Since this is an unmaintained pipeline I'd like to explore rolling up whatever it contributes to node-test-commit into the containered builds if possible. I can't quite figure out what the purpose of this thing is and why it's so darn complicated.

From what I can glean from the bash and the way it's invokved, all it's doing is running the worker tests when invoked from node-test-commit.

It's passed `CI_JS_SUITES=default` and `TEST_ARGS=--worker`. The console output shows it doing:

- `make -j1 bench-addons-build`
- `make -j 4 build-ci`
- `python tools/test.py -j 4 -p tap --logfile test.tap --mode=release --flaky-tests=dontcare --worker`

I think `bench-addons-build` is unnecessary for node-test-commit so it's really just a `build-ci` and the custom test.py, which is very similar to the way we invoke debug and withoutssl runs in the containered jobs (with a bonus that the configs are so much simpler).

It's also got `if [ "$TEST_ARGS" = "--worker" ] && [ "$NODEJS_MAJOR_VERSION" -lt "10" ]; then` so it's limited to Node >= 10. This should go into VersionSelectorScript.groovy (there's more of this logic for s390 and running v8-updates, those should be pulled out too).

So, I believe that if I pull out a `--worker` invocation of test.py into a containered job, put a >=10 restriction in VersionSelectorScript.groovy then I could remove this from node-test-commit. If there's anyone around that has more of a clue about this and believes this assessment is wrong, please speak up now. I'm really unhappy with this weight around our necks and would like it gone. It can be left as is for whatever nightly jobs its doing, but I also wouldn't mind removing those too in due course.

richardlau:
> 
> 
> Since this is an unmaintained pipeline I'd like to explore rolling up whatever it contributes to node-test-commit into the containered builds if possible. I can't quite figure out what the purpose of this thing is and why it's so darn complicated.

It's complicated because it has been (ab)used in different contexts beyond its original purpose. The job was originally written to allow one-off testing of things not tested by the regular CI runs (e.g. want to run the internet or pummel test suites? Or run without intl (this was before the containerized jobs)). Somewhere along the line as requests came in to test more things as part of the regular/nightly CI and instead of adding new jobs the `custom-suites` jobs was just chained into the other jobs called with the appropriate parameters:
e.g.: 
[...]

PR #1865, opened 2019-07-18
node-test-commit matrix document

rvagg:
I was writing up some basic notes on our standard CI matrix and figured it might make a good doc for our repo even though it's not likely to stay updated (hence the caveats at the top).

It's highlighted a couple of things already:

1. node-test-commit-custom-suites-freestyle is really far too complex for what it contributes to node-test-commit and could be easily refactored out into the containered jobs (from what I can tell) https://github.com/nodejs/build/issues/1864
2. The debian9-docker-armv7 label was missing from node-test-commit-linux-arm, it should have been there since Node 12 but I guess this was my oversight. I've added it and am testing it out https://ci.nodejs.org/job/node-test-commit-arm/25379/nodes=debian9-docker-armv7/console
3. The containered jobs config could be refactored into a script in this repo. As the list of them as grown, the copypasta has increased, there's a standard set of things that need to be executed for each of them and they could be put into a single script, or at least a collection of scripts which would make it easier to diagnose problems like the recent one that was highlighted where debug builds aren't running the full compliment of tests.

@joaocgreis the Windows matrix doesn't seem to be using VersionSelectorScript.groovy, I've tried my best to reproduce what I think the rules are but a sanity check from you would be appreciated.

### DIFF
--- a/doc/node-test-commit-matrix.md
+++ b/doc/node-test-commit-matrix.md
@@ -49,28 +49,46 @@ This is assumed correct as of the date of last commit. If you notice a discrepan
     - smartos18-64 (Node >= 12)
   - **node-test-commit-windows-fanned**
     - **node-compile-windows**
-      * Version selection: `(!(ENGINE!="ChakraCore") || !label.endsWith("-arm")) && (!(NODE_MAJOR_VERSION.toInteger()>=6) || !label.startsWith("win-vs2013")) && (!(NODE_MAJOR_VERSION.toInteger()>=10) || !(label.startsWith("win-vs2015") || label.startsWith("win-vcbt2015"))) && (!(NODE_MAJOR_VERSION.toInteger()<8) || !label.startsWith("win-vs2017")) && (!(NODE_MAJOR_VERSION.toInteger()<10) || label!="win-vs2017-x86")`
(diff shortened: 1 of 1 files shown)


rvagg:
Need some Windows sanity check before I can merge this I think.

joaocgreis:
@rvagg pushed a fixup with some adjustments, feel free to adjust as you see fit.

Windows doesn't use the VersionSelectorScript because of two issues (at least):
- Would have to separate what can build and what should be used to build the official releases. For example, for Node 8, we want to use both VS2015 and VS2017 to build in CI and we have those two versions as well in the release Jenkins ([where VS2015 should be used](https://github.com/nodejs/build/blob/master/doc/windows-visualstudio-supported-versions.md#official-releases)).
- We support running (including building addons) in more versions than we support compiling core, and we want to test those in CI.

Separating the test jobs in CI as we have now for Node >= 11 is not a bad way of organizing it. So, even though VersionSelectorScript would probably be a better way I don't think adapting it is urgent at the moment.

rvagg:
thanks @joaocgreis, just having the windows matrix outlined like this is helpful

rvagg:
merged

PR #1869, opened 2019-07-21
fedora27-x64: allow multicast with firewalld

santigimeno:
So `libuv` multicast join tests pass.
Refs: https://github.com/libuv/libuv/pull/2185

### DIFF
--- a/ansible/roles/jenkins-worker/tasks/main.yml
+++ b/ansible/roles/jenkins-worker/tasks/main.yml
@@ -148,6 +148,18 @@
   command: systemctl restart iptables
   when: "'rhel72-s390x' in inventory_hostname"
 
+- name: firewalld | allow ipv4 multicast traffic
+  command: firewall-cmd --permanent --direct --add-rule ipv4 filter INPUT 0 -m pkttype --pkt-type multicast -j ACCEPT
+  when: "'fedora27-x64' in inventory_hostname"
+
+- name: firewalld | allow ipv6 multicast traffic
+  command: firewall-cmd --permanent --direct --add-rule ipv6 filter INPUT 0 -m pkttype --pkt-type multicast -j ACCEPT
+  when: "'fedora27-x64' in inventory_hostname"
+
+- name: firewalld | restart so rules take effect
+  command: systemctl restart firewalld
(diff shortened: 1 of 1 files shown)


rvagg:
@santigimeno is it just 27 that's the problem? Is fedora-latest-x64 OK?

27 is supposed to be remove from CI, it's pretty old, we just need a strategy to get a newer one on there and have has soem trouble.

I'll get this merged and deployed to whatever we have, but if it's not impacting the newer one then it might be redundant pretty soon.

rvagg:
@santigimeno did you edit /etc/hosts on the machine you were on?

```
TASK [jenkins-worker : add ::1 to /etc/hosts for ipv6 compat] ************************************************************************************************
changed: [test-rackspace-fedora27-x64-1]
ok: [test-digitalocean-fedora27-x64-1]
```

So one of the machines had its /etc/hosts not match what it was supposed to for ipv6 purposes. If you changed that and it was part of getting your required multicast functionality then it might not work now it's reset.

rvagg:
@santigimeno and second follow-up, this is failing on test-digitalocean-fedora27-x64-1 because `firewall-cmd` isn't installed. I've added `firewalld` to the install list for fedora27 but it's still failing because it has to be started, but perhaps the fact that it's not installed is good thing and means the tests will pass on this machine? Should I make it so that a failure is acceptable if `firewall-cmd` doesn't exist? Perhaps a better approach to this whole thing might even be to remove firewalld?

santigimeno:
@rvagg 

> is it just 27 that's the problem? Is fedora-latest-x64 OK?

It was just on fedora 27 (`fedora-last-latest-x64`)

> did you edit /etc/hosts on the machine you were on?

From what I remember no, I did not. I just played around with `iptables` and `firewall-cmd` commands.

Also with regards to `firewalld` being installed or not. I think the issue is that if `firewalld` is running with the `public` zone as the only active zone (which is the default configuration), no external traffic is allowed except for some protocols (dns, dhcp, etc), so multicast won't work as the traffic comes from the external network interfaces. So yes, both options would work: make a failure acceptable if no `firewalld` or remove `firewalld` completely. I'd prefer the latter myself.

PR #1885, opened 2019-08-13
jenkins: release 12.x on centos7-ppcle

sam-github:
13.x has been releasing on centos7-ppcle for months. Start releasing
12.x on centos7-ppcle now. 10.x and below is not changed.

We get a supported operating system and expanded platform support:
1. The current release system, ubuntu14.04, has been EOL since April.
2. The devtoolset-6 compiler tools are available on centos7, and allow
building binaries that run on MORE old linux libc versions than the
current binaries build with gcc.


sam-github:
I forgot to mention: we don't want to do this between the sec releases on the 15th, and the next 12.x on   Aug 20, its not enough time to confirm its working.

I suggest landing this after the 09-03 12.x release, see https://github.com/nodejs/Release/issues/435

sam-github:
@nodejs/build any concerns about this?

I'm calling it out because we don't usually change the O/S that we use to build a version except at the start of a new release line.

In this case, by using devtools, we are going to get better platform support, fixing complaints about our current builds not running on some older platforms -- they will after this change.

sam-github:
Once landed, test by kicking the nightly build.

PR #1887, opened 2019-08-14
jenkins: 10.x/s390 building with devtoolset-6

sam-github:
Start using devtoolset-6 to compile 10.x instead of gcc-4.9.

This will fix problems where gcc-4.9 builds either don't run on older
RHEL, or will run, but only after installing the gcc-4.9 runtime
libraries.

This has the effect of increasing the quality of platform coverage.

### DIFF
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -58,12 +58,6 @@ elif [ "$SELECT_ARCH" = "S390X" ]; then
   echo "Setting compiler for Node version $NODEJS_MAJOR_VERSION on s390x"
 
   if [ "$NODEJS_MAJOR_VERSION" -gt "9" ]; then
-    export PATH="/data/gcc-4.9/bin:$PATH"
-    export LD_LIBRARY_PATH="/data/gcc-4.9/lib64:$LD_LIBRARY_PATH"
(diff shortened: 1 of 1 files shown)


sam-github:
I'm manually testing with a build on test-linuxonecc-rhel72-s390x-1. It seems to be going well (as expected), but its quite slow, because builds with this toolset haven't been  ccached (also as expected).

mhdawson:
As an FYI it would also be possible to do a test using the CI by making a copy of the CI job and modifying the part that pulls in the grovy selector to pull in the version in this PR

mhdawson:
This does change the compiler from 4.9 to 6 but we believe that is a limited risk as we already compile with gcc 6 for 10.x on AIX. Overall we believe the benefit outweighs the risks of making this switch.

mhdawson:
One last thing is that we think now is the right time as we'll get some test coverage when the upcoming 10 rc goes out

sam-github:
Post-merge checks:
- v10.x (should select devtoolset for compiler): https://ci.nodejs.org/job/node-test-commit-linuxone/15205/
- v8.x (should be unchanged): https://ci.nodejs.org/job/node-test-commit-linuxone/15206/
- v12.x (should be unchanged): https://ci.nodejs.org/job/node-test-commit-linuxone/15207/

Issue #1889, opened 2019-08-15
ci-release: no space on pi device


BethGriggs:
When running the builds for v8.16.1 I am hitting:

```
21:04:22 Building remotely on release-requireio_rvagg-debian9-armv6l_pi1p-1 (pi1-docker pi1-raspbian-stretch) in workspace /home/iojs/build/ws
...
21:04:23 java.nio.file.FileSystemException: /home/iojs/.jenkins/cache/jars/5C: No space left on device
21:04:23 	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:91)
21:04:23 	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
```

https://ci-release.nodejs.org/job/iojs+release/nodes=pi1-docker/4633/console

sam-github:
I can't ssh into that machine:

```
build/hosts (build-10-with-devtoolset-s390x $%) % ssh release-requireio_rvagg-debian9-armv6l_pi1p-1                                                                          
zsh:1: parse error near `}'
ssh_exchange_identification: Connection closed by remote host
build/hosts (build-10-with-devtoolset-s390x $%) % grep -A 5 release-requireio_rvagg-debian9-armv6l_pi1p-1 ~/.ssh/config                                                      
Host release-requireio_rvagg-debian9-armv6l_pi1p-1
  HostName 192.168.2.40
  IdentityFile ~/.ssh/nodejs_build_release
  User pi
  ProxyCommand ssh -i {{ ansible_ssh_private_key_file }} -W %h:%p -p 2222 jump@vagg-arm.nodejs.org                                                                           

```

It looks like the .ssh/config has some ansible garbage in it, so I hacked it up, worked a bit better, but still wants a password:
```
 build/hosts (build-10-with-devtoolset-s390x $%) % ssh release-requireio_rvagg-debian9-armv6l_pi1p-1                    
The authenticity of host '[vagg-arm.nodejs.org]:2222 ([203.217.18.114]:2222)' can't be established.
ECDSA key fingerprint is SHA256:cQOh+H/IRjq/rFvpoqcgfFfneK7H2FRqo69rfzeS+Bs.
Are you sure you want to continue connecting (yes/no)? yes
Warning: Permanently added '[vagg-arm.nodejs.org]:2222,[203.217.18.114]:2222' (ECDSA) to the list of known hosts.
jump@vagg-arm.nodejs.org's password: 

[...]

Issue #1483, opened 2018-09-03
Local SSH configuration to access require.io hosts broken by Ansible 2.6


joaocgreis:
When running `ansible-playbook playbooks/write-ssh-config.yml` with Ansible 2.6, the resulting SSH config file contains the templates `{{ ansible_ssh_private_key_file }}` instead of the resolved value.

I bisected this to https://github.com/ansible/ansible/commit/dae737c8b714f2c6d28663ef0afeab10e6d3a667 . On our side, https://github.com/nodejs/build/blob/master/ansible/plugins/library/ssh_config.py seems to be able to receive the `hostinfo` argument and work with it, but with templates are unresolved.

I can't investigate further at the moment, so if anyone wants feel free to take this. This is also a good first issue for anyone looking to join the Build WG. It's easy to work with a local clone of Ansible by following https://github.com/ansible/ansible/blob/devel/hacking/README.md#env-setup .

PR #1890, opened 2019-08-16
update arm jump host proxycommand

rvagg:
Updating the jump host command, `-W` isn't working anymore due to some restrictions I've placed on the jump host but `nc` will work just fine now. Further to #1889.

This is not replacing `{{ ansible_ssh_private_key_file }}` in the .ssh/config output of playbooks/write-ssh-config.yml and hasn't been for a while but I don't know why. Would appreciate some help tracking this down if someone has time.

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -21,7 +21,7 @@ become_user         = root
 become_method       = sudo
 
 [hosts:requireio]
-ansible_ssh_common_args = -o ProxyCommand='ssh -i {{ ansible_ssh_private_key_file }} -W %h:%p -p 2222 jump@vagg-arm.nodejs.org'
+ansible_ssh_common_args = -o ProxyCommand='ssh -i {{ ansible_ssh_private_key_file }} -p 2222 jump@vagg-arm.nodejs.org nc -q0 %h %p'
 become_method           = sudo
 
 [hosts:smartos]


mhdawson:
Worked on the two release raspberry pis but not release-requireio-osx1010-x61-1

sam-github:
I'm checking this. Can someone confirm this is correct way to work-around the ansible bug:

`s/{{ ansible_ssh_private_key_file }}/~/ssh/nodejs_build_test/`

mhdawson:
I believe so I just have:

ProxyCommand ssh -i ~/.ssh/nodejs_build_release -p 2222 jump@vagg-arm.nodejs.org nc -q0 %h %p

sam-github:
@mhdawson Its a bit more complicated, they key that is substituted in has to be either ~/.ssh/nodejs_build_release or ~/.ssh/nodejs_build_test depending on the host... :-(. I guess that's why it would be good if ansible did this correctly, it should have that knowlege. I don't ATM have the time to do the context sensitive replacement (manually, with a script, or with ansible).

joaocgreis:
I investigated this a bit but wasn't able to take it all the way: https://github.com/nodejs/build/issues/1483

sam-github:
OK, so it would be better to use ansible 2.4? We should mention that in the README -- I wonder if some of the warnings I see from ansible-playbook are because I'm using an ansible version that doesn't match the one the playbooks were developed for.

rvagg:
> OK, so it would be better to use ansible 2.4

I avoid running it on my Mac these days where Python 3 is available and the Ansible that ships with brew has that baked in. I run from my Linux box which is still defaulting to Python 2. There's a bunch of other niggles in here that cause problems.

See https://github.com/nodejs/build/pull/1399 for the stalled effort to address this. Someone needs to take this over.

PR #1892, opened 2019-08-16
ansible,doc: add how to run a single command

joaocgreis:
Add an example of how to run a command directly with Ansible.

@sam-github I was reminded of this by your comment at https://github.com/nodejs/node/issues/25789#issuecomment-521301912. I think this is what you were looking for, hope this helps. There are frequently issues with some of the hosts, but this should take it most of the way.


joaocgreis:
Landed in https://github.com/nodejs/build/commit/3fc38164a03b6d12f0db49f140e77525b8928b44

PR #1905, opened 2019-08-27
fix: remove logo from coverage.nodejs.org

Trott:
Recent changes to the main website style sheet (which is loaded fover
coverage.nodejs.org) make the logo awfully big. The logo is superfluous
anyway, so instead of fixing it, let's just remove it so other problems
don't arise in the future.


rvagg:
will this be auto deployed on next coverage run?

XhmikosR:
We can readd the logo, it's a simple fix that can be applied on each case. Or we add back the dimensions in the CSS with a comment why they are needed.

Trott:
Argh! A different, but equally obtrusively large, logo is back. @XhmikosR If you want to fix this via CSS, that would be 💯 by me. I don't think the logo is particularly high-value on this page, but I also don't want to always be removing logos from this page for the rest of my life. 😀

XhmikosR:
I've fixed it in my PR and switched to the SVG logo, so check that out :)

PR #1918, opened 2019-09-17
jenkins: use devtoolset-6 on 10.x centos7-ppcle

sam-github:
10.x uses gcc-4.9 currently on ubuntu, but devtoolset-6 is a better
choice because it creates binaries that will run on older systems than
those created by ubuntu/gcc-4.9.

Since gcc-4.9 doesn't exist on centos7, ci is failing test builds
on centos7 for 10.x. Using devtoolset-6 will fix the failing builds.

### DIFF
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -24,7 +24,7 @@ if [ "$SELECT_ARCH" = "PPC64LE" ]; then
 
   case $NODE_NAME in
     *centos7* )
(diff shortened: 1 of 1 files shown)


sam-github:
ci: https://ci.nodejs.org/job/node-test-commit-plinux/27921/

sam-github:
ci passed using devtoolset-6: https://ci.nodejs.org/job/node-test-commit-plinux/27921/nodes=centos7-ppcle

PR #1919, opened 2019-09-18
jenkins: release 10.x on centos7-ppcle

sam-github:
ci is passing tests with 10.x on centos7-ppcle, see #1918, so start
building releases on it.

@nodejs/lts


sam-github:
Ok, @BethGriggs any ideas how I should test this? Perhaps 10.x nightlies tonight will be built on the new release machines, so I can run them? Or, @mhdawson the napi failures you found might go away with this change, since they are using the nightlies on ubuntu1404 machines.

sam-github:
I think switching to include might make things look weird, but from a different angle. And I'm getting used to exclude. Stockholm syndrome?

mhdawson:
Unless there is something new in the nightlies we won't get one.  I'll kick one off.

PR #1927, opened 2019-09-26
jenkins: use ccache with devtoolset on ppc64le

sam-github:
Because devtoolset setup works by putting the devtoolset bin/ on the
front of PATH the compiler is called directly instead of via the ccache
wrappers.

Change select-compiler.sh to explicitly use ccache, as it does for s390x.


sam-github:
It doesn't look like ubuntu was using ccache either for ppc64le

sam-github:
I reverted this, it broke node-test-commit-v8-linux: https://github.com/nodejs/build/issues/1940#issuecomment-537102717

I'll come back to it when I have time.

richardlau:
> 
> 
> It doesn't look like ubuntu was using ccache either for ppc64le

I found out via https://github.com/nodejs/node/pull/31628 that it was by virtue of ccache symbolic links.

PR #1929, opened 2019-09-27
ansible: improve Python 3 compatibility

cclauss:
As requested at https://github.com/nodejs/build/pull/1399#issuecomment-522687404, this PR cherrypicks key modifications from #1399 to improve the Python 3 compatibility of our build process.
* __git checkout refack/pipenv+python3 -- .flake8 .gitignore ansible/plugins/library/remmina_config.py ansible/plugins/library/ssh_config.py setup/www/tools/metrics/country-lookup.py jenkins/scripts/coverage/generate-index-html.py__
* __code ansible/plugins/inventory/nodejs_yaml.py__
    * cherrypick some change but not others
    * sync __valid["arch"]__ with https://github.com/nodejs/node/blob/master/configure.py
* __flake8 .__  # placate flake8

PR #1930, opened 2019-09-27
build: on push GitHub Action to lint Python code

cclauss:
Use GitHub Actions to lint Python code with flake8 on both Python 2 and Python 3.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,24 @@
+name: on push
+on: [push]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 2
+      matrix:
+        python-version: [2.7, 3.7]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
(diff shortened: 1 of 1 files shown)


sam-github:
gh actions are the new hotness, it seems, but I don't know that yet a third ci system added to jenkins and travis is really supportable, but we can leave that decision for a bit. I'm sure it will come up.

In the meantime, RSLGTM too .

rvagg:
👍 from me, we're not using travis here so we're not doubling up. The more we exercise GitHub Actions to understand its limitation the better.

rvagg:
this is failing https://github.com/nodejs/build/commit/303e626f9d550c70606d95e3cae6ba8287d93793/checks
But I see now that the commit has an additional comment that it's blocked by #1929 (this didn't show up in the PR btw @cclauss, just in the commit summary). If this is the case I suppose we can just wait for to land, presumably soon?

rvagg:
nope, #1929 already landed, so the failure is unexpected then. failed on pip upgrade pip:

```
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
```

cclauss:
Runtime issues resolved in #1934

PR #1934, opened 2019-09-30
GitHub Actions: Comment out requirements.txt for now

cclauss:
tap2junit can be added once https://github.com/nodejs/admin/issues/413 is resolved.


rvagg:
how do we get GitHub Actions to run on individual pull requests? It'd be nice to see this successfully passing in here. Is that a setting we need to enable somwehere?

PR #1935, opened 2019-09-30
Update nodejs.org config

### LINKED ISSUE #1932: Coverage website looks weird
https://coverage.nodejs.org/ looks odd, the header image is huge and styles are mostly missing. I think this is happening because https://nodejs.org/layouts/css/styles.css 404s, but I can’t tell whether that’s a build or a website issue (there were some recent changes to the nodejs.org repo that likely affect this but I’m not sure where the fix should end up, here or there). /cc @XhmikosR

XhmikosR:
Redirect `/layouts/css/styles.css` to `/static/css/styles.css`

**This is untested!**

Fixes #1932 and similar issues.

/CC @rvagg @richardlau @addaleax

### DIFF
--- a/setup/www/resources/config/nodejs.org
+++ b/setup/www/resources/config/nodejs.org
@@ -354,6 +354,7 @@ server {
     rewrite ^/feed/vulnerability/?$                               https://$server_name/en/feed/vulnerability.xml permanent;
 
     # Asset rewrites
+    rewrite ^/layouts/css/styles\.css$                            https://$server_name/static/css/styles.css permanent;
     rewrite ^/(?!static/)?favicon\.ico$                           https://$server_name/static/images/favicons/favicon.ico permanent;
     rewrite ^/apple-touch-icon.*\.png$                            https://$server_name/static/images/favicons/apple-touch-icon.png permanent;
     rewrite ^/trademark-policy.pdf$                               https://$server_name/static/documents/trademark-policy.pdf permanent;


setup/www/resources/config/nodejs.org
```diff
@@ -354,8 +354,10 @@ server {
     rewrite ^/feed/vulnerability/?$                               https://$server_name/en/feed/vulnerability.xml permanent;
 
     # Asset rewrites
+    rewrite ^/layouts/css/styles\.css$                            https://$server_name/static/css/styles.css permanent;
     rewrite ^/(?!static/)?favicon\.ico$                           https://$server_name/static/images/favicons/favicon.ico permanent;
-    rewrite ^/apple-touch-icon.*\.png$                            https://$server_name/static/images/favicons/apple-touch-icon.png permanent;
+    rewrite ^/(?!static/)?favicon\.png$                           https://$server_name/static/images/favicons/favicon-32x32.png permanent;
+    rewrite ^/(?!static/)?apple-touch-icon.*\.png$                https://$server_name/static/images/favicons/apple-touch-icon.png permanent;
```
rvagg: why did this one need to change? was there apple-touch-icon*.png files in other directories other than / and /static/?

XhmikosR: There are requests to `https://nodejs.org/static/apple-touch-icon.png` which isn't handled without this change. I noticed this on https://github.com/nodejs/build/pull/1936/files#diff-5322f429725c524a1cbc315351962ec4L24

XhmikosR: Actually, wait I think my regex is wrong

XhmikosR: We basically need the regex to work for paths with and without `static`

XhmikosR:
@rvagg can you please check again?

Basically we need both `https://nodejs.org/apple-touch-icon.png` and `https://nodejs.org/static/apple-touch-icon.png` to redirect to `https://nodejs.org/static/images/favicons/apple-touch-icon.png`.

Similarly for `favicon.ico` and `favicon.png`.

rvagg:
Sorry but I'm going to have to defer this for ~8h or so unless someone else with access and confidence in this code wants to do it. I'm heading for bed and breaking the website when I clock off is not something I want to be responsible for.

XhmikosR:
Alright, I kept only the CSS redirect in this PR so that we reduce the chances to break something. I'll make another PR for the rest of the redirects.

rvagg:
live

PR #1933, opened 2019-09-30
Update path to style sheet for coverage.nodejs.org

richardlau:
Fixes; https://github.com/nodejs/build/issues/1932
Refs: https://github.com/nodejs/nodejs.org/pull/2544

PR #1936, opened 2019-09-30
Update generate-index-html.py

XhmikosR:
nodejs.org backports and fixes

~~So, if the index.html is autogenerated, does this mean that https://github.com/nodejs/benchmarking/pull/289 and https://github.com/nodejs/benchmarking/pull/290 are moot?~~

/CC @Trott @richardlau

### DIFF
--- a/jenkins/scripts/coverage/generate-index-html.py
+++ b/jenkins/scripts/coverage/generate-index-html.py
@@ -9,23 +9,23 @@ with open('out/index.csv') as index:
 with open('out/index.html', 'w') as out:
   out.write('''
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <meta charset="utf-8">
     <title>Node.js Code Coverage</title>
 
-    <link rel="dns-prefetch" href="http://fonts.googleapis.com">
-    <link rel="dns-prefetch" href="http://fonts.gstatic.com">
+    <link rel="dns-prefetch" href="https://fonts.googleapis.com">
+    <link rel="dns-prefetch" href="https://fonts.gstatic.com">
 
     <meta name="author" content="Node.js Foundation">
     <meta name="robots" content="index, follow">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
https://github.com/nodejs/build/pull/1936/files?w=1

Live preview: <https://stoic-agnesi-bfb346.netlify.com/>

Should be ready now. Feel free to squash when merging

richardlau:
> 
> 
> nodejs.org backports and fixes
> 
> So, if the index.html is autogenerated, does this mean that [nodejs/benchmarking#289](https://github.com/nodejs/benchmarking/pull/289) and [nodejs/benchmarking#290](https://github.com/nodejs/benchmarking/pull/290) are moot?
> 

AFAIK benchmarking.nodejs.org and coverage.nodejs.org are separate sites.

XhmikosR:
You are right, I confused it with coverage. Sorry, I jumped to this as fast as I could and I needed more coffee.

But I think everything should be fine now. I've submitted a couple of PRs to the other 2 projects.

XhmikosR:
@Trott here it is.

This PR includes #1933, so better get that merged first.

nschonni:
@XhmikosR would there be a lot of CSS changes to use an HTML table? Just saw you fix some of the semantics in here, but I can also just raise an issue on the coverage repo if that makes more sense

XhmikosR:
If I did change the semantics that would be a bug :)

I just normalized stuff and fixed the HTML validation errors for the time being. See the non-whitespace diff https://github.com/nodejs/build/pull/1936/files?w=1

That being said, I think that first we need to come up with a solution for the CSS. I don't think it's a good practice to load the nodejs.org CSS because it can break at any time like it already has (I wasn't even aware this was a thing and people did this :P)

After that, we can switch to a table for sure. The amount of CSS depends on how flexible we can be with styling, but from a quick look, I don't think it needs a lot.

I could try doing this in the next days assuming I find a way to develop locally. Just make an issue and CC me and we'll pick it up there.

nschonni:
> If I did change the semantics that would be a bug :)

I just meant fixing the `<main>` :)
I'll open up an issue on this repo

mhdawson:
Don't think this is controversial so landing so we get fixup sooner than later.

mhdawson:
Somehow the UI  landed before I could paste the meta data, sorry.

PR #1937, opened 2019-09-30
Nodejs org redir

XhmikosR:
**Needs review and #1935**

I notice we have `https://nodejs.org/static/apple-touch-icon.png` or `https://nodejs.org/static/favicon.png`. This patch attempts to handle those requests and redirect them to the existent files.


rvagg:
for reference, this is one location where some of those were used https://github.com/nodejs/build/pull/1936/files#diff-5322f429725c524a1cbc315351962ec4L24, linked from discussion in #1935

PR #1939, opened 2019-10-01
Attempt to fix GitHub language detection

XhmikosR:
This is the recommended way of solving the issue, but AFAICT it doesn't work, at least not on my fork.

Basically, because the config filenames end with `.org` GitHub autodetects them as Org https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3576

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/setup/www/resources/config/*.org linguist-language=Nginx


rvagg:
will try it on master and force push it out if it doesn't work

XhmikosR:
I think GitHub caches this stuff, so I'm not sure when it's invalidated. Maybe leave it there for a while and see if it helps. Otherwise we can contact GH support, but I do believe the root cause is this.

rvagg:
nope, didn't work, but there is an alternative - rename the files, they don't need to be named as they are, see setup/www/tasks/nginx.yaml where they are copied into place, the names are somewhat arbitrary and could just have an appropriate suffix.

XhmikosR:
Alright, I'll leave that to someone else who knows the codebase better.

PR #1942, opened 2019-10-02
ansible: update debian8 support

rvagg:
I reprovisioned test-rackspace-debian8-x64-1 due to some odd descripancies pointed out in https://github.com/nodejs/citgm/pull/751, I had to fix a few minor things to get it to run fully in ansible now that debian8 is EOL. I'm unsure if the descripancies are entirely sorted out, it's difficult to determine but maybe a citgm run or two could tell.

We should probably phase debian8 out, but I wasn't brave enough to just pull the rug unanimously, so restoring support was probably worthwhile.

### DIFF
--- a/ansible/roles/baselayout/tasks/partials/repo/debian8.yml
+++ b/ansible/roles/baselayout/tasks/partials/repo/debian8.yml
@@ -1,8 +1,8 @@
 ---
 
-- name: "debian8 | add debian-backports" # has systemd-coredump if needed
+- name: "debian8 | add jessie-backports" # has systemd-coredump if needed
   apt_repository:
-    repo: 'deb http://ftp.debian.org/debian jessie-backports main'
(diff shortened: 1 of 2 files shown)


rvagg:
oh, not EOL, my mistake, we have till mid-2020: https://wiki.debian.org/LTS

rvagg:
And, I discovered something along the way, the Oracle Java binaries are now not something we can support with Ansible. We did use the webupd8 ppa for Ubuntu and Debian but it's been discontinued as discussed in the update at the top of this post http://www.webupd8.org/2012/09/install-oracle-java-8-in-ubuntu-via-ppa.html

For this one I downloaded it manually, unpacked and symlinked it in the right places (`/usr/bin/java` is the critical piece for our systemd configs).

Not sure the best path here, probably need to migrate fully to openjdk I suppose. I think there were good reasons we were using the Oracle binaries but they've done their Oracle thing and made that unworkable.

PR #1945, opened 2019-10-05
ansible: upgrades for docker hosts (1804 hosts & containers, remove debian7-armv7, swap Alpine 3.8 for 3.10)

rvagg:
This is a WIP, I'm still working through it. There will be an accompanying secrets update too.

* Change Docker hosts to Ubuntu 18.04, x64 and the ARMv7 ones on Scaleway, they were all Ubuntu 16.04 and it's time to update.
* Change x64 Docker containers to be mainly Ubuntu 18.04, this mainly impacts the sharedlibs containers. They've all been 16.04.
* Remove Wheezy / Debian 7 from the ARMv7 Scaleway hosts. It's becoming too difficult to keep them properly updated and Node 8 is nearly EOL so there shouldn't be any great risk in this.
* Introduce Alpine 3.10 and remove 3.8, switch alpine-latest-x64 and alpine-last-latest-x64 in Jenkins appropriately.

Need to upgrade the rests of the hosts but I'm running basic tests now with these new configs before fully committing.


rvagg:
I think this might be all done now, docker hosts all on 18.04 which has mostly meant destroying the old ones and provisioning new ones from scratch - there were different reasons in each provider that meant it was optimal to do it that way rather than reimaging what was there.

ansible/plugins/filter/filters.py
```diff
@@ -43,11 +43,6 @@ def match_key(value, dictionary, raise_error=True, feedback_name='os'):
     return False
 
 
-def latest_version(versions):
-    versions.sort(key=lambda s: map(int, s.split('.')))
```
rvagg: THIS was the Python3 problem that's been haunting me, stopping me from doing any of this Docker update stuff on my Mac, having to do it all on my Linux machine instead. I've lost the error, although it's been mentioned a couple of times in this repo already, something about `<` not being compatible with `map`.
So I've avoided the Python2/3 dumpster fire entirely and used good 'ol reliable Bash to get the job done, see below.

rvagg:
merging because https://github.com/nodejs-private/secrets/pull/47 got signoffs and this is live in our infra now anyway

BethGriggs:
Since roughly around the time that this PR landed I seem to be getting test failures on `alpine-last-latest-x64` for v10.x. Should CI be running on `alpine-last-latest-x64` for v10.x? The test failures are `parallel/test-process-euid-egid`, `parallel/test-process-uid-gid` (https://ci.nodejs.org/job/node-test-commit-linux/30230/console). I took a quick look to see if the tests are missing some fixes on v10.x but it didn't look like that was the case.

rvagg:
ouch .. the configs for `alpine-last-latest-x64` should be equivalent to what were used as `alpine-latest-x64`, this just shuffled them down a version. But, because they are now running on an Ubuntu 18.04 host, the kernel version will have changed, that could be what's to blame here. This might require its own issue on nodejs/node, I'll have a look and open one if it's reliable repeatable.

PR #1947, opened 2019-10-09
coverage: add the needed CSS from the main site

XhmikosR:
This unties the site from the upstream CSS

**I need some help actually copying the CSS file. Any help welcome.**

Preview: <https://relaxed-ptolemy-556131.netlify.com/>

Refs https://github.com/nodejs/nodejs.org/pull/2635

TODO:

- [ ] Adapt the python script to copy styles.css too
- [ ] Move the other assets too like favicons, logo etc?

Note that the CSS is only the required. Which means that if more elements are added at some point, their rules should be added manually here again.

### DIFF
--- a/jenkins/scripts/coverage/generate-index-html.py
+++ b/jenkins/scripts/coverage/generate-index-html.py
@@ -3,6 +3,10 @@
 
 import datetime
 
+with open('jenkins/scripts/coverage/styles.css') as in_file:
+  with open('out/styles.css') as out_file:
+    out_file.write(in_file.read())
+
 with open('out/index.csv') as index:
   index_csv = filter(lambda line: line, index.read().split('\n'))
 
--- a/jenkins/scripts/coverage/generate-index-html.py
+++ b/jenkins/scripts/coverage/generate-index-html.py
@@ -25,50 +25,7 @@ with open('out/index.html', 'w') as out:
     <link rel="icon" sizes="32x32" type="image/png" href="https://nodejs.org/static/images/favicons/favicon-32x32.png">
 
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600&display=fallback">
-    <link rel="stylesheet" href="https://nodejs.org/static/css/styles.css">
-    <style>
-        #logo { margin-bottom: 1rem; }
-        #badge {
-          margin-top: 1rem;
-          margin-bottom: 1rem;
-        }
-        #badge, a {
-          border: none;
-        }
-        main { margin-bottom: 2rem; }
-        .table-header,
-        .table-row {
-          box-sizing: border-box;
-          display: flex;
-          width: 100%;
-          padding: 2px 10px;
-        }
-        .table-header { font-weight: bold;}
-        .table-header > div,
-        .table-row > div {
-          flex-grow: 1;
-          width: 100px;
-        }
-        .table-row:nth-child(even) { background-color: #eee; }
-        .sha .cell-value { font-family: Consolas, "Liberation Mono", Menlo, Courier, monospace; }
-        .cell-header { display: none; }
-        @media screen and (min-width: 690px) and (max-width: 850px) {
-          .table-header > div:nth-child(n+3),
-          .table-row > div:nth-child(n+3) {
-            flex-grow: 0.2;
-          }
-          .table-header > div:first-child,
-          .table-row > div:first-child {
-            flex-grow: 0.4;
-          }
-        }
-        @media screen and (max-width: 690px) {
-          .cell-header { display: block; font-weight: bold; }
-          .table-header { display: none; }
-          .table-row { display: block; }
-          .table-row > div { width: 100%; text-align: center; margin-bottom: 12px;}
-        }
-    </style>
+    <link rel="stylesheet" href="styles.css">
   </head>
   <body>
   <header>
--- a/jenkins/scripts/coverage/styles.css
+++ b/jenkins/scripts/coverage/styles.css
@@ -1,0 +1,227 @@
+html {
+  scroll-behavior: smooth;
+}
+
+@media (prefers-reduced-motion: reduce) {
+  html {
+    scroll-behavior: auto;
+  }
+}
+
+body {
+  box-sizing: border-box;
+  font: 400 20px/1.5 "Source Sans Pro", "Open Sans", Roboto, "San Francisco", Helvetica, Arial, sans-serif;
+  color: #333;
+  margin: 0;
+  background-color: #fff;
+  display: -webkit-box;
+  display: flex;
+  -webkit-box-orient: vertical;
+  -webkit-box-direction: normal;
+  flex-direction: column;
+  min-height: 100vh;
+}
+
+header,
+#main,
+footer {
+  display: -webkit-box;
+  display: flex;
+}
+
+h1,
+h3 {
+  font-weight: 400;
+}
+
+a,
+a:link,
+a:active {
+  color: #43853d;
+  text-decoration: none;
+  border-radius: 2px;
+}
+
+a:hover {
+  color: #fff;
+  background-color: #43853d;
+}
+
+strong {
+  font-weight: 600;
+}
+
+p a {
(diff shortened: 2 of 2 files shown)


XhmikosR:
Can someone help me out with the above 2 TODO? Mostly the python script change to include the new styles.css.

XhmikosR:
I still think we should also move the other assets like the logo and favicons so that this is completely independent from the main site. It will also be faster since there won't be any redirects.

XhmikosR:
@rvagg maybe you can give me a hand here?

rvagg:
python aint' my thing, the original author isn't around anymore, and we've lost a bunch of Python folks, so I'm not sure who else to call on other than @cclauss.

XhmikosR:
Can we get some help here please?

cclauss:
> Adapt the python script to copy styles.css too

Which Python script?

XhmikosR:
I guess https://github.com/nodejs/build/blob/master/jenkins/scripts/coverage/generate-index-html.py?

cclauss:
Something like:
´´´python
with open('jenkins/scripts/coverage/styles.css') as in_file:
  with open('out/styles.css') as out_file:
    out_file.write(in_file.read())
```

XhmikosR:
Can you push to this branch? People with push rights can push to a fork's PR branch.

I cannot try it nor confirm the fix, hence why I suggest someone else to do it.

cclauss:
That code seem like the desired functionality? I am on an iPhone so I will not push for several hours.

XhmikosR:
There's no hurry :) I just want someone to test this beforehand, and also take the responsibility of the change :P

mhdawson:
The similar change for benchmarking broken things because the SCP copies need to be updated as well. Just an FYI that before this lands we should double check if a similar change is needed.

.

XhmikosR:
So, any news? Anyone who can help with the python script?

rvagg:
@nodejs/coverage-admins @nodejs/testing we seem to have a void in our working knowledge of how coverage.nodejs.org is built and updated. benchmarking.nodejs.org is similar but gets more attention. This PR appears to be touching code that was originally authored by folks who are no longer with us or whose attention is elsewhere. Perhaps we need to do some more radical surgery here and replace what's being done here with code that more people understand and can maintain? (IMO the choice of Python for our tooling should be a last resort, we don't have many active folks who are handy enough with Python). Is coverage.nodejs.org even something that should be actively maintained as it is or are we migrating to this new third-party hosted tool?

Trott:
> Is coverage.nodejs.org even something that should be actively maintained as it is or are we migrating to this new third-party hosted tool?

I think we need to keep it for a while, although it would be nice to eventually rely on codecov.io instead, I think. @bcoe

cclauss:
Sorry.  I will complete this one.

I do agree that in general we should keep converting Python code to JS code.

XhmikosR:
Should I move the rest of the assets like favicons to the repo too?

bcoe:
@rvagg @Trott I think we can consider migrating to codecov.io sometime in the new year, but there are a few bugs on their end that I'm working with them to sort out:

1. we don't yet see block level coverage.
2. we need to figure out how to combine our C++ and JavaScript coverage output (this feature might already exist, but @Trott and need to wire it up).
3. it seems like we lose line counts when reports are merged, it would be nice if we could see both how many times a line is executed, and which uploaded coverage it's executed for.

CC: @thomasrockhu, do you think you'll have any cycles in the new year to address the block coverage issues?

Trott:
This is blocking other work (particularly https://github.com/nodejs/nodejs.org/pull/2635) so I'm adding to the meeting agenda in the hopes that we can move this forward.

XhmikosR:
Do note that the current live site on https://coverage.nodejs.org has its footer broken already.

sam-github:
I'm not clear on what it takes to move this forward, or who is involved with it.

@mhdawson do you know who uses coverage? Does this help? Can we give someone who is interested enough permissions to maintain it, like @XhmikosR ?

XhmikosR:
I don't want to maintain it :P

This is just the proper way moving forward so that child sites do not use the upstream CSS thus avoiding breakage.

This patch just needs a few lines in the python script to copy the newly added styles.css file.

Personally, I'd even add all the assets like the favicons too, but that's for another day.

sam-github:
So, I've never personally looked at the coverage site (until an hour ago). I've no idea who maintains it, or how it works, what a "child site" is, what the "upstream CSS" is, or what would happen if this PR merged (if anything, maybe some kind of follow steps are needed?). 

I'm sorry, I'm sure that is not so satisfying, but basically, you'll need to find someone who is involved with maintaining that site to review this. Do you know who the right person to bug is?

If the site isn't maintained by anybody, it unfortunately might need to be removed, or just left as-is if it offers some value.

XhmikosR:
I don't understand why "I" have to do all this?

I simply submitted a patch to fix something that's known to break.

I don't have the time nor the knowledge to do something more.

cclauss:
@XhmikosR I can not push to your fork but if you can please paste [this code](https://github.com/nodejs/build/pull/1947#issuecomment-547775609) __just above line 6__ of the Python file?  We can then see if that achieves your objective.

XhmikosR:
Weird, if you have push rights to the build repo you should be able to push
to the pull request branch. I can try tomorrow although the path seems off
judging by the other code (Jenkins isn't used in the other places)

cclauss:
I am not in @nodejs/build but I believe that you and I can work this together.

XhmikosR:
@cclauss done, can you test it?

cclauss:
It is not clear to me how to test this because `jenkins/scripts/coverage/generate-index-html.py` needs to be executed in a Jenkins run and this PR only does GitHub Action runs, not Jenkins runs.

Ideas from the other on how to test this change?

rvagg:
this thread keeps on making it clear that there is no owner for this stuff. It's probably time to remove all of the scripting, and probably styling (directory listings are perfectly usable), from it, anything that carries a maintenance burden that there's nobody to shoulder.

XhmikosR:
Maybe @mhdawson knows more?

Either way, this patch is a good change for as you can see the current live site has its footer already broken.

mhdawson:
I don't really think we need to go deleting things. The main reason I've not landed this is that nothing is currently broken, the PR is a suggested improvement.  

When I land I want to have time to validate things are ok.  That has just not lined up in terms of having to do it at a good time, and since nothing is actually broken I've not made it a priority.

mhdawson:
Landing this may just work, but will have to check first the scripts copy everything from the directory of specific files.  

To be honest given that the end goal is to migrate to codecov.io I'd prefer to just leave things as are as opposed to having to spend time to update and possibly investigate if things are broken.

mhdawson:
I'll also add that I've not been completely ignoring the coverage site as I recently did a cleanup after it was reported that we were not getting data...

cclauss:
What was https://github.com/nodejs/build/pull/1947#issuecomment-575237313 all about??

mhdawson:
@cclauss that is newish info that I'd not picked up.

mhdawson:
I just checked and the full out directory should be copied over.  That means that this should likely be ok.

Given that I suggest we just land and then check its ok (I think it runs 4 times a day from the cron job). Not a great time since tomorrow is my last day before heading off on vacation but I assume somebody else can back it out if there is a problem and I can't get to it.

mhdawson:
Coverage run so that we find out if its ok sooner than later: https://ci.nodejs.org/view/Node.js%20Daily/job/node-test-commit-linux-coverage-daily/415/

mhdawson:
The copy failed unfortunately. Going to move copy of styles file to the ci job instead.

mhdawson:
New CI do see if its ok now: https://ci.nodejs.org/view/Node.js%20Daily/job/node-test-commit-linux-coverage-daily/

mhdawson:
One more time after tweaking job to get the copy to the right directory: https://ci.nodejs.org/view/Node.js%20Daily/job/node-test-commit-linux-coverage-daily/417/

mhdawson:
All good now.

PR #1948, opened 2019-10-09
Losslessly compress ns-arm-cluster.jpg.

XhmikosR:
Use `jpegoptim -sp`.

116227 -> 112718 bytes


sam-github:
Is making it 3%  smaller going to make a difference? Where is this file used?

XhmikosR:
In README.md. 3% is not so much, I know. I just personally don't like having extra bytes, that is all.

Issue #1955, opened 2019-10-11
OSX builds take _very_ long


BridgeAR:
Each release build is waiting on OSX to finish. All other builds finish significantly faster than OSX. It would be great if this could be looked at so that they finish around about at the same time.

rvagg:
I believe this is fixed, installed ccache manually, also #1956. Still a bit more work needed on the ansible scripts for macos, things have marched on a bit further than when it was last used I think.

BridgeAR:
Thanks a lot!

PR #1956, opened 2019-10-14
ansible: macos: /usr/localopt->/usr/local/opt & install ccache

rvagg:
I'm not sure of the backstory here, this path was introduced in the original PRs:

Ref: https://github.com/nodejs/build/pull/971
Ref: https://github.com/nodejs/build/pull/1347

But afaik /usr/localopt has never been a thing. /usr/local/opt is where this stuff is at.

The macos machines all seem to be set up to have /usr/localopt in start.sh and /etc/paths, but because /usr/local/opt is in there I believe it just _mostly works_.

https://github.com/nodejs/build/issues/1955 reports slow release builds on macos, but that's because there's no ccache. I tried running the playbook but it failed at doing something in /home, so there's more work to be done on that. I don't even know if it installs ccache by `brew` but I did that manually for now to get it going.

### DIFF
--- a/ansible/roles/baselayout/tasks/main.yml
+++ b/ansible/roles/baselayout/tasks/main.yml
@@ -95,9 +95,9 @@
     - "{{ packages[os|stripversion]|default('[]') }}"
     - "{{ common_packages|default('[]') }}"
 
-- name: Check whether /etc/paths contains "/usr/localopt/ccache/libexec" (macos)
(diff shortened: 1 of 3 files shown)


rvagg:
Added a new change that installs `ccache` as well, I don't believe it's anywhere in the homebrew install process on these machines.

With https://github.com/nodejs-private/secrets/pull/48 and these changes I get through an ansible run on release-macstadium-macos10.11-x64-1.

PR #1962, opened 2019-10-16
doc: move aix7.2 setup docs

rvagg:
@sam-github can you fill me in on the purpose & status of this doc? it refers to the aix61 ansible setup but we also seem to have aix7.1 and aix7.2 setup stuff in ansible/ already? Is this doc old and redundant now or worth preserving? If so, I think a more appropriate place is doc/non-ansible-configuration-notes.md like I've done in this PR.

Ref: #1620


doc/non-ansible-configuration-notes.md
```diff
@@ -216,3 +216,95 @@ Also ensure that /usr/local/bin is included in the `JENKINS_PATH` set in /etc/in
 [`ansible.intro_windows`]: http://docs.ansible.com/ansible/intro_windows.html
 [newer Ansible configuration]: https://github.com/nodejs/build/tree/master/ansible
 [stand-alone]: https://github.com/nodejs/build/tree/master/setup/windows
+
+## AIX 7.2
+
+To set up, basically:
+
+1. Run ansible
+2. Run through ansible/aix61-standalone/manualBootstrap.md, doing the steps marked as
```
sam-github: That file doesn't exist!

rvagg: does now thanks to #1620
https://github.com/nodejs/build/blob/master/ansible/aix61-standalone/manualBootstrap.md

Issue #1965, opened 2019-10-17
No nightly tarballs for AIX


gabrielschulhof:
In https://ci.nodejs.org/job/node-test-node-addon-api-new/nodes=aix61-ppc64/781/console it looks like it's failing to download the appropriate Node.js release tarball.

Looks like the nightlies for AIX have been missing for a short while: https://github.com/nodejs/build/issues/1957#issuecomment-543013729

rvagg:
missing for 13.0.0-rc\* builds too. @sam-github do you have access to the release machine?

sam-github:
Yes I have access, and I believe/hope this is a symptom of https://github.com/nodejs/build/issues/1957#issuecomment-543460591

sam-github:
duplicate of https://github.com/nodejs/build/issues/1957

sam-github:
I built nightlies for 13.x: https://github.com/nodejs/build/issues/1957#issuecomment-543893348

Is that what you need @gabrielschulhof ?

gabrielschulhof:
@sam-github I'll check if it causes the build for https://github.com/nodejs/node-addon-api/pull/562 to go green.

gabrielschulhof:
@sam-github looks like it moved past the point where it was failing to find the nightly tarball.

gabrielschulhof:
Looks like the build will be green 👍 Thanks!

PR #1966, opened 2019-10-18
promote: add expected_assets for v13.x (and v12.x)

rvagg:
v12.x is on the server but never made it here for some reason. This adds it and also adds v13.x expected assets as a copy of v12.x. @nodejs/releasers you're going to be prompted with this when promoting v13.x releases, are you, or @nodejs/build, aware of any proposed or actual change in distributed assets for 13.x? afaik there are none, the ICU thing was resolved by just changing how the main binaries are build, not adding anything new.


rvagg:
Running this against v13.0.0-rc.2 directories gives:

```
... Checking assets
... Expecting a total of 39 assets for v13.x
... 0 assets waiting in staging
... 38 assets already promoted
 ⚠  The following assets are expected for v13.x but are currently missing from staging:
    • node-v13.0.0-rc.2-aix-ppc64.tar.gz
    Promote if you are certain this is the the correct course of action
```

Hopefully the AIX thing is just #1965.

Issue #1964, opened 2019-10-17
Supporting multiple versions of xcode


AshCripps:
From disscusion here: https://github.com/nodejs/node/pull/29493#issuecomment-530862404

What is the opinion on multple xcode version?

From a bit of research you can use xcode-select to change versions inside select-compiler.sh but this requires sudo root to do so which the iojs user on the build machines does not have.

This gives two options:
 * Edit the sudoers file to whitelist xcode-select for the iojs user, and then edit select-complier to call xcode-select.
or
* Have multiple machines running different versions of xcode to run the builds

rvagg:
I think it's safe to assume a single build will occur on a machine at one time. To be safe we should probably have a default version that is always set at the end of a run regardless of what happens, because there are numerous jobs that use the osx machines other than standard test builds, like citgm and libuv, it'd be annoying for them to have one compiler for one run and another for the next.

We could create a sudoers special-case for xcode-select, we already do that on some hosts that use Docker to switch environments, giving narrow access to a certain command https://github.com/nodejs/build/blob/ee7d97fb9d7d7f3e647b4ff39d1e2c1dee6b6828/ansible/roles/jenkins-worker/tasks/partials/docker-host.yml#L34-L38

More machines will be a bigger problem, so we should try and solve this for single machines. We need to handle this for release builds too.

rvagg:
How do we *get* the multiple versions of xcode onto these machines? We currently have to resort to this abomination for older versions already https://github.com/nodejs/build/blob/master/ansible/roles/package-upgrade/files/install-xcode.sh, only George has experience with that one, so I guess one of us needs to upskill here, upgrade it for 10.9, 10.10, and get ansible to do the multiple-versions thing.

sam-github:
If we are going the sudoers and less machines route, then @AshCripps is going to take a shot at figuring this out. He's not in the buildwg yet, not even a node collaborator, but he's working on our team and once he's made enough contributions to demonstrate his usefulness, we'll nominate him.

rvagg:
oh, awesome, help managing our macstadium setup is certainly something we're in need of

AshCripps:
Old versions of xcode can be found here - https://developer.apple.com/download/more/ 

You can scroll through the list select just the command line tools options and grab the link from there.

The link can be curled in the same way as is done in install-xcode.sh

[...]

PR #1967, opened 2019-10-18
[WIP] ansible: Edit select-compiler.sh to select different versions of xcode

AshCripps:
Edited select-compiler.sh to change the xcode version used based on which version of node is being built.

Works on my local machine but needs to be checked to ensure the paths are correct on the build machines. 

Will also require sudoers being edited to allow for xcode-select to run with sudo permissions.

Discussion here - https://github.com/nodejs/build/issues/1964

### DIFF
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -13,6 +13,8 @@
 # way to do ccache is to push `/path/to/gcc-version` on to the front of PATH,
 # then push a `/path/to/ccache/wrappers` in front of the compiler path.
 
+
+
 if [ "$DONTSELECT_COMPILER" != "DONT" ]; then
   NODE_NAME=${NODE_NAME:-$HOSTNAME}
   echo "Selecting compiler based on $NODE_NAME"
@@ -22,14 +24,15 @@ if [ "$DONTSELECT_COMPILER" != "DONT" ]; then
     *aix* ) SELECT_ARCH=AIXPPC ;;
     *x64* ) SELECT_ARCH=X64 ;;
     *arm64* ) SELECT_ARCH=ARM64 ;;
+    *macos* ) SELECT_ARCH=MACOS ;;
   esac
 fi
 
 # get node version
-if [ -z ${NODEJS_MAJOR_VERSION+x} ]; then
-  NODE_VERSION="$(python tools/getnodeversion.py)"
-  NODEJS_MAJOR_VERSION="$(echo "$NODE_VERSION" | cut -d . -f 1)"
-fi
+ if [ -z ${NODEJS_MAJOR_VERSION+x} ]; then
+   NODE_VERSION="$(python tools/getnodeversion.py)"
+   NODEJS_MAJOR_VERSION="$(echo "$NODE_VERSION" | cut -d . -f 1)"
+ fi
 
 if [ "$SELECT_ARCH" = "PPC64LE" ]; then
   # Set default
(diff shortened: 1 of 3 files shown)


jenkins/scripts/select-compiler.sh
```diff
@@ -112,4 +113,23 @@ elif [ "$SELECT_ARCH" = "AIXPPC" ]; then
...
+  elif [ "$SELECT_ARCH" = "MACOS" ]; then
+
+  echo "Setting compiler for Node version $NODEJS_MAJOR_VERSION on macos"
+
+  if [ "$NODEJS_MAJOR_VERSION" -ge "13" ]; then
+    # set xcode version to latest
+    sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
+    echo "set xcode to latest - At least version 10"
+  elif [ "$NODEJS_MAJOR_VERSION" -ge "12" ]; then
+    sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
```
mhdawson: How does this select a different version? ` /Applications/Xcode.app/Contents/Developer` looks identical to the section for 13?

AshCripps: Atm it is identical becauase 12 and 13 both build on 10 +. The actual min. xcode doesnt change to version 10 until node 13 (according to BUILDING.md) so the question is do we want to use latest for these versions or hardcode in a stable version and only change it if required (security patches etc.). 

For example `/Applications/Xcode.11.1.app/Contents/Developer` for node 13 and then something like `/Applications/Xcode10.x.app/Contents/Developer` for node 12. 

Depends if we want as few xcodes on the machine as possible or if we want a stable xcode for each node release?

sam-github: stable for each release, decided at time of first `.0.0` for that release line, never changed afterwards.

jenkins/scripts/select-compiler.sh
```diff
@@ -112,4 +113,23 @@ elif [ "$SELECT_ARCH" = "AIXPPC" ]; then
...
+    echo "set xcode to latest - at least version 8"
```
sam-github: I think prev line should have been 8

jenkins/scripts/select-compiler.sh
AshCripps: Are these documented anywhere as to which version is used for each release?

sam-github:
@AshCripps Would be good to add:
1. updates to https://github.com/nodejs/build/blob/master/ansible/MANUAL_STEPS.md#macos to describe how the sudoers needs to be setup
2. setup... how we will get the xcodes on the machine. There is a setup script: ./ansible/roles/package-upgrade/files/install-xcode.sh , can it do multiple versions?

AshCripps:
>     2. setup... how we will get the xcodes on the machine. There is a setup script: ./ansible/roles/package-upgrade/files/install-xcode.sh , can it do multiple versions?

ATM it currently does 2 differenet versions based on what version of macos is on the box. So it should be easy enough to change it ive already tracked down where URLs for the different versions are so should be a simple change to install multiple versions

jenkins/scripts/select-compiler.sh
sam-github: in the BUILDING.md on the relevant release branch of nodejs/node, if documented, if not.... https://github.com/nodejs/build/issues/1522 :-(

rvagg: We're locked in to 8 for Node 12. We've historically locked in for the entirety of a release line. That could change and it would be lower risk on macOS than others, but not zero risk and the TSC would have to sign off on it and Build WG would shoulder the blame if things break. So .. for now consider it locked to Xcode 8 for Node 12.

AshCripps:
https://github.com/nodejs/build/blob/97351053b2e723a0ec138c4fa3c56d6109913c3f/ansible/roles/package-upgrade/files/install-xcode.sh#L7-L15

The above lines don't work on the new nearform machines so the install-xcode script will have to be reworked to install xcode on the newer machines.

Also I found this blog post https://derflounder.wordpress.com/2018/06/10/updated-xcode-command-line-tools-installer-script-now-available/ which looks remarkably similar to what we have and the original file predates install-xcode.sh by about 3 years but the repo is under the MIT license but I don't see any link to the author or repo?

AshCripps:
Ive edited the install-xcode script to allow for the installing of xcode - there are some problems here
* Firstly the xcodes will need to be hosted on a server somewhere - the apple site requires a login to download the files and this is tricky on command line
* Secondly it seems the xip files have certificates that expire as some old downloads I have from the other week I can no longer unzip
* These are full blown xcodes on the new oses - this is required for notorization but also makes xcode-select easier IMO but they are big files ~10GB

Ive also edited the MANUALSTEPS.md to include the sudo line needed to run xcode-select

Next step is to sort out the select compiler script to change xcodes based on the OS version and what version of node is being built

rvagg:
We have a spot on ci.nodejs.org where we put files required by our infra. See aix61-standalone/manualBootstrap.md for some examples of using https://ci.nodejs.org/downloads/. I don't see anything in Ansible but there was a time when I was doing ARMv7 compiler stuff on there too and I see the files are still there.
There might be problems with making these publicly available as there are probably redistribution restrictions on the files. Can you see anything associated with the files that says anything about redistribution? I don't have any great ideas for limiting distribution without opening up whole new security holes or setting up new dedicated resources.

rvagg:
> the xip files have certificates that expire

Oh, what? so even if we stored these files they will stop being extractable? So we're going to have to put ourselves at the mercy of Apple continue to supply these files?

What's the structure inside them? Is it practical to unpack and pack them up again into a format that we can properly archive for ourselves?

AshCripps:
> What's the structure inside them? Is it practical to unpack and pack them up again into a format that we can properly archive for ourselves?

Not sure, when you extract them you got a .app file which is actually a directory I believe as the xcode path becomes `/Applications/Xcode.10.3.app/Contents/Developer`

Ill have to try packing it up myself and see if someone else can reuse it.

AshCripps:
Closing this, dont think this will be needed as notorization has forced us to update the release machines to new OS

Issue #1970, opened 2019-10-18
Compilers are too old on many CI machines


targos:
Found with https://github.com/nodejs/node/pull/30028

Expected: g++ 6.3.0 or clang++ 8.0.0

V8 requires C++14 support, meaning we need at least g++ 5.x everywhere.

## node-test-commit-v8-linux

* `infra-softlayer-ubuntu1404-x64-2`: `(CXX=g++, 4.9.4)`
* `test-nearform_intel-ubuntu1604-x64-1`: `(CXX=g++, 5.4.0)`

## node-test-commit-freebsd

* `test-digitalocean-freebsd11-x64-1`: `(CXX=ccache c++, 6.0.0)`

## node-test-commit-custom-suites-freestyle

* `test-rackspace-ubuntu1604-x64-1`: `(CXX=ccache g++, 5.4.0)`

## node-test-commit-arm

* `test-packetnet-ubuntu1604-arm64-2`: `(CXX=ccache g++, 5.4.0)`

## node-test-commit-linux-containered

* `test-joyent-ubuntu1804_sharedlibs_container-x64-4`: `(CXX=ccache g++, 5.4.0)`
* `test-digitalocean-ubuntu1804_sharedlibs_container-x64-5`: `(CXX=ccache g++, 5.4.0)`
* `test-digitalocean-ubuntu1804_sharedlibs_container-x64-10`: `(CXX=ccache g++, 5.4.0)`
* `test-digitalocean-ubuntu1804_sharedlibs_container-x64-1`: `(CXX=ccache g++, 5.4.0)`
* `test-softlayer-ubuntu1804_sharedlibs_container-x64-5`: `(CXX=ccache g++, 5.4.0)`
* `test-digitalocean-ubuntu1804_sharedlibs_container-x64-7`: `(CXX=ccache g++, 5.4.0)`

## node-test-commit-linux

* `test-rackspace-debian8-x64-1`: `(CXX=ccache g++, 4.9.2)`
* `test-rackspace-ubuntu1604-x64-2`: `(CXX=ccache g++, 5.4.0)`

rvagg:
> node-test-commit-v8-linux
> `infra-softlayer-ubuntu1404-x64-2`: (CXX=g++, 4.9.4)

[...]

PR #1973, opened 2019-10-21
jenkins: only test FreeBSD 11 on <= Node.js 12 thanks to clang 6

rvagg:
Ref: https://github.com/nodejs/build/issues/1970
Ref: https://github.com/nodejs/node/pull/30020

FreeBSD 11 comes with clang 6.0.0. `configure.py` states we have a minimum of 8.0.0 (though not BUILDING.md). V8 7.9 @ https://github.com/nodejs/node/pull/30020 is having problems with older gcc but it compiles OK on FreeBSD 11, with just one test failing.

Two questions for this PR:

1. Is clang 8.0.0 a _real_ minimum requirement or just a guess? Should it disqualify FreeBSD 11?
2. What is the likelihood of V8 7.9 going into Node 13.x? I've crafted this PR to exclude it from >=13 because of this assumption, but maybe that's not a safe assumption if 7.9 is going to break real compiler minimums, not just what we state in BUILDING.md (i.e. someone can compile 13.x today with their gcc 5, but in a month they won't be able to because 7.9 is there and having problems).

Also we don't have anything newer than FreeBSD 11 so maybe we should ditch all our FreeBSD machines and put in some fresh new 12 ones and start from scratch?

@targos @richardlau @nodejs/build @nodejs/platform-freebsd

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -79,6 +79,7 @@ def buildExclusions = [
   
   // FreeBSD -----------------------------------------------
   [ /^freebsd10/,                     anyType,     gte(11) ],
+  [ /^freebsd11/,                     anyType,     gte(13) ],
 
   // Source / headers / docs -------------------------------
   [ /^osx1010-release-sources$/,      releaseType, gte(11) ],


richardlau:
> 
>     1. Is clang 8.0.0 a _real_ minimum requirement or just a guess? Should it disqualify FreeBSD 11?
> 

It was bumped up to 8.0.0 (previously 3.4.2) for Node.js 12 to match the LLVM version in XCode 8 for macOS: https://github.com/nodejs/node/pull/26719#discussion_r266254951

I don't know if it's a real minimum in general. Do V8 document their minimum compiler requirements anywhere?

rvagg:
I don't think they do, historically it's been up to us to bump into the edgecases and bump it on our end. @targos?

targos:
I'm not aware of documented compiler requirements.
This page says that Chromium (and I assume V8) supports C++14 but not C++17 yet: https://chromium-cpp.appspot.com

C++14 is supported in [Clang 3.4 and later](https://clang.llvm.org/cxx_status.html).

I'm not too concerned about using Clang 6 for now. We'll see when we have issues with it.

PR #1986, opened 2019-10-23
Remove Ubuntu 12.04 from CI

rvagg:
It's still active for Node 8 testing, which will cease at the end of the year. 12.04 has been out of standard support since 2017 and even left extended (paid) support earlier this year. It's not been as painful as Debian or CentOS but let's remove it anyway.

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -24,7 +24,6 @@ def buildExclusions = [
   [ /debian8-x86/,                    anyType,     gte(10) ], // 32-bit linux for <10 only
   [ /debian8/,                        anyType,     gte(13) ],
   [ /^ubuntu1804/,                    anyType,     lt(10)  ], // probably temporary
-  [ /^ubuntu1204/,                    anyType,     gte(10) ],
   [ /^ubuntu1404-32/,                 anyType,     gte(10) ], // 32-bit linux for <10 only
   [ /^ubuntu1404-64/,                 anyType,     gte(12) ],
   [ /^ubuntu1604-32/,                 anyType,     gte(10) ], // 32-bit linux for <10 only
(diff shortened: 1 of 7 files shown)


jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -24,7 +24,6 @@ def buildExclusions = [
   [ /debian8-x86/,                    anyType,     gte(10) ], // 32-bit linux for <10 only
   [ /debian8/,                        anyType,     gte(13) ],
   [ /^ubuntu1804/,                    anyType,     lt(10)  ], // probably temporary
-  [ /^ubuntu1204/,                    anyType,     gte(10) ],
```
richardlau: We need to make sure that all jobs referencing these nodes are updated before committing this otherwise they will start scheduling jobs against them (since this script is "don't allow if" so if the line is removed it is allowed).

rvagg:
Removed references to ubuntu12 from the following jobs' config:

* archive-libuv-test-pipeline
* archive-node-stress-single-test
* archive-node-test-commit-unified
* archive-node-test-known-issues
* archive-node-test-npm
* chakracore-test-linux
* citgm-abi-smoker
* citgm-abi-smoker-forward-compat
* citgm-smoker
* citgm-smoker-nobuild
* citgm-smoker-pipeline
* citgm-smoker-private
* citgm-smoker-rebuild
* citgm-smoker-yarn
* libuv-test-commit-linux
* llnode-continuous-integration
* llnode-pipeline
* node-inspect
* node-stress-single-test
* node-test-commit-linux
* node-test-commit-linux-coverage-daily
* node-test-napi-modules-citgm
* node-test-napi-modules-citgm-v8.x
* node-test-node-addon-api
* node-test-node-addon-api-new
* readable-stream-continuous-integration
* readable-stream-pipeline
* string_decoder-continuous-integration
* string_decoder-pipeline

PR #1987, opened 2019-10-23
Remove SmartOS 14 from CI

rvagg:
Hasn't been needed since Node 6. The last 32-bit SmartOS and saves 5 machines.

### DIFF
--- a/doc/node-test-commit-matrix.md
+++ b/doc/node-test-commit-matrix.md
@@ -41,8 +41,6 @@ This is assumed correct as of the date of last commit. If you notice a discrepan
     - ppcbe-ubuntu1404 (Node < 8)
     - ppcle-ubuntu1404 (Node < 13)
   - **node-test-commit-smartos**
-    - smartos14-32 (Node < 8)
-    - smartos14-64 (Node < 8)
     - smartos15-64 (Node >= 8 < 10)
     - smartos16-64 (Node >= 8 < 12)
     - smartos17-64 (Node >= 10 < 12)
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -46,7 +46,6 @@ def buildExclusions = [
   [ /vs2017/,                         anyType,     lt(10)  ],
 
   // SmartOS -----------------------------------------------
-  [ /^smartos14/,                     anyType,     gte(8)  ],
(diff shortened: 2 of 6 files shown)


doc/node-test-commit-matrix.md
```diff
@@ -41,8 +41,6 @@ This is assumed correct as of the date of last commit. If you notice a discrepan
     - ppcbe-ubuntu1404 (Node < 8)
```
nschonni: Probably on your backlog, but this also looks like it's retired. Same with some old Windows vs2013 stuff from a quick find in files

joaocgreis: VS2013 is still supported for building add-ons with Node.js v8. I think we have a policy of keeping the machines around for 6 months after the release lines that use them become EOL, so VS2013 can be removed in July 2020. I may scale it down to only 1 machine though.

nschonni: Spun up #1997 to remove the ubuntu PPCBE, but not sure if it's right

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -46,7 +46,6 @@ def buildExclusions = [
   [ /vs2017/,                         anyType,     lt(10)  ],
 
   // SmartOS -----------------------------------------------
-  [ /^smartos14/,                     anyType,     gte(8)  ],
```
richardlau: https://ci.nodejs.org/job/node-test-commit-smartos/28512/console started runs on smartos14 for Node.js master since this line was removed but the job wasn't updated to remove the smartos14 labels:
```console
12:28:29 -------------------------------------------------------------------
12:28:29 Node.js major version: 14
12:28:29 Node.js version: 14.0.0
12:28:29 Skipping smartos15-64 for Node.js 14
12:28:29 Skipping smartos16-64 for Node.js 14
12:28:29 Skipping smartos17-64 for Node.js 14
12:28:29 Running nodes
12:28:29 Triggering node-test-commit-smartos » smartos14-64
12:28:29 Triggering node-test-commit-smartos » smartos14-32
12:28:29 Triggering node-test-commit-smartos » smartos18-64
```

All of the `smartos14-*` hosts are currently offline so the build will never complete.

richardlau: @mhdawson has [updated the jobs](https://github.com/nodejs/build/issues/2004).

Issue #1992, opened 2019-10-23
Select compiler broken on AIX


mhdawson:
@rvagg  this change seems to have broken select compiler on AIX. 
 https://github.com/nodejs/build/pull/1985

Failure is:
```
+ rm -rf build
+ git clone https://github.com/nodejs/build.git
Cloning into 'build'...
+ . ./build/jenkins/scripts/select-compiler.sh
/tmp/jenkins4646137741077972430.sh[50]: 0403-057 Syntax error at line 125 : `=~' is not expected.
Build step 'Conditional steps (multiple)' marked build as failure
```
https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-new/nodes=aix61-ppc64/811/console

mhdawson:
@sam-github unless we can fix this quickly can you back out the previous change? It will block all releases.

sam-github:
Leaving open til a build passes.

Kicked off: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-new/820/

(I've never run this job before, I limited it to aix61, perhaps someone cat take a look that I did it correctly).

richardlau:
> 
> 
> Leaving open til a build passes.
> 
> Kicked off: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-new/820/
> 
> (I've never run this job before, I limited it to aix61, perhaps someone cat take a look that I did it correctly).

Unfortunately that build failed while downloading the Node.js binary... probably being affected by https://github.com/nodejs/build/issues/1993.

sam-github:
Ouch. Well, I'll wait a bit before trying again.

rvagg:
do you not have Bash on AIX? if you do, then just make sure `#!/bin/bash` is at the top of the script blocks in Jenkins and it'll all run in Bash and this will be resolved. I already sorted out one instance of bad scripting in node-test-commit-v8-linux yesterday that had `#/bin/bash` at the top and was running Dash instead, the `!` got this sorted. Can we just do the same on AIX?

richardlau:
`Bash` does exist on AIX.

rvagg:
<img width="854" alt="Screen Shot 2019-10-24 at 9 46 17 am" src="https://user-images.githubusercontent.com/495647/67439851-2ecded00-f643-11e9-9c19-80b0191bbd36.png">

[...]

PR #1994, opened 2019-10-23
jenkins: use /bin/sh syntax in select-compiler.sh

sam-github:
Fixes: https://github.com/nodejs/build/issues/1992

### DIFF
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -1,4 +1,10 @@
-#!/bin/bash
+# Usage:
+#   .../node % . ./build/jenkins/scripts/select-compiler.sh
+#
+# This file is sourced from the CWD of node by CI build scripts to select the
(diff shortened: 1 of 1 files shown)


sam-github:
I reproduced the original syntax error with /bin/dash on Linux, and this fixes it (and uses the same case syntax as the rest of the file). I'm in process of trying to repro and proves this fixes it on actual AIX.

sam-github:
I verified that the original could not be sourced on AIX, and that this can be after conversion to case statements.

@nodejs/build can I fasttrack this onto master?

mhdawson:
+1 to fasttrack since I think AIX Is currently broken.

sam-github:
https://ci.nodejs.org/job/node-test-commit/32083/

^-- kicked off a test run on master to confirm nothing was broken by this

richardlau:
Belated LGTM.

Issue #1953, opened 2019-10-10
Prepare to forward foundation.nodejs.org to openjsf.org


brianwarner:
Hi Node build WG,

I would like to open an exploratory conversation about a 301 redirect from foundation.nodejs.org to openjsf.org.  The goal is to eliminate any potential sources of confusion about which Foundation is currently active.

I believe one path to a graceful transition would be:
1. Go over the current site and determine if anything remains to be rehosted on openjsf.org
2. Move the current site somewhere it can be easily accessed if something gets missed, such as nodearchive.openjsf.org (we can do this on our end)
3. Add a 301 redirect from foundation.nodejs.org to openjsf.org

Does anybody have any thoughts about this?

mhdawson:
Makes sense to me.  I had through the foundation.nodejs.org component was now hosted by the Foundation as opposed to the Node.js project but I'm not the expert on that front.

brianwarner:
My understanding is that the foundation should be able to do everything except the DNS redirects, since that's delegated to CloudFlare.  But also, I wanted to make sure everyone was aware and that I'm not missing anything subtle before doing something drastic. :-)

rvagg:
This is fine, we currently point foundation.nodejs.org (CNAME) to some host maintained by the LF so what's on there and how to manage it is in your hands already. So I believe steps 1 and 2 are out of our hands. We can just point it to one of our servers and do a full redirect of `foundation.nodejs.org/*` to `openjsf.org/`. Let us know when that's good to go and we can make that happen. I'll prepare a PR to change the web server config files and then it'll just be a DNS switch.

brianwarner:
Checking back in one more time before we start the process of the transition.  Thanks @rvagg and @mhdawson - would anybody else like to weigh in on this?

mhdawson:
not me.

rvagg:
#1999 for config change, we can deploy this prior to making the actual switch, then the switch is just a DNS update.

brianwarner:
I think at this point we're ready to go on this.  @rvagg, could I ask your assistance on adding the redirect?  Once that's redirected and replicated, I'll take care of the other parts.

rvagg:
I briefly forgot that we control DNS too @brianwarner when I spoke to you about this last. But I've merged the change to the server(s) and updated DNS and the redirect looks like it's in place.

PR #1999, opened 2019-10-25
www: add foundation.nodejs.org redir

rvagg:
Ref: https://github.com/nodejs/build/issues/1953

sanity check please!

We currently have a CNAME for this domain pointing to some LF server that manages it. The plan here is to change it to an A to point to our main server (also proxying via Cloudflare so they'll actually do all the redirecting). Then we get to redirect `foundation.nodejs.org/*` and not worry about redirecting dangling links. i.e. all old /whatever.html URLs will go to openjsf.org/.

### DIFF
--- a/ansible/www-standalone/resources/config/nodejs.org
+++ b/ansible/www-standalone/resources/config/nodejs.org
@@ -169,6 +169,21 @@ server {
 
 server {
     listen *:80;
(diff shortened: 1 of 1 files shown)


rvagg:
landed & deployed @ 1eaa6bf

PR #2003, opened 2019-10-25
ansible: add new smartos18 machines

mhdawson:


rvagg:
oh, actually no, I think you'll want to rename the old ones because this way you need to make new Jenkins nodes and new secrets, if you just change the IP of existing ones you can reuse everything else

mhdawson:
@rvagg I don't have time to undo quickly enough if something goes wrong. I want to add and them without removing the original machines so that reverting can be quick if we have problems.

I personally don't see the problem with having machines named with -2 after the switch instead of -1 and if really necessary that can be fixed up later.

mhdawson:
@rvagg I should also have mentioned that they are already added/running...

Issue #2001, opened 2019-10-25
New Linter strategy


rvagg:
The linter jobs have been stuck on FreeBSD 10 since we first introduced them. We need to get rid of FreeBSD 10, it's way out of support and we don't have anyone who is interested in maintaining them anyway. There was an aborted attempt to move them to Ubuntu machines but there were specifics about the various `make lint` commands that caused problems with that (we got ourselves into a BSD corner). It seems that we're past that and the linter(s) are all nicely BSD+GNU. So it's time to move them or have a new strategy.

There are two possible paths I want to put forward:

1. Run the lint jobs on Ubuntu machines (the jenkins-workspace machines might be a good choice). This would mean (1) making sure there is a `node` on those machines, and (2) reconfiguring the job to use the new machines.
2. Run the liters in Docker using some of the infra that comes in #1982 - namely the ability to execute a particular container that gets pulled from Docker Hub. I've set up a linter Dockerfile here: https://github.com/rvagg/node-ci-containers/tree/master/node-linter (in Docker Hub as `rvagg/node-ci-containers:node-linter`, again, that 'rvagg' will change). You just run it with your directory mounted: `docker run -v $(pwd):/home/iojs/workspace rvagg/node-ci-containers:node-linter`. This gets our linting logic out of Jenkins and into a form that's accessible and usable by collaborators and gives us an easier upgrade path in the future.

When this is done, I want to purge FreeBSD 10 from our infra. There's some urgency to this too btw since we need to move some Joyent resources around in the next few weeks and there's a FreeBSD 10 / linter machine involved in that.

jbergstroem:
FWIW: there's nothing about the linters that is tied to FreeBSD; it was just a personal preference when I originally set it up. 

> Run the liters in Docker using some of the infra that comes in #1982 - namely the ability to execute a particular container that gets pulled from Docker Hub. 

I like this approach. It makes it simpler to drop in any shared infra from other CI's or github actions.

sam-github:
caveat: I may be missing something about the history and requirements, for example, I'm honestly not sure if `git node land` requires a green Travis run. I don't think I've ever seen Travis red and Jenkins  green, so haven't run into that :-).

Why are we setting this up in jenkins? Can linting be left to be done only on Travis?

Travis only is simpler (delete the jobs, Travis does `make lint-py doc-only lint` already), and for the lint jobs in particular, they only need one platform and Travis already gives immediate feedback on a PR without collaborator intervention.

[...]

Issue #793, opened 2017-07-12
Add core-validate-commit to node-test-pull-request


mhdawson:
Was thinking we should probably add core-validate-commit to node-test-pull-request to automate checks on the commit message.

joaocgreis:
+1

I remember some discussion around this, but I can't find it. @evanlucas do you remember? If it was simply not ready at the time, is it now?

gibfahn:
The job is there, it just needs to be included (as long as it's working). I see that @evanlucas has access to it.

https://ci.nodejs.org/job/node-test-commitmsg/

gibfahn:
>@Trott thanks for the suggestion. Have installed it now. Do you use it when reviewing as well or just as a sanity check for the final land ?
>
>I also wonder if we could add running this to the node-test-pull-request in addition to the linter ?

@mhdawson in reply to https://github.com/nodejs/node/commit/d5b397c9b63507a5dcf6d1da47df2a090d3d73a0#commitcomment-23061905

I use it as a sanity check before landing. It'll fail before then anyway as it won't find the `PR-URL` or `Reviewed-By` lines. I think @refack has it set up as a pre-commit hook (so you can't commit without it being correct), though I could be misremembering.

mcollina:
I am -1 with this change as proposed. This will fail for all the PRs on the first go, because they lack reviews and a PR url. It will just increase the amount of nit-picking, as we will likely require to fix those details by the author, while we currently fix them on landing.

If we can disable those checks for the job, then I am +1.

refack:
> I think @refack has it set up as a pre-commit

`pre-push`, and it has a `if ($2.includes('/nodejs/node'))`.
I don't want to tie my hands completely.

evanlucas:
> It'll fail before then anyway as it won't find the PR-URL or Reviewed-By lines. 

You can pass the `--no-validate-metadata` flag and it won't fail from a missing PR-URL or missing Reviewed-By.

The job that is there works pretty well. It can be run against a single commit or all of the commits for a PR. It becomes tricky with fixup commits though...

fhinkel:
`core-validate-commit` is great! But iIrc it complains about Chromium backports because a line with the URL is too long.

evanlucas:
@fhinkel I’ll try to get around to fixing that.

sam-github:
Closing as stale, but if anyone wants to take this up feel free to reopen.

richardlau:
[...]

PR #2008, opened 2019-10-27
ansible: prepare jenkins-workspace machines for linting duty

rvagg:
Ref: https://github.com/nodejs/build/issues/2001

Also included this is a reprovisioned jenkins-workspace-1, it started having problems 2 weeks back and has locked up twice now. I just made a new one and re-attached the big network block device we're using for /home from the old one so it's good to go.

I've run this against the hosts, I've also switched the label for the linter job in Jenkins from `linter` to `jenkins-workspace` so they've already taken over duty from the FreeBSD 10 machines which we need to retire.

The script has changed, a few old bits are removed and the BSD-compatibility is gone too:

```sh
#!/bin/bash -ex

# Lint with python3 in new branches
# Refs: https://github.com/nodejs/build/issues/1631
if [[ "$NODEJS_MAJOR_VERSION" -ge "12" ]]; then
  make lint-py-build PYTHON=python3 || true
  make lint-py PYTHON=python3
fi
    
make lint-py-build PYTHON=python2 || true
make lint-ci PYTHON=python2 || { 
  cat test-eslint.tap | grep -v '^ok\|^TAP version 13\|^1\.\.' | sed '/^\s*$/d' && 
  exit 1; }
```

### DIFF
--- a/ansible/playbooks/jenkins/worker/create.yml
+++ b/ansible/playbooks/jenkins/worker/create.yml
@@ -68,42 +68,13 @@
   roles:
     - linux-perf
 
-#
-# Set up linter servers
-#
-
-- hosts:
-   - test-joyent-freebsd10-x64-2
-   - test-rackspace-freebsd10-x64-1
-
-  tasks:
-    - name: install lint-related packages
-      package:
-        name: "{{ package }}"
-        state: present
-      loop_control:
-        loop_var: package
-      with_items: [ "node", "npm" ]
-
-    - name: install core-validate-commit
-      npm:
-        name: "core-validate-commit"
-        global: yes
-        state: present
-        production: yes
-
-    - name: periodically update core-validate-commit
-      cron:
-        special_time: daily
-      # ksh does different stdout/err routing
-        job: "npm update -g core-validate-commit > & /dev/null"
-
 # Ensure node is not installed anywhere but the linter servers
 - hosts:
   - test
   - release
-  - "!test-joyent-freebsd10-x64-2"
-  - "!test-rackspace-freebsd10-x64-1"
+  - "!test-packetnet-ubuntu1604-x64-1"
+  - "!test-packetnet-ubuntu1604-x64-2"
+  - "!test-softlayer-ubuntu1604-x64-1"
   tasks:
     - name: remove node and npm packages
       when: not os|startswith("win") and not os|startswith("zos")
--- a/ansible/roles/jenkins-workspace/tasks/main.yml
+++ b/ansible/roles/jenkins-workspace/tasks/main.yml
@@ -106,3 +106,31 @@
   command: "git config gc.auto 0"
   args:
     chdir: "~binary_tmp/binary_tmp.git/"
+
+- name: Add nodesource signing key
+  apt_key:
+    url: https://deb.nodesource.com/gpgkey/nodesource.gpg.key
+    state: present
+
+- name: Add nodesource repo
+  apt_repository:
+    repo: deb https://deb.nodesource.com/node_12.x xenial main
+    state: present
+
+- name: Install node
+  package:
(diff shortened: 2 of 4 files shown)


ansible/roles/jenkins-workspace/tasks/main.yml
```diff
@@ -106,3 +106,31 @@
...
+  apt_repository:
+    repo: deb https://deb.nodesource.com/node_12.x xenial main
+    state: present
+
+- name: Install node
+  package:
+    name: nodejs
+    state: present
+
+- name: Upgrade pip2
```
rvagg: I was getting errors with `make lint-py-build` on both Python 2 and 3, this upgrade seems to have fixed it. I'm hoping this is the right thing to do. Can anyone tell me what would happen if there's a python-pip or python3-pip update that comes in via apt? Would this upgrade get written over and need to be run again?

cclauss: `pip` _should_ point to `pip2` but you can be more assured by using `pip2` instead of `pip`.

In situations where I need absolute clarity, I tend to run pip as a module from the Python that I care about as in:
```
python3 -m pip install --upgrade pip
python3 -m pip install flake8
# or even better if multiple Python 3 instances might be installed on a single box...
python3.8 -m pip install --upgrade pip
python3.8 -m pip install flake8
```

rvagg: :thumbsup: switched to `executable: pip2`, I'll trust Ansible to get it right beyond that and the playbook runs with this change.

ansible/playbooks/jenkins/worker/create.yml
```diff
@@ -102,8 +102,9 @@
 - hosts:
   - test
   - release
-  - "!test-joyent-freebsd10-x64-2"
-  - "!test-rackspace-freebsd10-x64-1"
+  - "!test-packetnet-ubuntu1604-x64-1"
+  - "!test-packetnet-ubuntu1604-x64-2"
+  - "!test-softlayer-ubuntu1604-x64-1"
```
joaocgreis: There is a "Set up linter servers" section just above this, that runs exclusively on the two hosts being removed here. Shouldn't that section be completely removed? There are some bits about core-validate-commit there that I don't know if are still relevant in some scenario.

rvagg: Yeah, missed that, also missed core-validate-commit. There is a job: https://ci.nodejs.org/job/node-test-commitmsg/ that runs it but it hasn't had any runs since the last build cleanup (~1 week). Does that mean it's disabled? Maybe this is all handled happily in Travis now and in node-core-utils. @nodejs/tooling can you weigh in here? Can we remove node-test-commitmsg from Jenkins and the associated server resources for running it?

richardlau: We are definitely running core-validate-commit as part of node-core-utils and on Travis for PRs and that should be sufficient for the normal PR workflows for core.

An old issue, https://github.com/nodejs/build/issues/793, proposed adding https://ci.nodejs.org/job/node-test-commitmsg/ to `node-test-pull-request` but that didn't happen.

rvagg: Integrating core-validate-commit into my .ci.yml proposal; a failing v8.x: https://ci.nodejs.org/job/rv-test-commit-linux-docker/129/label=docker-host-x64,linux_x64_container_suite=lint-commit/console, ci.yml entry (see 'tests->lint-commit' block): https://github.com/rvagg/io.js/blob/rvagg/testing-containers-v8.x/.ci.yml

I might archive that Jenkins job since it's not in use, it can be unarchived if there's reason to before it eventually gets deleted.

PR #2009, opened 2019-10-27
Ansible: Add baselayout for macos10.13 thru 10.15

cclauss:
Prepare for testing on macOS High Sierra, Mojave, and Catalina.


rvagg:
We try to keep ansible/ to machines that we _actually_ manage so it doesn't get out of hand. I don't mind the name changes in here but I'm dubious about the rest because it's pretty obvious what needs to be added. We have other platforms where it's mostly copy-paste too. @AshCripps maybe you can take some of this commit into your work?

AshCripps:
Closing this one for, will include the parts of this pr that is required when adding the new nearform machines #1695

PR #2012, opened 2019-10-27
ansible,jenkins: remove freebsd10

rvagg:
Really just in the inventory here, also need to remove from Jenkins jobs, remove the Jenkins node configs and unprovision the resources.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -85,7 +85,6 @@ hosts:
         fedora27-x64-1: {ip: 159.203.117.50}
         fedora30-x64-1: {ip: 178.62.236.249}
         fedora30-x64-2: {ip: 159.203.98.84}
-        freebsd10-x64-1: {ip: 159.203.59.134, user: freebsd}
         freebsd11-x64-1: {ip: 45.55.90.237, user: freebsd}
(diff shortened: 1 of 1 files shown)


rvagg:
have removed freebsd10 references from:

archive-libuv-test-pipeline
archive-node-stress-single-test
archive-node-test-known-issues
libuv-test-commit-freebsd
node-stress-single-test
node-test-commit-freebsd

Wasn't aware we had _two_ linter jobs, a pipeline one at https://ci.nodejs.org/job/node-linter/ which I've swapped out `linter` for `jenkins-workspace`. That's distinct from node-test-linter which is a classic job. node-linter appears to be a dependency of node-test-pull-request-lite-pipeline. Both still run. A bit of a mess, consolidation would be good.

Have removed the 'linter' label from https://ci.nodejs.org/computer/test-rackspace-freebsd10-x64-1/ and https://ci.nodejs.org/computer/test-joyent-freebsd10-x64-2/ to flush out anything else that might depend on these, won't delete the machines yet unless something unexpected turns up, this is all a bit tangled.

rvagg:
removed from rackspace, DO, Joyent, Jenkins and the firewall, all gone

PR #2013, opened 2019-10-27
fix: Remove CentOS 5 config notes

nschonni:
Move anchor links to the bottom as well


rvagg:
good catch, I guess that's where the `-i` flag would have come in handy for my `git grep`.

nschonni:
There are also some references in https://github.com/nodejs/build/blob/master/jenkins/xml/citgm-smoker.xml#L221 but I'm not really sure if those XML files are used or just backup snapshots

rvagg:
We can clean those up, but they're really only there for informational purposes. I'm dubious about their utility and would like to know if anyone who doesn't already have jenkins-admins access has even looked at them.

It might be more interesting if we had a script that automatically updated this repo as configs changed on the server(s). That way we could get some nice history of versions, which would be great because the history feature in Jenkins has been screwy for the last few months in Jenkins and I'm not sure why.

Issue #2005, opened 2019-10-26
can we give ashcripps access to help ansible smartos machines?


sam-github:
@mhdawson would like me to help with the ansibling of smartos machines this coming monday. I haven't found an issue that exactly says what's happening, it just seems there is lots of mention of smartos.

If its a lot of work, it might be a good thing to get @AshCripps to help us.

He is newish to the community build WG, but he is part of our team at IBM, and its intended that he spend a bunch of his time here.

I'm not sure at what point exactly he can request to join the build-wg. He meets the criteria of being known to current members (@mhdawson @richardlau and me), and of significant employment repercussion should he do something malicious.

So, just as a heads up, we hope he will be able to join the WG soonish, and perhaps can get limited access to the smartos machines this monday. I've already mostly gotten him setup with ansible, though not with any build secrets, in prep for him running ansible against the nearform minis.

@rvagg and other @nodejs/build WG members, thoughts on 

1. temp access monday for smartos provisioning?
2. whether its too soon to join the build WG? Should we wait until he has a few more PRs in?

richardlau:
> I'm not sure at what point exactly he can request to join the build-wg. He meets the criteria of being known to current members (@mhdawson @richardlau and me), and of significant employment repercussion should he do something malicious.
> 

😅 I'm not currently a member of the Build WG... but since you're not the first person to assume I am let's fix that: https://github.com/nodejs/build/pull/2006 😄.

rvagg:
> temp access monday for smartos provisioning?

Personally very hesitant on this one, DO and Joyent have some of key resources on machines hosted in them. Infra access to both of those services gives you pretty much everything since you can get access into the disks that hold all the crown jewels. From a security perspective there's really no such thing as "temporary access" unless we're willing to regenerate all secrets.

Will take further discussion about expanding infra to email where we tend to handle this kind of thing.

> whether its too soon to join the build WG? Should we wait until he has a few more PRs in?

No concern here, hop on board. We have a lot of test- machines in Joyent to reprovision, having nodejs_build_test access lets you do at least some of the work. It would just need some in infra to provision and setup firewall access to ci.nodejs.org.

rvagg:
In the list of machines we need reprovisioned, there is one that I'm concerned about because it will likely involve more yak shaving than it shoud.

[...]

PR #2015, opened 2019-10-29
ansible: update arm_cross servers

joaocgreis:
Rebuilt all 3 cross compilers. Azure was very outdated, Joyent asked us to move to another datacenter.

Refs: https://github.com/nodejs/build/issues/2005#issuecomment-546931181

Tried using Ubuntu 18.04 but it's not straightforward because gcc 4.9.4 is not easily available. The current machines (16.04) should work until Node v10 becomes EOL.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -38,7 +38,7 @@ hosts:
         smartos15-x64-1: {ip: 72.2.113.195}
         smartos17-x64-1: {ip: 72.2.114.225}
         smartos18-x64-1: {ip: 72.2.119.185}
-        ubuntu1604_arm_cross-x64-1: {ip: 72.2.114.100, user: ubuntu}
+        ubuntu1604_arm_cross-x64-1: {ip: 165.225.151.28, user: ubuntu}
 
     - requireio:
         rvagg-debian9-armv6l_pi1p-1: {ip: 192.168.2.40, user: pi}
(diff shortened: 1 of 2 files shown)


joaocgreis:
@nodejs/build any objection to landing this?

ansible/inventory.yml
```diff
@@ -74,7 +74,7 @@ hosts:
   - test:
 
     - azure:
-        msft-ubuntu1404-x64-1: {ip: node-msft-cross-compiler-1.cloudapp.net}
+        msft-ubuntu1604_arm_cross-x64-1: {ip: nodejs.eastus2.cloudapp.azure.com, user: ubuntu}
```
rvagg: the naming is non-standard btw, should remove "msft" from here since "azure" goes into the name already and the pattern is supposed to be: "type-os/dist+version[_extrainfo]-arch-X"

joaocgreis: I would like the name to remain `azure_msft`. Azure is where the machines are hosted, however it is not Azure that is sponsoring us but Microsoft directly.

There are 2 machines that have only `msft`, because they are physically located in my office. They're meant to be only temporary until we can get ARM64 Azure VMs and I can't guarantee they'll remain here, so just `msft` is fine for those 2.

This is covered in our current documentation: https://github.com/nodejs/build/tree/master/ansible#naming , I took that to mean `provider_sponsor` when the two are different, as with the Raspberry Pi's.

rvagg: oh yeah, you're right, all good then

joaocgreis:
Thanks!

Landed in https://github.com/nodejs/build/commit/f66cbfe53d1ffac5ab24afd2c15106096c0bcc64

PR #2020, opened 2019-10-31
www: dist-perms cron.daily script to fix ~dist permissions

rvagg:
After 5 days change ownership of downloadable resources to `root`, this locks them down and they can't be modified in any way by releasers. I'd like to change this to a shorter window but will wait till after Node 8 is EOL since Node 8 still has a Raspberry Pi builder so has reasonable delays in getting all assets promoted.


targos:
I think this broke the release process.

I have the following error when I try to promote v13.1.0:

```
mkdir: cannot create directory ‘/home/dist/nodejs/release/v13.1.0’: Permission denied
```

`/home/dist/nodejs/release` belongs to `root`.

rvagg:
Fixed @ https://github.com/nodejs/build/pull/2031, live on the server now, very sorry for the inconvenience @targos!

PR #2024, opened 2019-11-01
ansible: update metrics scripts with latest on server

rvagg:
this is what's running on the server for nodejs.org/metrics, just getting it updated here


ansible/www-standalone/tools/metrics/country-lookup.py
```diff
@@ -29,7 +28,7 @@
         country = georec.country.iso_code
       if georec.subdivisions.most_specific.iso_code:
         region = georec.subdivisions.most_specific.iso_code
-  except Exception:
+  except:
```
cclauss: https://realpython.com/the-most-diabolical-python-antipattern/

PR #2031, opened 2019-11-06
www: fix /dist permissions fixer, ignore top-level directories

rvagg:
Ref: https://github.com/nodejs/build/pull/2020#issuecomment-549964893

`-mindepth 2` excludes the top directories, including `/home/dist/nodejs/release/` which is what's caused the problem with 13.1.0.


targos:
Thanks

PR #2034, opened 2019-11-06
Python: Avoid bare exceptions in country-lookup.py

cclauss:
The problem with these is that you can put a syntax error in between the try: and except: and it will be extremely difficult to spot.  This is discussed in PEP8, Zen of Python, and many other places.


sam-github:
@nodejs/build can we fast-track, or land? PRs to nodejs/build all fail travis without it.

PR #2037, opened 2019-11-06
ansible: make warnings about secrets informative

sam-github:
Describe what it is that is being looked for (a git checkout of
nodejs-private/secrets), and the locations that are automatically
searched.

### DIFF
--- a/ansible/plugins/inventory/nodejs_yaml.py
+++ b/ansible/plugins/inventory/nodejs_yaml.py
@@ -138,21 +138,29 @@ def get_secrets_path():
             return path
         else:
             print("WARNING: NODE_BUILD_SECRETS defined but not a directory", file=sys.stderr)
+            print("It must be the path to a local checkout of https://github.com/nodejs-private/secrets", file=sys.stderr)
             return None
 
     path = os.path.realpath('../../secrets/build/')
     if os.path.isdir(path):
         return path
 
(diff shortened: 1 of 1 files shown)


sam-github:
Up until recently, I assumed the secret that was being searched for was was the `secret:` var from a host_vars file, like [this one](https://github.com/nodejs/build/blob/09a0a64283c93c6b2ce4352f4605714a24e8b013/ansible/host_vars/test-marist-zos13-s390x-template#L1).

ansible/plugins/inventory/nodejs_yaml.py
```diff
@@ -138,21 +138,27 @@ def get_secrets_path():
...
         else:
             print("WARNING: NODE_BUILD_SECRETS defined but not a directory", file=sys.stderr)
+            print("It must be the path to a local checkout of https://github.com/nodejs-private/secrets", file=sys.stderr)
             return None
 
     path = os.path.realpath('../../secrets/build/')
     if os.path.isdir(path):
         return path
 
+    print("WARNING: ../../secrets/build/ is not a directory", file=sys.stderr)
```
rvagg: I'm not sure about these 3 warnings, it means that if you have it in the last checked location you'll see two of them, but that location isn't a problem so doesn't deserve to trigger warnings.
Maybe these should be printed at the end, "I checked here, here and here but didn't find it".

sam-github:
@rvagg, made it not a warning, PTAL

Issue #1859, opened 2019-07-02
No pgrep on AIX


Trott:
This shows up on AIX in CI:

```console
22:06:11 + pgrep node
22:06:11 /tmp/jenkins372064123762390528.sh[28]: pgrep:  not found.
```

It passes anyway because the test is for output. Since no `pgrep` means no output, the job interprets that as "no stray `node` processes running" and thinks everything is OK.

I'm not sure if the answer is to install `pgrep` on AIX or if the answer is to rewrite the check to use `ps` and `grep` instead of `pgrep`. @nodejs/build-files @nodejs/platform-aix

bnoordhuis:
I don't know if pgrep is available for AIX. Using plain grep is finicky because `grep -E node$` won't match on AIX and just `grep node` is kind of dangerous.

Perl to the rescue!
```
$ perl -MFile::Basename -e '
  my @r = ();
  for (qx[ps ax]) {
    my @F = split;
    push @r, $F[0] if $F[4] && basename($F[4]) eq "node"
  }
  print "@r\n" if @r
'
```

Trott:
> I don't know if pgrep is available for AIX. Using plain grep is finicky because `grep -E node$` won't match on AIX and just `grep node` is kind of dangerous.

I don't think it's dangerous here as all we're doing is "check if there are any `node` processes and if so, assume they are left over from tests and that there is therefore a bug in the tests that leaves processes running". It doesn't actually *do* anything with those processes. The doing-something-about-it happens in the `Makefile`  but there it's `ps awwx | grep Release/node`. 

I'm not sure where `pgrep` is coming from. It's not in the `Makefile`. I imagine Jenkins config somewhere?

richardlau:
> I'm not sure where `pgrep` is coming from. It's not in the `Makefile`. I imagine Jenkins config somewhere?

Yes:

![image](https://user-images.githubusercontent.com/5445507/60521632-81277e80-9cdf-11e9-9610-aba98177a8c7.png)

I don't know the history but I suspect it was copied from the job for another platform. It looks similar to https://github.com/nodejs/build/blob/master/jenkins/scripts/node-test-commit-diagnostics.sh -- I also don't know why it's not calling that script (with any appropriate platform specifics encoded in the script).

mhdawson:
I'll confirm this was a copy from another job.

Trott:
[...]

PR #2041, opened 2019-11-07
jenkins: edit test-commit-diagnostics.sh to solve missing pgrep

AshCripps:
resolves https://github.com/nodejs/build/issues/1859 using https://github.com/nodejs/build/issues/1859#issuecomment-507563309

This gets around AIX not having pgrep but does imply that all other build machines that use this script have perl installed which me and @sam-github are not sure of.

### DIFF
--- a/jenkins/scripts/node-test-commit-diagnostics.sh
+++ b/jenkins/scripts/node-test-commit-diagnostics.sh
@@ -39,4 +39,11 @@ mv ${DIAGFILE} ${DIAGFILE}-OLD || true
 tail -c 20000000 ${DIAGFILE}-OLD > ${DIAGFILE} || true
 rm ${DIAGFILE}-OLD || true
 set -x
-pgrep node || true
+perl -MFile::Basename -e '
+  my @r = ();
+  for (qx[ps ax]) {
+    my @F = split;
+    push @r, $F[0] if $F[4] && basename($F[4]) eq "node"
+  }
+  print "@r\n" if @r
+' || true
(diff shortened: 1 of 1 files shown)


jenkins/scripts/node-test-commit-diagnostics.sh
```diff
@@ -39,4 +39,11 @@ mv ${DIAGFILE} ${DIAGFILE}-OLD || true
...
 set -x
-pgrep node || true
+perl -MFile::Basename -e '
+  my @r = ();
+  for (qx[ps ax]) {
+    my @F = split;
+    push @r, $F[0] if $F[4] && basename($F[4]) eq "node"
+  }
+  print "@r\n" if @r
+' || true
```
sam-github: is the `|| true` needed anymore? pgrep would fail if the process isn't found, perl will just list what it found and exit with sucess.

AshCripps: Removed

sam-github:
@nodejs/build What do you all think? If perl is everywhere (and I think its virtually a requirement for linux systems) this should be more portable.

@rvagg you in particular probably have the best sense of whether this is going to work - though we could always just try and see.

rvagg:
just try and see, I think `perl` should be everywhere but I don't know much about the perl, looks standard enough.

something like `ps a | grep node | grep -v grep` is what I'd probably use in any case, maybe simpler? I don't mind either way as long as it works and is simple(ish)

sam-github:
Test build: https://ci.nodejs.org/view/Node.js%20Daily/job/node-daily-master/1747/

PR #2042, opened 2019-11-07
add bzip2-devel

sam-github:
To fix issue 1. in https://github.com/nodejs/node/issues/30152


sam-github:
Needs to be followed through with before landing, see https://github.com/nodejs/build/issues/2043

sam-github:
yes,  commit message is also draft.

this doesn't work, bzip2-devel has to be under another OS key. @AshCripps got it working, he'll open a new PR with a real commit message once the second test machine is ansibled.

PR #2051, opened 2019-11-19
ansible: move IP of {unencrypted,backup-www}.nodejs.org

rvagg:
The good people at Joyent have manually moved unencrypted.nodejs.org, which also serves as backup-www in our load balancer pool. It's up and running in a new location because the old datacenter is being closed. DNS and load balancer have this new change and it appears to be in good order.

The second infra server, "backup", is still to be moved, will keep this PR open until that one has a new IP and is confirmed working.


rvagg:
pushed a second update that changes the IP of the backup server. We're now fully in Joyent's us-west.

PR #2056, opened 2019-11-20
ansible: add new AIX7.1 machines to CI

AshCripps:
This is a draft pass at adding new AIX72 to our test ci.

Ive edited the anisble to install the packages that use to be done manually.

One machine is commented out as I dont have access atm and so its not currently ansibled yet

Finally Ive cleaned up and edited the manual steps .md with additions needed and ive also split it into 6.1 and 7.2 install steps as both have some differences.

These machines aren't ready yet as I dont have CI access so they are missing the jenkins secret.

### DIFF
--- a/ansible/aix61-standalone/manualBootstrap.md
+++ b/ansible/aix61-standalone/manualBootstrap.md
@@ -9,19 +9,115 @@ modules will be installed.
 ```bash
 chfs -a size=+1300000 /opt
 chfs -a size=+250000 /
-chfs -a size=917504 /tmp
+chfs -a size=+917504 /tmp
 ```
 
-AIX72: unneeded: -1 and -2 and aix7-2 have different sizes, and 7.2 is big
-enough already:
+For the systems without ramdisks `/home` needs to be set to the same size as `/home/iojs/build` would be on a ramdisk machine
 
-               6.1-1          6.1-2     7.2-2
-	/opt: 3670016 x 512, 2228224  31457280
-	/:    3538944 x 512, 3407872   4194304
+```bash
+chfs -a size=23000000 /home
+```
+
+## Edit SSH config
+
+Add the following two lines to `/etc/ssh/sshd_config`:
+
+```
+PasswordAuthentication no
+ChallengeResponseAuthentication no
+```
+
+## Remove failedlogin file
+
+If /etc/security/failedlogin is growing without bounds on AIX6.1, remove
+it with a cron job, use `crontab -e`, and add one line:
+```sh
+0 12 * * * /usr/bin/rm -f /etc/security/failedlogin
+```
+
+# AIX 7.2 Install
+
+Most packages should be installed via ansible.
+
+If there are any missing they should be installed via yum
+
+What you do need to install manually is **gcc** and **ccache**
+
+## gcc 6.3.x on AIX 7.2
+
+```bash
+mkdir -p /opt/gcc-6.3 && cd /opt/gcc-6.3
+curl -L https://ci.nodejs.org/downloads/aix/gcc-6.3-aix7.2.ppc.tar.gz | /opt/freeware/bin/tar -xzf -
+```
+
+## ccache 3.7.4 on AIX 7.2
+
+```bash
+mkdir -p /opt/ccache-3.7.4 && cd /opt/ccache-3.7.4
+curl -L https://ci.nodejs.org/downloads/aix/ccache-3.7.4.aix7.2.ppc.tar.gz | /opt/freeware/bin/tar -xzf -
+```
+## Enable the AHA fs
+
+For AIX 7.2 and 6.1, needed for the file watcher unit tests.
+
+Add the following to /etc/filesystems:
+
+```
+/aha:
+        dev             = /aha
+        vfs             = ahafs
+        mount           = true
+        vol             = /aha
+```
+
+and then:
+
+```bash
+mkdir /aha
+mount /aha
+```
+
+## Install XL compilers
+
+1. Download 16.1.0 packages from: https://testcase.boulder.ibm.com (username:
+   xlcomp4, password: ask @mhdawson)
+2. scp them to target:/opt/ibm-xlc
+3. on target:
+```bash
+cd /opt/ibm-xlc
+uncompress 16.1.0.3-IBM-xlCcmp-AIX-FP003.tar.Z
+uncompress IBM_XL_C_CPP_V16.1.0.0_AIX.tar.Z
+installp -aXYgd ./usr/sys/inst.images -e /tmp/install.log all
+inutoc
+installp -aXgd ./ -e /tmp/install.log all
+```
+4. Find compilers in `/opt/IBM/xl[cC]/16.1.0/bin/`
+
+
+
+#### tap2junit
+
+Should be installed via ansible but if needed to be done manually use the following:
+
+AIX72:
+```
+  python  -m pip install --upgrade pip pipenv git+https://github.com/nodejs/tap2junit
+  python3 -m pip install --upgrade pip pipenv git+https://github.com/nodejs/tap2junit
+  ln -s /opt/freeware/bin/tap2junit /usr/bin/tap2junit
+```
+
+Note: probably only the py3 was needed
+
+```bash
+python -m pip install --upgrade pip pipenv git+https://github.com/nodejs/tap2junit.git
+ln -s /opt/freeware/bin/tap2junit /usr/bin/tap2junit
+```
+
+
+# AIX 6.1 Install
 
 ## Install required packages
 
-AIX72: unneeded, package were installed by yum
 
 #### curl, unzip
 
@@ -57,8 +153,6 @@ rm -rf gcc
 
 #### git-tools
 
-AIX72: unneeded, installed with yum
-
 ```bash
 mkdir git-tools
 cd git-tools
@@ -73,8 +167,6 @@ rm -rf git-tools
 
 #### git
 
-AIX72: unneeded, installed with yum
-
 ```bash
 LIBPATH=/usr/lib curl --insecure -O https://ci.nodejs.org/downloads/aix/git-2.8.1-1.aix6.1.ppc.rpm
 rpm -ivh git-2.8.1-1.aix6.1.ppc.rpm --force
@@ -82,7 +174,6 @@ rpm -ivh git-2.8.1-1.aix6.1.ppc.rpm --force
 
 #### openssl, openssh
 
-AIX72: unneeded, installed with yum
 
 Download and scp to the machine:
 https://www-01.ibm.com/marketing/iwm/iwm/web/reg/pick.do?source=aixbp&lang=en_US
@@ -109,7 +200,6 @@ installp -Y -qaXgd . openssl openssh
 
 #### gettext, java, make
 
-AIX72: unneeded, installed by ansible in /home/iojs/
 
 ```bash
 LIBPATH=/usr/lib curl --insecure -O https://public.dhe.ibm.com/aix/freeSoftware/aixtoolbox/RPMS/ppc/gettext/gettext-0.19.7-1.aix6.1.ppc.rpm
@@ -126,7 +216,6 @@ rpm -i make-3.82-1.aix5.3.ppc.rpm
 
 #### libtool
 
-AIX72: unneeded, should not be installed
 
 Download and scp to the machine:
 http://www.bullfreeware.com/affichage.php?id=1458#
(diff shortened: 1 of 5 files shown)


AshCripps:
This is now good to go.

We are only adding the AIX7.1 machines currently as we are still waiting on confirmation of how many 7.2 machines we will get.

There are two test machines and one release machines

The next step will be connecting them to the CI

ansible/aix61-standalone/manualBootstrap.md
```diff
@@ -9,19 +9,136 @@ modules will be installed.
 ```bash
 chfs -a size=+1300000 /opt
 chfs -a size=+250000 /
-chfs -a size=917504 /tmp
+chfs -a size=+917504 /tmp
```
sam-github: The original was intended, it sets /tmp to a size that is required for citgm. Using a + just makes it bigger, but there is no need to make it bigger if its big enough already,.

ansible/aix61-standalone/manualBootstrap.md
```diff
@@ -9,19 +9,136 @@ modules will be installed.
...
+
+# AIX 7.2 Install
+
+Most packages should be installed via ansible.
+
+If there are any missing they should be installed via yum
+
+What you do need to install manually is **gcc** and **ccache**
+
+## gcc 6.3.x on AIX 7.2
```
sam-github: We are installing gcc6 from yum now, aren't we, so this can be removed?

ansible/aix61-standalone/manualBootstrap.md
```diff
@@ -9,19 +9,136 @@ modules will be installed.
...
+and then:
+
+```bash
+mkdir /aha
+mount /aha
+```
+
+## Install XL compilers
+
+1. Download 16.1.0 packages from: https://testcase.boulder.ibm.com (username:
```
sam-github: On the pdp machines, I think it might be easier to ask pdp if they can install them. Could be wrong, but they offered during machine provisioning.

ansible/aix61-standalone/manualBootstrap.md
```diff
@@ -9,19 +9,136 @@ modules will be installed.
...
+inutoc
+installp -aXgd ./ -e /tmp/install.log all
+```
+4. Find compilers in `/opt/IBM/xl[cC]/16.1.0/bin/`
+
+
+
+#### tap2junit
+
+Should be installed via ansible but if needed to be done manually use the following:
```
sam-github: I think it was done by ansible on aix7, wasn't it?

ansible/aix61-standalone/manualBootstrap.md
```diff
@@ -9,19 +9,136 @@ modules will be installed.
...
+
+## ccache 3.7.4 on AIX 7.2
+
+```bash
+mkdir -p /opt/ccache-3.7.4 && cd /opt/ccache-3.7.4
+curl -L https://ci.nodejs.org/downloads/aix/ccache-3.7.4.aix7.2.ppc.tar.gz | /opt/freeware/bin/tar -xzf -
+```
+## Enable the AHA fs
+
+For AIX 7.2 and 6.1, needed for the file watcher unit tests.
```
sam-github: AIX7.2 --> "AIX7"?

jenkins/scripts/select-compiler.sh
```diff
@@ -105,6 +105,22 @@ elif [ "$SELECT_ARCH" = "AIXPPC" ]; then
       ;;
   esac
 
+  case $NODE_NAME in
+    *aix71* )
```
sam-github: If it's too early to test this on AIX7.2, then this is fine, but I have reason to hope this will eventually be `*aix7* )`.

ansible/plugins/inventory/nodejs_yaml.py
```diff
@@ -45,10 +45,10 @@
     'type': ('infra', 'release', 'test'),
 
     # providers - validated for consistency
-    'provider': ('azure', 'digitalocean', 'joyent', 'ibm', 'linuxonecc',
+    'provider': ('azure', 'digitalocean', 'joyent', 'linuxonecc',
                  'macstadium', 'marist', 'mininodes', 'msft', 'osuosl',
                  'rackspace', 'requireio', 'scaleway', 'softlayer', 'voxer',
-                 'packetnet', 'nearform')
+                 'packetnet', 'pcc', 'nearform')
```
sam-github: typo? Dawson suggested `use "pdp" to stand for "Power Development Cloud"`

jenkins/scripts/select-compiler.sh
```diff
@@ -105,6 +105,22 @@ elif [ "$SELECT_ARCH" = "AIXPPC" ]; then
       ;;
   esac
 
+  case $NODE_NAME in
```
sam-github: don't need a new `case...`, add the block to the existing one at line 91, and this `*aix71* )` block would get inserted after line 105 (the `;;`)

ansible/inventory.yml
```diff
@@ -131,6 +134,11 @@ hosts:
         win10-arm64-1: { vs: '2017' }
         win10-arm64-2: { vs: '2017' }
 
+    - pcc:
```
sam-github: "pdp"

ansible/inventory.yml
```diff
@@ -71,6 +71,9 @@ hosts:
     - marist:
         zos13-s390x-1: {ip: 148.100.36.135, user: Unix1}
 
+    - pcc:
```
sam-github: "pdp"

ansible/aix61-standalone/manualBootstrap.md
AshCripps: I think I did this because that wasnt big enough but I cant remember why it wasnt big enough

ansible/aix61-standalone/manualBootstrap.md
AshCripps: I never installed these compilers so your call on whether to remove instructions and get them to do it

ansible/aix61-standalone/manualBootstrap.md
AshCripps: yes think so too

jenkins/scripts/select-compiler.sh
AshCripps: Am I right in thinking we have a 7.2 machine in ci? if so this seperation is needed because it wont have yum installed gcc6

ansible/plugins/inventory/nodejs_yaml.py
AshCripps: on the call we had with monday they said they got this order from pdp - they are actually `Poughkeepsie Client Center" I believe

ansible/aix61-standalone/manualBootstrap.md
sam-github: We should figure it out. The originals were those `+` statements, which were probably what was run, but doesn't help even a little bit when setting up new machines, where we need to know what the minimum required size is, not how much bigger the last machines disks needed to be!

I think I set the /tmp there so that all the machines would have the same /tmp size, so, should be big enough. If its not, then we need a new (absolute) size.

ansible/aix61-standalone/manualBootstrap.md
sam-github: fair enough. Leave as-is for now, and I'll adjust if necessary if/when we do libuv builds on these machines.

ansible/aix61-standalone/manualBootstrap.md
sam-github: so the whole section can be removed.

jenkins/scripts/select-compiler.sh
sam-github: we do not! :-) I setup 1 single 7.2 machine, but we can't add it until we have 3. And given what we now know about using yum to install gcc6, I'll want to revisit this. But you don't have to deal with this now. Forget I mentioned it, its unrelated to AIX7.1.

ansible/plugins/inventory/nodejs_yaml.py
sam-github: Internal conversation ended with "lets just call this `ibm`".

AshCripps:
@sam-github Addressed your comments PTAL

sam-github:
@richardlau PTAL -- I think this is purely additive, and mostly docs, so it should be possible to land without disturbing any existing builds.

ansible/aix61-standalone/manualBootstrap.md
```diff
@@ -12,16 +12,114 @@ chfs -a size=+250000 /
...
+        0509-022 Cannot load module make_64.
+        0509-150   Dependent module /opt/freeware/lib/libintl.a(libintl.so.8) could not be loaded.
+        0509-022 Cannot load module .
+```
+
+The fix is as following:
+
+```sh
+sudo rm /usr/lib/libiconv.a && sudo ln -s /opt/freeware/bin/libiconv.a /usr/lib
+```
```
richardlau: Is it possible to do this through ansible instead of a manual step?

ansible/aix61-standalone/manualBootstrap.md
```diff
@@ -12,16 +12,114 @@ chfs -a size=+250000 /
...
+mkdir -p /opt/gcc-6.3 && cd /opt/gcc-6.3
+curl -L https://ci.nodejs.org/downloads/aix/gcc-6.3-aix7.2.ppc.tar.gz | /opt/freeware/bin/tar -xzf -
+```
+
+## ccache 3.7.4 on AIX 7.2
+
+```bash
+mkdir -p /opt/ccache-3.7.4 && cd /opt/ccache-3.7.4
+curl -L https://ci.nodejs.org/downloads/aix/ccache-3.7.4.aix7.2.ppc.tar.gz | /opt/freeware/bin/tar -xzf -
+```
```
richardlau: Missing blank line before the next section.

ansible/aix61-standalone/manualBootstrap.md
```diff
@@ -243,6 +316,66 @@ pages earlier and not wait for the sync demon:
...
+
+LIBPATH=/usr/lib curl -L --insecure -O https://ci.nodejs.org/downloads/aix/gmake-dep.tar.gz
+gunzip -d gmake-dep.tar.gz
+tar -xf gmake-dep.tar
+```
+
+
+## Install python3
+
+AIX72: uneeded, installed by yum
```
richardlau: This line is unnecessary since it's still in the AIX 6.1 section? (Other instances seem to have been removed.)

ansible/plugins/inventory/nodejs_yaml.py
```diff
@@ -45,10 +45,10 @@
     'type': ('infra', 'release', 'test'),
 
     # providers - validated for consistency
-    'provider': ('azure', 'digitalocean', 'joyent', 'ibm', 'linuxonecc',
+    'provider': ('azure', 'digitalocean', 'joyent', 'linuxonecc',
                  'macstadium', 'marist', 'mininodes', 'msft', 'osuosl',
                  'rackspace', 'requireio', 'scaleway', 'softlayer', 'voxer',
-                 'packetnet', 'nearform')
+                 'packetnet', 'ibm', 'nearform')
```
richardlau: Any particular reason why `ibm` was moved here? It was already in the set so this change is not needed, or if it is changed maybe this should be sorted alphabetically?

ansible/inventory.yml
```diff
@@ -131,6 +134,11 @@ hosts:
         win10-arm64-1: { vs: '2017' }
         win10-arm64-2: { vs: '2017' }
 
+    - ibm:
+        aix71-ppc64_be-1: {ip: 129.33.196.197, user: b9s010a}
+        aix71-ppc64_be-2: {ip: 129.33.196.198, user: b9s010a}
```
richardlau: Should we attempt to keep providers sorted alphabetically? (It isn't quite before this change but almost is.)

ansible/aix61-standalone/manualBootstrap.md
sam-github: Its not safe to do, it could really damage the system. I'm not sure how badly, but possibly make ever executable unrunnable. I did it because the upstream yum packages don't seem to do the right thing here and had to figure out a way to make make build, and only after manual verification that the .so's (which are stored _inside_ static archives) were bitwise identical. It could be useful to add some info here: `ar x libiconv.a` on the two archives will extract  the .so from them. In this system (but not the aix7.2 I worked on 2 months ago, from the same yum package repo) the .a file in /opt was a pure superset of the one in /lib, so symlinking works.

ansible/plugins/inventory/nodejs_yaml.py
AshCripps: Good catch my fault, mustve moved it when I changed it to pcc and forgot to move it back

ansible/aix61-standalone/manualBootstrap.md
AshCripps: good catch

ansible/inventory.yml
AshCripps: I agree

PR #2067, opened 2019-11-27
ansible: set JOBS to 4 for AIX

sam-github:
Some of the builds were taking the 20 CPUS to mean run with JOBS=20, but
the machines don't have enough memory to run at that kind of build
concurrency. Throttle to 4. Some jobs were doing this already, but by
setting it in the global env, it should be inherited by all jobs.

### DIFF
--- a/ansible/aix61-standalone/resources/S20jenkins
+++ b/ansible/aix61-standalone/resources/S20jenkins
@@ -15,6 +15,7 @@ start )
               export NODE_TEST_DIR=$HOME/tmp; \
               export NODE_COMMON_PIPE="$HOME/test.pipe"; \
               export OSTYPE=AIX61; \
+              export JOBS=4; \
               export GIT_SSL_CAINFO="$HOME/ca-bundle.crt"; \
               export SSL_CERT_FILE="$HOME/ca-bundle.crt"; \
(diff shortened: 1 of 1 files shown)


ansible/aix61-standalone/resources/S20jenkins
```diff
@@ -15,6 +15,7 @@ start )
               export NODE_TEST_DIR=$HOME/tmp; \
               export NODE_COMMON_PIPE="$HOME/test.pipe"; \
               export OSTYPE=AIX61; \
+              export JOBS=4; \
```
richardlau: https://ci.nodejs.org/job/node-test-commit-aix/ has:

```bash
export JOBS=5
```

I'm okay with 5 or 4.

richardlau:
For some further background see https://github.com/nodejs/node/issues/27186#issuecomment-558839150.

richardlau:
@AshCripps @sam-github Do we also need to set `JOBS` for the AIX 7.1/7.2 systems, which use [`ansible/roles/jenkins-worker/templates/aix.rc2.j2`](https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/templates/aix.rc2.j2)?

sam-github:
@richardlau This PR was triggered by builds failing with out of memory, I don't think we need to make the changes anywhere that the problem is not being seen. There isn't any relationship between AIX7 mem and cpu provisioning, and AIX6.1, different machines, different host, different resource amounts allocated.

PR #2073, opened 2019-12-01
feat: add link to codecov.io

bcoe:
## What is this?

We are now collecting combined coverage on codecov.io for Windows and Linux (@Trott plans to also add additional coverage for other Linux runtimes).

What's nice about this project, is that it will allow us to eventually hit 100% coverage on the Node.js project, despite some code paths being conditional on platform.

codecov

### DIFF
--- a/jenkins/scripts/coverage/generate-index-html.py
+++ b/jenkins/scripts/coverage/generate-index-html.py
@@ -28,6 +28,13 @@ with open('out/index.html', 'w') as out:
     <link rel="stylesheet" href="https://nodejs.org/static/css/styles.css">
     <style>
         #logo { margin-bottom: 1rem; }
+        #badge {
+          margin-top: 1rem;
+          margin-bottom: 1rem;
+        }
+        #badge, a {
+          border: none;
+        }
         main { margin-bottom: 2rem; }
         .table-header,
         .table-row {
@@ -72,8 +79,13 @@ with open('out/index.html', 'w') as out:
   <div id="main">
     <div class="container">
       <h1>Node.js Nightly Code Coverage</h1>
+      Nightly combined JavaScript coverage, across Windows, Linux, and a variety of runtimes,
+      can be found on <a href="https://codecov.io/gh/nodejs/node">codecov.io</a>:
+      <a href="https://codecov.io/gh/nodejs/node">
+        <img id="badge" alt="Combined coverage" src="https://codecov.io/gh/nodejs/node/branch/master/graph/badge.svg" />
+      </a>
       <h3>
-        Node.js Core&nbsp;&nbsp;<a href="https://github.com/nodejs/node">&rarr;</a>
+        Nightly Linux Coverage&nbsp;&nbsp;<a href="https://github.com/nodejs/node">&rarr;</a>
       </h3>
       <main>
         <div class="page-content">
--- a/jenkins/scripts/coverage/generate-index-html.py
+++ b/jenkins/scripts/coverage/generate-index-html.py
@@ -79,13 +79,13 @@ with open('out/index.html', 'w') as out:
   <div id="main">
     <div class="container">
       <h1>Node.js Nightly Code Coverage</h1>
-      Nightly combined JavaScript coverage, across Windows, Linux, and a variety of runtimes,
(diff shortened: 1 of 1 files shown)


bcoe:
CC: @Trott, @joaocgreis, @mhdawson, @rvagg.

jenkins/scripts/coverage/generate-index-html.py
```diff
@@ -72,8 +79,13 @@
...
     <div class="container">
       <h1>Node.js Nightly Code Coverage</h1>
+      Nightly combined JavaScript coverage, across Windows, Linux, and a variety of runtimes,
+      can be found on <a href="https://codecov.io/gh/nodejs/node">codecov.io</a>:
+      <a href="https://codecov.io/gh/nodejs/node">
+        <img id="badge" alt="Combined coverage" src="https://codecov.io/gh/nodejs/node/branch/master/graph/badge.svg" />
+      </a>
       <h3>
-        Node.js Core&nbsp;&nbsp;<a href="https://github.com/nodejs/node">&rarr;</a>
+        Nightly Linux Coverage&nbsp;&nbsp;<a href="https://github.com/nodejs/node">&rarr;</a>
```
rvagg: Should this be more descriptive? The above text tells you to look elsewhere for "nightly ... linux" coverage but this is also "nightly linux coverage". Is this is just stuff we run ourselves? Should we describe the difference a bit more?

bcoe: Perhaps, "Nightly Linux Only".

As it sits today, the main difference is:

* we host it ourselves.
* it does not combine coverage from Windows, Linux, etc.
* it has block coverage, whereas the combined coverage is line coverage only (this is a bug with codecov which I believe is being worked on).

Trott: > it has block coverage,

That should be branch coverage, and yeah, that one's a big deal for me. But I might be the only one who cares, to be honest! The combined Windows coverage is A+++++ and I'm looking forward to also get a non-Intl build merged in there as well, as that will also cover some otherwise-uncovered code. 🎉

rvagg: I'm OK with whatever you two think appropriate, just flagging that it _seems_ a bit, but I don't tend to consume these pages myself anyway.

jenkins/scripts/coverage/generate-index-html.py
```diff
@@ -72,8 +79,13 @@
   <div id="main">
     <div class="container">
       <h1>Node.js Nightly Code Coverage</h1>
+      Nightly combined JavaScript coverage, across Windows, Linux, and a variety of runtimes,
```
Trott: It's not yet across "a variety of runtimes", is it? It's just Linux and Windows for now?

jenkins/scripts/coverage/generate-index-html.py
bcoe: I think the branch coverage is useful for helping encourage new contributions, once we have this on `codecov.io`; and have seen it working for a few weeks, I'd advocate retiring the self-hosted builds...

I talked to codecov.io, and there's literally just an `if` statement that turns off block coverage for JavaScript. CC: @thomasrockhu.

jenkins/scripts/coverage/generate-index-html.py
bcoe: removed the variety blurb for now.

jenkins/scripts/coverage/generate-index-html.py
bcoe: I went ahead and called the linux run, "Nightly Branch Coverage", since this is the main reason we want it to stick around for the time being.

bcoe:
@Trott @richardlau @joaocgreis @rvagg, I think this is ready to land; I don't have permission on this repository.

mhdawson:
Looks like this could be closed since its been merged.

mhdawson:
@bcoe does it include C/C++ coverage as well? I see the JS coverage but did not find the C/C++ after looking quickly.  This is in relation to retiring the existing builds.

bcoe:
@mhdawson it does not yet, it totally could; my only concern is I believe the reports would be the combined view, vs., being split out quite as nicely.

Issue #1840, opened 2019-06-18
arm-fanned perma-build-failing


Trott:
arm-fanned task git-nodesource-update-reference is red eight times in a row. 

```console
remote: Compressing objects: 100% (91555/91555), done.        
21:26:48 Receiving objects:   0% (1/394400)   
fatal: write error: Bad file descriptor
21:26:48 fatal: index-pack failed
21:26:49 debug1: client_input_channel_req: channel 0 rtype exit-signal reply 0
21:26:50 Build step 'Execute shell' marked build as failure
21:26:50 Collecting metadata...
21:26:50 Metadata collection done.
21:26:50 Notifying upstream projects of job completion
21:26:50 Finished: FAILURE
```
  
arm-fanned combined with "Bad file descriptor" makes me automatically suspect NFS trouble, but that may be a guess from anecdote and ignorance. @rvagg

rvagg:
taken arm-fanned offline, time for some whole-cluster maintenance I think, it's been a couple of months.

rvagg:
@refack: there's 2 pi's marked offline by you back in April: test-requireio_securogroup-debian9-armv6l_pi1p-2 "Keep offline to save space" and test-requireio_davglass-debian9-armv6l_pi1p-1 "Keep offline to safe space" - can you explain what this means?

test-requireio_davglass-debian9-armv6l_pi1p-1 has been dodgy in the past, I wouldn't be surprised if it's a perma-flake at the hardware level, but I'm not sure if I can interpret these messages as "Rod needs to check the hardware" or some other thing that can just be re-enabled?

@Trott they're enabled again after an update and some reboots. Seems to be a permafail on test.parallel/test-worker-debug on master though. Is this a known issue?

sam-github:
@nodejs/workers See above, test-worker-debug permafailing on arm, shall we mark it as flaky?

Trott:
@rvagg @sam-github test-worker-debug is a known-flaky on Windows, but I don't recall it being much of an issue on ARM. https://github.com/nodejs/node/issues/28106

refack:
> there's 2 pi's marked offline by you back in April "Keep offline to save space"

Since we don't test master and v12 on PI1, I wanted to see if having two machines offline will stave off filling up the shared storage, and/or cause any significant backlog.
I think it's a no on both - no significant backlog on PI1s, but we had hit storage overflow twice since then.

tl;dr AFAICT we can permanently retire 2 PI1 machines, if we want.

[...]

PR #2077, opened 2019-12-06
pi: reprovision all Pi's with Raspbian Buster, add docs

rvagg:
Raspbian Stretch to Raspbian Buster. I also discovered that a sound driver update sometime this year was the cause of boot instability, so that's now disabled at boot and they're running much smoother now.

Have updated the additional docs in here for Pi's to be current and make the setup (somewhat) reproducable.

arm-fanned is back online now after its hiatus, still ironing out some minor hiccups though.

Ref: https://github.com/nodejs/build/issues/1840
Ref: https://github.com/nodejs/node/issues/30786

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -23,6 +23,7 @@ become_method       = sudo
 [hosts:requireio]
 ansible_ssh_common_args = -o ProxyCommand='ssh -i {{ ansible_ssh_private_key_file }} -p 2222 jump@vagg-arm.nodejs.org nc -q0 %h %p'
 become_method           = sudo
+ansible_python_interpreter = /usr/bin/python3
 
 [hosts:smartos]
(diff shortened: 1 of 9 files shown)


rvagg:
I also think that arm-fanned wasn't properly handling Node <12. I enabled git in the parent job so it could get the `NODEJS_MAJOR_VERSION` to perform its subjob branching. That may have been coming from node-test-commit already but it was only checking for `NODE_MAJOR_VERSION` which I don't think is anywhere. Regardless, it meant that you couldn't run arm-fanned independently and get the appropriate version running. So now node-test-binary-arm and node-test-binary-arm-12+ are selected accordingly.
When doing that, I also noticed that node-test-binary-arm was selecting wheezy on the Pi2's which was removed some time ago, so now it's only wheezy on the Pi1's and stretch on the rest which means wheezy can be properly retired when Node 8 goes away I think.

PR #2079, opened 2019-12-06
ansible: fix up release-centos6

rvagg:
I had to rebuild this machine, it wasn't connected to Jenkins and I couldn't ssh into it and force reboot didn't do anything.. Later I noticed the IP address we had wasn't the one for the machine. I don't _think_ it got changed when I ordered a rebuild in SoftLayer, but I'm also not sure what the deal is with this new address anyway because it wasn't in ci-release's firewall.

Anyway, rebuild resulted in yaks to shave, see ansible changes here.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -56,7 +56,7 @@ hosts:
         ubuntu1604-armv7l-2: {ip: 212.47.234.107}
 
     - softlayer:
-        centos6-x64-1: {ip: 169.62.77.228}
+        centos6-x64-1: {ip: 50.97.245.10}
 
     - linuxonecc:
         rhel72-s390x-1: {ip: 148.100.110.65}
--- a/ansible/roles/baselayout/tasks/main.yml
+++ b/ansible/roles/baselayout/tasks/main.yml
@@ -146,7 +146,7 @@
   when: "os == 'centos6'"
   yum:
     disable_gpg_check: true
-    name: "devtoolset-2-gcc,devtoolset-2-gcc-c++"
+    name: "devtoolset-2-gcc,devtoolset-2-gcc-c++,devtoolset-2-binutils"
     state: present
 
 - name: remove fortune from login shells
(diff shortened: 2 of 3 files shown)


rvagg:
also, some of the reprovisioned smartos machines weren't in the ci-release firewall so were not connected. So I've done that now and we have most machines connected to ci-release now and it seems to be in a happier state.

rvagg:
release-osuosl-centos7-ppc64_le-1 wasn't in the firewall either. I think that they haven't been added to the iptables save file because I rebooted ci-release and it would have loaded its save.

for the record, here's my workflow for messing with the firewall on both CI hosts. I don't use `iptables-save` directly cause it feels too brute force to me and I find it makes it difficult to _remove_ hosts, but it could be used here instead:

* if I need to remove a host I run `iptables -L --line-numbers > /tmp/iptables.numbers` and find the number of the host I need to remove. If I need to remove more than one I start from the highest number and work downward (cause they'll shuffle otherwise!). Remove with `iptables -D jnlp X` where X is the number.
* manually edit /etc/iptables/rules.v4 and remove whatever I removed with `-D` and add what I want added. I then copy any lines I add and run them prefixed with `iptables` from the commandline. Then I know any future `iptables-save` and `iptables-load` will do what we need.

mhdawson:
@rvagg which smartos release machines. I'd added them to the firewall, made sure they were connected and then did test builds. Just to say I'm not sure how they would not be in the firewall?

mhdawson:
This is the doc that we have for how to add the firewall entries: https://github.com/nodejs/build/blob/master/jenkins/doc/add-worker-firewall-entry.md

If it needs to be something else we should probably update taht.

rvagg:
I think it was the reboot, they weren't saved properly - perhaps there was a problem with the automatic save that's supposed to happen on shutdown, I'll try and find time to look into it and work out why it's not beign updated or maybe we update our docs to force save of iptables/rules.v4.

rvagg:
yep, thanks

PR #2086, opened 2019-12-10
ansible: remove hardcoded cert location for AIX jenkins

AshCripps:
This was causing build failures on the new AIX 7.1 machines as they couldnt perform a git pull.

ive updated the machines with this and have restarted the agents but will need someone to kick off a test build of node-test-commit-aix71 to ensure it worked.

### DIFF
--- a/ansible/roles/jenkins-worker/templates/aix.rc2.j2
+++ b/ansible/roles/jenkins-worker/templates/aix.rc2.j2
@@ -20,7 +20,7 @@ start )
               export HOME=/home/{{server_user}}; \
               export NODE_TEST_DIR=$HOME/tmp; \
               export NODE_COMMON_PIPE="$HOME/test.pipe"; \
-              export OSTYPE=AIX7; \
+              export OSTYPE=AIX72; \
         /home/{{server_user}}/jdk8u192-b12-jre/bin/java -Xmx128m -Dorg.jenkinsci.plugins.gitclient.Git.timeOut=30 -jar "$HOME/slave.jar" \
           -secret {{secret}} \
           -jnlpUrl {{jenkins_url}}/computer/{{inventory_hostname}}/slave-agent.jnlp >$jenkins_log_file 2>&1 &'
--- a/ansible/roles/jenkins-worker/templates/aix.rc2.j2
+++ b/ansible/roles/jenkins-worker/templates/aix.rc2.j2
@@ -20,9 +20,7 @@ start )
               export HOME=/home/{{server_user}}; \
               export NODE_TEST_DIR=$HOME/tmp; \
               export NODE_COMMON_PIPE="$HOME/test.pipe"; \
-              export OSTYPE=AIX72; \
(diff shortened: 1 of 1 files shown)


sam-github:
https://ci.nodejs.org/job/node-test-commit-aix71/3/

ansible/roles/jenkins-worker/templates/aix.rc2.j2
```diff
@@ -20,9 +20,7 @@ start )
               export HOME=/home/{{server_user}}; \
               export NODE_TEST_DIR=$HOME/tmp; \
               export NODE_COMMON_PIPE="$HOME/test.pipe"; \
-              export OSTYPE=AIX72; \
-              export GIT_SSL_CAINFO="$HOME/ca-bundle.crt"; \
-              export SSL_CERT_FILE="$HOME/ca-bundle.crt"; \
+              export OSTYPE=AIX7; \
```
sam-github: I'm not sure where the `OSTYPE` env var is used, do you? Maybe it should be left as-is in this PR, I see you've another PR to change "aix72" to "aix7", would this be better there?

richardlau: Anyone know what uses `OSTYPE` and whether we need to have a version number in there (as opposed to just `AIX`)?

sam-github: I got confused, its "rhel72" that is changing to "rhel7"... sorry! I'd be tempted to leave this single line alone, it does no harm, is possibly unused, and sometime we can track it down and figure out if it is dead code or not, but that's unrelated to the cert location fix, which is understood, and we can just do now in this PR.

AshCripps: I only changed it because I thought it might have had something to do with select-compiler, but it doesnt Ill happily swap it back

AshCripps: @sam-github @richardlau Im not 100% sure but there is a reference to `OSTYPE` in the makefile here https://github.com/nodejs/node/blob/024b6588313a0e32595317936ee279dec7a5a756/Makefile#L12

Could this be what its used for? and if so why are we setting it here when it gets set in the makefile?

sam-github: `$(TARNAME)-$(OSTYPE)-$(ARCH).tar.gz` ... so, it may be wrong as it is, but its unrelated to the cert problems, lets leave this cleanup, and the decision about whether its a good idea to later.

sam-github: though also, its initialized with `OSTYPE := $(uname ...` in Makefile, so I'm not sure the env is even used anymore. Either way, its unrelated, so lets not change now.

AshCripps:
@sam-github @richardlau ive fixed up and removed the OSTYPE reference as not needed PTAL

Issue #2084, opened 2019-12-09
N-API testing Windows failures 


mhdawson:
We've had failures across a number of machines over the last few days which look most like machine problems. For example: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-new/1272/

I can't find those machines in the inventory, so not sure how to figure out. Is there a PR I did not find to add them?

Looking at the tags we were using win10 and win2012r2-vs2015 but those are not used by the current testing.

So I'm also not sure what we should be running the N-API tests on.

I tried:

win-vs2015
win-vs2017
win-vs2019
win-vcbt2015.

The win-vcbt2015 ones seem to fail consistently so I removed that.

@joaocgreis can you let me know what you think we should be testing on.

The test pulls down the release binaries for 8,x, 10.x, 12.x 13.x, master and runs the node-addon-api tests.

richardlau:
https://github.com/nodejs/build/issues/1996 is tracking changes to the Windows CI machines being made by @joaocgreis.

mhdawson:
Also wondering if we expect vs2019 to work with 8.x?  https://ci.nodejs.org/job/node-test-node-addon-api-new/nodes=win-vs2019/1282/console

richardlau:
> 
> 
> Also wondering if we expect vs2019 to work with 8.x? https://ci.nodejs.org/job/node-test-node-addon-api-new/nodes=win-vs2019/1282/console

Not out of the box. VS2019 support was added in node-gyp 5 which isn't in the version of npm that ships with the most recent Node.js 8.x.

mhdawson:
I'm thinking the version selector script should be filtering that out, but will need to check if there is something special to the N-API job that prevents that.

richardlau:
> 
> 
> I'm thinking the version selector script should be filtering that out, but will need to check if there is something special to the N-API job that prevents that.

Agreed. If you look in https://ci.nodejs.org/job/node-test-node-addon-api-new/1282/console you can see `win-vs2017` being filtered out by https://github.com/nodejs/build/blob/39cacfead86952db488ef76d88ab0ab3e405b38c/jenkins/scripts/VersionSelectorScript.groovy#L44 so it would seem to make sense to do the same for `win-vs2019`. I've raised https://github.com/nodejs/build/pull/2085.

mhdawson:
@richardlau thanks. I went to look at the selector script saw 2019 was missing and was happy to see your PR.

[...]

PR #2098, opened 2019-12-17
Enable jenkins testing on AIX71

AshCripps:
Start to enable testing to be done on the new AIX71 machines.

I cant remember if @sam-github added the label to node-test-commit-aix or not

also does `anyType` cover _both_ test and release or just test?


richardlau:
> I cant remember if @sam-github added the label to node-test-commit-aix or not

I don't think that has been done as otherwise builds would be running on the AIX 7.1 machines now.

> also does `anyType` cover _both_ test and release or just test?

It covers both. Or more technically it covers any setting of `buildType`: https://github.com/nodejs/build/blob/6bcc28ea91d2c767533869bcc264bb544a839b52/jenkins/scripts/VersionSelectorScript.groovy#L4-L5

AshCripps:
> I don't think that has been done as otherwise builds would be running on the AIX 7.1 machines now.

Yeah I checked our internal checklist and it was the next step for him to do.


> It covers both. Or more technically it covers any setting of `buildType`:

Thought so.

richardlau:
> > > also does `anyType` cover both test and release or just test?
> >
> > It covers both. Or more technically it covers any setting of `buildType`:
> 
> Thought so.

Furthermore currently the only type checked is `releaseType` (`buildType` is `release`). @joaocgreis is planning some changes (https://github.com/nodejs/build/issues/2084#issuecomment-564098021) that will also add a `testType` if we want to have different rules for test only.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -61,6 +61,7 @@ def buildExclusions = [
 
   // AIX61 -------------------------------------------------
   [ /aix61/,                          anyType,     lt(6)   ],
+  [ /aix71/,                          anyType,     lt(14)   ],
```
sam-github: nit: there appears to be one extra space in this line...

sam-github:
@AshCripps I think you might not have noticed:
> I think we also need to add a line to exclude RELEASE from aix71 for lt(15) (everything)

richardlau:
> 
> 
> @AshCripps I think you might not have noticed:
> 
> > I think we also need to add a line to exclude RELEASE from aix71 for lt(15) (everything)

We don't want to build 14 on AIX 7.1?

sam-github:
@richardlau yes, we _want_ to, but we do this in order, usually:
```
for each V in (14, 13, 12, 10)
  enable test builds for V on NEW MACHINE
  wait for tests to be green
  stop release builds on OLD MACHINE
  start release builds on NEW MACHINE
  // setup in .groovy must ensure that ONLY one machine gets selected to do release builds on a V
  wait for no reports of problems for users
end
```

I don't know, I guess we cloud start doing testing and releases on the 7.1 machines for _all_ release lines `> = 10` immediately, but we're generally more cautious.

I pushed a fixup commit.

@richardlau LGTY with above context? I could add the aix61 build label to node-test-commit-aix as soon as this lands.

richardlau:
SGTM. Thanks for the context.

AshCripps:
> @AshCripps I think you might not have noticed:
> 
> > I think we also need to add a line to exclude RELEASE from aix71 for lt(15) (everything)

Sorry @rsam completely blanked that thanks for the catch!

Issue #1996, opened 2019-10-24
Windows CI machine updates


joaocgreis:
@nodejs/build 

I will change and rebuild some/most of the Windows machines we have connected to CI to add VS2019 to the matrix and better balance other versions for the current needs.

I will try to keep `node-test-commit-windows-*` working all the time, but expect jobs to queue for longer periods when the number of machines is reduced.

I will try to keep other jobs working as well, but please let me know if something is not working as expected.

joaocgreis:
I'm planning to include the VS version in the host names, as metadata in the OS section (similar to `arm_cross` machines). This reinforces the restriction that only one VS version should be installed and makes the machines easier to manage, both because Ansible is made a bit simpler and we can know with a glace what version is installed.

Please let me know if there are any objections. (cc @rvagg)

richardlau:
> I will try to keep other jobs working as well, but please let me know if something is not working as expected.

@BethGriggs is seeing Windows CI failures on v10.x against releases that weren't seen at the time the releases were made (https://github.com/nodejs/node/issues/30639#issuecomment-559453240). Two Windows nodes have been marked offline in https://github.com/nodejs/build/issues/2070.

joaocgreis:
The new machines are in place and the Windows job has been updated. PR coming soon with build repo changes.

Resuming jobs before the change does not work, I expect this to be an issue in the next few days. I'll rebuild the jobs I see failing.

I've updated the libuv job, there are a few others that I still have to get to. Please let me know if there are issues in other jobs.

joaocgreis:
Many compile jobs have failed in the last day. I noticed this happens after a V8 canary compile job runs, every compile job of the same type that runs after it fails. I'm still investigating.

For now, if you see a compile job failing with a lot of errors like
```
openssl.def : error LNK2001: unresolved external symbol ...
```
please take the machine that ran it offline.

joyeecheung:
I think it might be related to https://github.com/nodejs/node/pull/30647, though from the build logs it might not be considering both `node_use_node_code_cache` and `node_use_node_snapshot` appeared to be true in the VS2019 builds. I'll try to see if reverting it makes a difference.

joaocgreis:
The issue of failed compile jobs has been fixed by deleting all clcache data when compilation fails. If this starts happening again, we'll probably have to disable clcache completely, unless someone has the bandwidth to investigate further.

richardlau:
[...]

Issue #2097, opened 2019-12-17
cmake on CITGM infrastructure


BridgeAR:
Some modules partially use cmake and seem to fail due to the lack of it on our infrastructure.

Would it be possible to add cmake to fix that?

targos:
Refs: https://github.com/nodejs/citgm/issues/770

rolftimmermans:
I'm a maintainer of zeromq.js and we use CMake to build the node add-on. Is there any way I can assist to get CMake added to the build systems/images?

sam-github:
If the machines have a working registry system (yum, apt, brew), and that system has a sufficiently up-to-date cmake, then this should be not too difficult, but like all things build, it takes a bit of legwork:

- List affected jobs: ...
- List affected build workers: ...
- List what version of cmake is available, if any, on their OS level: ...
- List exact name of the package (per host), and add it to the relevant ansible yml file: ...
- Run ansible on each of the hosts to install cmake
- Trigger builds of all affected jobs (see)

@rolftimmermans We don't onboard people into build very quickly because of the amount of root privs in CI it gives, but much of the work above doesn't require any privileges.

If someone took this to the stage where I just had to run ansible-playbook on a PR they put together, I'd find an hour to do that, but otherwise, I'm working on a few other build tasks ATM and won't have time to start this. Someone else might have time.

richardlau:
We do have some machines with cmake installed as we have two running libuv jobs:
- https://ci.nodejs.org/view/libuv/job/libuv-test-commit-linux-cmake/
- https://ci.nodejs.org/view/libuv/job/libuv-test-commit-windows-cmake/

I haven't looked (yet) at how those jobs specify which nodes to run on and whether there's overlap with the hosts the CITGM jobs currently run on.

joaocgreis:
Installed CMake on all Windows machines (included in PR https://github.com/nodejs/build/pull/2099).

BridgeAR:
What's the status here?

AshCripps:
cmake should now be available on all test macs

richardlau:
cmake is now on the IBM platforms in the CI (we're using cmake to build libuv).

bl-ue:
Stable bots don't seem all that useful 🤔

PR #2099, opened 2019-12-18
Windows Updates

joaocgreis:
This is all already in use by CI, but not yet by the release Jenkins. The release job need to be adjusted while landing this.

Notable changes:
- VS2019 support
- Windows now uses the VersionSelectorScript - it is a bit complex as expected, but it manages to filter for building node, building add-ons, and choose what runs where depending on what labels are available while always avoiding invalid combinations
- Moved most scripts from CI jobs to the build repo - this allows us to reuse parts of the scripts for different jobs, gives us some history for long-term changes, and it is easy to change to a fork for testing if needed
- Test building in debug mode
- Test building the MSI
- Python 3 added to all Windows machines
- (EDIT) CMake added to all Windows machines
- Windows moved to the main create worker playbook
- Some documentation updates

Re-created all Windows machines to balance available versions optimizing for current master.

Refs: https://github.com/nodejs/build/issues/1996


doc/windows-visualstudio-supported-versions.md
```diff
@@ -34,6 +37,9 @@ Supported versions for building Node.js from source.
 | v9              | 2015, VCBT2015, 2017                |
 | v10             | 2017 <sup>[5]</sup>                 |
 | v11             | 2017                                |
+| v12             | 2017, 2019 (flag) <sup>[8]</sup>    |
+| v13             | 2017, 2019 <sup>[9]</sup>           |
+| v13             | 2017, 2019                          |
```
richardlau: Is this line supposed to be `v14`?

joaocgreis: Yes, fixed, thanks!

joaocgreis:
Added a commit to install CMake (https://github.com/nodejs/build/issues/2097) and rebased.

richardlau:
> 
> 
> Added a commit to install CMake (#2097) and rebased.

It's not part of the Native Desktop Workload for Visual Studio? We have https://ci.nodejs.org/job/libuv-test-commit-windows-cmake/ already running in the CI.

joaocgreis:
As I understand it, using CMake from VS is a bit of an hack. Needs to be accessed by full path, so location will change with VS version. Also, it's not present in VS2015. Having it from Chocolatey should be a better option, more so if we start using it to build Node at some point.

ansible/roles/baselayout-windows/tasks/main.yml
```diff
@@ -17,10 +25,14 @@
...
     win_command: 'git config --global core.longpaths true'
+  - name: set Git user name
+    win_command: 'git config --global user.name "{{inventory_hostname}}"'
+  - name: set Git user email
+    win_command: 'git config --global user.email "ci@iojs.org"'
 
 # Necessary for junit output in Jenkins jobs
 - name: install tap2junit
-  win_command: 'pip2 install tap2junit'
+  win_command: 'pip3 install --upgrade tap2junit'
```
rvagg: I think on others we've done both pip2 and pip3 to make sure it's available on both versions of python, might that be necessary here too?

doc/node-test-commit-matrix.md
```diff
@@ -137,3 +101,92 @@ This is assumed correct as of the date of last commit. If you notice a discrepan
...
+
+### Combinations being tested
+
+- Run JS tests on all supported Windows Versions.
+- Run add-ons tests on all supported Visual Studio versions (preferably spanning all supported Windows Versions).
+- If necessary, run more combinations for the Visual Studio version that is used to build the official releases.
+
+The following combinations are used:
+
+#### Node.js version 8
```
rvagg: this section will be redundant in ~10 days, is it even worth merging this in?

doc/node-test-commit-matrix.md
```diff
@@ -137,3 +101,92 @@ This is assumed correct as of the date of last commit. If you notice a discrepan
...
+
+| Binaries produced by: | Run in Windows version: | Test add-ons with: |
+|-----------------------|-------------------------|--------------------|
+| VCBT 2015             | 10                      | VCBT 2015          |
+| VS 2015               | 2008R2                  | VS 2017            |
+|                       | 2012R2                  | VS 2013, VS 2015   |
+| VS 2015 (x86)         | 2012R2                  | VS 2015            |
+| VS 2017               | 2016                    | VS 2017            |
+
+#### Node.js versions 9 to 12
```
rvagg: no more 9, maybe just '10 & 12'

jenkins/scripts/ci-node-version.sh
```diff
@@ -0,0 +1,20 @@
...
+
+source env.properties
+
+set +x
+export GIT_SSH_COMMAND="ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i $JENKINS_TMP_KEY"
+set -x
+
+# Retry once
+git archive --format=tar --remote="$TEMP_REPO" $TEMP_BRANCH src/node_version.h -o node_version.h.tar ||
+git archive --format=tar --remote="$TEMP_REPO" $TEMP_BRANCH src/node_version.h -o node_version.h.tar
```
rvagg: this is interesting, fetching a single file from a remote repo? I get protocol problems trying this out locally so can't seem to replicate this behaviour.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -38,11 +45,32 @@ def buildExclusions = [
...
+  [ /vs2015(-\w+)?$/,                 testType,    gte(10)       ],
+  [ /vcbt2015(-\w+)?$/,               testType,    gte(10)       ],
+  [ /vs2017(-\w+)?$/,                 testType,    lt(8)         ],
+  [ /vs2019(-\w+)?$/,                 testType,    lt(13)        ],
+  [ /vs2015-x86$/,                    testType,    gte(10)       ], // compile x86 only once
+  [ /vs2017-x86$/,                    testType,    ltGte(10, 14) ],
+  [ /vs2019-x86$/,                    testType,    lt(14)        ],
+  // VS versions supported to build add-ons
+  [ /vs2013-COMPILED_BY/,             testType,    gte(9)        ],
+  [ /vs2015-COMPILED_BY/,             testType,    noVer         ],
```
rvagg: 👍 good idea to include these with `noVer` for completeness and documentation purposes, I like

jenkins/scripts/ci-node-version.sh
joaocgreis: I also had issues with http repos, but this works for our temp repo (ssh). This is already being used by node-test-commit-windows-fanned (running in jenkins-workspace).

ansible/roles/baselayout-windows/tasks/main.yml
joaocgreis: Isn't that redundant? tap2junit is invoked by the build scripts, so we have control over what exactly is called. If we have both, that would either just ignore one (the one not first in the path) or call an arbitrary one depending on which one is first in the path. Since tap2junit supports Python 3, I figured we could just go with that.

@cclauss do you know if we need both?

cclauss: In this case we just need a single instance of tap2junit and I would prefer that Python 3 is the way that we install it.

ansible/roles/visual-studio/tasks/main.yml
```diff
@@ -4,26 +4,51 @@
 # Install Visual Studio
 #
 
-- when: vs == '2013'
-  block:
-  # TODO: Ensure no other versions are installed
-  - name: install Visual Studio 2013
-    include_tasks: "partials/vs2013.yml"
+- name: install Visual Studio 2013
```
cclauss: Do we still want to test and support 2013?  https://docs.microsoft.com/en-us/visualstudio/releases/2019/servicing#support-for-older-versions-of-visual-studio

ansible/roles/visual-studio/tasks/partials/vs2019.yml
```diff
@@ -0,0 +1,28 @@
+---
+
+#
+# Install Visual Studio 2017
```
cclauss: 2019?

ansible/roles/visual-studio/tasks/main.yml
joaocgreis: VS2013 is used to test add-ons for Node.js v8. We should keep the machines available for some time after a Node version goes EOL, but anyway I'd like this to be in the commit history for documentation (we might need to test an old issue or something).

ansible/roles/baselayout-windows/tasks/main.yml
joaocgreis: @rvagg I plan to land this as is, let me know if I missed something.

joaocgreis:
Thanks for the reviews! This needs some work to land, iojs+release needs to have the same changes as the test release job, and the jobs in CI need to be changed to point at master. I'll do it later this week.

joaocgreis:
Thanks!

Adapted the jobs on both Jenkins instances.

Landed in https://github.com/nodejs/build/compare/1dd97fa746c3a22a5a09ee8990aff8ed14862091...36202a24f7d19729bea1e0efb0fe7045b2c42dc6

Issue #2110, opened 2019-12-19
aix: add AIX7.1 build machines to CI


sam-github:
Plan is a gradual addion of AIX7.1 to test configs, and then an even more gradual switch over of the release builds from AIX6.1 to AIX7.1.

See this for details:
- https://github.com/nodejs/build/pull/2098#issuecomment-567672131

This is a tracking issue, to reference from pull requests, and to comment on as changeover occurs, if problems occur, or to notify interested people what is happening.

sam-github:
I added `aix-7.1` to  the build labels for https://ci.nodejs.org/job/node-test-commit-aix/

Building:
- 14.x: https://ci.nodejs.org/job/node-test-commit-aix/27417/ (should build on 7.1)
- 12.x: https://ci.nodejs.org/job/node-test-commit-aix/27418/ (should not build on 7.1 for the moment)

richardlau:
> 
> 
> I added `aix-7.1` to the build labels for https://ci.nodejs.org/job/node-test-commit-aix/
> 
> Building:
> 
>     * 14.x: https://ci.nodejs.org/job/node-test-commit-aix/27417/ (should build on 7.1)
> 
>     * 12.x: https://ci.nodejs.org/job/node-test-commit-aix/27418/ (should not build on 7.1 for the moment)

`aix71-ppc64` would follow the existing labelling convention.

sam-github:
@richardlau Am I greatly misunderstanding the groovy setup?

https://github.com/nodejs/build/blob/48e99171cb768e4483a3d1afc5eca4345accf9b0/jenkins/scripts/VersionSelectorScript.groovy#L64 looks to me like it excludes aix7.1 for lt 14, but its running:
```
13:07:51 Commit message: "Working on v12.14.1"
...
13:07:56 Triggering node-test-commit-aix » aix61-ppc64
13:07:56 Triggering node-test-commit-aix » aix-7.1
```

- https://ci.nodejs.org/job/node-test-commit-aix/27418/console

... And I think I see  the problem, I selected build label `aix-7.1`, but the groovy script needs to apply to `aix71-ppc64`...

I changed the build label selected for the job.

richardlau:
@sam-github I think the job isn't using the VersionSelectorScript -- most likely because previously we haven't had to exclude anything on AIX before.

sam-github:
I'll go check... but I think it worked this last time:

[...]

PR #2111, opened 2019-12-24
jenkins: use AIX7.1 for 13.x and 14.x

AshCripps:
v14 testing is looking green so this pr will enable testing on v13.

Ive also included changes to allow for releasing 14 from 7.1. LMT if anyone feels this is too soon/incorrect.

refs: #2110

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -61,8 +61,9 @@ def buildExclusions = [
 
   // AIX ppc64 ---------------------------------------------
   [ /aix61/,                          anyType,     lt(6)   ],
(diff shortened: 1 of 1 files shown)


sam-github:
I'd change the commit message to be more generic and enumerate the changes in the description, since it only describes one thing that the commit does, and add exclusion of aix61 for `gt 13`. Or maybe one commit for each change since they are so independent:
- 13.x test with 7.1
- 14.x release with 7.1
- 14.x exclude test with 6.1

AshCripps:
Ive split the commits up, PTAL.

sam-github:
- v13.x: https://ci.nodejs.org/job/node-test-commit-aix/27612/

PR #2113, opened 2019-12-30
jenkins: add gt operator to VersionSelectionScript

AshCripps:
Add the gt operator to VersionSelectionScript.


rvagg:
Wait, this was merged kind of fast. Is this because the AIX thing broke?
There's a clear delineation between lt & gte since they leave no grey area between, a gt operator adds additional ambiguity. The AIX entry should have been changed to use `gte(14)` instead of introducing this new thing.

AshCripps:
@rvagg Yes this was done to fix AIX, I can revert if you like to remove this ambiguity as I have no preference on which operators are used.

PR #2124, opened 2020-01-03
chore: Remove Windows 2008 for EOL

### LINKED ISSUE #1663: FYI: Windows 7 end-of-extended-support to be 2020-01-14
Refs: https://support.microsoft.com/en-us/help/4487594
Refs: https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet#section-2

Microsoft has announced that on January 14, 2020 it will end it's support for Windows 7 and Windows Server 2008 R2 (`WINVER === 0x0601`).

Should we recommend that node12 minimal supported version should be Windows8.1/Server2012R2 (`WINVER === 0x0603`)?

/CC @nodejs/platform-windows @joaocgreis

nschonni:
EOL is January 14th 2020
Closes #1663


joaocgreis:
This LGTM, but needs to wait for the actual EOL date to land.

This can only land at the same time or after the machines are removed. Need to be removed from Jenkins and also from the secrets.

The core repo should also be updated, though I don't think it's urgent.

nschonni:
@joaocgreis feel free to redo this when you're cleaning up the machines if it's easier. Also a ping now that the EOL dates has passed :)

sam-github:
What's the next steps here?

nschonni:
@sam-github I think the machines would need to be actually decommissioned. This is just the config related bits that would that would change after that happens.

joaocgreis:
Landed in https://github.com/nodejs/build/commit/cf0f93b72a18757fbb820b233206b9f4f0bbeba7

Thanks!

I've disabled the machines in Jenkins (by removing the labels). If all goes well, I'll remove them next week.

PR #2137, opened 2020-01-20
Update generate-index-html.py

XhmikosR:
Add image dimensions in badge


sam-github:
@mhdawson can you review?

sam-github:
@XhmikosR If this landed, what would it do? Its a python script, does it need to be run? Is it already run as part of a job somewhere? Is this related to some issue somewhere else?

XhmikosR:
@sam-github the coverage site is generated via this script. This patch just adds the image dimensions so that there are no reflows.

There's no related issue, I just noticed it.

PR #2143, opened 2020-01-24
use adoptopenjdk on Linux/s390x as it has a JIT

sxa555:
The java being used for the jenkins agent is a java 8 with a HotSpot JVM which does not have JIT compiler. In order to speed up the jenkins processes running on the machine I propose switching to an AdoptOpenJDK/OpenJ9 JVM which does have a JIT. This PR will make that change.

(May require further adjustments to the `jenkins-worker` directories if the full path has to be encoded somewhere)

Current output from one of the build machines shows it using the Zero VM

```
[linux1@test-ibm-rhel7-s390x-3 ~]$ /usr/bin/java -version
openjdk version "1.8.0_232"
OpenJDK Runtime Environment (build 1.8.0_232-b09)
OpenJDK 64-Bit Zero VM (build 25.232-b09, interpreted mode)
[linux1@test-ibm-rhel7-s390x-3 ~]$
```

Looking for reviews on the concept - the code hasn't been tested yet ;-) Am tempted to do a bit of an overhaul of this to put the JVMs that are downloaded outside the repos into `/usr/lib/jvm` and standardise the two builds that we will now have from adoptopenjdk but that can wait until later ...
Signed-off-by: Stewart Addison <sxa@uk.ibm.com>

### DIFF
--- a/ansible/roles/java-base/tasks/main.yml
+++ b/ansible/roles/java-base/tasks/main.yml
@@ -67,7 +67,7 @@
     url: https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u242-b08_openj9-0.18.1/OpenJDK8U-jdk_s390x_linux_openj9_8u242b08_openj9-0.18.1.tar.gz
     dest: /tmp/
 - name: unarchive java s390x
-  when: os == "rhel7" arch == "s390x"
+  when: os == "rhel7" and arch == "s390x"
   unarchive:
     src: /tmp/OpenJDK8U-jdk_s390x_linux_openj9_8u242b08_openj9-0.18.1.tar.gz
     remote_src: yes
--- a/ansible/roles/java-base/tasks/main.yml
+++ b/ansible/roles/java-base/tasks/main.yml
@@ -61,6 +61,26 @@
     dest: "/usr/bin/java"
     state: link
 
+- name: download java on s390x
+  when: os == "rhel7" and arch == "s390x"
+  get_url:
+    url: https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u242-b08_openj9-0.18.1/OpenJDK8U-jdk_s390x_linux_openj9_8u242b08_openj9-0.18.1.tar.gz
+    dest: /tmp/
+- name: unarchive java s390x
+  when: os == "rhel7" arch == "s390x"
+  unarchive:
+    src: /tmp/OpenJDK8U-jdk_s390x_linux_openj9_8u242b08_openj9-0.18.1.tar.gz
+    remote_src: yes
+    dest: /opt
+  tags: java
+
+- name: symlink java s390x
+  when: os == "rhel7" and arch == "s390x"
+  file:
+    src: "/opt/jdk8u242-b08/bin/java"
+    dest: "/usr/bin/java"
+    state: link
+
 # if this fails you want to check in vars/main.yml and add package name
 # as appropriate -- try to use generic os family if available.
 


sam-github:
Builds are slow, I'd be surprised if java performance was an issue, but I'm regularly surprised. Could we try it on one machine by modifying it by hand, and characterize the performance gain? If we don't notice, then maybe its not worth it.

I'm always a bit concerned by incremental creep in the complexity of the setup, and it was nice to have systems with java built-in, that didn't need special java setup tasks to run after the fact, but that's not an objection, if CI responsiveness is noticeably improved, its worth the effort.

https://github.com/nodejs/build/tree/master/ansible/roles/jenkins-worker/templates is the jenkins start scripts, they usually have hard coded references to the java location, for example, the rhel systems affected by this PR are https://github.com/nodejs/build/blob/1eaa6bf57cc948d8638bfd2e7338bb1078b13052/ansible/roles/jenkins-worker/templates/systemd.service.j2#L26

Note that systemd.service isn't specific to s390x, so all rhel archs would need to have the same locations for java, hopefully adoptopenjdk exists for all of them. If not, then somehow the ansible scripts will have to do different things on different rhel architectures, and maybe parameterize the systemd.service (complexity creep).

sxa555:
Since my PR creates the symlink to `/usr/bin/java` that should be ok. We could arguable reduce complexity by moving everything to using adoptopenjdk as the boot jdk ;-) It's not immediately clear from the PR that initially introduced adoptopenjdk for ppc64le why (Just a reference to something on a build WG call).

Ref performance, yes I agree it would be good to validate it. I raised this mostly on the basis that I saw quite a bit of a performance degradation on another project when I accidentally used a non-JIT enabled JVM on a Linux/s390x system during the final "archiving artefacts" stage. Of course, the Node jobs may not see such a change but it would be worth testing it. I can work with @AshCripps or yourself to test it on one of the machines.

sam-github:
> We could arguable reduce complexity by moving everything to using adoptopenjdk as the boot jdk ;-) 

I'd be OK with that. If every system does the same thing, it can often get easier to maintain (though when that means building gcc from src on every platform, that makes things worse, not better, but this is just a curl). Java is one of those things a lot of the platforms I've worked on seem to not have adequate packages for, so just uniformly curling and putting in place a single java runtime sounds fine to me, particularly if it reduces the ansible per-platform condiational behaviour.

ansible/roles/java-base/tasks/main.yml
```diff
@@ -61,6 +61,27 @@
     dest: "/usr/bin/java"
     state: link
 
+ - name: download java on s390x
+   when: os == "rhel72" and arch == "s390x"
```
richardlau: Aren’t the new machines `rhel7` (without the `2`)?

AshCripps: Yes they are

sxa555: Do you have a conditional that's used elsewhere for "Linux" generally? That's what I was aiming for (Or basically "s390x & !zos")

ansible/roles/java-base/tasks/main.yml
```diff
@@ -61,6 +61,27 @@
...
+   get_url:
+     url: https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u242-b08_openj9-0.18.1/OpenJDK8U-jdk_s390x_linux_openj9_8u242b08_openj9-0.18.1.tar.gz
+     dest: /tmp/
+
+ - name: unarchive java s390x
+   when: os == "rhel72" arch == "s390x"
```
AshCripps: Still say `72` insted of `7` here and above

sxa555:
[test-ibm-rhel7-s390x-4](https://ci.nodejs.org/computer/test-ibm-rhel7-s390x-4) has had the JVM updated to adoptopenjdk jdk8u242-b08 / OpenJ9.

The first two JDK14 test-commit jobs which ran on the new JVM ran in 5m28 and 4m38 which. The first one would have been sinject to JIT warm-up but 4m38 (job 18634) for a passing job is definitely quicker than anything else that's ran in the last 20 or so. There have been a couple of occurrences of it running in 4m43 but most are over 5minutes. We'll see going forward but it's certainly looking like there's no disadvantage.

sxa555:
Every run of [node-test-commit-linuxone on rhel7](https://ci.nodejs.org/job/node-test-commit-linuxone/nodes=rhel7-s390x/buildTimeTrend) that has run on test-ibm-rhel7-s390x-4 since we switched the JVM (Other then the first one warming the JIT) has run in under 4 minutes unless there has a been a switch of the node version that has been used (presumably it has to resync some stuff. The runs on the other machine (test-ibm-rhel7-s390x-1) are almost all over 6 minutes in execution time.

image

sxa555:
`-4` with the alternate JVM in place is running the V8 jobs in about 15 minutes (full ccached run) compared to what looks like ~20 minutes on `-3`. I'll be honest I didn't expect it to make that much of a difference and I haven't looked into the timings inside the job in detail but that's quite a boost if it's purely down to the agent JVM and therefore I can't see any reason not to deploy this on all the machines.

image

sam-github:
LGTM - @sxa555 do you need anyone to land the PR and re-ansible the rhel-s390x machines, or can you do all that? If not, maybe @AshCripps can help?

sxa555:
I don't have access to anything at the moment, so Ash will have to do it :-P

AshCripps:
I need to reansible the rhel machines anyway so I will include this and do them tomorrow. @sam-github will need you to do the release machine however once the test machines are done.

mhdawson:
@sxa555 thanks for investigating this.

ansible/roles/java-base/tasks/main.yml
```diff
@@ -61,6 +61,26 @@
     dest: "/usr/bin/java"
     state: link
 
+- name: download java on s390x
+  when: os == "rhel7" and arch == "s390x"
+  get_url:
+    url: https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u242-b08_openj9-0.18.1/OpenJDK8U-jdk_s390x_linux_openj9_8u242b08_openj9-0.18.1.tar.gz
+    dest: /tmp/
+- name: unarchive java s390x
+  when: os == "rhel7" arch == "s390x"
```
AshCripps:
```suggestion
  when: os == "rhel7" and arch == "s390x"
```

AshCripps:
All machines reansibled and rejoined to jenkins.

@sam-github can you please reanisble the release machine when you get a chance?

sam-github:
Done. 

> java.vm.info=JRE 1.8.0 Linux s390x-64-Bit Compressed References 20200122_505 (JIT enabled, AOT enabled)>OpenJ9   - 51a5857d2>OMR      - 7a1b0239a>JCL      - 8cf8a30581 based on jdk8u242-b08

in https://ci-release.nodejs.org/computer/release-ibm-rhel7-s390x-1/systemInfo

PR #2154, opened 2020-01-30
coverage: remove copy of styles

mhdawson:
Remove copy of styles from script, added it instead to
the job that runs/generates the coverage output


mhdawson:
Going to land so that I can test that it fixes the job.  Will back out
if there are objections.

mhdawson:
CI to validate: https://ci.nodejs.org/view/Node.js%20Daily/job/node-test-commit-linux-coverage-daily/

mhdawson:
Ok, all good now.

PR #2156, opened 2020-02-04
doc: remove test-commit-matrix

sam-github:
@rvagg I've mixed feelings on this, I leave it to you.

On the one hand:

> This is assumed correct as of the date of last commit. If you notice a discrepancy with actual ci.nodejs.org runs, please update this document.

I don't think anyone is updating it, and having to manually sync state from ci to a markdown document is doomed to failure as a long-term plan, so that argues for deletion.

On the other hand, some of the info here, particularly the "Windows Test Matrix" section does give a high-level view of how/why the CI is organized the way it is, I can see if being useful even if the actual details are out of date.

This PR just deletes it, but I'd like your opinion:
1. Do nothing, close this PR and leave the file as-is?
2. Delete it?
3. Add a note to the top saying its not being updated, but contains useful information?
4. Delete the out-of-date parts and leave the free text?
5 ... ?


rvagg:
No strong feelings.

Weakly held opinion is to keep it in place for longer and then delete at some point in the future when it's objectively ancient and nobody is updating it. It's only 6 months old and the kind of thing that someone will do in a flurry of activity (I can imagine myself doing it, or being pulled in by someone else doing it).

But yeah, stale docs suck, we just don't have a good replacement for this even if it's not quite accurate.

PR #2161, opened 2020-02-04
ansible: avoid duplicate keys

richardlau:
I'm working on resurrecting https://github.com/nodejs/build/pull/1476 but as a GitHub Actions workflow. 

Current yamllint failures are mostly whitespace and I'll deal with those in a separate PR but this is the one non-whitespace issue found.

cc @AshCripps @sam-github

### DIFF
--- a/ansible/roles/jenkins-worker/tasks/main.yml
+++ b/ansible/roles/jenkins-worker/tasks/main.yml
@@ -77,14 +77,7 @@
   command: systemctl enable iptables
   when: "'rhel7-s390x' in inventory_hostname"
 
-- name: Firewall | check for firewalld
-  raw: stat /usr/sbin/firewalld
-  register: has_firewalld
-  failed_when: has_firewalld.rc > 1
-  when: "'rhel7-s390x' in inventory_hostname"
-
 - name: Firewall | remove firewalld
-  when: has_firewalld.rc == 0
   raw: yum remove -y firewalld
   when: "'rhel7-s390x' in inventory_hostname"
 
@@ -128,7 +121,7 @@
     content: "127.0.0.1 localhost\n::1 localhost\n"
     dest: /etc/ipnodes
 
-- name: Resolver | set DNS resolving priorities for zos 
+- name: Resolver | set DNS resolving priorities for zos
   when: os|startswith("zos")
   copy:
     content: "LOOKUP LOCAL DNS"


ansible/roles/jenkins-worker/tasks/main.yml
```diff
@@ -84,9 +84,8 @@
   when: "'rhel7-s390x' in inventory_hostname"
 
 - name: Firewall | remove firewalld
-  when: has_firewalld.rc == 0
   raw: yum remove -y firewalld
-  when: "'rhel7-s390x' in inventory_hostname"
+  when: "'rhel7-s390x' in inventory_hostname and has_firewalld.rc == 0"
```
sam-github: What is the behaviour for duplicate keys? If its "last key wins", then this isn't a refactor, its adding a condition. Did you run it  on a host to confirm it works?

richardlau: I don't know what the behaviour is for duplicate keys but the linting is flagging it. I've not run it. Is yum happy to remove something that isn't installed (`has_firewalld.rc` is set in the `Firewall | check for firewalld` check immediately above)?

sam-github:
```
% ssh test-ibm-rhel7-s390x-1 sudo yum remove -y firewalld && echo YES
Loaded plugins: product-id, search-disabled-repos, subscription-manager
This system is not registered with an entitlement server. You can use subscription-manager to register.
No Match for argument: firewalld
No Packages marked for removal
YES
```

richardlau: So do we need the `check for firewalld` at all then?

sam-github: It looks suspiciously unnecessary to me. If its last-key wins, the stat resiult was probably always being ignored. Hard to know without trying, though.

AshCripps: Not 100% sure but it seems that last key _does_ win https://github.com/ansible/ansible/pull/56933/files#diff-c495d684aecaea8f523caa00ab4ed915R74-R75

AshCripps:
@richardlau Have you managed to test this yet?

ansible/roles/jenkins-worker/tasks/main.yml
sam-github, line 88:
```suggestion
  when: "'rhel7-s390x' in inventory_hostname"
```

richardlau:
Updated to remove the `has_firewalld.rc` check and the task that set it. Have run the playbook against `test-ibm-rhel7-s390x-4`.

PR #2184, opened 2020-02-18
jenkins: use aix7.1 to release on 10.x

sam-github:


sam-github:
@richardlau @AshCripps @BethGriggs -- I think this is correct, from now on, we don't use 6.1 on 10 and greater -- which means nowhere. Even though we don't build anything less than 10 anymore, I left the clause in because the machines are still present. We use 7.1 for test+release on 10 and greater -- which means everywhere.

LGTY? This should affect the next 10.x release.

Issue #2190, opened 2020-02-19
suggestion: we enable the "stale" bot on nodejs/build


sam-github:
stale bot: https://github.com/apps/stale

Its used for a few other repos associated with Node.js (like libuv).

IMO, it would make a non-judgemental way to clean up stale issues. Note that we have 182 issues, and about 3/8th of them (judging by page counts) have not been updated since ***2018***. I think this is starting to snowball, there are so many, we inadvertently create duplicates, and we can't easily seperate the "this really needs doing" from the "wouldn't it be nice if ..." issues. Even going through all those issues would take time away from actually making progress on them.

Note that the stale bot won't close anything if anyone makes a comment, _any comment_. Its process is basically:

- Make a comment saying the issue is stale after a configurable number of days of inactivity.
- Close stale issues after a configurable number of days passes after the issue was marked stale.

Everyone gets a chance to disagree on staleness by posting, some labels can be designated as "never stale", and of course its github.... we can reopen issues if we ever need to.

bnb:
I would love to see this bot be able to be opt-in for repos across the organization. There are several repos that could benefit from it.

richardlau:
Another option is the GitHub Actions stale action: https://github.com/actions/stale

sam-github:
Since we are already using actions, I assume that would be lower barrier to entry.

@nodejs/build - any comments pro or con?

sam-github:
In case we do decide to try this: https://github.com/nodejs/build/pull/2196

trivikr:
Prior discussions with some consensus on stale action or stale bot but no follow-ups:
* Stale bot https://github.com/nodejs/node/issues/28798
* Stale action https://github.com/nodejs/node/issues/29232

Configs finalized for nodejs/node:
* `daysUntilStale`: 60
* `daysUntilClose`: 365

https://github.com/nodejs/node/issues/29232#issuecomment-523187880

Trott:
I'm +1 on at least trying this in the Build repo and seeing how it works for us.

mhdawson:
I'm +1.  In the worst case we can re-open issues and the barrier to keeping an issue open (commenting) is low.  Even though we may end up with closed issues for things we still want to do, if we have not gotten to then in a long time and we don't remember them without an issue then we probably have higher priorities anyway.

sam-github:
[...]

PR #2196, opened 2020-02-24
workflow: add stale action

sam-github:
Setup to mark stale after 90 days, and close 15 days after marked stale
unless the issue is commented on, or has the `stale` label removed.

Fixes: https://github.com/nodejs/build/issues/2190

-----

Draft - awaiting some comment on https://github.com/nodejs/build/issues/2190 to agree this is useful. I've never configured a GH action before, I'd appreciate a review by someone who has!

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,17 @@
+name: "Close stale issues"
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days'
+        stale-issue-label: 'stale'
+        exempt-issue-label: 'never stale'
+        days-before-stale: 300
+        days-before-close: 30


.github/workflows/stale.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days'
+        stale-issue-label: 'stale'
+        days-before-stale: 90
```
sam-github: Just a suggested value. With 90, every issue will get a nudge every 3 months, and if its still relevant whoever is interested just has to comment with a "I still plan to do this" or "current state is ..." (or anything else, or just remove the label), and it will get another 90 days of being ignored.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days'
+        stale-issue-label: 'stale'
```
sam-github: It wasn't clear whether this label has to be created, or the plugin will create it. I guess I'll find out if we decide to try this out.

rvagg:
I can't muster a strong opinion on this tbh, the open issues don't bother me so much on this repo since we're not so much building code (where open issues and prs bother me more) but managing infra (where status and information is more useful). If you think this will add value then I'm fine with working around it where it gets in my way, being able to label and/or bump is easy enough.

mhdawson:
No objections from those who were at the build WG meeting this week. We should move forward with it.

richardlau:
Do we want an [`exempt-issue-label`](https://github.com/actions/stale/blob/0d6a8e531449ecb97c9b6046b6b5d8eeac8b5115/action.yml#L21-L22)?

AshCripps:
> Do we want an [`exempt-issue-label`](https://github.com/actions/stale/blob/0d6a8e531449ecb97c9b6046b6b5d8eeac8b5115/action.yml#L21-L22)?

I think its a good idea as some issues can be open for a long time for a reason (waiting on people time to free up, etc)

PR #2202, opened 2020-03-04
 jenkins: osx10.15 release for 13.x 

rvagg:
This lands on top of #2199, after we get that one landed landed, https://github.com/nodejs/node/pull/31459 landed and a few nightlies out without any problems.

See timeline in https://github.com/nodejs/build/pull/2199#issuecomment-594253577


rvagg:
cleaned up the doc commit difference, it's just the versionselectorscript changes

PR #2203, opened 2020-03-04
jenkins: osx10.15 release for all versions 

rvagg:
This lands on top of #2202, after we get https://github.com/nodejs/node/pull/31459 backported to 12.x and 10.x which happens after we successfully get a 13.x release or two out.

See timeline in https://github.com/nodejs/build/pull/2199#issuecomment-594253577

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -101,11 +101,11 @@ def buildExclusions = [
   [ /sharedlibs_shared/,              anyType,     lt(9)   ],
 
   // OSX ---------------------------------------------------
-  [ /^osx1010/,                       anyType,     gte(11) ],
-  [ /^osx1011/,                       releaseType, lt(11)  ],
-  [ /^osx1011/,                       releaseType, gte(13) ],
+  [ /^osx1010/,                       testType,    gte(11) ],
+  [ /^osx1010/,                       releaseType, allVer  ],
   [ /^osx1011/,                       testType,    gte(14) ],
-  [ /^osx1015/,                       releaseType, lt(13)  ],
+  [ /^osx1011/,                       releaseType, allVer  ],
+  // osx1015 enabled for all, and builds all releases to support notarization
 
   // FreeBSD -----------------------------------------------
   [ /^freebsd10/,                     anyType,     gte(11) ],


rvagg:
Rebased, should be ready to go but would really appreciate sanity checking. What I intend:

* osx 10.10 is only used to test Node 10, no releases
* osx 10.11 is only used to test Node 10, 12, 13, no releases
* osx 10.15 tests all versions and builds releases for all versions

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -101,11 +101,11 @@ def buildExclusions = [
   [ /sharedlibs_shared/,              anyType,     lt(9)   ],
 
   // OSX ---------------------------------------------------
-  [ /^osx1010/,                       anyType,     gte(11) ],
-  [ /^osx1011/,                       releaseType, lt(11)  ],
-  [ /^osx1011/,                       releaseType, gte(13) ],
+  [ /^osx1010/,                       testType,    gte(11) ],
+  [ /^osx1010/,                       releaseType, noVer   ],
```
sam-github:
```suggestion
  [ /^osx1010/,                       releaseType, allVer   ],
```
For "all versions" don't release on 10.10

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -101,11 +101,11 @@ def buildExclusions = [
...
   [ /^osx1011/,                       testType,    gte(14) ],
-  [ /^osx1015/,                       releaseType, lt(13)  ],
+  [ /^osx1011/,                       releaseType, noVer   ],
```
sam-github:
```suggestion
  [ /^osx1011/,                       releaseType, allVer   ],
```

for "all versions", don't release on 10.11

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -101,11 +101,11 @@ def buildExclusions = [
...
+  [ /^osx1015/,                       releaseType, allVer  ],
```
sam-github:
```suggestion
  [ /^osx1015/,                       releaseType, noVer  ],
```

Exclude 10.15 for release build on "no versions" -- which is a no-op, it should just be removed.

richardlau: So I've been reminded that these are exclusions, so `noVer` and `allVer` should be the other way around?

rvagg:
Thanks fellas! This script is starting to feel more like a footgun. Inverted those `noVer` and `allVer`s, ptal.

richardlau:
> 
> 
> Thanks fellas! This script is starting to feel more like a footgun. 

Yeah, I think it's easier to think of inclusions (e.g. run tests on 14 and above) rather than exclusions (e.g. don't run tests on 13 and below) but the number of entries in the table makes me hesitant to flip the conditions. On the other hand making it inclusions would make retiring hosts easier as we would gradually shrink the inclusions for the host and then remove altogether. At the moment with exclusions we have to be very careful removing an entry from the file as no entry implies run everywhere.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -101,11 +101,11 @@ def buildExclusions = [
...
   // OSX ---------------------------------------------------
-  [ /^osx1010/,                       anyType,     gte(11) ],
-  [ /^osx1011/,                       releaseType, lt(11)  ],
-  [ /^osx1011/,                       releaseType, gte(13) ],
+  [ /^osx1010/,                       testType,    gte(11) ],
+  [ /^osx1010/,                       releaseType, allVer  ],
   [ /^osx1011/,                       testType,    gte(14) ],
-  [ /^osx1015/,                       releaseType, lt(13)  ],
+  [ /^osx1011/,                       releaseType, allVer  ],
+  [ /^osx1015/,                       releaseType, noVer   ],
```
sam-github:
```suggestion
```

Confusingly does nothing, I suggest deleting.

rvagg: Joao decided to use this pattern in the Windows section to be explicit about what's doing what. I figure that including these brings us closer to that explicit state that we might achieve by flipping this from exclude to include; `noVer` are essentially inline documentation so there's no confusion for future editors.
Do you actually find this confusing re osx, or just re how the script works?

sam-github: Confusion is relative, I find it more confusing than

> // Releases for all OS X versions are done on OS 10.15 with approriate target levels, to support notarization

Reading code that superficially might have an effect, but on consideration, does not tends to make me ask myself "did the author intend to write code that does nothing?", so there's that. If you find it a helpful reminder, though, go for it.

rvagg: No, that's fine, have replaced it with a comment. It's inconsistent anyway since it's `releaseType` and really it's `anyType noVer`.

MylesBorins:
Looks like this is good to go! We are planning an LTS release for 10.x and 12.x next week so would be good to get this landed ASAP so we can build an RC with the notary stuffs

PR #2218, opened 2020-03-11
ansible: add cmake for AIX

richardlau:
Install `cmake` on AIX from the [AIX Toolbox for Linux Applications](https://www.ibm.com/support/pages/aix-toolbox-linux-applications-downloads-alpha).

Refs: https://github.com/nodejs/build/issues/2097
Refs: https://github.com/libuv/libuv/pull/2731

cc @AshCripps @sam-github

### DIFF
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -47,7 +47,7 @@ packages: {
   ],
 
   aix: [
-    'bash,gcc-c++,gcc6-c++,tar,unzip,git,make,sudo',
+    'bash,cmake,gcc-c++,gcc6-c++,tar,unzip,git,make,sudo',
   ],
 
   debian7: [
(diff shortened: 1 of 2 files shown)


richardlau:
I've rerun ansible on `test-ibm-aix71-ppc64_be-1` and `test-ibm-aix71-ppc64_be-2`. `cmake` is installed in `/opt/freeware/bin`:

```
bash-5.0$ /opt/freeware/bin/cmake --version
cmake version 3.14.3

CMake suite maintained and supported by Kitware (kitware.com/cmake).
bash-5.0$
```

mhdawson:
If that works, then works for me.

richardlau:
`/opt/freeware/bin` isn't currently on the `PATH` for AIX. I'll add it as part of `select-compiler.sh` so we don't have to do it in every job that needs `cmake`.

PR #2226, opened 2020-03-17
jenkins: new workspace machines, ubuntu1804

rvagg:
2 new machines on packet.net for jenkins-workspace. These are running Ubuntu 18.04, the old ones are 16.04 and it's time to upgrade.

I haven't added the extra 1Tb disk onto these machines, they have a bit over 200G of disk to spare. I'm a bit hesitant here because the workspace machines get up over 200G in their ~iojs/build/workspace directories, _but_, as @mmarchini has discovered, they've also been running OOM errors which I'm pretty certain is due to their massive workspaces. I've cleaned the two active packet.net machines out and they're up and running fine again (I think). So my inclination with these new machines is to leave disk as is and we need to regularly do a cleanup or come up with some way to force Jenkins to do better cleanup. Manual isn't ideal because you need to take the node offline first so you're not interrupting existing work, and it takes _ages_ if you have ~200G of tiny files.

The SoftLayer jenkins-workspace machine is still running but is currently disabled in Jenkins. It's got some OS problems and needs to be removed. It's currently set as our primary binary_tmp git host for moving cross-compiled binaries around so I'm hesitant to touch it just yet.

So I think the plan is: watch these machines take some of the load for a day or two, remove the old packet.net workspace machines, watch these new ones take even more load, provision an 18.04 SoftLayer host and get it going, switch over binary_tmp to one of these new machines, make sure that works properly, remove old SoftLayer host.


ansible/roles/github-bot/templates/environment-file.j2
```diff
@@ -11,4 +11,4 @@ JENKINS_JOB_CITGM={{ envs.jenkins_job_citgm }}
 JENKINS_BUILD_TOKEN_CITGM={{ envs.jenkins_build_token_citgm }}
 JENKINS_JOB_NODE={{ envs.jenkins_job_node }}
 JENKINS_BUILD_TOKEN_NODE={{ envs.jenkins_build_token_node }}
-JENKINS_WORKER_IPS=147.75.70.237,147.75.73.189,169.60.150.88,51.15.200.62,51.15.218.201,147.75.74.174,147.75.193.230,163.172.186.154
+JENKINS_WORKER_IPS=147.75.70.237,147.75.73.189,169.60.150.88,51.15.200.62,51.15.218.201,147.75.74.174,147.75.193.230,163.172.186.154,147.75.66.203,147.75.81.67
```
richardlau: Does the bot need to be redeployed to pick up the new IP addresses?

rvagg: did it manually

rvagg:
have disabled test-packetnet-ubuntu1604-x64-1 and test-packetnet-ubuntu1604-x64-2, let's see how these two new ones take the load

rvagg:
A bit bumpy but we seem to be stable now.

I also cloned bare repos on both of these machines in /home/iojs/git/io.js.reference so checkouts should be quicker and take up less space, I've also symlinked /Users to /home because in some jobs it's set up to be /Users/iojs/git/io.js.reference because they were designed to run on macOS. This is optional so I'm not going to put anything in Ansible for it yet.

We'll need to keep a closer eye on space on these things, occasionally taking one offline and clearing out /home/iojs/build/workspace. Doing a `git fetch origin` inside the bare mirror directory is probably a good idea too to minimise the diff.

Issue #2230, opened 2020-03-17
`test-softlayer-debian9-x64-1` filesystem is read-only, leading to failing tests


mmarchini:
```
09:57:53  > git config remote.origin.url git@github.com:nodejs/node.git # timeout=10
09:57:53 ERROR: Error fetching remote repo 'origin'
09:57:53 hudson.plugins.git.GitException: Failed to fetch from git@github.com:nodejs/node.git
09:57:53 	at hudson.plugins.git.GitSCM.fetchFrom(GitSCM.java:909)
09:57:53 	at hudson.plugins.git.GitSCM.retrieveChanges(GitSCM.java:1131)
09:57:53 	at hudson.plugins.git.GitSCM.checkout(GitSCM.java:1167)
09:57:53 	at hudson.scm.SCM.checkout(SCM.java:505)
09:57:53 	at hudson.model.AbstractProject.checkout(AbstractProject.java:1205)
09:57:53 	at hudson.model.AbstractBuild$AbstractBuildExecution.defaultCheckout(AbstractBuild.java:574)
09:57:53 	at jenkins.scm.SCMCheckoutStrategy.checkout(SCMCheckoutStrategy.java:86)
09:57:53 	at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:499)
09:57:53 	at hudson.model.Run.execute(Run.java:1853)
09:57:53 	at hudson.matrix.MatrixRun.run(MatrixRun.java:153)
09:57:53 	at hudson.model.ResourceController.execute(ResourceController.java:97)
09:57:53 	at hudson.model.Executor.run(Executor.java:427)
09:57:53 Caused by: hudson.plugins.git.GitException: Command "git config remote.origin.url git@github.com:nodejs/node.git" returned status code 255:
09:57:53 stdout: 
09:57:53 stderr: error: could not lock config file .git/config: Read-only file system
09:57:53 
09:57:53 	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2430)
09:57:53 	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2360)
09:57:53 	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2356)
[...]

PR #2231, opened 2020-03-17
doc: add steps to fix read-only fs

mmarchini:
Fixes: https://github.com/nodejs/build/issues/2230

### DIFF
--- a/doc/jenkins-guide.md
+++ b/doc/jenkins-guide.md
@@ -210,11 +210,14 @@ If that happens, follow the steps below:
 touch foo                      # to confirm system is read-only, don't proceed if this succeeds
 sudo df                        # to determine the device for `/`
 sudo e2fsck -y /dev/mmcblk0p2  # replace mmcblk0p2 with proper device for `/`
-touch foo                      # this should succeed now
 ```
 
-If the second touch fails, follow instructions below to 
+After running the steps above, follow instructions below to 
 [restart the machine](#restart-the-machine).
+
+Note: the occurrence of read-only root filesystem indicates there's probably a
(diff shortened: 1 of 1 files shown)


doc/jenkins-guide.md
```diff
@@ -200,6 +201,21 @@ launchctl start org.nodejs.osx.jenkins
...
+### Read-only filesystem
+
+Sometimes a failure in the filesystem will cause it to enter a read-only state.
+If that happens, follow the steps below:
+
+```bash
+touch foo                      # to confirm system is read-only, don't proceed if this succeeds
+sudo df                        # to determine the device for `/`
+sudo e2fsck -y /dev/mmcblk0p2  # replace mmcblk0p2 with proper device for `/`
+touch foo                      # this should succeed now
```
rvagg: I'm not sure this is going to universally work, `mount -o remount,rw /` is what was traditionally required here, maybe ext4 improves the situation though?

doc/jenkins-guide.md
```diff
@@ -200,6 +201,21 @@ launchctl start org.nodejs.osx.jenkins
...
+```
+
+If the second touch fails, follow instructions below to
```
rvagg: I think a reboot after a ro root should be mandatory, it's chaos for state management and will likely leave you in an inconsistent place.

I'd go as far as to say that an ro root is suggestive of bigger problems that might need further investigation; maybe a reprovision of the machine.

mmarchini: Updated the instructions to reboot afterwards and to open an issue to investigate what caused the root fs to become read only.

PR #2234, opened 2020-03-18
jenkins: exclude new jenkins-workspace machines from removing `node`

rvagg:
forgot this second bit, it's needed so `node` doesn't get removed after it's installed, we need it on these workspace machines to run the linter

we need to find a better way of doing this than listing the machines, using the label might be good if we can work out a way to


rvagg:
messed up, had "softlayer" instead of "packetnet" so it didn't work! have pushed a fix

rvagg:
working now https://ci.nodejs.org/job/node-test-linter/33468/

PR #2255, opened 2020-03-31
doc: fix typo on workspace path

mmarchini:


rvagg:
some workspaces are in `ws`, not `workspace`, maybe it's just release jenkins that's configured this way, but this isn't a universal unfortunately!

mmarchini:
`rm -rf ~/build/*/node-test-commit*` would probably work but I don't feel comfortable putting that on the docs :)

rvagg:
yeah, and there's often other stuff inside ~/build/ but it's also usually deletable, a `rm -rf ~/build/*` is often OK as long as it's not running anything! Taking minimal steps is probably safest though.

PR #2261, opened 2020-04-03
docs: move remaining ansible docs to doc/

sam-github:
Some were already there (ansible-testing.md), move the rest, its easier
to remember that "all docs are doc/" (except for the top-level... but at
least the repository root contents are pretty prominently displayed).


sam-github:
@nodejs/build too boring to comment on, or to controversial? ;-) PTAL

richardlau:
I'm +0 on the change. I can see the logic in all docs under `doc/`. I'm also lazy and like the fact that the contents of the current `ansible/README.md` shows up under the file list when I go to https://github.com/nodejs/build/tree/master/ansible.

sam-github:
There is a certain "locality of reference" for the current location, but I keep having to search around to figure out where the manual steps are (hint: not in https://github.com/nodejs/build/blob/master/doc/non-ansible-configuration-notes.md), and `ls doc` would be an easier way to find them. Maybe its just because I have such a terrible memory.

mhdawson:
I think the locality of reference is such that there should be a README.md in ansible directory. If it does not have the info currently included it should at least have links to the relevant documents if they are somewhere else.

sam-github:
ok, I find having the docs not be in doc/ confusing, but apparently I'm the only one. I'll abandon this.

Issue #2242, opened 2020-03-22
Use devtoolset-8 on Centos/RHEL for Node.js 14+


richardlau:
Requires access to the release machines. Install devtoolset-8 on the Centos/RHEL machines and use it to build releases for Node.js 14+.

Refs: https://github.com/nodejs/build/issues/2168

## Install devtoolset-8 on centos7:

- [x] release-osuosl-centos7-ppc64_le-1
- [x] test-osuosl-centos7-ppc64_le-1 
- [x] test-osuosl-centos7-ppc64_le-2 
- [x] release-digitalocean-centos7-x64-1 : 
- [x] release-packetnet-centos7-arm64-1: 
- [x] test-packetnet-centos7-arm64-1 
- [x] test-packetnet-centos7-arm64-2 
- [x] test-rackspace-centos7-x64-1 
- [x] test-softlayer-centos7-x64-1:

## Install on rhel7

- [x] release-ibm-rhel7-s390x-1 
- [x] test-ibm-rhel7-s390x-1 
- [x] test-ibm-rhel7-s390x-2 
- [x] test-ibm-rhel7-s390x-3 
- [x] test-ibm-rhel7-s390x-4

## Use devtoolset-8 for 14.x: PR #2262

select-compiler.sh:
- [x] rhel7-s390x
- [x]  centos7-ppc64_le 
- [x]  centos7-64
- [x] centos7-arm64

VersionSelectorScript.goovy
- n/a rhel7-s390x
- n/a centos7-ppc64_le: ubuntu1404 already excluded for gte 14.x
- [x]  centos7-64
- [x] centos7-arm64

Jenkins config changes:
- (n/a) centos7-ppc64_le
- (n/a) centos7-s390x
- [x]  centos7-64:
   - [x] ci: machines labelled
   - [x] ci: config selected for job
   - [x] ci-release: machines labelled
   - [x] ci-release: config selected for job
- [x] centos7-arm64
   - [x] ci: machines labelled
   - [x] ci: config selected for job
   - [x] ci-release: machines labelled
   - [x] ci-release: config selected for job

test builds:
- node-test-commit-linux: (x64)
  - [x] 12.x: https://ci.nodejs.org/job/node-test-commit-linux/34154/
  - [x] master: https://ci.nodejs.org/job/node-test-commit-linux/34153/
- node-test-commit-arm: (arm64)
[...]

PR #2262, opened 2020-04-03
ansible: install devtoolset-8 on centos7

sam-github:
Refs: https://github.com/nodejs/build/issues/2242

### DIFF
--- a/ansible/roles/baselayout/tasks/partials/repo/centos7.yml
+++ b/ansible/roles/baselayout/tasks/partials/repo/centos7.yml
@@ -12,19 +12,11 @@
     state: present
 
 
-- name: centos7 | install scl for devtoolset-6
-  when: "arch != 'arm64' and arch != 'ppc64'"
+- name: centos7 | install scl for devtoolset-6, devtoolset-8
+  when: "arch != 'arm64'"
   yum:
     name: centos-release-scl
     state: present
-
-- name: centos7 | ppc64 | install sclo7 devtoolset-6
-  when: "arch == 'ppc64'"
-  yum_repository:
-    name: sclo7-devtoolset-6
-    description: cbs.centos.org sclo7-devtoolset-6-rh-release
-    baseurl: https://cbs.centos.org/repos/sclo7-devtoolset-6-rh-release/$basearch/os/
-    gpgcheck: no
 
 - name: centos7 | aarch64 | install sclo repo
   when: "arch == 'arm64'"
@@ -40,7 +32,8 @@
     dest: "/etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-SIG-SCLo"
     mode: 0644
 
-- stat: path=/usr/local/bin/python3.7
+- stat:
+    path: /usr/local/bin/python3.7
   register: build_python37
 
 - name: centos7 | ppc64 | download python 3.7
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -43,7 +43,7 @@ packages: {
   centos7_x64: ['git2u','centos-release-scl',], # centos-release-scl is required to enable SCLo
                                                # but we do it manually in partials/repo/centos7.yml for arm64
   centos7: [
-    'bzip2-devel,openssl-devel,ccache,gcc-c++,devtoolset-6,sudo,git,devtoolset-6-libatomic-devel,zlib-devel,libffi-devel',
+    'bzip2-devel,openssl-devel,ccache,gcc-c++,devtoolset-6,sudo,git,devtoolset-6-libatomic-devel,zlib-devel,libffi-devel,devtoolset-8,devtoolset-8-libatomic-devel',
   ],
 
   aix: [
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -42,11 +42,11 @@ packages: {
   # centos-release-scl is required to enable SCLo but we do it manually in
   # partials/repo/centos7.yml for arm64
   centos7_arm64: ['git'], # git2u not available for aarch64 (yet)
-  centos7_x64: ['git2u,centos-release-scl'],
-  centos7_ppc64: ['git'],
+  centos7_x64: ['devtoolset-6-libatomic-devel,git2u,centos-release-scl'],
+  centos7_ppc64: ['devtoolset-6-libatomic-devel,git'],
 
   centos7: [
-    'bzip2-devel,openssl-devel,ccache,gcc-c++,devtoolset-6,sudo,devtoolset-6-libatomic-devel,zlib-devel,libffi-devel,devtoolset-8,devtoolset-8-libatomic-devel',
+    'bzip2-devel,openssl-devel,ccache,gcc-c++,devtoolset-6,sudo,zlib-devel,libffi-devel,devtoolset-8,devtoolset-8-libatomic-devel',
   ],
 
   aix: [
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -39,11 +39,14 @@ packages: {
     'ccache,git2u,gcc-c++,devtoolset-6,sudo', # even need gcc on centos6 so ccache has symlinks
   ],
 
+  # centos-release-scl is required to enable SCLo but we do it manually in
+  # partials/repo/centos7.yml for arm64
   centos7_arm64: ['git'], # git2u not available for aarch64 (yet)
-  centos7_x64: ['git2u','centos-release-scl',], # centos-release-scl is required to enable SCLo
-                                               # but we do it manually in partials/repo/centos7.yml for arm64
+  centos7_x64: ['git2u,centos-release-scl'],
+  centos7_ppc64: ['git'],
+
   centos7: [
-    'bzip2-devel,openssl-devel,ccache,gcc-c++,devtoolset-6,sudo,git,devtoolset-6-libatomic-devel,zlib-devel,libffi-devel,devtoolset-8,devtoolset-8-libatomic-devel',
+    'bzip2-devel,openssl-devel,ccache,gcc-c++,devtoolset-6,sudo,devtoolset-6-libatomic-devel,zlib-devel,libffi-devel,devtoolset-8,devtoolset-8-libatomic-devel',
   ],
 
   aix: [
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -21,9 +21,11 @@ def buildExclusions = [
   // Linux -------------------------------------------------
   [ /^centos6/,                       releaseType, lt(8)   ],
   [ /^centos6/,                       anyType,     gte(12) ],
-  [ /centos[67]-(arm)?(64|32)-gcc48/, anyType,     gte(10) ],
-  [ /centos[67]-(arm)?(64|32)-gcc6/,  anyType,     lt(10)  ],
-  [ /centos6-32-gcc6/,                releaseType, gte(10) ], // 32-bit linux for <10 only
+  [ /^centos[67]-(arm)?(64|32)-gcc48/,anyType,     gte(10) ],
+  [ /^centos[67]-(arm)?(64|32)-gcc6/, anyType,     lt(10)  ],
+  [ /^centos[67]-(arm)?(64|32)-gcc6/, anyType,     gte(14) ], // 14.x: gcc6 builds stop
+  [ /^centos7-(arm)?(64)-gcc8/,       anyType,     lt(14)  ], // 14.x: gcc8 builds start
+  [ /^centos6-32-gcc6/,               releaseType, gte(10) ], // 32-bit linux for <10 only
   [ /^centos7-64/,                    releaseType, lt(12)  ],
   [ /debian8-x86/,                    anyType,     gte(10) ], // 32-bit linux for <10 only
   [ /debian8/,                        anyType,     gte(13) ],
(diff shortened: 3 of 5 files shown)


ansible/roles/baselayout/tasks/partials/repo/centos7.yml
```diff
@@ -12,20 +12,12 @@
...
 
-- name: centos7 | install scl for devtoolset-6
-  when: "arch != 'arm64' and arch != 'ppc64'"
+- name: centos7 | install scl for devtoolset-6, devtoolset-8
+  when: "arch != 'arm64'"
   yum:
     name: centos-release-scl
     state: present
 
-- name: centos7 | ppc64 | install sclo7 devtoolset-6
```
sam-github, line 21: In case you wonder -- this doesn't work anymore, the repo is 404, but the above works now for ppc64.

sam-github:
I wish I could move this back to draft... I'm finding its a bit WIP as I work through the various rhel/centos systems and need to patch the ansible.

jenkins/scripts/select-compiler.sh
```diff
@@ -143,6 +154,10 @@ elif [ "$SELECT_ARCH" = "X64" ]; then
   echo "Setting compiler for Node version $NODEJS_MAJOR_VERSION on x64"
 
   case $nodes in
+    centos7-x64 )
```
sam-github, line 157: This and the arm64 below aren't correct. It appears that https://ci.nodejs.org/job/node-test-commit-linux/ works differently from the other platforms above, which base the compiler decision on machine name (test/release-blah-blah_-x) + the nodejs version.

These ones take a different approach, its based on configuration name (`$nodes`), so a new configuration will need to be set for devtools-8 (probably centos7-gcc8). I'll have to work out how that relates to nodejs release version, if at all.

If anyone has some historic context and understands how this should work for devtools-8, feel free to chime in, otherwise I'll wing it.

rvagg: Historical context is probably that the ones using `$NODE_NAME` are the original form of this script which was limited to just the IBM platforms, so selection was a fairly straightforward affair and it ignored use of VersionSelectorScript, hence the liberal use of version selecting in here. The ones using `$nodes` are when we expanded it to more platforms and use labels instead to get broad selection and match how we're selecting nodes in our jobs and it uses VersionSelectorScript.

The use of labels lets us repurpose the same machine for different things and combine this script with VersionSelectorScript so we can switch its roles. Ideally `$NODEJS_MAJOR_VERSION` and anything like it should be removed from this script so version switching was isolated only to VersionSelectorScript, having it all in a single place (IMO!).

How you do this: assign the same machine(s) multiple labels indicating different compiler configurations (as in here where you have a `-gcc48` postfix. The job gets all those labels, then VersionSelectorScript tells the job which ones to select. By the time you get to this script, you have all the info you need for selecting the compiler contained within the label so you should just need to branch on that and not do anything with the Node.js version.

sam-github: OK, -gcc6 means devtools-6, I should keep the convention, use -gcc8 for devtools-8?

And for example, the VersionSelectorScript needs to change for 14.x to exclude centos7-x64-gcc6 for release (and test?), and allow centos7-x64-gcc8 for release (and test)?

rvagg: yeah, you should just be able to extend the centos block in VersionSelectorScript: https://github.com/nodejs/build/blob/050fb9733697acb3dd5bb53891c2818e2c335bab/jenkins/scripts/VersionSelectorScript.groovy#L24-L27
excluding centos6 from the optionality (`[67]`).

then in the centos7 machines, add a new label to each of them: https://ci.nodejs.org/label/centos7-64-gcc6/

then in linux-test-commit you can add that new label to the list:

Screenshot 2020-04-07 15 54 22

There may also be some stuff inline in node-test-commit-linux that matches labels that will need tweaking, better check the script invocations in there.

If you want to tinker with this, copy node-test-commit-linux to a temporary one, point the select-compiler.sh run command to point to your fork/branch and do the same for VersionSelectorScript so it's pulling from a copy you're editing. Be aware that there's some caching that can get in the way when you're editing VersionSelectorScript live so you may need to add a query string on the end and change it each time you edit (`?1` etc.).

rvagg: and re gcc6 and gcc8 vs devtools-6 and devtools-8, it's not critical but probably better to use gcc version because this same scheme can apply outside of centos

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -21,9 +21,11 @@ def buildExclusions = [
   // Linux -------------------------------------------------
   [ /^centos6/,                       releaseType, lt(8)   ],
   [ /^centos6/,                       anyType,     gte(12) ],
-  [ /centos[67]-(arm)?(64|32)-gcc48/, anyType,     gte(10) ],
-  [ /centos[67]-(arm)?(64|32)-gcc6/,  anyType,     lt(10)  ],
-  [ /centos6-32-gcc6/,                releaseType, gte(10) ], // 32-bit linux for <10 only
+  [ /^centos[67]-(arm)?(64|32)-gcc48/,anyType,     gte(10) ],
+  [ /^centos[67]-(arm)?(64|32)-gcc6/, anyType,     lt(10)  ],
```
sam-github, line 25: Just added the `^` for consistency.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -21,9 +21,11 @@ def buildExclusions = [
...
+  [ /^centos[67]-(arm)?(64|32)-gcc6/, anyType,     gte(14) ], // 14.x: gcc6 builds stop
+  [ /^centos7-(arm)?(64)-gcc8/,       anyType,     lt(14)  ], // 14.x: gcc8 builds start
```
sam-github, line 27: Here we are stopping using one set of labels after 14.x, and switching to a new set. I think its intentional to do  anyType, its not intended that we test with gcc6/aka devtoolset-6, is it?

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -48,6 +51,9 @@ def buildExclusions = [
   [ /^cross-compiler-armv[67]-gcc-4.9/, anyType,   lt(10)  ],
   [ /^cross-compiler-armv[67]-gcc-4.9/, anyType,   gte(12) ],
   [ /^cross-compiler-armv[67]-gcc-6/, anyType,     lt(12)  ],
+  [ /^cross-compiler-armv[67]-gcc-6/, anyType,     gte(14) ],
+  [ /^cross-compiler-armv[67]-gcc-8/, anyType,     lt(14)  ],
+  [ /^ubuntu1604-arm64/,              anyType,     gte(14) ],
```
sam-github, line 56: Unrelated to devtools8, but related to 14.x requirements, see comment https://github.com/nodejs/build/issues/2242#issuecomment-611190279

I'm not sure if we are comfortable with this, but there is a case to be made that it should be excluded for gte(12.x), cc: @nodejs/releasers , because it went EOL on the 2019-april-25th, https://ubuntustudio.org/2019/04/ubuntu-studio-16-04-lts-reaches-end-of-life-eol/, 2 days after `(tag: v12.0.0) 2019-04-23, Version 12.0.0 (Current)`, so we should have excluded it from test then.

sam-github:
@nodejs/build @nodejs/releasers  Please take a careful look at this, I think I got it all right, see checklist in https://github.com/nodejs/build/issues/2242. Some of the builds were kicked off recently as repeats of last night to make sure I build with default configs selected, so are ongoing, but went well last night.

I did builds on my job-forks for 12.x (to make sure I didn't break existing releases) and master (for 14.x).

I'm trying to be careful, and the changes aren't huge, but there is a non-zero chance of breaking things. Once I merge, I'll kick off builds for multiple branches in ci and i-release (obviously not real release builds, just test runs).

releasers: Is there any more releases planned for the next few days that this could impact?

build-wg and anyone: this reflects the 14.x plan? PTAL

ansible/roles/baselayout/vars/main.yml
```diff
@@ -40,10 +40,12 @@ packages: {
   ],
 
   centos7_arm64: ['git'], # git2u not available for aarch64 (yet)
-  centos7_x64: ['git2u','centos-release-scl',], # centos-release-scl is required to enable SCLo
+  centos7_x64: ['devtoolset-6-libatomic-devel', 'git2u','centos-release-scl',], # centos-release-scl is required to enable SCLo
```
rvagg, line 43: Not critical, just FYI: I _think_ it might be more efficient to run these as comma-separated strings rather than an array of strings. When it's a CSV I think it does them all at once, when they're separate array elements it does them one after the other. I can't be sure but my impression from flattening some of these in the past is that you get a speed-up from batching.

jenkins/scripts/VersionSelectorScript.groovy
rvagg: ack
Yes let's just leave it as a clean break even for testing. We can expand later if we see a need but we had had a problem with exploding permutation complexity _just because we can_ in the past, and let's try to resist that urge.

jenkins/scripts/VersionSelectorScript.groovy
rvagg: ok, but _are you sure_ there are no labels used anywhere with some cruft in front? I don't think there but we've been caught with overly strict regexes in the past (in select-compler.sh trying to identify all the benchmark machines iirc was the latest).

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -48,6 +51,9 @@ def buildExclusions = [
   [ /^cross-compiler-armv[67]-gcc-4.9/, anyType,   lt(10)  ],
   [ /^cross-compiler-armv[67]-gcc-4.9/, anyType,   gte(12) ],
   [ /^cross-compiler-armv[67]-gcc-6/, anyType,     lt(12)  ],
+  [ /^cross-compiler-armv[67]-gcc-6/, anyType,     gte(14) ],
+  [ /^cross-compiler-armv[67]-gcc-8/, anyType,     lt(14)  ],
```
rvagg, line 55: oh this is interesting, do we have that running? I see in https://github.com/nodejs/build/blob/master/ansible/roles/cross-compiler/files/cc-selector.sh I set it up to possibly work and I did do a toolchain for gcc-8 but don't recall ever getting it properly running.

This line should drop the `[67]` and just go with `7`. I think we probably should have done that with the gcc-6 one too since we dropped armv6 for Node 12 but there may be unintended consequences with doing that.

jenkins/scripts/VersionSelectorScript.groovy
rvagg: ack 👍

jenkins/scripts/VersionSelectorScript.groovy
rvagg: I've added the `cross-compiler-armv7-gcc-8` to the cross compilers in [ci-release](https://ci-release.nodejs.org/label/cross-compiler-armv7-gcc-8/) and [ci](https://ci.nodejs.org/label/cross-compiler-armv7-gcc-8/) but they'll need to be added to iojs+release and whatever we do for node-test-binary-arm. I'll be interested to see if they actually work and I'll also need to revisit the configs I set up to make the crosstools! Probably even needs a re-do with a newer gcc-8: https://github.com/rvagg/rpi-newer-crosstools/blob/master/x64-gcc-8.3.0.config

Or maybe we should pull this bit out of this PR and handle it separately?

sam-github: I'll pull it

sam-github: PTAL

ansible/roles/baselayout/vars/main.yml
sam-github: I'll consider that as a followup refactor, or next time I touch this, but I don't want to reansible all the machines now, which I'd feel duty bound to do if I made such a change.

sam-github: actually, I misread, its only x64. changed it.

jenkins/scripts/VersionSelectorScript.groovy
mhdawson: I assume we still have some linux platforms building/testing on gcc6  thought right ?

sam-github: For 14.x, specifically? I don't know. For earlier releases, yes.

If you want I can take can try to find the compiler levels for all machines 14.x are tested on, but I spent 10 minutes looking, and its not documented anywhere AFAICT. Figuring it out would take a combination of finding all linux builds in https://ci.nodejs.org/view/Node.js%20Daily/job/node-daily-master/1899/, reading the console logs, then sshing into the machines (if they don't explicitly set the version, many don't, I have to look... for example ubuntu 18.04 is using gcc7, I just checked). If https://github.com/nodejs/node/pull/32715#issuecomment-611595712 lands, then the ssh won't be needed, just reading the top of all the console logs.

AIX builds on gcc6.

richardlau: From the CI for https://github.com/nodejs/node/pull/32715 (https://ci.nodejs.org/job/node-test-pull-request/30565/) the following Linux jobs are using gcc6:

- [debian9-64](https://ci.nodejs.org/job/node-test-commit-linux/34121/nodes=debian9-64/consoleText) is using 6.3.0
- [ubuntu1604-64](https://ci.nodejs.org/job/node-test-commit-linux/34121/nodes=ubuntu1604-64/consoleText) is using 6.5.0

sam-github:
Landed in a6dd558...a042232

Issue #2264, opened 2020-04-06
retrospective on download outage


Trott:
For picking up once the current issue is addressed and behind us, a retrospective to figure out how to make it better next time would be a good thing, I think.

Refs: https://github.com/nodejs/node/issues/32683
Refs: https://github.com/nodejs/node/issues/32683

MattIPv4:
An official status page for NodeJS would likely be helpful in situations like this for a couple of reasons:

- A canonical place for NodeJS folks to communicate & acknowledge the issue
- Make the community aware, reducing the number of issues created in the org & beyond

Created #2265 to look in more detail at the process of creating a status page.

Trott:
No one has been added to @nodejs/build-infra in literally years. 5 years or something like that? We need to break that choke-point. It's OK if there are only 4 people in the group, but it's got to be 4 highly-active people. Right now, my guess is that we have 1 highly-active person, 2 moderately-active people, and 1 effectively-inactive person.

I would like us to think about who we might trust with this level of access. I'm surprised we haven't added @sam-github to the list, to be honest.

Trott:
This, of course, also raises the issue of if we are taking a risky approach by relying entirely on volunteers. Assuming the Foundation would be willing to pay to have someone on staff doing the most critical work, should we consider that? (We'll have to carefully define the most critical work.)

Trott:
On-call schedule? No, wait, hear me out! The idea is that we may not trust (for example) me with the keys to the kingdom, but maybe we trust me with the cell phone numbers of the people who do have those keys. And we trust me to not abuse this and to only notify people in a real crisis. Maybe we can have someone on-call to figure out who is in a not-terrible time zone and try to get them on the phone if we have an outage like this one.

I'm not sold on this being a great idea, but I'm trying to get all the ideas out anyway.

AshCripps:
I think the crux of the issue with build infra is coverage, we have 1 US, 2 Europe and 1 AUS which gives us decent coverage of time zones - but the European members are often harder to contact. Adding @sam-github would be a good start but still leaves a big void in the middle of the timezones. Obviously this is tricky due to the security concerns around the level of access this role gives.

MattIPv4:
Having a proper system for an emergency contact makes sense, either through an on-call system of trusted volunteers (build-infra) or a paid position that comes with the responsibility of being okay with being woken up in the middle of the night if the world falls apart.

sam-github:
[...]

PR #2268, opened 2020-04-07
www: less agressive nightly promotion & CDN cache purge

rvagg:
Ref: #2264

Just slowing things down a bit. Not fixing anything specific but being more cautious because a hypothetical potential for cache invalidation _every minute_. I can't see a means by which this could be caused, but slowing down the checks in cron will help. Will comment about implications of each thing inline.

This is active on nodejs.org right now but can be easily changed if there's feedback here.

### DIFF
--- a/ansible/www-standalone/tasks/site-setup.yaml
+++ b/ansible/www-standalone/tasks/site-setup.yaml
@@ -110,11 +110,9 @@
     dest: /etc/crontab
     line: "{{ item }}"
   with_items:
-    - '*/5 *   * * *   dist    /home/staging/tools/promote/promote_nightly.sh nodejs'
-    - '*/5 *   * * *   dist    /home/staging/tools/promote/promote_nightly.sh iojs'
-    - '*  *    * * *   nodejs  /home/nodejs/check-build-site.sh nodejs'
-    - '*  *    * * *   nodejs  /home/nodejs/check-build-site.sh iojs'
-    - '*  *    * * *   root    /home/nodejs/cdn-purge.sh'
+    - '*/30 *    * * *   dist    /home/staging/tools/promote/promote_nightly.sh nodejs'
+    - '*/5  *    * * *   nodejs  /home/nodejs/check-build-site.sh nodejs'
+    - '*/5  *    * * *   root    /home/nodejs/cdn-purge.sh'
     - '* */4   * * *   nodejs  /home/nodejs/sync-benchmarking.sh'
   tags: setup
 


ansible/www-standalone/tasks/site-setup.yaml
```diff
@@ -110,11 +110,9 @@
     dest: /etc/crontab
     line: "{{ item }}"
   with_items:
-    - '*/5 *   * * *   dist    /home/staging/tools/promote/promote_nightly.sh nodejs'
-    - '*/5 *   * * *   dist    /home/staging/tools/promote/promote_nightly.sh iojs'
-    - '*  *    * * *   nodejs  /home/nodejs/check-build-site.sh nodejs'
-    - '*  *    * * *   nodejs  /home/nodejs/check-build-site.sh iojs'
-    - '*  *    * * *   root    /home/nodejs/cdn-purge.sh'
+    - '*/30 *    * * *   dist    /home/staging/tools/promote/promote_nightly.sh nodejs'
```
rvagg, line 113: This checks for auto-promotable assets and moves them from staging to dist. This is everything but releases that get signed, so nightlies, v8-canary, RC, test, and others. Release machines upload resources to staging and every 5 minutes this script will look for which ones can be promoted and will do so. If at least one thing was promoted, it will queue a request for a cache purge (it doesn't perform the purge itself). So it's theoretically possible to purge CDN cache every 5 minutes if we have a new resource to promote every 5 minutes.

Slowing to every 30 minutes. Which means we're more likely to promote most of any of these drops in one or two goes. But it slows down the time between the drops and when they might show up. This might get frustrating for someone who's doing an RC, for example, watching their builds complete but not show up on nodejs.org so they can announce them. So there's a balance here.

ansible/www-standalone/tasks/site-setup.yaml
```diff
@@ -110,11 +110,9 @@
...
+    - '*/5  *    * * *   nodejs  /home/nodejs/check-build-site.sh nodejs'
```
rvagg, line 114: This checks whether index.json is newer than the website's index.html, and if it is, triggers a full site rebuild which in turn triggers a cache purge. This is done so that the website has the latest release on its front page without that having to be triggered by nodejs/nodejs.org.

ansible/www-standalone/tasks/site-setup.yaml
```diff
@@ -110,11 +110,9 @@
...
+    - '*/5  *    * * *   root    /home/nodejs/cdn-purge.sh'
```
rvagg, line 115: The actual cache purge script, it looks at its "queue" (just a temp file that any one of the above processes might create) and if there's a purge queued then it does a `curl` to CloudFlare to request a full purge. By reducing from every 1 minute to every 5 minutes we're reducing the potential for no caching at the edges.

rvagg:
We only check if a purge is requested that frequently. Purges get queued by various processes, either website rebuilds or /download/ resource promotion. But since we have nightlies and v8-canary builds that run every day, it ends up being at least one purge per day.

Issue #2249, opened 2020-03-25
Fedora-latest-x64 failures for 13.x releasee


MylesBorins:
https://ci.nodejs.org/job/node-test-commit-linux/33730/nodes=fedora-latest-x64/console
https://ci.nodejs.org/job/node-test-commit-linux/33721/nodes=fedora-latest-x64/

Two examplese of the same failure

rvagg:
can't tell exactly what this is, could be oom for compiler maybe? they're only 2G of memory but they're certainly not alone in that. I've done an update and full cleanup on both machines and restarted them, 🤞

PR #2269, opened 2020-04-07
www: log cdn purge calls

rvagg:
Ref: #2264

We don't have logs of our calls to Cloudflare's cache purge API calls we make when we promote new /download/ assets or publish new website content, so this is adding such logging. Along the way, I think I've found one possible "bug" that may provide a partial explanation to our outage problems, but I'll explain that in #2249 

Additional comments inline.

### DIFF
--- a/ansible/www-standalone/resources/scripts/build-site.sh
+++ b/ansible/www-standalone/resources/scripts/build-site.sh
@@ -44,6 +44,7 @@ docker run \
   -v /home/nodejs/.npm:/npm/ \
   node:latest \
   bash -c " \
+    apt-get update && apt-get install -y rsync && \
     addgroup nodejs --gid ${nodeuid} && \
     adduser nodejs --uid ${nodeuid} --gid ${nodegid} --gecos nodejs --disabled-password && \
     su nodejs -c ' \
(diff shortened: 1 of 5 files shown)


ansible/www-standalone/resources/scripts/build-site.sh
```diff
@@ -44,6 +44,7 @@ docker run \
   -v /home/nodejs/.npm:/npm/ \
   node:latest \
   bash -c " \
+    apt-get update && apt-get install -y rsync && \
```
rvagg, line 47: this is a holdover from some website publish debugging we were doing a few months ago, changed on server but not in ansible, I'm unsure if this is necessary but it's there for now!

ansible/www-standalone/resources/scripts/cdn-purge.sh.j2
```diff
@@ -27,6 +27,8 @@ if ! [ -f /tmp/cdnpurge.$site ]; then
   exit 0
 fi
 
+reason="$(cat /tmp/cdnpurge.$site | tr '\n' ' ')"
```
rvagg, line 30: Previously this file was a zero-length signal file, now it should contain a "reason", but it may also contain multiple reasons if more than one request was made before we check it, hence the flattening of multiple lines into one. We use this in the log, below.

PR #2275, opened 2020-04-08
Use system python3 on centos7-ppc64

sam-github:
It install python 3.6.8.

Ansible's previously built python3 wasn't capable of running node's
./configure, it lacked sufficient capabilities (like bzip2).


richardlau:
So was the `python3` not previously available on centos7-ppc64? Wondering why we were building Python 3.7 from source in the first place.

sam-github:
Its difficult to go back in time to see the pkg repos of the time. Either it wasn't available (most likely), or we tried to bump from 3.6 to 3.7.

PR #2288, opened 2020-04-14
docs: Update DigitalOcean logo

MattIPv4:
Noticed the DO logo in the readme was our rather old one from before our rebrand.

This updates the EPS file to be the one provided on the DO press age download and updates the tier 1 providers image with the new logo as well.


MattIPv4:
Is this technically docs (assets for docs)? Should I update commit name to indicate docs?

rvagg:
whoops, that was quite a while ago wasn't it? thanks for the fix

Issue #2265, opened 2020-04-06
Status page for NodeJS


MattIPv4:
## Latest status for the new status page

The experimental and very W.I.P status page is now available at https://status.nodejs.org/

We are currently working on establishing the policy for what should be posted on the status page and who should have access to it.

### Open tasks & questions

- [What is an Incident?](https://github.com/nodejs/build/issues/2303): #2303
- [Who should have access?](https://github.com/nodejs/build/issues/2304): #2304

Any discussion not related to these specific topics should be kept in this issue (unless a dedicated issue is then created for said topic).

### Draft PRs

- [build: Docs & customisation backup](https://github.com/nodejs/build/pull/2299): #2299
- [website: Adding embed & link to status page](https://github.com/nodejs/nodejs.org/pull/3126): nodejs/nodejs.org#3126

These PRs should not be landed until the status page is ready to go completely.

---

## Original issue

Hey folks,

**The current crisis ([nodejs.org intermittently returning 500s](https://github.com/nodejs/build/issues/2264)) has prompted a thought that perhaps having a proper status page for NodeJS would be a great canonical location to be able to communicate with the community that the NodeJS folks are aware of issues.**

**This would help with reducing the number of issues being created for the incident and would reassure the wider NodeJS community that folks here are looking into any issues.**

I'm not proposing any automation here (unless someone wishes to add metrics) -- automation of status pages is often frowned upon, as it then isn't that helpful, only showing what folks already know, instead of acting as an acknowledgement that humans on the project know something is wrong.

As NodeJS is an established open source project, it should be possible to get an open-source license from Atlassian to use StatusPage: https://www.atlassian.com/software/views/open-source-license-request -- I have previously done this for cdnjs (status.cdnjs.com) and it works perfectly.

---

**As a POC, assuming blessing from NodeJS folks, I would be happy to reach out to Atlassian and start the licensing process. Once acquired, I'm quite happy to stand up a POC status page on status.node.js.org (js.org, not the node-owned domain) and build out the page (styling, components, etc.).**

Once that's done, the page can then be reviewed and if happy with it, it can be moved over to status.nodejs.org (all that's required is a CNAME to the statuspage site), and folks that need access can be invited.

I think it'd be best to invite as many NodeJS folks as possible, so that during an incident hopefully someone will be around with access to post updates to the status page.

daniel-white:
It would be great if there was twitter account that one could follow the status updates akin to githubstatus.

MattIPv4:
[...]

PR #2299, opened 2020-04-20
StatusPage

MattIPv4:
This PR adds all the customisations used on the new status page to the repo as a backup. These should be updated whenever changes get made to the status page itself.

### DIFF
--- a/doc/statuspage.md
+++ b/doc/statuspage.md
@@ -1,0 +1,58 @@
+# StatusPage
+
+This document relates to the [StatusPage](https://www.statuspage.io/) for Node.js, which can be
+found at [status.nodejs.org](https://status.nodejs.org).
+
+## Access
+
+The following people have access to the StatusPage control panel. New accounts can be added via the
+[User Management](https://admin.atlassian.com/s/f6b1a2fe-1302-40aa-9a3f-4f7523370639/users) page.
+
+Atlassian StatusPage has three groups available, `site-admins` that have full administration access
+in Atlassian, `statuspage-administrators` that have full administration access specifically for the
+StatusPage product, and `statuspage-users` who have regular operational access to StatusPage.
+
+Atlassian organizations also have a separate concept of org-wide administrators, that have full
+access to everything (this is technically different from `site-admins` as that is scoped to the
+"site", but this org only has one site, so they are effectively the same).
+
+| GitHub                                         | Site Role          | Org Admin |
+|------------------------------------------------|--------------------|-----------|
+| [@MattIPv4](https://github.com/MattIPv4)       | `site-admins`      | Yes       |
+| OpenJSF Operations                             | `site-admins`      | Yes       |
+| [@MylesBorins](https://github.com/MylesBorins) | `statuspage-users` | No        |
+
+## Twitter
+
+There is also a Twitter account, [@NodejsStatus](https://twitter.com/NodejsStatus), for the
+status page.
+
+This account is run via the
+[Twitter integration](https://manage.statuspage.io/pages/rxy2rhgm8q1n/twitter) on StatusPage and
+will automatically tweet any incident updates (if tweeting is selected for an incident update).
+
+## Customisations
+
+The [statuspage](../statuspage) directory in this repository contains a backup of the major
+customisations applied to the Node.js status page. These should be updated whenever the
+customisations are changed on the status page.
+
+[colors.md](../statuspage/colors.md) contains the custom color set for the whole status page
+(including emails for subscribers) in
+[Your Page > Customize page and emails > Colors](https://manage.statuspage.io/pages/rxy2rhgm8q1n/design#colors-container).
+
+[styles.scss](../statuspage/styles.scss) stores the custom styling that is applied to all web pages
+of the Node.js StatusPage site. This is controlled from the
+[Customize HTML & CSS](https://manage.statuspage.io/pages/rxy2rhgm8q1n/full-customize) view
+accessed from the button top-right on
+[Your Page > Customize page and emails > Customize status page](https://manage.statuspage.io/pages/rxy2rhgm8q1n/design#design-container).
+
+[header.html](../statuspage/header.html) contains the custom HTML used for the header on the
+status page. It is important to note that this replaces the default header (logo & subscribe
+button), so using this should be avoided if possible. Like the custom styling, this is managed from
+the [Customize HTML & CSS](https://manage.statuspage.io/pages/rxy2rhgm8q1n/full-customize) page.
+
+[footer.html](../statuspage/footer.html) contains custom HTML injected into the footer of the
+status page website. This doesn't replace any default content and is a good place for custom
+scripting if needed. This is also controlled from the
+[Customize HTML & CSS](https://manage.statuspage.io/pages/rxy2rhgm8q1n/full-customize) page.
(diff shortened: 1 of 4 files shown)


richardlau:
@nodejs/build Do we think this is the right place for this (as opposed to a separate repository)? My fear is it will end up like the benchmarking website scripts and coverage scripts which live here in build but are essentially not being actively maintained. Committing these here would also have the barrier of anyone wanting to maintain them (@MattIPv4, for example) having to join the Build WG in order to get write permissions to merge in changes.

MattIPv4:
I posted about this in the build IRC/Slack and the initial suggestion was that here might be easier vs having to create a new repo. I feel these files likely will see very little change going forward, as the design will probably remain rather static. So, with that in mind, I'm not too worried about access, I can just make a PR if it ever needs to be changed.

Equally though, I wouldn't object to having a dedicated repo for it (like [cdnjs/statuspage](https://github.com/cdnjs/statuspage)) so that I (and others) could just push changes up whenever.

richardlau:
My concerns aren't blocking... while I'd prefer they were kept separate if the other members of the team are happy for these files to live in this repository that's fine by me.

AshCripps:
Thought I agree it will just get forgotten about in build, will there actually be that many changes to warrant a seperate repo?

MattIPv4:
I don't think there will be a crazy number of changes, as I said above I hope it'll pretty much just stay as it is. I imagine there will be some refinements though, both to improve accessibility etc. as well as to ensure the design is kept in-line with what's happening on nodejs.dev.

sam-github:
I don't mind them being here rather in yet-another-repo, but there should be a doc/ file that describes where the data came from, so that they can get re-backed up/confirmed to reflect the current state. Also useful would be info on how to use the backups, if restore from backup is necessary. If its super obvious, just a link to the relevant status admin page is likely enough.

Also, the status page is still in the experimenting-with-it phase, but just as a heads up, its not going to get used without docs on how and when to use it, so eventually that policy needs to be added, too.

I don't know if access control has been considered yet, but ideally membership in a GH team would be sufficient to get access, and that team would be another build infra team, in the build-wg README (synced with `ncu-team sync`).

And while I'm thinking of the TBD list... also needed would be changes to nodejs/nodejs.org to point to the status page.

Maybe this PR should be WIP and commits for the other necessaries can be added until its ready to go live, then we merge?

sam-github:
As for @richardlau 's pref for another repo, I can see how that keeps it in its own place, unentangled with anything in the build repo. But then, its so small, maybe that isn't worth the admin overhead (it would need a contributing.md, a governance, worse case, maybe even chartering as a wg so that there is a membership policy). Having it here makes that a bit simpler. Also, the motivation for it was that build (I think?) would be expected to update it under certain conditions, so the policies might be here. I think its 6 of one vs a half-dozen of the other where it should be, but note, once the content is in this PR, or even merged, moving it out is pretty easy if we decide that makes things easier.

MattIPv4:
> I don't mind them being here rather in yet-another-repo, but there should be a doc/ file that describes where the data came from, so that they can get re-backed up/confirmed to reflect the current state. Also useful would be info on how to use the backups, if restore from backup is necessary. If its super obvious, just a link to the relevant status admin page is likely enough.

Makes sense, will write this in a bit. It's pretty simple: there is an admin page for colors, then another for custom css/html.

> Also, the status page is still in the experimenting-with-it phase, but just as a heads up, its not going to get used without docs on how and when to use it, so eventually that policy needs to be added, too.

Will appreciate some help with that once we get to that stage. :)

> I don't know if access control has been considered yet, but ideally membership in a GH team would be sufficient to get access, and that team would be another build infra team, in the build-wg README (synced with ncu-team sync).

There is discussion in #2265 about access. Imo, as many people as possible assuming they know what they're doing.

> And while I'm thinking of the TBD list... also needed would be changes to nodejs/nodejs.org to point to the status page.

Would you like me to raise a draft PR over there now to add the status embed and links in relevant places?

> Maybe this PR should be WIP and commits for the other necessaries can be added until its ready to go live, then we merge?

Sure, I think it makes sense to make this *the* status page PR :)

doc/statuspage.md
```diff
@@ -0,0 +1,57 @@
...
+|------------------------------------------------|--------------------------------|--------|
+| [@MattIPv4](https://github.com/MattIPv4)       | nodejs-status@mattcowley.co.uk | Owner  |
+| [@brianwarner](https://github.com/brianwarner) | operations@openjsf.org         | Member |
+
+## Twitter
+
+There is also a Twitter account, [@NodejsStatus](https://twitter.com/NodejsStatus), for the
+status page.
+
+This account is run by [@MattIPv4](https://github.com/MattIPv4) via nodejs-status@mattcowley.co.uk,
```
mhdawson, line 29: Is there a reason this is not going to be a repo under the node.js org?

doc/statuspage.md
```diff
@@ -0,0 +1,57 @@
...
+
+| GitHub                                         | Email (for StatusPage access)  | Role   |
+|------------------------------------------------|--------------------------------|--------|
+| [@MattIPv4](https://github.com/MattIPv4)       | nodejs-status@mattcowley.co.uk | Owner  |
+| [@brianwarner](https://github.com/brianwarner) | operations@openjsf.org         | Member |
+
+## Twitter
+
+There is also a Twitter account, [@NodejsStatus](https://twitter.com/NodejsStatus), for the
+status page.
```
mhdawson, line 27: We should have the account be under the auspices of the project/OpenJS Foundation.  Is that the plan?

doc/statuspage.md
MattIPv4: Hm?

doc/statuspage.md
MattIPv4: Yeah, once we've figured out access, I think it makes sense to make ops@jsf the owner of twitter & statuspage.

mhdawson:
> Also, the status page is still in the experimenting-with-it phase, but just as a heads up, its not going to get used without docs on how and when to use it, so eventually that policy needs to be added, too.

I'll echo this, that before we say this is "live" we need a clearly defined process for how outages are added/resolved, what expectations are on who will do what etc.

Ideally I think what would work best is if it was based on an issue in the node.js repo (for greatest visibility) and then being approved by two approvals from Node.js collaborators.  We trust our collaborators to push code so it quite likely makes sense to trust them with decide what is a reportable incident and when it is resolved.  Even better if after the approvals simply adding a tag (which any collaborator can do) would result in the incident being pushed to the status page. If this could not be automated, it could be done manually to start.

MattIPv4:
> Ideally I think what would work best is if it was based on an issue in the node.js repo (for greatest visibility) and then being approved by two approvals from Node.js collaborators.

I briefly mentioned this in #2265, but essentially that'd need someone to write a custom solution to integrate with their API. Not something I'm familiar with really.

I would personally recommend having a policy in place and then a set of individuals with access to the StatusPage control panel, as the UI there is probably much easier to use than writing a issue based solution to emulate all of that.

doc/statuspage.md
mhdawson: My question is if there is a reason that the twitter account is run by you personally versus being something managed by the project with you being one of the people who can post to it.

mhdawson: But I guess that is covered below.

MattIPv4: Yeah, I made it at the same time as setting up the status page itself. Going forward, no one should actually need to ever post manually to it, as status page automation will do that, so haning it over the ops@jsf makes sense once we're ready to go.

AugustinMauroy:
any news here ?

PR #2302, opened 2020-04-21
ansible,www: switch from 13.x nightly to 14.x nightly

rvagg:
Not sure why this is so far out of sync but this switches out 13.x-staging nightly for 14.x-staging. But to be honest I'm not sure of the value of these current-staging nightlies. @nodejs/release do you want to weigh in here? Would they be missed if we dropped them? Do you know of anyone even using these non-master nightlies (did you even know they existed?).


targos:
I know they exist but I never use them.

BethGriggs:
I would not be against dropping them.

rvagg:
removed the current-nightly, we're just down to master-nightly and v8-canary-nightly, and this is live now.

PR #2306, opened 2020-04-22
doc: link to other "build" efforts in nodejs/

sam-github:


sam-github:
Are there more? This is in part to diffuse the expectation that if a node release is done (snap, for example), that it is obviously and only the build WG that can make it happen. Also, because I couldn't find the unofficial builds when I went looking and had to ask @richardlau .

README.md
```diff
@@ -24,6 +24,19 @@ missing please open an issue. If you are interested in joining, please read
...
+
+The Build WG isn't responsible for all "build" related activity, there are
+several other groups working on specific aspects of build:
+
+- https://github.com/nodejs/release - Builds the official Node.js releases.
+  Works very closely with the Build WG. Supporting Release is one of the
+  most important goals of the Build WG.
+- https://github.com/nodejs/docker-node - Builds the official Docker images
+  for Node.js. A consumer of the builds produced by Release.
+- https://github.com/nodejs/unofficial-builds/ - Builds unofficial binaries for
```
richardlau, line 37:
```suggestion
- https://github.com/nodejs/unofficial-builds - Builds unofficial binaries for
```
for consistency.

README.md
```diff
@@ -24,6 +24,19 @@ missing please open an issue. If you are interested in joining, please read
 [GOVERNANCE.md][] to understand the process and reasoning we use for granting access
 to the resources we manage.
 
+## Related groups
+
+The Build WG isn't responsible for all "build" related activity, there are
+several other groups working on specific aspects of build:
+
+- https://github.com/nodejs/release - Builds the official Node.js releases.
```
richardlau, line 32:
```suggestion
- https://github.com/nodejs/release - Builds the official Node.js releases hosted on https://nodejs.org/.
```
?

rvagg:
Depends on how expansive you want to be with this, what's the purpose?

* We still have good connections with NodeSource re their Linux distributions and they're still one of the primary ways Node is consumed: https://github.com/nodesource/distributions
* https://snapcraft.io/node has NodeSource's name on it but it's really just me and it's a great way to consume Node on a dev computer, I can strongly recommend it.
* https://github.com/nvm-sh/nvm has close connections to what we do, we collaborate with @ljharb on many things related to this realm and it accounts for a very high percentage of downloads from nodejs.org (thanks in large part to Travis). Very popular with devs.
* node-gyp is closely related and there's strong crossover in people with Build.

But of course, you make a list like this and you'll get others showing up - what about the Windows nvm clones? What about the people that maintain the distribution for some obscure *BSD?

Needs a limiting principle probably.

sam-github:
Limiting to github.com/nodejs/... would be pretty clear cut.

A bit more mushy would be the things the build-wg might reasonably do, if it had infinite resources.

So, nvm and the like would be a stretch, since its building user-facing tooling, but docker builds, snap builds, debian builds, the unofficial builds, the nodesource builds, those are all similar in scope, all things someone might come looking for the build-wg to do, but at least right now would get redirected there to contribute.

PR #2311, opened 2020-04-27
doc: fix typo in macos release notes

AshCripps:
When I ran this for the orka release machine, base64 needed `-D` instead of `-d` to decrypt

### DIFF
--- a/ansible/MANUAL_STEPS.md
+++ b/ansible/MANUAL_STEPS.md
@@ -104,7 +104,7 @@ As root:
 
 #### Signing certificates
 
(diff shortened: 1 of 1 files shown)


sam-github:
Keeping it woring on Orca makes sense, I guess it depends on which `base64` utility is first in the path, GNU uses `-d`: https://www.gnu.org/software/coreutils/manual/html_node/base64-invocation.html#base64-invocation

AshCripps:
Interesting this was on my personal mac not the orka machine

sam-github:
Ah, so it will depend on what people have installed, I bet OS X comes with a base64 that is BSDish, but if you brew install coreutils, you get GNU. Anyhow, if you check the path, and you are using the default OS X utility, then that seems like the right one to document.

PR #2313, opened 2020-04-29
ansible: add ibm(softlayer) 1804 jenkins-workspace, remove old 1604 workspaces

rvagg:
live now, old ones deleted (Michael needs to delete the old softlayer one)

I've gone with the "ibm" name now instead of "softlayer" since the softlayer brand seems all but dead now. @mhdawson let me know if that's not the appropriate thing to do here.


rvagg:
Edited github-bot config (latest commit) to add the new workspace and remove the old ones. There's some IP addresses in the list I can't account for in the inventory so I've left them alone but I think this could be trimmed further.

mhdawson:
Old device removed.

sam-github:
And FTR, we had trouble figuring out how to get perms for anybody but @mhdawson to do it... I can see the devices, and peform some actions, but I can't "Cancel" them, probably the same situation as @rvagg.

PR #2334, opened 2020-06-02
Drop VS2017 and remove EoL v13.x

targos:


doc/windows-visualstudio-supported-versions.md
```diff
@@ -60,3 +60,5 @@ These versions are used to build the official releases.
 10. Support for **building addons** with Visual Studio 2019 was added in v12.8.0 (node-gyp 5.0.0).
     - Pull Request: https://github.com/nodejs/node-gyp/pull/1762
     - Pull Request: https://github.com/nodejs/node/pull/28853
+11. Support for Visual Studio 2017 was removed in v15.0.0.
+    - Pull Request: https://github.com/nodejs/node/pull/33694
```
richardlau, line 64: So it's not new to this PR but the way GitHub renders markdown the bullet points are renumbered (to be sequential) and don't match the references to them above.

targos: I don't see it renumbered on https://github.com/nodejs/build/blob/master/doc/windows-visualstudio-supported-versions.md

targos: Oh actually I see, it just doesn't start at one... Let me fix that.

richardlau: It doesn't just not start at one. In the raw text the numbers are 5, 8, 9, 10, 11 (i.e. no 6 and 7) but they're rendered as 5, 6, 7, 8, 9:
image

targos:
@nodejs/build please merge to unblock https://github.com/nodejs/node/pull/33694 :)

PR #1450, opened 2018-08-14
doc: update ci-public security release matrix

maclover7:
Updating based on changes since the last time these screenshots were taken

PR #2336, opened 2020-06-02
doc: describe how to revert sec lockout

sam-github:


sam-github:
@richardlau WDYT, does this describe what you did? I think we both did it at the same time.

richardlau:
The additional text LGTM (I also didn't see a way to restore the old permissions from the history view). We might need to have a discussion on what the "normal" permissions are supposed to be. I was paranoid and took picture of the permissions prior to editing them for the lockdown. They looked as follows and you'll note there's several entries that weren't in the picture in the guide:

image

I did put all these back, but unfortunately it looks like because we both did it at the same time the changes I made were immediately overwritten by the ones you did (no blame there, the guide is either inaccurate or we'd somehow accrued a bunch of permissions that shouldn't have been there).

There was at least one previous attempt to update the guide which stalled out: https://github.com/nodejs/build/pull/1450

sam-github:
@richardlau I pushed your screenshot update into this PR

sam-github:
I retook the screenshot, with all the headers.

Issue #2330, opened 2020-05-26
AIX CI machines cannot access raw.githubusercontent.com


targos:
This is looking like some sort of infrastructure issue where the AIX CI machine doesn't appear to be able to download https://raw.githubusercontent.com/ URLs. 

e.g.

> 
> 
> Running AIX CI on v12.x to check if that works: https://ci.nodejs.org/job/node-test-commit-aix/30710/

(Note this build passed)

```
...
+ curl https://raw.githubusercontent.com/nodejs/build/master/jenkins/scripts/node-test-commit-pre.sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0+ bash -xe

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
...
  0     0    0     0    0     0      0      0 --:--:--  0:01:14 --:--:--     0
curl: (28) Failed to connect to raw.githubusercontent.com port 443: A remote host did not respond within the timeout period.
...
if [ -x /home/iojs/build/workspace/node-test-commit-aix/nodes/aix71-ppc64/./node ] && [ -e /home/iojs/build/workspace/node-test-commit-aix/nodes/aix71-ppc64/./node ]; then /home/iojs/build/workspace/node-test-commit-aix/nodes/aix71-ppc64/./node  tools/doc/versions.js out/previous-doc-versions.json; elif [ -x `which node` ] && [ -e `which node` ] && [ `which node` ]; then `which node`  tools/doc/versions.js out/previous-doc-versions.json; else echo "No available node, cannot run \"node  tools/doc/versions.js out/previous-doc-versions.json\""; exit 1; fi;
Unable to retrieve https://raw.githubusercontent.com/nodejs/node/master/CHANGELOG.md. Falling back to /home/iojs/build/workspace/node-test-commit-aix/nodes/aix71-ppc64/CHANGELOG.md.
...
```

The `tools/doc/versions.js` tool falls back to the local copy of the `CHANGELOG.md` for non-release commits (which is why only the release commit build failed) to accommodate users with poor/firewalled Internet connections.  

cc @nodejs/platform-aix

_Originally posted by @richardlau in https://github.com/nodejs/node/pull/33197#issuecomment-633946700_

[...]

PR #2342, opened 2020-06-05
jenkins: call checked out script if available

richardlau:
If `node-test-commit-diagnostics.sh` exists locally (e.g. we are in
a checked out copy of the build repository then call the local version
from `node-test-commit-pre.sh`, otherwise download and execute the
latest from GitHub (as we had been doing before this change).

Refs: https://github.com/nodejs/build/issues/2330

### DIFF
--- a/jenkins/scripts/node-test-commit-pre.sh
+++ b/jenkins/scripts/node-test-commit-pre.sh
@@ -39,5 +39,10 @@ if [ -n "${POST_REBASE_SHA1_CHECK}" ]; then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


richardlau:
Test builds:
AIX: https://ci.nodejs.org/view/All/job/node-test-commit-aix-richardlau/3/nodes=aix71-ppc64/ (test failure due to running out of disk space, but the part of the build that runs the script worked)
Linux: https://ci.nodejs.org/job/node-test-commit-linux-richardlau/4/ (✔️)

richardlau:
Test builds look okay. This is ready for review.

Issue #2345, opened 2020-06-09
z/OS hosts offline


richardlau:
All of our z/OS hosts are offline. I can ssh into them and have tried restarting the Jenkins agent, but the agent is unable to connect back to https://ci.nodejs.org. We see this in the agent logs (`/u/iojs/jenkins.log`:
```
Caused by: java.security.cert.CertPathValidatorException: The certificate issued by CN=USERTrust RSA Certification Authority, O=The USERTRUST Network, L=Jersey City, ST=New Jersey, C=US is not trusted; internal cause is:
        java.security.cert.CertPathValidatorException: Certificate chaining error
        at com.ibm.security.cert.BasicChecker.<init>(BasicChecker.java:111)
```

Which looks like it's related to those root certs that expired at the end of May: https://support.sectigo.com/articles/Knowledge/Sectigo-AddTrust-External-CA-Root-Expiring-May-30-2020 which also caused issues in https://github.com/nodejs/build/pull/2337.

rvagg:
Is it possible this might be fixed on the hosts? I removed the expired certificates from the chain but it looks like this host wants a complete chain. I think the only option there is to regenerate the cert from scratch.

richardlau:
I've reached out internally in IBM for advice. One of the things I'm looking at is whether we can update Java since what we're currently running was built in 2016.
```
IOJS:/u/iojs: >/usr/lpp/java/J8.0_64/bin/java -version
java version "1.8.0"
Java(TM) SE Runtime Environment (build pmz6480sr3fp22-20161213_02(SR3 FP22))
IBM J9 VM (build 2.8, JRE 1.8.0 z/OS s390x-64 Compressed References 20161209_329148 (JIT enabled, AOT enabled)
J9VM - R28_20161209_1345_B329148
JIT  - tr.r14.java.green_20161207_128946
GC   - R28_20161209_1345_B329148_CMPRSS
J9CL - 20161209_329148)
JCL - 20161213_01 based on Oracle jdk8u111-b14
IOJS:/u/iojs: >
```
(I don't think that was installed via ansible.)

richardlau:
After some investigation it appears that this should be resolved by updating the version of Java we are using to include APAR IJ05800 (which adds the "USERTrust RSA Certification Authority" to the Java's cacerts): https://www.ibm.com/support/pages/apar/IJ05800

[...]

Issue #2258, opened 2020-04-02
Errors running `playbooks/jenkins/worker/create.yml` playbook on z/OS


richardlau:
I've not been able to run the complete unaltered `playbooks/jenkins/worker/create.yml` playbook on z/OS. The first error encountered is:
```
TASK [jenkins-worker : set NODE_TEST_DIR permission and owner] *************************************************************************************************************
fatal: [test-marist-zos13-s390x-1]: FAILED! => {"changed": false, "gid": 0, "group": "IPGROUP", "mode": "0755", "msg": "chgrp failed: failed to look up group iojs", "owner": "IOJS", "path": "/u/iojs/tmp", "size": 8192, "state": "directory", "uid": 990005}
```

This task attempts to set the ownership and group to the same value (`{{ server_user }}`, i.e. `iojs`) but on z/OS there is no `iojs` group:
https://github.com/nodejs/build/blob/050fb9733697acb3dd5bb53891c2818e2c335bab/ansible/roles/jenkins-worker/tasks/main.yml#L48-L53

If I use `--start-at-task` to start at a later task in the playbook I get this error:
```
TASK [jenkins-worker : enable jenkins at startup - zos - part 1] ***********************************************************************************************************
fatal: [test-marist-zos13-s390x-1]: FAILED! => {"changed": false, "checksum": "549cb9fd20154201348340261de0e716f3a15a6c", "expected_checksum": "753bae1fd4cc32fb65f07a65c2ff03fcd769307c", "msg": "Copied file does not match the expected checksum. Transfer failed."}
```

richardlau:
Updated https://github.com/nodejs/build/pull/2346 with changes to fix running the playbook against our z/OS hosts. The basic idea is to force sftp for file transfers (to prevent the files being converted) and then either tag them as ASCII (where possible) or, if the file is used by the system, convert them to EBCDIC as tasks in the playbook.

Some ansible oddities while investigating this:
- The [template](https://docs.ansible.com/ansible/latest/modules/template_module.html) module has a parameter `output_encoding` but if I set an EBCDIC encoding (e.g. `cp500`) the resulting file is in EBCDIC but loses all line ending (so everything's on one line). Changing the `newline_sequence` parameter makes things worse.
- It doesn't seem possible to override the `scp_if_ssh` setting with `ansible_scp_if_ssh` (but `ansible_ssh_transfer_method` does work to specify the transfer method, e.g. sftp, used).
[...]

PR #2346, opened 2020-06-10
ansible: update Java on z/OS

richardlau:
First commit:
> ansible: fix z/OS ansible scripts
>
> Current versions of ansible checksum files before transferring them
to the remote host. Our ansible.cfg was forcing scp to be used for
transfer, which on z/OS results in an ASCII->EBCDIC conversion and
a mismatched checksum, causing ansible to error.
> 
>This commit sets the file transfer method for z/OS to sftp, which on
z/OS will transfer the file without conversion and fix the mismatched
checksums. The transferred files are either tagged as ASCII or
converted to EBCDIC if they are used by the system, e.g. the start
script.

Fixes: https://github.com/nodejs/build/issues/2258

Second commit:
> ansible: update Java on z/OS
>
> We need to use a newer service refresh of Java 8 on z/OS to one containing
an updated cacerts file that contains the root certificate to verify the
ones we have for our CI servers.

Fixes: https://github.com/nodejs/build/issues/2345

### DIFF
--- a/ansible/MANUAL_STEPS.md
+++ b/ansible/MANUAL_STEPS.md
@@ -25,6 +25,7 @@
 * [Raspberry Pi](#raspberry-pi)
   * [NFS boot](#nfs-boot)
 * [IBM i](#ibm-i)
+* [z/OS](#zos)
 
 ## Adding firewall entries for Jenkins workers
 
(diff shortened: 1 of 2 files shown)


richardlau:
I've pushed another commit (ordered first) that fixes https://github.com/nodejs/build/issues/2258. I've tested it (and both commits in this PR) on `test-marist-zos13-s390x-1` with ansible 2.9.9.

ansible/ansible.cfg
```diff
@@ -14,7 +14,6 @@ squash_actions      = apk
 # since we have all information available.
 [ssh_connection]
 ssh_args            = -F /dev/null -o ControlMaster=auto -o ControlPersist=60s
-scp_if_ssh          = True
```
mhdawson, line 17: This will affect all platforms right?  Seem like what we need for z/OS but not sure if we want the same for all platforms?

richardlau: I haven't found a way to unset this if set.

richardlau: Found a setting (`ssh_transfer_method`) so reverted this back to True.

richardlau:
Updated to use enhanced ASCII file tagging where possible. This reduces the number of written files (when we use `iconv` to convert we need an intermediate file). We still convert to EBCDIC the files used by the system, e.g. the `start.sh` script is called from `/etc/rc`.

richardlau:
Landed in 7887eb3a8246233714f175701fd75c65c65143d8...d4b074cd221cce329ee1f31292e3114845390a56.

(Sorry, messed up pushing order which is why this PR now looks like it didn't change anything.)

Issue #2315, opened 2020-04-30
Migrate off of Macstadium vsphere


AshCripps:
In the vsphere we currently have 6 active machines - 2 Release (10.10 and 10.11) and 4 test (2 10.10 and 2 10.11).

The current issue is that Orka doesn't support anything older than 10.13 as shown below:

```
Ash,

Unfortunately, we don’t currently support any macOS older than High Sierra 10.13 on Orka.  Our team spent considerable time attempting to make older macOS releases work on Orka but encountered significant challenges whereupon it was determined that a lower-level rewrite that was in the works made more sense. 

This is called the, “Bring your own image” model, and we are actively developing it but it is not scheduled for release in the next sprint due May 11.  I presume not being able to support these older instances is a deal breaker with regard to bringing the entire build apparatus over to Orka? 
```

So we have two options:
1) Wait for the new model to be released in May and hopefully add the machines back that way.
2) Stop using the machines (asides from tests only the 10.10 machine is used to build release sources but only on 10.x)

Thought @nodejs/build? Im personally for moving off these old machines.

rvagg:
I'm fine with either option.

We're using the old machines for testing and building source + header tarballs for 10.x. I don't believe breakage would be likely if we were to move the tarball creation to Linux, it would likely be an improvement if anything. Keeping them fixed was always just a matter of being cautious so I guess it depends on how cautious we want to be.

richardlau:
I think migrating off the release machine (I don't think that the macOS 10.11 release machine is used now?) should be low risk.

Losing the test machines would be unfortunate. Then again we've never tested as far back as we claim to support (i.e. macOS 10.7 for Node.js 10.x).

AshCripps:
with #2347 merged im going to wait a bit incase anything breaks and the will start the process of decomissioning the machines in vsphere

AshCripps:
To Do List:
- [x] Move off osx10.10 for 10.x release headers (https://github.com/nodejs/build/pull/2347)
- [ ] Remove osx10.11 labels from jobs
- [ ] Wait for a 12.x release to ensure nothing goes amiss
- [ ] Disable and remove osx10.11 from jenkins
- [ ] Remove osx10.10 labels from jobs
- [ ] Wait for a 10.x release to ensure nothing goes amiss
- [ ] Disable and remove osx10.10 from jenkins
- [ ] Contact macstadium to begin decommissing vsphere cluster

AshCripps:
[...]

PR #2347, opened 2020-06-11
jenkins: move 10.x release sources off osx

AshCripps:
I want to propose moving the build of 10.x release sources off osx onto linux with the rest of the release lines.

This will allow us to migrate off of our old macstadium vsphere (https://github.com/nodejs/build/issues/2315) as this is our only dependency upon osx 10.10.

Im not 100% sure if this is possible to test before merging.

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -113,8 +113,7 @@ def buildExclusions = [
   [ /^freebsd10/,                     anyType,     gte(11) ],
 
   // Source / headers / docs -------------------------------
-  [ /^osx1010-release-sources$/,      releaseType, gte(11) ],
-  [ /^centos7-release-sources$/,      releaseType, lt(12)  ],
+  [ /^centos7-release-sources$/,      releaseType, lt(10)  ],
 
   // -------------------------------------------------------
 ]


richardlau:
> Im not 100% sure if this is possible to test before merging.

We can produce test builds in the release-ci.

richardlau:
> Im not 100% sure if this is possible to test before merging.

Probably not without cloning the job in the release-ci to change where it picks up the script from.

AshCripps:
> Probably not without cloning the job in the release-ci to change where it picks up the script from.

Yeah I didnt think so, theres no release scheduled for a couple weeks from what I can see in the release repo so merging and running a couple test release-ci jobs to verify should be good enough

AshCripps:
ive ran a sample build before changes for comparison: https://ci-release.nodejs.org/job/iojs+release/6102/

build with changes: https://ci-release.nodejs.org/job/iojs+release/6103/

AshCripps:
Both builds pass and the outputs look the same.

Ive revert the job back to using master's versionselectorscript

rvagg:
I had a quick look and I don't think we're stating anywhere how the tarballs are built, so we're not _officially_ tied to their format. There will be a difference in the tarball format but it _should_ be an improvement in compatibility rather than introducing potential incompatibilities. But you never know what's going on out there in the wild, there's always surprises when we change the most minor things! 🤞 thanks for this @AshCripps

PR #2364, opened 2020-06-23
Container upgrade: Alpine 3.12, OpenSSL 1.1.1g, remove OpenSSL 1.0.2 and FIPS

rvagg:
Alpine 3.12 doesn't come with a Python 2 option and Node 10 and 12's `configure` don't seem to like Python 3 so I've added an exclusion to VersionSelectorScript. I've also had to symlink `python3` as `python` in those packages because we still force `PYTHON=python` in the commit-linux job.

### DIFF
--- a/ansible/roles/docker/templates/alpine312.Dockerfile.j2
+++ b/ansible/roles/docker/templates/alpine312.Dockerfile.j2
@@ -1,4 +1,4 @@
-FROM alpine:3.10
+FROM alpine:3.12
 
 ENV LC_ALL C
 ENV USER {{ server_user }}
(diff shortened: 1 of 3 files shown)


richardlau:
We can also remove openssl 1.0.2 and fips from node-test-commit-linux-containered (we'll need to keep the VersionSelector entries until after we update the job). I can do that, if there are no objections.

rvagg:
yep, was going to leave that for a separate job since the order of operations is important, more than happy for you to do it. the old FIPS too while you're there.

richardlau:
> 
> 
> We can also remove openssl 1.0.2 and fips from node-test-commit-linux-containered (we'll need to keep the VersionSelector entries until after we update the job). I can do that, if there are no objections.

I've updated the job.
Test builds:
10.x: https://ci.nodejs.org/job/node-test-commit-linux-containered/20965/
master: https://ci.nodejs.org/job/node-test-commit-linux-containered/20966/

Labels still exist on each of the containered nodes in Jenkins,
i.e. 

- https://ci.nodejs.org/label/ubuntu1804_sharedlibs_fips20_x64/
- https://ci.nodejs.org/label/ubuntu1804_sharedlibs_openssl102_x64/

and the VersionSelectorScript.groovy hasn't been updated. They're relatively harmless (as the job is no longer using them), so no rush to edit.

PR #2408, opened 2020-08-11
jenkins: add ARM64 Windows

joaocgreis:
Windows ARM64 cross-compilation support has recently been added to master/v14. This adds Windows ARM64 to the main Windows compile job in Jenkins. This allows us to move support for compilation for Windows ARM64 to Tier 2.

Running Node.js on ARM64 is still experimental, and we only have one machine connected to Jenkins at the moment. Thus, a checkbox is added in the main job to run the tests, disabled by default.

Test job: https://ci.nodejs.org/view/All/job/node-test-commit-windows-fanned-arm64/ . The changes in this job and the 3 subjobs that have the `-arm64` suffix need to land on the main jobs simultaneously with this PR.

cc @nodejs/build @nodejs/platform-windows

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -65,9 +65,10 @@ def buildExclusions = [
   [ /vcbt2015(-\w+)?$/,               testType,    gte(10)       ],
   [ /vs2017(-\w+)?$/,                 testType,    ltGte(8, 15)  ],
   [ /vs2019(-\w+)?$/,                 testType,    lt(13)        ],
-  [ /vs2015-x86$/,                    testType,    gte(10)       ], // compile x86 only once
+  [ /vs2015-x86$/,                    testType,    gte(10)       ], // compile arm64/x86 only once
   [ /vs2017-x86$/,                    testType,    ltGte(10, 14) ],
   [ /vs2019-x86$/,                    testType,    lt(14)        ],
+  [ /vs2019-arm64$/,                  testType,    lt(14)        ],
   // VS versions supported to build add-ons
   [ /vs2013-COMPILED_BY/,             testType,    gte(9)        ],
   [ /vs2015-COMPILED_BY/,             testType,    noVer         ],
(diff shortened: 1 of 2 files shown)


rvagg:
That's one nasty hack in the version selector to get the checkbox to work! Probably better than building too much more of the logic into Jenkins but it might not be a good pattern to encourage unless we can find a way to make it a bit cleaner.

Is the intention to leave the checkbox only at the fanned-arm64 job and its subjobs or does it get migrated all the way up the job call stack?

And what's the path to a more permanent CI presence? Do we have any word on Windows ARM64 hosting options, do we even imagine that possibility on the horizon or are we going to be in a similar position to our Pi cluster, with Windows laptops in offices?

joaocgreis:
> Is the intention to leave the checkbox only at the fanned-arm64 job and its subjobs or does it get migrated all the way up the job call stack?

Only at the `node-test-commit-windows-fanned` job and subjobs. If someone wants to run tests on ARM64 it's easy and reasonable to use a dedicated run for it, only at the Windows level. Note the `fanned-arm64` job is only for testing the changes until this lands. After, I plan to merge all the `-arm64` jobs into the original ones.

> And what's the path to a more permanent CI presence?

The plan is to have Windows ARM64 VMs on Azure as soon as they become available. There is no set date for that to happen, but we'd like to wait some time in order to avoid the laptops in offices option.

sxa:
@joaocgreis What would I need to do to run a local a build of this? Can I just give an `arm64` parameter to vcbuild on master on an x64 machine with a recent Visual Studio?

joaocgreis:
@sxa Visual Studio needs some optional components installed, listed in https://github.com/nodejs/node/blob/master/BUILDING.md#option-1-manual-install. In that list the first point mentions you need an ARM64 machine but that's not correct (a correction should land soon). At the moment, you can only build from x64.

Then, just add `arm64` to `vcbuild`. It's not possible to run tests on x64, but other parameters should be compatible. Support for `msi` should land soon (https://github.com/nodejs/node/pull/34009).

joaocgreis:
Landed in https://github.com/nodejs/build/commit/b03b57cf8b16d4ee9b524b8ea2db94b683b7f891

Will update the Jenkins jobs.

joaocgreis:
Jenkins jobs updated.

Test run: https://ci.nodejs.org/view/All/job/node-test-commit-windows-fanned/37850/

dennisameling:
@joaocgreis do you have any idea when we'll see Windows arm64 show up in unofficial/nightly builds? Those still seem to be x64 and x86-only: https://nodejs.org/download/nightly/v15.0.0-nightly20200922d71b467bbe/ and https://unofficial-builds.nodejs.org/download/release/v14.11.0/.

Please let me know if I can be of any help here 😊

PR #2410, opened 2020-08-13
doc: document validating apple certificates

AshCripps:
Add a set of instructions to validate the apple certs are still in date


mhdawson (approved):
LGTM

PR #2428, opened 2020-09-08
ansible: update `upgrade-jar.yml` to use a variable

AshCripps:
When the certs expired on the main CI we had to upgrade the jenkins
agents to use the newer certs. When I tried this on the macs it failed
due to a hard coded `/home` which doesn't exist on mac. Used the create
playbook as the inspiration for this change.

I need to test this on mac before I can merge.

### DIFF
--- a/ansible/playbooks/jenkins/worker/upgrade-jar.yml
+++ b/ansible/playbooks/jenkins/worker/upgrade-jar.yml
@@ -14,5 +14,28 @@
         mode: 0644
         force: yes
 
-    - name: restart jenkins
+    - name: Kill Jenkins process - AIX
+      shell: ps awwx | grep jenkins | grep -v grep | awk '{ print $1 }' | xargs kill
+      when: os|startswith("aix")
+
+    - name: Restart Jenkins agent - AIX
+      shell: /etc/rc.d/rc2.d/S20jenkins start
+      become: yes
+      become_user: root
+      when: os|startswith("aix")
+
+    - name: Stop Jenkins process - macos
+      when: os|startswith("macos")
+      command: launchctl unload /Library/LaunchDaemons/org.nodejs.osx.jenkins.plist
+
+    - name: Reload Jenkins process - macos
+      when: os|startswith("macos")
+      command: launchctl load /Library/LaunchDaemons/org.nodejs.osx.jenkins.plist
+
+    - name: Restart Jenkins service - generic
+      when:
+        - not os|startswith("aix")
+        - not os|startswith("ibmi")
+        - not os|startswith("macos")
+        - not os|startswith("zos")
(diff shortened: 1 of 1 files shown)


richardlau (approved):

richardlau:
FWIW the task in the main jenkins-worker playbook sets `force: true` so will always upgrade when run
https://github.com/nodejs/build/blob/d4b074cd221cce329ee1f31292e3114845390a56/ansible/roles/jenkins-worker/tasks/main.yml#L153-L160

The upgrade-jar playbook is more lightweight, but probably the "proper" way in ansible would be to use tags. But that's a whole other PR.

AshCripps:
@richardlau reckon adding `force: yes` here would be best as this is to force and upgrade?

> the "proper" way in ansible would be to use tags. But that's a whole other PR.

That sounds like a big refactoring for another day :)

richardlau:
> 
> 
> @richardlau reckon adding `force: yes` here would be best as this is to force and upgrade?

Sounds reasonable. I think ansible doesn't replace the file if it exists otherwise:
https://docs.ansible.com/ansible/latest/modules/get_url_module.html#parameter-force

richardlau (approved):

mhdawson (approved):
LGTM

AshCripps:
This is needs to be held off of merging - the agent isnt restarted on macos so needs some investigating/rewriting of the start.sh script

ansible/playbooks/jenkins/worker/upgrade-jar.yml
```diff
@@ -10,8 +10,32 @@
...
         url: "{{ jenkins_worker_jar }}"
-        dest: /home/{{ server_user }}/slave.jar
+        dest: '{{ home }}/{{ server_user }}/slave.jar'
         mode: 0644
+        force: yes
 
-    - name: restart jenkins
+    - name: Kill Jenkins process - AIX
+      shell: ps awwx | grep jenkins | grep -v grep | awk '{ print $1 }' | xargs kill
+      when: os|startswith("aix")
```
richardlau, line 19: Maybe we should implement `stop` in the rc.d script? https://github.com/nodejs/build/blob/2a48e7f4c4e91bff403e346885ffc20150b1e5a4/ansible/roles/jenkins-worker/templates/aix.rc2.j2#L25-L26

richardlau (approved):
Generally looks okay. Maybe in the future we could tag tasks in the `jenkins-worker` role and avoid repetition here?

ansible/playbooks/jenkins/worker/upgrade-jar.yml
AshCripps: hmm yeah that should be something we should look into, I never realised it was even defined in the script.

PR #2439, opened 2020-09-18
doc: add some AIX specific troubleshooting

AshCripps:
AIX often has much bigger disks than is being used - add some basic
steps to increase the size of the disk without limiting the amount of
builds that can run on the machine

### DIFF
--- a/doc/jenkins-guide.md
+++ b/doc/jenkins-guide.md
@@ -12,6 +12,7 @@ A guide on maintaining Node.js' Test and Release Jenkins clusters
 * [Solving problems](#solving-problems)
   * [Out of memory](#out-of-memory)
   * [Out of space](#out-of-space)
+    * [AIX Space Issues](#AIX-Space-Issues)
   * [General issues with Jenkins agent: "normal machines" edition](#general-issues-with-jenkins-agent-normal-machines-edition)
   * [Read-only filesystem](#read-only-filesystem)
   * [Restart the machine](#restart-the-machine)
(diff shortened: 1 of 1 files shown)


doc/jenkins-guide.md
```diff
@@ -160,6 +161,61 @@ then it is probably appropriate to clean out part of the worker's
 workspace (where Jenkins jobs are performed). To clean out part of the
 workspace, run `rm -rf ~/build/workspace/node-test-commit*`.
 
+#### IBM platforms
```
richardlau, line 164: I'd suggest just titling this section as `AIX` and then we'd just list further platforms under their own headings.

doc/jenkins-guide.md
```diff
@@ -160,6 +161,61 @@ then it is probably appropriate to clean out part of the worker's
...
+MAX PPs per VG:     32512
+MAX PPs per PV:     16256                    MAX PVs:        2
+LTG size (Dynamic): 512 kilobyte(s)          AUTO SYNC:      no
+HOT SPARE:          no                       BB POLICY:      relocatable
+PV RESTRICTION:     none                     INFINITE RETRY: no
+DISK BLOCK SIZE:    512                      CRITICAL VG:    no
+FS SYNC OPTION:     no
+```
+
+To increase the size of `/home` run: `chfs -a size=+XG /home` where X is the number of GBs to increase the system.
```
richardlau, line 212: Does this work as the default login (whatever we have defined in the inventory), or does it have to be run as root?

richardlau (approved):

doc/jenkins-guide.md
AshCripps: thats a good point im running as root currently - will update

doc/jenkins-guide.md
```diff
@@ -12,6 +12,7 @@ A guide on maintaining Node.js' Test and Release Jenkins clusters
 * [Solving problems](#solving-problems)
   * [Out of memory](#out-of-memory)
   * [Out of space](#out-of-space)
+    * [AIX Space Issues](#AIX-Space-Issues)
```
richardlau, line 15: Technically the anchor is all lower case, e.g. `#aix-space-issues`, but it seems to still link through to the correct place ¯\\_(ツ)\_/¯.

richardlau (approved):

doc/jenkins-guide.md
```diff
@@ -160,6 +161,61 @@ then it is probably appropriate to clean out part of the worker's
...
+DISK BLOCK SIZE:    512                      CRITICAL VG:    no
+FS SYNC OPTION:     no
+```
+
+To increase the size of `/home` run: `sudo chfs -a size=+XG /home` where X is the number of GBs to increase the system.
+
+If you hit an error about the maximum allocation of the logical volume you can increase the max with: 
+`sudo chlv -x <size> <logical volume>`
+
+*If you are unsure about any of the commands please ask a Red Hat/IBM member of the build working group for assistance*
```
mhdawson, line 217: I think we need to say that if we are increasing sizes, that those changes should be PR'd into the ansible templates so that new machines start of with the larger sizes.

AshCripps: We currently dont do any disk resizing in ansible as far as im aware. It would also be tricky and each machine from each provider is often different so you cant guarentee the sizes available.

mhdawson: @AshCripps, we used to have some some instructions to increase the default sizes in the manual steps for AIX.  I think we should try to capture somewhere, so that we don't have to re-debug issues after introducing new machines.

PR #2454, opened 2020-10-02
ansible: replace fedora27 with fedora32

rvagg:
A chore I've put off far too long. I was able to reimage both fedora27 VMs with fedora32, thanks DigitalOcean and Rackspace 👌. I've shuffled tags so fedora30 is now fedora-last-latest-x64 and fedora32 is fedora-latest-x64. The major problem encountered here is the choice that Fedora have made somewhere along the way to make /usr/bin/python default to python3! Which is a bit surprising since there seems to have developed a consensus of using that for python2 and forcing python3 use through the executable with that name. Since fedora30 python2 hasn't come installed so we have to install it in ansible to make everything work (yes this sucks, we'll rid ourselves of it eventually, there's too many baked in assumptions). So the script now has to redirect /usr/bin/python to python2 for things to work nicely.

Test runs:

* master https://ci.nodejs.org/job/node-test-commit-linux/37330/
* v14.x https://ci.nodejs.org/job/node-test-commit-linux/37331/
* v12.x https://ci.nodejs.org/job/node-test-commit-linux/37332/
* v10.x https://ci.nodejs.org/job/node-test-commit-linux/37333/

So far I'm seeing some `ERROR: Unable to retrieve changeset` git errors on fedora30 machines which I don't understand. Even after cleaning out their workspaces.


AshCripps (approved):

rvagg:
I have to head off for the night so I'd appreciate if someone could keep an eye on those test jobs to make sure there are no major test or compile surprises with fedora32. Restarting the fedora30 failures for master and 14.x might be good to to make sure those aren't persistent errors. I don't want to hold up CI with persistent failures which is why we do these test runs. If they're going to get in the way of normal CI runs then maybe we remove the two fedora labels from node-test-commit-linux until we can figure out what's wrong.

rvagg:
CI looking solid a day later

Issue #2476, opened 2020-11-05
too many threads on libuv-test-aix-cmake


vtjnash:
I'm seeing the identical problem as https://github.com/nodejs/build/issues/2402 (@rvagg), but on the https://ci.nodejs.org/job/libuv-test-commit-aix-cmake/250/ worker with the missing JOBS variable. Thanks!

mhdawson:
@AshCripps can you take  a look at this one ?

AshCripps:
Added the variable https://ci.nodejs.org/job/libuv-test-commit-aix-cmake/253/ passed

richardlau:
@AshCripps We'll need to set `JOBS` as well for the AIX nodes in Jenkins. [`ansible/roles/jenkins-worker/templates/aix.rc2.j2`](https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/templates/aix.rc2.j2) is probably the best place to do so based on the Jenkins start scripts for other platforms. 

FWIW it looks like the failing job (https://ci.nodejs.org/job/libuv-test-commit-aix-cmake/250/) was the only one to run on the new AIX image (`test-ibm-aix71-ppc64_be-4`) as Jenkins appears to prefer `test-ibm-aix71-ppc64_be-2`/`test-ibm-aix71-ppc64_be-1`: https://ci.nodejs.org/job/libuv-test-commit-aix-cmake/nodes=aix71-ppc64/buildTimeTrend
The new image probably either has more CPU's or [SMT](https://en.wikipedia.org/wiki/Simultaneous_multithreading) per CPU pushing the build over the edge with `-j` without a capped number.

AshCripps:
@richardlau the machines have the same amount of CPUs. question is why would we need to set it for these ones and not the older ones?

richardlau:
> @richardlau the machines have the same amount of CPUs. question is why would we need to set it for these ones and not the older ones?

@AshCripps 🤷:
```
$ ssh test-ibm-aix71-ppc64_be-2  'python -c "import multiprocessing; print(multiprocessing.cpu_count())"'
16
$ ssh test-ibm-aix71-ppc64_be-4  'python -c "import multiprocessing; print(multiprocessing.cpu_count())"'
32
$
```
or with `node`:
```
$ ssh test-ibm-aix71-ppc64_be-2 '/home/iojs/build/workspace/node-test-commit-aix/nodes/aix71-ppc64/out/Release/node -p "os.cpus().length"'
16
[...]

PR #2477, opened 2020-11-09
ansible: set JOBS to 6 on AIX

AshCripps:
refs: #2476

### DIFF
--- a/ansible/roles/jenkins-worker/templates/aix.rc2.j2
+++ b/ansible/roles/jenkins-worker/templates/aix.rc2.j2
@@ -17,6 +17,7 @@ start )
               export NODE_TEST_DIR=$HOME/tmp; \
               export NODE_COMMON_PIPE="$HOME/test.pipe"; \
               export OSTYPE=AIX72; \
+              export JOBS={{ jobs_env }}; \
         /usr/bin/java -Xmx128m -Dorg.jenkinsci.plugins.gitclient.Git.timeOut=30 -jar "$HOME/slave.jar" \
           -secret {{secret}} \
           -jnlpUrl {{jenkins_url}}/computer/{{inventory_hostname}}/slave-agent.jnlp >$jenkins_log_file 2>&1 &'
--- a/ansible/roles/jenkins-worker/vars/main.yml
+++ b/ansible/roles/jenkins-worker/vars/main.yml
@@ -145,5 +145,5 @@ jobs_variants: {
   armv6l: '1',
   armv7l: '2',
   arm64: '3',
-  smartos: '2'  
+  smartos: '2'
 }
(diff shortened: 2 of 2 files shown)


ansible/roles/jenkins-worker/templates/aix.rc2.j2
```diff
@@ -17,6 +17,7 @@ start )
               export NODE_TEST_DIR=$HOME/tmp; \
               export NODE_COMMON_PIPE="$HOME/test.pipe"; \
               export OSTYPE=AIX72; \
```
AshCripps, line 19: why do we set this to `AIX72`?

richardlau: I have no idea what `OSTYPE` is used for. Originally AIX 7.2 was what we were going to use on the current machines but we rolled back to AIX 7.1 for minimum supported version/compatibility reasons -- I suspect this wasn't adjusted when we switched to AIX 7.1.

ansible/roles/jenkins-worker/templates/aix.rc2.j2
```diff
@@ -17,6 +17,7 @@ start )
...
+              export JOBS=4; \
```
richardlau, line 20: Do we want to make this something like 
```suggestion
              export JOBS={{ jobs_env }}; \
```
and set `server_jobs` in the inventory or add a variant in
https://github.com/nodejs/build/blob/ae1841834dfd8d941f356b38f187c07e39131642/ansible/roles/jenkins-worker/vars/main.yml#L142-L149

https://github.com/nodejs/build/blob/ae1841834dfd8d941f356b38f187c07e39131642/ansible/roles/jenkins-worker/tasks/main.yml#L7-L25
This would potentially allow us to set different values on machine basis.

ansible/roles/jenkins-worker/vars/main.yml
```diff
@@ -144,5 +144,6 @@ jobs_variants: {
   armv6l: '1',
   armv7l: '2',
   arm64: '3',
-  smartos: '2'
+  smartos: '2',
+  aix: '4'
```
richardlau, line 148: Keep this is alphabetical order?

richardlau (approved):

ansible/roles/jenkins-worker/vars/main.yml
```diff
@@ -141,8 +141,9 @@ scaleway_armv7: {
 # primarily for raspberry pi workers, some arm64 workers may need `server_jobs` defined in inventory
 # to override these low numbers
 jobs_variants: {
+  aix: '4',
   armv6l: '1',
   armv7l: '2',
   arm64: '3',
-  smartos: '2'
+  smartos: '2'
```
richardlau, line 148:
```suggestion
  smartos: '2'
```

richardlau (approved):

mhdawson (approved):
LGTM

AshCripps:
both `-3` and `-4` are updated to have `$JOBS` of 6

richardlau (approved):

Issue #2487, opened 2020-11-29
github-bot playbook errors on `install apt-transport-https` task


phillipj:
With the goal of upgrading Node.js running on the github-bot server (https://github.com/nodejs/build/pull/2479), I'm trying to use the [`create-github-bot.yml` playbook](https://github.com/nodejs/build/blob/master/ansible/playbooks/create-github-bot.yml).

After having setup secrets & SSH key, I'm able to ping the host and get past the initial `check if secrets are properly set` task successfully.

Sadly it explodes shortly after with the following error:

```
$ ansible-playbook --check playbooks/create-github-bot.yml

... snipped for brevity

TASK [bootstrap : install apt-transport-https] ******************************************************************************************************************************************
fatal: [infra-rackspace-debian8-x64-1]: FAILED! => {"msg": "The conditional check 'has_apt_transport.rc == 1' failed. The error was: error while evaluating conditional (has_apt_transport.rc == 1): 'dict object' has no attribute 'rc'\n\nThe error appears to be in '/..../build/ansible/roles/bootstrap/tasks/partials/debian8.yml': line 12, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: install apt-transport-https\n  ^ here\n"}
```

Which again references this part: https://github.com/nodejs/build/blob/dc94f7e911ea6d2ea90a28a9cf2966a06f2f12f2/ansible/roles/bootstrap/tasks/partials/debian8.yml#L7-L14

Does this look familiar to anyone?

phillipj:
Diving into the docs and noticing the `check for apt-transport-https` got skipped by ansible, I decided to just run the playbook without `--check`. I didn't find an obvious way to make the `raw` module run in check mode, even with `check_mode: yes` added 🤷‍♂️ 

Bottom line: this is only an issue when running in check mode. Works as expected otherwise.

PR #2499, opened 2020-12-12
ansible: fix needed to run debian8 bootstrap in check mode

phillipj:
Running playbooks using the `bootstrap` role on machines running debian8 failed when in check mode:

```
$ ansible-playbook --check playbooks/create-github-bot.yml

TASK [bootstrap : check for apt-transport-https] ****************************************************************************************************************************************
skipping: [infra-rackspace-debian8-x64-1]

TASK [bootstrap : install apt-transport-https] ******************************************************************************************************************************************
fatal: [infra-rackspace-debian8-x64-1]: FAILED! => {"msg": "The conditional check 'has_apt_transport.rc == 1' failed. The error was: error while evaluating conditional (has_apt_transport.rc == 1): 'dict object' has no attribute 'rc'\n\nThe error appears to be in '/..../build/ansible/roles/bootstrap/tasks/partials/debian8.yml': line 12, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: install apt-transport-https\n  ^ here\n"}
```

The `install apt-transport-https` task failed due to the fact that the preceding `check for apt-transport-https` did not define the `has_apt_transport` result because it got skipped.

Ansible skips that task by default because it uses a `raw` command and it's therefore unknown wheter or not that command has a side effect.

~~By setting `check_mode: no` we tell ansible it's okey to run that command also when running playbooks in check mode (`--check`).~~

Instead of using `raw`, we ended up wanting to use `stat` for the purpose of checking if a file exists or not. Primarily because it's more self descriptive and ansible knows it safe to execute in all scenarios, so it doesn't need to be _forcibly_ run in check mode.

Refs https://github.com/nodejs/build/issues/2487

### DIFF
--- a/ansible/roles/bootstrap/tasks/partials/debian8.yml
+++ b/ansible/roles/bootstrap/tasks/partials/debian8.yml
@@ -5,12 +5,12 @@
 #
 
 - name: check for apt-transport-https
-  raw: stat /usr/lib/apt/methods/https
+  stat:
+    path: /usr/lib/apt/methods/https
   register: has_apt_transport
-  failed_when: has_apt_transport.rc > 1
 
 - name: install apt-transport-https
-  when: has_apt_transport.rc == 1
+  when: has_apt_transport.stat.exists == False
   raw: apt-get install -y apt-transport-https
 
 - name: install dbus


richardlau (approved):

richardlau:
I don't suppose anyone knows why `raw` is being used instead of `stat`?

phillipj:
Feels like using built-in `stat` would make it less confusing for me who's not using ansible frequently.

I'd be more than happy to switch `raw -> stat` if no one can think of good reasons to keep `raw` around.

rvagg:
you could try using `stat`, but I think the idea here is that these bootstrap tasks are done early and with as few assumptions about what the base system can support as possible, `raw` being the safest choice. But since this is `debian8` specific then you could try using `stat` and if it works on the most basic fresh system you have then that'd be :thumbsup:.

phillipj:
Ahhh thanks for the elaborate answer @rvagg 👍🏻 I surely wouldn't have guessed that was the reason for `raw`..

I tried `stat` on a fresh debian8 installation and it worked like a charm. So I amended & force pushed the use of that instead of keeping `raw` combined with forcibly running in check mode.

richardlau (approved):

PR #2500, opened 2020-12-13
chore: Add Shellcheck CI action

nschonni:
Setup the CI to see if this makes sense to add here before trying to fix any of the things it's flagging.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -19,3 +19,5 @@ jobs:
       uses: ludeeus/action-shellcheck@1.0.0
       with:
         check_together: 'yes'
+      env:
+        SHELLCHECK_OPTS: -e SC1004 -e SC1007 -e SC1090 -e SC1091 -e SC2001 -e SC2002 -e SC2006 -e SC2010 -e SC2012 -e SC2034 -e SC2035 -e SC2046 -e SC2086 -e SC2091 -e SC2115 -e SC2129 -e SC2148 -e SC2154 -e SC2155 -e SC2162 -e SC2164 -e SC2166 -e SC2196 -e SC2209 -e SC2219 -e SC2242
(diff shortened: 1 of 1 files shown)


nschonni:
I disabled all the failing checks so they could be resolved later.
Included a commit that ran `shfmt -w .` to address a few auto-fixable things, but could rebase that out to keep this simpler

nschonni:
@richardlau does this make sense?

richardlau:
> 
> 
> @richardlau does this make sense?

I like the idea of linting for consistency but as it stands this looks like too much churn (in terms of the number of files being changed). It's also going to be difficult to update all the places these scripts are actually deployed. For example, for things under `ansible/*-standalone` @rvagg has previously informed me that most of that is a record of what's on the machines but it's mostly manual synced and we have no guarantee it those can actually be deployed via ansible. Other stuff under `ansible` would require the CI hosts to be reansibled which is going to take a lot of someone's time. 

(I'm also not a fan of replacing spaces with tabs but if there's agreement from the other @nodejs/build members I'm willing to cede.)

nschonni:
> I like the idea of linting for consistency but as it stands this looks like too much churn (in terms of the number of files being changed).

It could possibly be batched by folder before adding the CI




> (I'm also not a fan of replacing spaces with tabs but if there's agreement from the other @nodejs/build members I'm willing to cede.)

No preference from me, but it looks like it can be set https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd#L59-L61, just the default is tabs

mhdawson:
I favor spaces instead of tabs and also share the concern about changing things that may be a lot of effort to test/validate.

In other repos when we've turned on linting there has been an option to only lint on code which has been changed. Is that an option?

nschonni:
Ok, rebased out the `shfmt` commits and left it with all failing rules disabled

Trott (approved):

mhdawson (approved):
LGTM, given that it won't fail on any existing files I'm willing to give it a try.

richardlau (approved):

Issue #2491, opened 2020-12-03
z/OS changes for https://github.com/libuv/libuv/pull/3060/


richardlau:
This task is to log required changes to build/test https://github.com/libuv/libuv/pull/3060/

- [ ] Download and build https://github.com/ibmruntimes/zoslib (tbd do this as part of the ansible set up or in the libuv test job)
- [ ] Set `CXX` environment variable. Currently we set `CC` to `/bin/xlc` https://github.com/nodejs/build/blob/c6d76d0bdf38b0f5997c002fa72bc21a7ee02ca6/ansible/roles/jenkins-worker/templates/zos_start.j2#L18. @zsw007 has informed me that IBM have been using `xlclang`/`xlclang++` which I believe we have on our z/OS CI hosts.

richardlau:
I've made a temporary clone of https://ci.nodejs.org/view/libuv/job/libuv-test-commit-zos-cmake as https://ci.nodejs.org/view/libuv/job/libuv-test-commit-zos-cmake-richardlau/ to test out changes. For now I'm limiting the changes to the Jenkins job.

## Cloning issue

The first issue I've run into is cloning https://github.com/ibmruntimes/zoslib. Attempting to clone with the https URL results in an SSL verification error:
e.g. https://ci.nodejs.org/view/libuv/job/libuv-test-commit-zos-cmake-richardlau/5/BUILDTYPE=Debug,nodes=zos-s390x/console
```
+ /u/iojs/git-wrapper clone https://github.com/ibmruntimes/zoslib --depth 1
Cloning into 'zoslib'...
fatal: unable to access 'https://github.com/ibmruntimes/zoslib/': SSL certificate problem: unable to get local issuer certificate
```
there's a note about git on z/OS not supporting https URLs in https://github.com/nodejs/build/blob/ae1841834dfd8d941f356b38f187c07e39131642/ansible/roles/jenkins-worker/tasks/main.yml#L328-L330 
but I'm wondering if we can supply certificates like we used to do on AIX 6.1: https://github.com/nodejs/build/blob/4671676f9885e8e58b9e676025a40a37d0f9cfbd/ansible/aix61-standalone/resources/S20jenkins#L19-L20
for now I've passed `-c http.sslVerify=false` to the clone command to temporarily skip SSL verification. 

FWIW https://github.com/ibmruntimes/zoslib says to do:
```
$ git clone git@github.com:ibmruntimes/zoslib.git zoslib
```
[...]

PR #2505, opened 2020-12-24
ansible: z/OS updates

richardlau:
Refs: https://github.com/nodejs/build/issues/2491

- **ansible: install tap2junit on z/OS**
- **ansible: provide root certificates for z/OS**
The versions of git and Python on z/OS have outdated root certificates
that result in downloads over SSL failing. Provide more recent root
certificates from https://curl.se/docs/caextract.html so we do not have
to disable SSL certification verification.

These have been deployed to `test-marist-zos24-s390x-1` and `test-marist-zos24-s390x-2`. 
Test build: https://ci.nodejs.org/view/libuv/job/libuv-test-commit-zos-cmake-richardlau/47/


mhdawson (approved):
LGTM

AshCripps (approved):

PR #2519, opened 2021-01-08
ansible: remove reference to old AIX71 machines

AshCripps:
Our DOU expired at the end of 2020 so these machines have now been
removed. The machines have already been replaced.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -35,7 +35,6 @@ hosts:
         centos7-x64-1: {ip: 138.68.12.105}
 
     - ibm:
-        aix71-ppc64_be-1: {ip: 129.33.196.199, user: b9s010a}
         aix71-ppc64_be-2: {ip: 169.48.19.173}
         rhel7-s390x-1: {ip: 148.100.86.101, user: linux1}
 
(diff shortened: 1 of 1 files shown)


richardlau (approved):

ansible/inventory.yml
```diff
@@ -35,7 +35,6 @@ hosts:
         centos7-x64-1: {ip: 138.68.12.105}
 
     - ibm:
-        aix71-ppc64_be-1: {ip: 129.33.196.199, user: b9s010a}
         aix71-ppc64_be-2: {ip: 169.48.19.173}
```
nschonni, line 38: Should this be remove if you're removing the reference below

nschonni: Although with a different IP...

richardlau: No, these are release machines and the ones below are separate test machines.

mhdawson: The second one needs to stay.  Its different because it is under the `release` tree as opposed to the `test` tree which is the case for the ones lower down.

mhdawson (approved):
LGTM

Issue #2534, opened 2021-02-03
AIX libuv build fails with "Member libiconv.so.2 is not found in archive"


richardlau:
https://ci.nodejs.org/view/libuv/job/libuv-test-commit-aix-cmake/301/nodes=aix71-ppc64/console
```
    Building C object CMakeFiles/cmTC_2220e.dir/testCCompiler.c.o
    /opt/freeware/bin/gcc    -o CMakeFiles/cmTC_2220e.dir/testCCompiler.c.o   -c /home/iojs/build/workspace/libuv-test-commit-aix-cmake/nodes/aix71-ppc64/out/cmake/CMakeFiles/CMakeTmp/testCCompiler.c
    Could not load program /opt/freeware/bin/gcc:
    	Dependent module /usr/lib/libiconv.a(libiconv.so.2) could not be loaded.
    	Member libiconv.so.2 is not found in archive 
    gmake[1]: *** [CMakeFiles/cmTC_2220e.dir/build.make:66: CMakeFiles/cmTC_2220e.dir/testCCompiler.c.o] Error 255
    gmake[1]: Leaving directory '/home/iojs/build/workspace/libuv-test-commit-aix-cmake/nodes/aix71-ppc64/out/cmake/CMakeFiles/CMakeTmp'
    gmake: *** [Makefile:121: cmTC_2220e/fast] Error 2
```

Most probably related to running the ansible playbook against the AIX hosts yesterday (https://github.com/nodejs/build/issues/2531#issuecomment-771661131).

richardlau:
[`node-test-commit-aix`](https://ci.nodejs.org/job/node-test-commit-aix/) builds are passing.

[`citgm-smoker`](https://ci.nodejs.org/job/citgm-smoker/) is failing:
e.g. https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2611/nodes=aix71-ppc64/testReport/junit/(root)/citgm/bufferutil_v4_0_3/
```
 gmake: Entering directory '/home/iojs/tmp/citgm_tmp/a5b373ee-2e18-4683-9af5-0ba24d1c83a3/bufferutil/build'
   CC(target) Release/obj.target/bufferutil/src/bufferutil.o
 gmake: Leaving directory '/home/iojs/tmp/citgm_tmp/a5b373ee-2e18-4683-9af5-0ba24d1c83a3/bufferutil/build'
[...]

PR #2535, opened 2021-02-03
jenkins: unset LIBPATH on AIX 7.1

richardlau:
It appears that either Jenkins and/or the JDK on AIX is setting the
`LIBPATH` environment variable which is breaking `gcc`/`g++`. Some
jobs already `unset LIBPATH` -- now do this consistently in the
`select-compiler.sh` script.

This should allow us to avoid having to do the rather horrible manual 
hack in https://github.com/nodejs/build/blob/master/ansible/MANUAL_STEPS.md#fix-missing-shared-objects

Refs: https://github.com/nodejs/build/issues/2534

cc @nodejs/platform-aix


richardlau:
@AshCripps FWIW I haven't touched the AIX72 bit because AFAIK we're not currently running any jobs on AIX 7.2 that use `gcc`/`g++`.

AshCripps:
> @AshCripps FWIW I haven't touched the AIX72 bit because AFAIK we're not currently running any jobs on AIX 7.2 that use `gcc`/`g++`.

We find out once I add the new machines if we get any failures

AshCripps (approved):

Issue #2547, opened 2021-02-19
Nightly build of AIX broken


mhdawson:
From the Node-api tests which use the nightly downloads. Seems to have been broken since Feb 13.

@gibfahn do you know what is going on there?

https://ci.nodejs.org/job/node-test-node-addon-api-new/nodes=aix71-ppc64/3609/console
```
+ [ AIXPPC = S390X ]
+ [ AIXPPC = IBMI72 ]
+ [ AIXPPC = AIXPPC ]
+ echo Setting compiler for Node version 16 on AIX7.1
Setting compiler for Node version 16 on AIX7.1
+ [ 16 -gt 9 ]
+ export PATH=/opt/ccache-3.7.4/libexec:/opt/freeware/gcc6/bin:/opt/freeware/bin:/home/iojs/build/workspace/node-test-node-addon-api-new/nodes/aix71-ppc64/node-v16.0.0-nightly2021021388d9268d08-aix-ppc64/bin:/usr/bin:/etc:/usr/sbin:/usr/ucb:/home/iojs/bin:/usr/bin/X11:/sbin:.
+ export CC=gcc CXX=g++ CXX_host=g++
+ unset LIBPATH
+ echo Compiler set to 6.3
Compiler set to 6.3
+ return
+ node -v
terminate called after throwing an instance of 'std::system_error'
  what():  Invalid argument
/tmp/jenkins6443153357633553911.sh[58]: 6553704 Abort
Build step 'Conditional steps (multiple)' marked build as failure
Sending e-mails to: michael_dawson@ca.ibm.com gabriel.schulhof@intel.com Arunesh.Chandra@microsoft.com
Collecting metadata...
Metadata collection done.
Notifying upstream projects of job completion
Finished: FAILURE
```

richardlau:
I suspect that's when we switched to building the releases on AIX 7.2.
cc @AshCripps

AshCripps:
yeah thats likely - we switched to building on 7.2 for master which means it wont run on 7.1 anymore

mhdawson:
@AshCripps 

Well need a section similar to this in the job

```since the groovy selector script does not quite end up doing what we need.
if [ $OS == "sunos" ]; then
  SMARTOS_VERSION=`grep 'release' /etc/pkgsrc_version | awk '{print \$2}'`
   if [[ "${SMARTOS_VERSION:0:4}" = "2016" && ${NODEJS_MAJOR_VERSION} = "10" ]]; then
[...]

PR #2555, opened 2021-03-02
jenkins: update select-compiler to GCC8 for AIX on 16.x

AshCripps:


richardlau (approved):

jenkins/scripts/select-compiler.sh
```diff
@@ -99,12 +99,10 @@ elif [ "$SELECT_ARCH" = "AIXPPC" ]; then
...
     *aix72* )
       echo "Setting compiler for Node version $NODEJS_MAJOR_VERSION on AIX7.2"
-
       if [ "$NODEJS_MAJOR_VERSION" -gt "9" ]; then
-        export LIBPATH=/opt/gcc-6.3/lib/gcc/powerpc-ibm-aix7.2.0.0/6.3.0/pthread/ppc64:/opt/gcc-6.3/lib
-        export PATH="/opt/ccache-3.7.4/libexec:/opt/gcc-6.3/bin:/opt/freeware/bin:$PATH"
+        export PATH="/opt/ccache-3.7.4/libexec:/opt/freeware/bin:$PATH"
         export CC="gcc" CXX="g++" CXX_host="g++"
-        echo "Compiler set to 6.3"
+        echo "Compiler set to 8.3"
```
mhdawson, line 105: Is this intended? It means that if we start using 7.2 for older builds it will use the newer compiler level which I'm not sure would be intended.

AshCripps: yes it is turns out the 7.2 machines only have 8.3 installed and have been building with it since I switched them on. However saying that im not sure if building with 8.3 would break people trying to run with 6.3? 

Thoughts @richardlau ?

richardlau: We've already established that binaries built on AIX 7.2 don't run on AIX 7.1 (https://github.com/nodejs/build/issues/2547#issuecomment-782107370) -- is not building with gcc 6 the issue there?

I've always been a bit hesitant building with gcc 6 on AIX 7.2 as we have to go out of our way to get it on there as the default gcc installation on AIX 7.2 via the AIX toolbox is 8.

AshCripps: the 7.2 vs 7.1 is not GCC related thats an pure AIX issue, but if the older lines say we build with 6.3 on 7.2 I reckon we should at least test they run on a system with GCC6 when built with GCC8 - I cant see why they wouldnt work

AshCripps: turns out gcc6 is installed so ive added back the check for gcc6

mhdawson (approved):
LGTM

richardlau:
I don't object to keeping them seperate, but aren't the aix71 and aix72 sections the same now?

AshCripps:
> I don't object to keeping them seperate, but aren't the aix71 and aix72 sections the same now?

Not quite the AIX7.1 gcc6 section points to a different path

richardlau (approved):

AshCripps:
ive got AIX7.2 now compiling with GCC16 on 15.x and below - https://ci.nodejs.org/job/node-test-commit-aix-AshCripps/nodes=aix72-ppc64/30/

This can now do with rereviews - @richardlau @mhdawson 

Ive also updated the `node-test-commit-aix` job to set the path correctly to find `tap2junit` on aix7.2

richardlau (approved):

Issue #2507, opened 2020-12-30
Install Python 3 everywhere in CI


targos:
This is a requirement for https://github.com/nodejs/node/pull/36691
There are a lot of hosts that only have Python 2 installed.

For example, in `ci-release`, only the Windows builds work if we remove Python 2 support: https://ci-release.nodejs.org/job/iojs+release/6555/

cclauss:
I do not have access rights for the link above...  How many hosts are we talking about that only have legacy Python?

AshCripps:
It is a fair few of the host - interestingly the mac builds failed in that release build and they most certainty do have python3 installed.

cclauss:
Python 2 died over a year ago (not even security updates) so it is a security exposure to keep those hosts online.

Can you please paste some details on how Py3 on macOS is failing?

targos:
<details>

```
15:41:54 PRODUCTSIGN_CERT=Developer ID Installer: Node.js Foundation
15:41:54 CODESIGN_CERT=Developer ID Application: Node.js Foundation
15:41:54 NOTARIZATION_ID=build@iojs.org
15:41:54 PACKAGEMAKER=~/PackageMaker.app/Contents/MacOS/PackageMaker
15:41:54 
15:41:54 [EnvInject] - Variables injected successfully.
15:41:54 [ws] $ /bin/bash -ex /var/folders/zq/929cmt8n33j9_8x7b06pbzqh0000gp/T/jenkins490912561010356266.sh
15:41:54 + RELEASE_URLBASE=https://nodejs.org/download/test/
15:41:54 + '[' Xtest == Xtest ']'
15:41:54 + disttype=custom
15:41:54 + CUSTOMTAG=test20201230f8775b51b5
15:41:54 + RELEASE_URLBASE=https://nodejs.org/download/test/
15:41:54 + CONFIG_FLAGS='--download-path=/Users/iojs/node-icu/ --verbose'
15:41:54 + '[' -z x ']'
15:41:54 + mkdir -p /Users/iojs/node-icu/
15:41:54 + make -j 2 pkg-upload DESTCPU=x64 ARCH=x64 DISTTYPE=custom DATESTRING=20201230 COMMIT=f8775b51b5 CUSTOMTAG=test20201230f8775b51b5 RELEASE_URLBASE=https://nodejs.org/download/test/ 'CONFIG_FLAGS=--download-path=/Users/iojs/node-icu/ --verbose'
15:41:59 rm -f -r out/macos
[...]

PR #2564, opened 2021-03-10
ansible: add python3 to AIX hosts

richardlau:
I've run this on all of our test/release AIX 7.1 and 7.2 hosts. It installs Python 3 (3.7.9 at this time) into `/opt/freeware/bin`, e.g. `/opt/freeware/bin/python3`. Crucially it leaves the existing Python 2 installation alone (so `python` still runs Python 2). 

On the AIX 7.2 machines there is already a Python 3.6.12 installation in `/opt/bin/python3`. We end up with three versions of Python installed (2.7.18, 3.6.12 and 3.7.9) but they're in different directories and the python3 binary we pick up depends on the path. If we've run through select-compiler.sh and prepend `/opt/freeware/bin` to the path `python3` will be 3.7.9. https://github.com/nodejs/build/blob/f95880c7e40a1d547afdbfeb9376c476bd6cd5e7/jenkins/scripts/select-compiler.sh#L103

I haven't found a CI job that will actually use Python 3 yet as they either directly call `python configure ...` (i.e. invoke `python` which is Python 2) or they run `make run-ci` which defaults the `PYTHON` make variable to `python` (again Python 2). We'll probably need to directly run `configure ...` (as there is Python autodetection/selection logic in the `configure` script) and/or set the default `PYTHON` in the makefile to `python3` in master/16.x onwards.

cc @nodejs/platform-aix

### DIFF
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -54,7 +54,7 @@ packages: {
   ],
 
   aix: [
-    'bash,cmake,curl,gcc-c++,gcc6-c++,tar,unzip,git,make,sudo,python-setuptools',
+    'bash,cmake,curl,gcc-c++,gcc6-c++,tar,unzip,git,make,sudo,python-setuptools,python3',
   ],
 
   ibmi: [


mhdawson (approved):
LGTM

AshCripps (approved):

richardlau:
Refs: https://github.com/nodejs/build/issues/2507

Issue #2552, opened 2021-02-28
Replacing machines from Joyent data center


mhdawson:
We found out late Friday that the machines in the Joyent data center needed to be moved by next week. These have been moved: 

ip | hostname | changes | halted old?
-- | -- | -- | --
147.75.88.51 | test-joyent-ubuntu1804-x64-1 | updated firewall on ci | yes
147.75.88.54 | test-joyent-smartos18-x64-3 | updated firewall on ci | yes
147.75.88.55 | test-joyent-smartos18-x64-4 | updated firewall on ci | yes
147.75.88.53 | release-joyent-smartos18-x64-2 | updated firewall on ci-release | yes
147.75.88.56 | release-joyent-ubuntu1804_docker-x64-1 | updated firewall on ci-release | yes
147.75.88.58 | test-joyent-ubuntu1604_arm_cross-x64-1 | updated firewall on ci | yes
147.75.88.52 | release-joyent-ubuntu1604_arm_cross-x64-1 | updated firewall on ci-release | yes
147.75.88.59 | test-joyent-smartos17-x64-3 | updated firewall on ci | yes
147.75.88.60 | release-joyent-smartos17-x64-2 | updated firewall on release-ci | yes
147.75.88.61 | test-joyent-smartos17-x64-4 | updated firewall on ci | yes
147.75.88.57 | infra-joyent-ubuntu1604-x64-1-new/unencrypted | updated cloudflare | yes
147.75.88.62 | infra-joyent-debian10-x64-1/grafana | updated cloudflare, updated telegraf configs on many machines | yes
139.178.83.227 | infra-joyent-smartos15-x64-1-new | sniff test that looks ok | yes

The smartos machines below 17 will not be moved over.

We'll need to:
- [x] Update the ansible inventory file with the new IPs
- [X] remove smartos16 from node-test-commit-smartos in test ci
- [X] remove smartos 16 from node-test-addon-api-new
- [x] remove smartos 15, 16 from the libuv tests - FYI to libuv team asking that they do that - https://github.com/libuv/libuv/issues/3121
- [x]  remove the smartos 15,16 machines from the CIs and inventory
- [ ] figure out how we can reboot/manage the new machines (ie is there a management system we'll have access to)
- [x] See if there is a good way to test the new release machines in before the next 10, 12 releases (which are the only 
       ones that we current release on for smartos)

jbergstroem:
[...]

PR #2568, opened 2021-03-12
ansible: update Joyent hosted IP addresses

richardlau:
Refs: https://github.com/nodejs/build/issues/2552

cc @mhdawson @jbergstroem 

Some oddities:
- ~I can't find a replacement for `test-joyent-ubuntu1804_docker-x64-1` in the table in https://github.com/nodejs/build/issues/2552. It also looks like all the hosts in https://ci.nodejs.org/search/?q=test-joyent that are containers (i.e. have the word "container" in their node names) are offline.~ Removed in this PR.
- It looks like `test-joyent-ubuntu1604_arm_cross-x64-1` and `release-joyent-ubuntu1604_arm_cross-x64-1` were removed from the inventory in https://github.com/nodejs/build/pull/2290/files#diff-6faf346b88ef5cbd5f6733c35ef37f0884f3fcc946a55c5b811843b652421f6d and possibly did not need to be migrated as we now cross compile in docker containers. The build history for https://ci.nodejs.org/computer/test-joyent-ubuntu1604_arm_cross-x64-1/builds is empty (i.e. no recent builds).
- ~`infra-joyent-debian10-x64-1/grafana` (from the table) doesn't appear to match anything in the inventory.~ Added in this PR.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -17,6 +17,7 @@ hosts:
         ubuntu1804-x64-3: {ip: 157.245.7.159, alias: metrics}
 
     - joyent:
+        debian10-x64-1: {ip: 147.75.88.62, alias: grafana}
         smartos15-x64-1: {ip: 139.178.83.227, alias: backup}
         ubuntu1604-x64-1: {ip: 147.75.88.57, alias: unencrypted}
 
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -17,8 +17,8 @@ hosts:
         ubuntu1804-x64-3: {ip: 157.245.7.159, alias: metrics}
 
     - joyent:
-        smartos15-x64-1: {ip: 165.225.151.21, alias: backup}
-        ubuntu1604-x64-1: {ip: 165.225.149.55, alias: unencrypted}
+        smartos15-x64-1: {ip: 139.178.83.227, alias: backup}
+        ubuntu1604-x64-1: {ip: 147.75.88.57, alias: unencrypted}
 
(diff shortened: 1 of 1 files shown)


jbergstroem:
> infra-joyent-debian10-x64-1/grafana (from the table) doesn't appear to match anything in the inventory.

It was never added to inventory; perhaps add it as part of the list?

richardlau:
> > infra-joyent-debian10-x64-1/grafana (from the table) doesn't appear to match anything in the inventory.
> 
> It was never added to inventory; perhaps add it as part of the list?

done.

mhdawson:
In terms of test-joyent-ubuntu1604_arm_cross-x64-1 and release-joyent-ubuntu1604_arm_cross-x64-1 I noticed that after the fact. I think we should just confirm with @rvagg and then spin those down.

For `test-joyent-ubuntu1804_docker-x64-1` we may have missed that since the machine do not show up in the CI (instead the container ones do.  Luckily we have a number of equivalent machines so nothing is blocked but we might have lower capacity than we want for the docker based testing. At this point I think we should look to add an additional docker host on one of the other infra's versus trying to bring that one back.

mhdawson (approved):
LGTM

richardlau:
I've removed `test-joyent-ubuntu1804_docker-x64-1` from the inventory -- the old host is gone and we don't currently have a new one. I'll open a separate issue(s) to track removing the associated containers from Jenkins (we still have the equivalents at SoftLayer and DigitalOcean) and whether we want to replace it.

Issue #2533, opened 2021-01-29
"baselayout : run ccache installer" fails running ansible against LinuxONE/smartos


richardlau:
Running the create worker ansible playbook against our LinuxONE hosts errors on the `baselayout : run ccache installer` task:
```
TASK [baselayout : run ccache installer] *******************************************************************************************************************
fatal: [test-ibm-rhel7-s390x-1]: FAILED! => {"msg": "The conditional check 'os == \"rhel7\" and arch == \"s390x\" and not has_ccache.stat.exists' failed. The error was: error while evaluating conditional (os == \"rhel7\" and arch == \"s390x\" and not has_ccache.stat.exists): 'dict object' has no attribute 'stat'\n\nThe error appears to be in '/home/rlau/sandbox/github/build/ansible/roles/baselayout/tasks/main.yml': line 202, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: run ccache installer\n  ^ here\n"}
fatal: [test-ibm-rhel7-s390x-2]: FAILED! => {"msg": "The conditional check 'os == \"rhel7\" and arch == \"s390x\" and not has_ccache.stat.exists' failed. The error was: error while evaluating conditional (os == \"rhel7\" and arch == \"s390x\" and not has_ccache.stat.exists): 'dict object' has no attribute 'stat'\n\nThe error appears to be in '/home/rlau/sandbox/github/build/ansible/roles/baselayout/tasks/main.yml': line 202, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: run ccache installer\n  ^ here\n"}
fatal: [test-ibm-rhel7-s390x-3]: FAILED! => {"msg": "The conditional check 'os == \"rhel7\" and arch == \"s390x\" and not has_ccache.stat.exists' failed. The error was: error while evaluating conditional (os == \"rhel7\" and arch == \"s390x\" and not has_ccache.stat.exists): 'dict object' has no attribute 'stat'\n\nThe error appears to be in '/home/rlau/sandbox/github/build/ansible/roles/baselayout/tasks/main.yml': line 202, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: run ccache installer\n  ^ here\n"}
[...]

PR #2569, opened 2021-03-12
ansible: fix scripts for LinuxONE and SmartOS

richardlau:
Move non-package manager ccache tasks into partial tasks to
avoid the different platforms interfering which each other.

Fixes: https://github.com/nodejs/build/issues/2533

This refactors by replacing `ansible/roles/baselayout/tasks/ccache.yml` and
platform specific ccache tasks in `ansible/roles/baselayout/tasks/main.yml`
with platform specific versions under `ansible/roles/baselayout/tasks/partials/ccache`.


richardlau:
One thing I've not changed is the version of ccache being installed, which on AIX, LinuxONE and SmartOS is not the latest (3.7.2 on AIX and 3.6 on LinuxONE and SmartOS).

The replaced `ansible/roles/baselayout/tasks/ccache.yml` was downloading ccache source from https://www.samba.org/ftp/ccache/ but ccache stopped uploading to there after ccache 3.6: https://lists.samba.org/archive/ccache/2019q2/001580.html

I'd prefer to look at that separately in a follow up PR if necessary.

AshCripps (approved):

Issue #2599, opened 2021-03-30
Request to install libcap on Linux CI machines


danbev:
There is currently a draft PR https://github.com/nodejs/node/pull/37727 which if accepted will add a build time dependency for Linux for `libcap`. This issue would like to request libcap be installed on the CI machines that use Linux.

The pull request includes updates to the github workflows and used the following command:
```console
$ sudo apt-get install -y libcap-dev
```

rvagg:
Is that going to mean that `libcap` will be a runtime dependency for Node now? I'm imagining it's ubiquitous by now, but maybe not, do you know?

mhdawson:
@rvagg I think @danbev linked to it statically, through:

```shell
      [ 'OS in "linux"', {
        'ldflags': [ '-Wl,-Bstatic -Wl,--whole-archive -lcap -Wl,--no-whole-archive -Wl,-Bdynamic' ],
      }],
```

but I was wondering a bit about that.

rvagg:
OK, interesting. A test release build will also be interesting from a size perspective! But we'd have to get libcap on there too.

@danbev this is going to require a bit of ansible work unfortunately. Probably narrowing down the hosts where we run 16+ (or whatever base this is expected to flow back to) and updating the list of installed packages to include libcap-dev or its alternative name on the various platforms. Would you mind having a go at that? I can't tell you off the top of my head which file(s) to look at but I think there's a variables file in baselayout that might be the best place to start, you'll find it split up across OS+version.

richardlau:
We'd probably need to do the containers too.

danbev:
> Would you mind having a go at that? 

Yeah I'll give it a shot, thanks.

Issue #2531, opened 2021-01-28
Errors running Ansible across current cluster


rvagg:
I decided to just run our ansible worker create script across our cluster today in light of the `sudo` security flaw that was announced. In the process I've encountered a bunch of errors that we should probably look in to. Recording here as a list so we can tick them off. I'd appreciate help in dealing with these.

I didn't run any updates on:

 * Windows - shouldn't need it, I think they all do auto updates and regular restarts anyway
 * IBM platforms (including ibmi, aix, rhel-s390x, zos) I'll let @nodejs/build IBM folks deal with that
 * macOS - @AshCripps would you mind doing these? I'm a little afraid to just run the scripts against the existing infra.

Errors:

## Failed 'not secret' (Jenkins secret not in secrets/inventory.yml)

 - [x] test-digitalocean-debian9-x64-1
 - [x] ~test-digitalocean-centos5-x86-1~ Removed
 - [x] ~test-softlayer-centos5-x64-2~ Removed
 - [x] ~test-softlayer-centos5-x64-1~ Removed
 - [x] ~release-digitalocean-centos5-x64-1~ Removed
 - [x] ~release-digitalocean-centos6-x86-1~ Removed https://github.com/nodejs/build/pull/2886
 - [x] ~release-softlayer-centos5-x86-1~ Removed
 - [x] test-digitalocean-ubuntu1404-x64-1
 - [x] ~test-digitalocean-ubuntu1404-x86-1~ Removed https://github.com/nodejs/build/pull/2886
 - [x] test-digitalocean-ubuntu1604-x86-2
 - [x] test-digitalocean-ubuntu1804-x64-1
 - [x] ~test-rackspace-ubuntu1204-x64-1 (should remove)~
 - [x] ~test-requireio_rvagg-ubuntu1404-arm64_odroidxu-1 (should remove)~
 - [x] ~test-requireio_rvagg-ubuntu1404-arm64_odroidxu-2 (should remove)~
 - [x] ~test-requireio_rvagg-ubuntu1404-arm64_odroidxu3-1 (should remove)~
 - [x] ~release-scaleway-ubuntu1604-armv7l-1 (should remove)~
 - [x] ~release-scaleway-ubuntu1604-armv7l-2 (should remove)~

## Unreachable

 - [x] ~test-joyent-smartos18-x64-1~ Removed
 - [x] ~test-joyent-smartos18-x64-2~ Removed
[...]

PR #2607, opened 2021-04-06
ansible: add libcap-dev package to Linux distros

danbev:
This commit adds the libcap-dev package to linux distributions to
support the usage of Linux capabilities in Node.js.

Fixes: https://github.com/nodejs/build/issues/2599
Refs: https://github.com/nodejs/node/pull/37727


ansible/roles/baselayout/vars/main.yml
```diff
@@ -35,17 +35,17 @@ common_packages: [
 packages: {
   centos6_x64: ['centos-release-scl'], # only available on x86_64, 32-bit is from https://copr.fedorainfracloud.org/coprs/mlampe
   centos6: [
-    'ccache,git2u,gcc-c++,devtoolset-6,sudo', # even need gcc on centos6 so ccache has symlinks
+    'ccache,git2u,gcc-c++,devtoolset-6,sudo,libcap-devel', # even need gcc on centos6 so ccache has symlinks
```
richardlau, line 38: We do not build `master`/16.x on centos 6 so we don't need to install the new package there.
https://github.com/nodejs/build/blob/371bb7626abe773a56c7c62e76d0a551ba1b9717/jenkins/scripts/VersionSelectorScript.groovy#L22-L23

ansible/roles/baselayout/vars/main.yml
```diff
@@ -57,27 +57,27 @@ packages: {
   ],
 
   debian7: [
-    'gcc-4.8,g++-4.8,sudo',
+    'gcc-4.8,g++-4.8,sudo,libcap-dev',
```
richardlau, line 60: AFAIK we do not have debian7 in the CI anymore and we can get rid of the `debian7` section (perhaps in a separate PR/commit).

ansible/roles/baselayout/vars/main.yml
```diff
@@ -146,10 +146,10 @@ packages: {
   ],
 
   ubuntu: [
-    'ccache,g++,gcc,g++-6,gcc-6,git,libfontconfig1,sudo,python3-pip',
+    'ccache,g++,gcc,g++-6,gcc-6,git,libfontconfig1,sudo,python3-pip,libcap-dev',
   ],
 
   ubuntu1404: [
-    'ntp,gcc-8,g++-8,gcc-6,g++-6,g++-4.8,gcc-4.8,g++-4.9,gcc-4.9,binutils-2.26',
+    'ntp,gcc-8,g++-8,gcc-6,g++-6,g++-4.8,gcc-4.8,g++-4.9,gcc-4.9,binutils-2.26,libcap-dev',
```
richardlau, line 153: We don't build `master`/16.x on Ubuntu 14.04 so we do not need to install the new package there.
https://github.com/nodejs/build/blob/371bb7626abe773a56c7c62e76d0a551ba1b9717/jenkins/scripts/VersionSelectorScript.groovy#L33-L34

ansible/roles/baselayout/vars/main.yml
danbev: Thanks, and sorry about being lazy. I'll update this and the others shortly.

ansible/roles/baselayout/vars/main.yml
danbev: I've opened #2608 for this.

richardlau:
I've updated this on top of the recently landed PR's that touch the ansible scripts have applied it to most of the non-containered Linux systems with the exception of hosts that we currently have issues with: https://github.com/nodejs/build/issues/2531

ansible/roles/docker/templates/ubi81.Dockerfile.j2
```diff
@@ -25,6 +25,7 @@ RUN dnf install --disableplugin=subscription-manager -y \
       python3 \
       openssl-devel \
       procps-ng \
+      libcap-devel \
```
richardlau, line 28: @danbev This is failing with:
```
No match for argument: libcap-devel
Error: Unable to find a match: libcap-devel
```

On one of the UBI 8 containers:
```
bash-4.4$ dnf search libcap
2021-04-12 09:39:17,189 [ERROR] dnf:22718:MainThread @logutil.py:194 - [Errno 13] Permission denied: '/var/log/rhsm/rhsm.log' - Further logging output will be written to stderr
Not root, Subscription Management repositories not updated
Extra Packages for Enterprise Linux Modular 8 - x86_64                                                                      510 kB/s | 557 kB     00:01
Extra Packages for Enterprise Linux 8 - x86_64                                                                               11 MB/s | 9.2 MB     00:00
Red Hat Universal Base Image 8 (RPMs) - BaseOS                                                                              2.7 MB/s | 778 kB     00:00
Red Hat Universal Base Image 8 (RPMs) - AppStream                                                                            18 MB/s | 5.2 MB     00:00
Red Hat Universal Base Image 8 (RPMs) - CodeReady Builder                                                                    47 kB/s |  15 kB     00:00
=============================================================== Name Exactly Matched: libcap ===============================================================
libcap.i686 : Library for getting and setting POSIX.1e capabilities
libcap.x86_64 : Library for getting and setting POSIX.1e capabilities
============================================================== Name & Summary Matched: libcap ==============================================================
libcap-ng-devel.i686 : Header files for libcap-ng library
libcap-ng-devel.x86_64 : Header files for libcap-ng library
=================================================================== Name Matched: libcap ===================================================================
libcap-ng.x86_64 : An alternate posix capabilities library
libcap-ng.i686 : An alternate posix capabilities library
bash-4.4$
```
so it looks like `libcap-devel` isn't available on UBI.

danbev: Hmm, from what I understand it is available but only if we register using the subscription manager and then find the pool-id of the repo, and then attach to it. After that it would be available, but that feels like something that we don't want to have to do. I'm going to try to figure out if there are other options.

danbev: Would compiling this be an option?
```console
$ dnf install -y git gcc make glibc-static
$ git clone https://github.com/mhiramat/libcap
$ cd libcap
$ make && make install
```
Which seems to compile and link properly:
```console
$ gcc -o cap -lcap -xc - <<HERE                                                      
> #include <sys/capability.h>                                                        
> #include <stdio.h>                                                                 
> int main(int argc, char** argv) {                                                  
>   printf("bajja\n");                                                               
>   return 0;                                                                        
> }                                                                                  
> HERE
$ ldd cap 
	linux-vdso.so.1 (0x00007ffff7ffa000)
	libcap.so.2 => /lib64/libcap.so.2 (0x00007ffff7bce000)
	libc.so.6 => /lib64/libc.so.6 (0x00007ffff780b000)
	/lib64/ld-linux-x86-64.so.2 (0x00007ffff7dd4000)

```

richardlau: @danbev We can give it a try 🤞

ansible/roles/docker/templates/ubi81.Dockerfile.j2
```diff
@@ -25,7 +25,10 @@ RUN dnf install --disableplugin=subscription-manager -y \
       python3 \
       openssl-devel \
       procps-ng \
+      glibc-static \
     && dnf --disableplugin=subscription-manager clean all
+RUN git clone https://github.com/mhiramat/libcap && cd libcap && make \
+    make install
```
richardlau, line 31:
```suggestion
RUN git clone https://github.com/mhiramat/libcap && cd libcap \
    && make install
```

richardlau:
This has been applied to the CI, but a test build of https://github.com/nodejs/node/pull/37727, https://ci.nodejs.org/job/node-test-pull-request/37434/, is failing on the centos, fedora and alpine hosts.

richardlau:
Perhaps at least on Fedora we need `libcap-static`?
```
[root@test-digitalocean-fedora32-x64-1 ~]# dnf list installed libcap*
Installed Packages
libcap.x86_64                                                               2.26-7.fc32                                                            @anaconda
libcap-devel.x86_64                                                         2.26-7.fc32                                                            @fedora
libcap-ng.x86_64
[root@test-digitalocean-fedora32-x64-1 ~]# dnf search libcap
Last metadata expiration check: 2:04:56 ago on Tue Apr 13 13:15:09 2021.
=============================================================== Name Exactly Matched: libcap ===============================================================
libcap.x86_64 : Library for getting and setting POSIX.1e capabilities
libcap.i686 : Library for getting and setting POSIX.1e capabilities
============================================================== Name & Summary Matched: libcap ==============================================================
libcap-devel.x86_64 : Development files for libcap
libcap-devel.i686 : Development files for libcap
libcap-ng-devel.i686 : Header files for libcap-ng library
libcap-ng-devel.x86_64 : Header files for libcap-ng library
libcap-ng-python3.x86_64 : Python3 bindings for libcap-ng library
libcap-static.i686 : Static libraries for libcap development
libcap-static.x86_64 : Static libraries for libcap development
=================================================================== Name Matched: libcap ===================================================================
libcap-ng.x86_64 : An alternate posix capabilities library
libcap-ng.i686 : An alternate posix capabilities library
libcap-ng-utils.x86_64 : Utilities for analyzing and setting file capabilities
[root@test-digitalocean-fedora32-x64-1 ~]#
```

`libcap-static` doesn't appear to be available on centos 7:
```
[root@test-softlayer-centos7-x64-1 ~]# yum list installed libcap*
Failed to set locale, defaulting to C
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * centos-sclo-rh: mirrors.raystedman.org
 * centos-sclo-sclo: mirror.sfo12.us.leaseweb.net
 * epel: sjc.edge.kernel.org
Installed Packages
libcap.x86_64                                                                  2.22-11.el7                                                             @base
libcap-devel.x86_64                                                            2.22-11.el7                                                             @base
libcap-ng.x86_64                                                               0.7.5-4.el7                                                             @base
[root@test-softlayer-centos7-x64-1 ~]# yum search libcap
Failed to set locale, defaulting to C
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * centos-sclo-rh: mirror.fileplanet.com
 * centos-sclo-sclo: mirror.sfo12.us.leaseweb.net
 * epel: sjc.edge.kernel.org
=================================================================== N/S matched: libcap ====================================================================
libcap-devel.i686 : Development files for libcap
libcap-devel.x86_64 : Development files for libcap
libcap-ng-devel.i686 : Header files for libcap-ng library
libcap-ng-devel.x86_64 : Header files for libcap-ng library
libcap-ng-python.x86_64 : Python bindings for libcap-ng library
compat-libcap1.i686 : Library for getting and setting POSIX.1e capabilities
compat-libcap1.x86_64 : Library for getting and setting POSIX.1e capabilities
libcap.i686 : Library for getting and setting POSIX.1e capabilities
libcap.x86_64 : Library for getting and setting POSIX.1e capabilities
libcap-ng.i686 : An alternate posix capabilities library
libcap-ng.x86_64 : An alternate posix capabilities library
libcap-ng-utils.x86_64 : Utilities for analyzing and setting file capabilities

  Name and summary matches only, use "search all" for everything.
[root@test-softlayer-centos7-x64-1 ~]#
```

danbev:
> This has been applied to the CI, but a test build of [nodejs/node#37727](https://github.com/nodejs/node/pull/37727), https://ci.nodejs.org/job/node-test-pull-request/37434/, is failing on the centos, fedora and alpine hosts.

I'll take a closer look at this now. Thanks!

danbev:
> Perhaps at least on Fedora we need `libcap-static`?

Yeah, it actually looks like this is required and I had it installed locally but don't remember installing it. I'll look into what can be done for centos and alpine.

danbev:
@richardlau I've updated https://github.com/nodejs/node/pull/37727 with a suggestion that only needs the header `sys/capability.h` from `libcap-devel`. If I recall the issue with the failures were when linking with `libcap` and I be interested to see if https://github.com/nodejs/node/pull/3772 compiles.

richardlau:
@danbev Could you rebase this and fix any merge conflicts?

danbev:
@richardlau I've updated this PR now :+1:

ansible/roles/docker/templates/ubi81.Dockerfile.j2
richardlau: @danbev We've lost this fixup.

danbev: Not sure how I did that but I've added it back now. Thanks

richardlau:
I've applied this onto the container images and the ubuntu1804 hosts (i.e. the ones failing from https://ci.nodejs.org/job/node-test-pull-request/39198/ due to missing `<sys/capability.h>`). Resumed that failing CI in https://ci.nodejs.org/job/node-test-pull-request/39212/

danbev:
Re-run of failing [node-test-commit-linux](https://ci.nodejs.org/job/node-test-commit-linux/42375/) :heavy_check_mark:

ansible/roles/docker/templates/ubuntu1804.Dockerfile.j2
```diff
@@ -24,7 +24,8 @@ RUN apt-get update && apt-get dist-upgrade -y && apt-get install -y \
     curl \
     python-pip \
     python3-pip \
-    libfontconfig1
+    libfontconfig1 \
+    libcap-dev
```
richardlau, line 28: Since the cross compile job is still failing to find `sys/capability.h` I presume the headers are being picked up from the cross compiler (https://github.com/rvagg/rpi-newer-crosstools) rather than the container.

danbev: Sorry, you've lost me here :) Could you explain what, if anything, we can do get this to work?

richardlau: Sorry, I have no idea how to get it to work 😞. I suspect for cross compiling the system headers can't be used because they will be for the host architecture.

danbev: I have an idea which is to copy the header to the sysroot that the cross compiler is using. I've tried this out locally using a container and it can then compile without complaining about not finding the header. This is an update to the DockerFile and I'm wondering how I could try this out. I'm guessing that just re-running the CI job would not be enough?

richardlau: Yeah, we'd need to rebuild the Docker image and restart the container.
@rvagg @sxa Any concerns over modifying the sysroot for the ARM cross-compiler?

rvagg: uhh, yeah that looks a bit weird, is libcap implemented simply as a header or is there a library behind it too? I think by copying that header into there says that you have a arm-rpi-linux-gnueabihf backing for whatever's in this header, is that true? has this been tested? These are going to be used for armv7l binary release builds, are we requiring users to have libcap installed to use our binaries everywhere now?

danbev: > uhh, yeah that looks a bit weird, is libcap implemented simply as a header or is there a library behind it too?

Yeah, sorry about that. I was not sure how to get the header into the sysroot and this was the only thing I could think of.

We only need the header in this case and we are not linking with libcap, so the requirement for someone building Node.js would be that they have the libcap-dev package installed.

rvagg: I guess if it's not being linked then that's a positive, but what about the contents of capability.h? I've had a quick look and I don't see anything obvious that might result in differences between architectures—is there any risk that we're going to be linking types or sizes that are going to be wrong for the arch we're cross compiling? I guess that's the most important question to answer and maybe it requires some testing of cross compiled binaries hitting the codepaths that are using this header?

danbev: I've taken a look at just using a `syscall` instead of using the function definition in `sys/capabilities.h` and perhaps that would avoid this trouble. I'm running CI tests for this now (https://ci.nodejs.org/job/node-test-pull-request/39333/).

PR #2608, opened 2021-04-06
ansible: remove debian7

danbev:
This is no longer in the CI so it should be alright to remove it.

Refs: https://github.com/nodejs/build/pull/2607

### DIFF
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -4,6 +4,7 @@
 # variables shared in baselayout
 #
 
+git_no_binpkg: [ ]
 git_version: 2.10.2
 
 ssh_config: /etc/ssh/sshd_config
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -4,7 +4,6 @@
 # variables shared in baselayout
 #
 
-git_no_binpkg: ['debian7']
 git_version: 2.10.2
 
 ssh_config: /etc/ssh/sshd_config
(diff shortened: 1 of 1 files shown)


AshCripps (approved):

ansible/roles/baselayout/vars/main.yml
```diff
@@ -4,7 +4,6 @@
 # variables shared in baselayout
 #
 
-git_no_binpkg: ['debian7']
```
richardlau, line 7: We'll need to check that removing `git_no_binpkg` doesn't cause an issue for https://github.com/nodejs/build/blob/f65a7e67e5b46bc735acbc1f4aff617715b2aa16/ansible/roles/baselayout/tasks/main.yml#L60

richardlau (approved):

mhdawson (approved):
LGTM

ansible/roles/baselayout/vars/main.yml
richardlau: Unfortunately removing it completely fails:
```
TASK [baselayout : check for a recent enough git] **********************************************************************************************************
fatal: [test-rackspace-debian8-x64-1]: FAILED! => {"msg": "The conditional check 'os in git_no_binpkg' failed. The error was: error while evaluating conditional (os in git_no_binpkg): 'git_no_binpkg' is undefined\n\nThe error appears to be in '/home/rlau/sandbox/github/build/ansible/roles/baselayout/tasks/main.yml': line 59, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: check for a recent enough git\n  ^ here\n"}
```

ansible/roles/baselayout/vars/main.yml
```diff
@@ -4,7 +4,6 @@
 # variables shared in baselayout
 #
```
richardlau, line 6:
```suggestion

git_no_binpkg: [ ]
```

PR #2613, opened 2021-04-08
ansible: update OpenSSL 3.0 to use quictls/openssl

danbev:
This commit update the version of OpenSSL 3.0 to use the quictls/openssl
fork. This for will be used until upstream OpenSSL includes support for
the QUIC protocol. At that point we can switch back to using upstream
again.

### DIFF
--- a/ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
+++ b/ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
@@ -58,7 +58,8 @@ ENV OPENSSL300DIR /opt/openssl-3.0.0
 
 RUN mkdir -p /tmp/openssl_3.0.0 && \
     cd /tmp/openssl_3.0.0 && \
-    git clone git@github.com:quictls/openssl.git && cd openssl && \
+    git clone https://github.com/quictls/openssl.git -b openssl-3.0.0-alpha14+quic --depth 1 && \
+    cd openssl && \
     ./config --prefix=$OPENSSL300DIR && \
     make -j 6 && \
     make install && \
--- a/ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
+++ b/ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
@@ -58,7 +58,7 @@ ENV OPENSSL300DIR /opt/openssl-3.0.0
 
 RUN mkdir -p /tmp/openssl_3.0.0 && \
     cd /tmp/openssl_3.0.0 && \
-    curl -sL https://www.openssl.org/source/openssl-3.0.0-alpha14.tar.gz | tar zxv --strip=1 && \
+    git clone git@github.com:quictls/openssl.git && cd openssl && \
     ./config --prefix=$OPENSSL300DIR && \
     make -j 6 && \
     make install && \


ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
```diff
@@ -58,7 +58,7 @@ ENV OPENSSL300DIR /opt/openssl-3.0.0
 
 RUN mkdir -p /tmp/openssl_3.0.0 && \
     cd /tmp/openssl_3.0.0 && \
-    curl -sL https://www.openssl.org/source/openssl-3.0.0-alpha13.tar.gz | tar zxv --strip=1 && \
+    curl -sL https://github.com/quictls/openssl/archive/refs/tags/openssl-3.0.0-alpha13.tar.gz | tar zxv --strip=1 && \
```
richardlau, line 61: Does this contain the quic patches?

https://github.com/quictls/openssl/releases/tag/openssl-3.0.0-alpha13 points to https://github.com/quictls/openssl/commit/88df2c0b3d6162971304c06a240deb9320c9ae67 and I don't see any of the quic commits in https://github.com/quictls/openssl/commits/openssl-3.0.0-alpha13 (compared to https://github.com/quictls/openssl/commits/openssl-3.0.0-alpha13+quic).

richardlau:
FWIW the OpenSSL project released 3.0.0-alpha14 today: https://mta.openssl.org/pipermail/openssl-announce/2021-April/000199.html
We could wait a day or two for the fork to catch up.

ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
danbev: I thought they would be included but perhaps they are not. I'll take a closer look. And yes we should hold off a day or tow so that we can update to alpha14 :+1:

danbev: Is it alright to use git clone instead?

richardlau: Go for it 😄

rvagg:
btw I've just updated to beta14 on the containers and there seems to be a couple of broken TLS tests -- for some reason the `curl <openssl> | tar` was broken even though I could make it work locally, something about "contents not gzip", so I just bumped the version and it ran again but it means we have a new version in the containers. PR coming soon but I've been working on getting the cross-compile containers working properly which is why I needed this fixed.

danbev:
@rvagg About the failing tests, I'm seeing two failure which are `test-crypto-dh-stateless.js` and `test-crypto-dh.js`. Are those the same failures you've run into?

richardlau:
@danbev those are the two tests failing in the CI: https://ci.nodejs.org/job/node-test-commit-linux-containered/26389/nodes=ubuntu1804_sharedlibs_openssl300_x64/

danbev:
@richardlau I've got a fix for one of then and I can skip the other until I've sorted it out. Should I create PRs against master for them?

richardlau:
@danbev Yes, please do.

danbev:
@richardlau The latest commit here is using git to clone `git@github.com:quictls/openssl.git` and the default branch is `openssl-3.0.0-alpha14+quic` which migth be nice for us so that we don't have to update it manually (hopefully). Would this work do you think?

richardlau:
@danbev I think so. I'll try to get https://github.com/nodejs/build/pull/2607 applied and merged first and then look at this one.

danbev:
> I'll try to get #2607 applied and merged first and then look at this one.

Awesome, thanks!

ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
```diff
@@ -58,7 +58,7 @@ ENV OPENSSL300DIR /opt/openssl-3.0.0
 
 RUN mkdir -p /tmp/openssl_3.0.0 && \
     cd /tmp/openssl_3.0.0 && \
-    curl -sL https://www.openssl.org/source/openssl-3.0.0-alpha14.tar.gz | tar zxv --strip=1 && \
+    git clone git@github.com:quictls/openssl.git && cd openssl && \
```
richardlau, line 61: This doesn't work as the docker image has no git credentials. Using a https://github.com/... URL would probably work.

richardlau:
```suggestion
    git clone https://github.com/quictls/openssl.git -b openssl-3.0.0-alpha15+quic --depth 1 && \
    cd openssl && \
```
@danbev Do you think we should test against both the quictls fork and non-forked versions of OpenSSL (instead of just replacing the non-forked with the fork)?

richardlau, line 61:
```suggestion
    git clone https://github.com/quictls/openssl.git -b openssl-3.0.0-alpha14+quic --depth 1 && \
    cd openssl && \
```
Use alpha14 for now as alpha15 breaks some Node.js tests (https://github.com/nodejs/node/issues/38373).

richardlau (approved):

AshCripps (approved):

richardlau:
Deployed this (with openssl3.0.0-alpha14+quic) to the docker hosts and started a test build: https://ci.nodejs.org/job/node-test-commit-linux-containered/26738/nodes=ubuntu1804_sharedlibs_openssl300_x64/

PR #2616, opened 2021-04-09
docker: update containers - gcc8 for cross, openssl3-a14

rvagg:
Builds on #2614, it turns out we didn't have gcc-8 ready in the cross compiler containers for host builds.

Also had to do the openssl-3-a14 update because a13 was having "not gzip content" errors (I couldn't replicate locally but upgrading to yesterday's release did the trick). I'm afraid this might have introduced new TLS failures from the one test run I've done so far. https://ci.nodejs.org/job/node-test-commit-linux-containered/26388/#showFailuresLink (test-crypto-dh-stateless & test-crypto-dh).

I'm also watching for some potential new failures from the gcc-8 armv7 builds, more runs need to shake out flakys. We'll also need to keep an eye on nightlies for the release version.

This is live now across the 4 containers in ci and 1 in ci-release.

### DIFF
--- a/ansible/roles/docker/templates/ubuntu1804_arm_cross.Dockerfile.j2
+++ b/ansible/roles/docker/templates/ubuntu1804_arm_cross.Dockerfile.j2
@@ -5,12 +5,14 @@ ENV LC_ALL=C
 RUN apt-get update \
   && apt-get dist-upgrade -y \
   && apt-get install -y software-properties-common \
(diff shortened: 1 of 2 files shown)


AshCripps (approved):

richardlau (approved):

richardlau:
Since this is already live I'll merge so that https://github.com/nodejs/build/pull/2607 can be rebased on top of it.

PR #2628, opened 2021-04-20
ansible: add extra M1 macs

AshCripps:
@nodejs/build-infra can I get `83.147.191.75` added to the test ci and `83.147.191.76` added to the release ci

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -67,6 +67,12 @@ hosts:
             ip: 83.147.191.69
             user: administrator
             ansible_python_interpreter: /usr/bin/python3
+        macos11.0-arm64-1:
+            ansible_python_interpreter: /usr/bin/python3
+            ip: 83.147.191.76
+            user: admin
+            remote_env:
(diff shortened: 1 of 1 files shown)


mhdawson (approved):
LGTM

richardlau:
The release machine has been added to the ci-release firewall. 

The [`iptables` playbook](https://github.com/nodejs/build/blob/master/ansible/playbooks/jenkins/host/iptables.yml) appears to still work (🎉) but it's erroring against the public CI host because there are stale host entries in secrets that don't have a matching entry in the inventory (i.e. presumably they've been removed but not cleared up from secrets, and also a reason for some of the failures in https://github.com/nodejs/build/issues/2531). I'll let you know when the test machine has been added.

AshCripps:
Both Machines added to ci

richardlau:
FTR I let @AshCripps know that the test machine has been added to the public CI before he added it 😄. I had to update the iptables playbook to filter out entries -- I'll open a PR with the changes.

richardlau (approved):

PR #2631, opened 2021-04-21
ansible: update iptables playbook

richardlau:
- Add a filter to skip over hosts that do not have an IP address, which
can happen if they have been removed from the main inventory but not
from secrets. Also filter out hosts that use hostnames instead of
numeric IP addresses (i.e. the Windows hosts).
Allow the proxy host look up when ansible is run in `--check` mode.

- Add filter for looking up IP addresses via `dig`. Add entries for
hosts that set their `ip` to a hostname instead of a numeric IP
address -- in the iptables rules we are adding these commented with
the supplied `ip` hostname rather than the name of the host.

### DIFF
--- a/ansible/playbooks/jenkins/host/iptables.yml
+++ b/ansible/playbooks/jenkins/host/iptables.yml
@@ -26,34 +26,29 @@
       - hostgroups[inventory_hostname] == hostvars[host].type
       # Following condition filters out entries without an IP address.
       - hostvars[host]['ansible_host'] is defined
-      # Following condition filters out entries that specify a hostname instead
-      # of an address (i.e. the Windows hosts). Ansible has a ipaddr filter but
-      # that requires an additional Python package (netaddr) to be installed.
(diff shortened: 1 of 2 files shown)


richardlau:
I've used this to update the firewalls for https://github.com/nodejs/build/pull/2628. 

I'll look at updating the documentation in https://github.com/nodejs/build/blob/master/ansible/MANUAL_STEPS.md#adding-firewall-entries-for-jenkins-workers. FWIW the ansible script updates the live iptables state, but `iptables-save > /etc/iptables/rules.v4` still needs to be run to persist the changes across reboots.

rvagg (approved):
I'm not familiar with this script but the changes seem sensible to me

AshCripps (approved):

mhdawson:
@richardlau should this land?

richardlau:
> @richardlau should this land?

Yes, done. I tested the playbook again after rebasing before merging.

PR #2633, opened 2021-04-23
metrics: Point to new metrics link

AshCripps:
Give a new link to where the current summaries are currently stored. This pr is meant to be a stop gap so people can access the data if they want until someone has time and patience to redesign the website and backfill the gaps in the data.

### DIFF
--- a/ansible/www-standalone/tools/metrics/public-index.html
+++ b/ansible/www-standalone/tools/metrics/public-index.html
@@ -326,6 +326,7 @@ pre code, pre tt {
   <body>
     <div class="container">
       <div class="body-content"><h1 id="nodejs-org-download-metrics">nodejs.org Download Metrics</h1>
+        <h2>  For "current" metrics go <a href="https://storage.googleapis.com/access-logs-summaries-nodejs/index.html">here.</a> Everything below is just historic data</h2>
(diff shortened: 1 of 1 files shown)


rvagg (approved):
seems fine to me. do you have access to put this in place on the web server (not with Ansible please)?

AshCripps:
No I dont so was going to ask someone else to do it for me.

richardlau (approved):

richardlau:
I've deployed this onto the [website](https://nodejs.org/metrics/) (copied the updated index into the `dist/metrics` directory and then manually queued/invoked a cdn-purge.
image

richardlau:
@rvagg Do you remember why we have https://github.com/nodejs/build/blob/master/ansible/www-standalone/tools/metrics/public-index.md (which hasn't been updated in this PR)? Are we supposed to update the markdown and then regenerate the HTML using https://github.com/rvagg/gfm2html (which appears to be archived)?

rvagg:
yep, https://github.com/nodejs/build/blob/f4fe95d7ea3e1e5459e9c243620f44017ef863f8/ansible/www-standalone/tools/metrics/public-index.html#L2

but not critical it works this way. Having them out of sync is a bit awkward but maybe we just remove the .md.

Issue #2626, opened 2021-04-19
Read-only file system on release CI


richardlau:
After landing https://github.com/nodejs/nodejs-dist-indexer/pull/15, I checked https://nodejs.org/download/nightly/, to see if I'd missed the nightly, and noticed that we haven't had a nightly build since 17 April but I know the master branch of the core repo has been updated since then.

From https://ci-release.nodejs.org/log/all
```
Apr 19, 2021 1:00:08 AM WARNING jenkins.model.lazy.LazyBuildMixIn newBuild
A new build could not be created in job iojs+release
java.io.IOException: Read-only file system
	at java.io.UnixFileSystem.createFileExclusively(Native Method)
	at java.io.File.createTempFile(File.java:2024)
	at hudson.util.AtomicFileWriter.<init>(AtomicFileWriter.java:142)
Caused: java.io.IOException: Failed to create a temporary file in /var/lib/jenkins/jobs/iojs+release
	at hudson.util.AtomicFileWriter.<init>(AtomicFileWriter.java:144)
	at hudson.util.AtomicFileWriter.<init>(AtomicFileWriter.java:109)
	at hudson.util.AtomicFileWriter.<init>(AtomicFileWriter.java:84)
	at hudson.util.AtomicFileWriter.<init>(AtomicFileWriter.java:74)
	at hudson.util.TextFile.write(TextFile.java:116)
	at hudson.model.Job.saveNextBuildNumber(Job.java:283)
	at hudson.model.Job.assignBuildNumber(Job.java:342)
	at hudson.model.Run.<init>(Run.java:322)
	at hudson.model.AbstractBuild.<init>(AbstractBuild.java:166)
	at hudson.matrix.MatrixBuild.<init>(MatrixBuild.java:79)
Caused: java.lang.reflect.InvocationTargetException
	at sun.reflect.GeneratedConstructorAccessor193.newInstance(Unknown Source)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at jenkins.model.lazy.LazyBuildMixIn.newBuild(LazyBuildMixIn.java:181)
	at hudson.model.AbstractProject.newBuild(AbstractProject.java:963)
	at hudson.model.AbstractProject.createExecutable(AbstractProject.java:1139)
	at hudson.model.AbstractProject.createExecutable(AbstractProject.java:138)
[...]

PR #2634, opened 2021-04-27
doc: add a disaster recovery plan

richardlau:
Initial version of a document containing pointers to how to recover critical parts of build infrastructure.

This is based on what we did to recover the release CI server last week (https://github.com/nodejs/build/issues/2626). It attempts to be descriptive rather than prescriptive because I don't expect we'll be able to predict the exact nature of a failure (should one occur).

The obvious takeaway for me after writing this is that we don't appear to have the nginx configuration for the two Jenkins servers backed up anywhere -- we were able to recover because @rvagg was able to rsync `/` from the old server in rescue mode. I'll look into what needs to be saved (maybe just rsnapshot all of `/etc/nginx`?) separately (mirroring updates to [./backup](https://github.com/nodejs/build/tree/master/backup) and this doc).

At least for the Jenkins configuration and iptables rules we have the backups and an alternative source to recover from if we lose the backup server(!) (even if there is potentially more work to recover the job configuration from the GitHub `jenkins-config-*` repositories).

cc @nodejs/build-infra

### DIFF
--- a/doc/disaster-recovery-plan.md
+++ b/doc/disaster-recovery-plan.md
@@ -1,0 +1,66 @@
+# Disaster Recovery Plan
+
+Notes and (hopefully) enough pointers to be able to recover critical bits of
+build infrastructure.
+
+## Table of Contents
+
+* [Jenkins Servers](#jenkins-servers)
+  * [Recovering Jenkins configuration](#recovering-jenkins-configuration)
+  * [Recovering iptables configuration](#recovering-iptables-configuration)
+
+## Jenkins Servers
+
+We have two Jenkins servers running the Stable (LTS) release of Jenkins:
+* ci.nodejs.org - The public testing CI. Accessible to anyone with a GitHub
+  login.
+* ci-release.nodejs.org - Restricted release CI, used to build binaries for
+  redistribution on nodejs.org. Access is restricted to releasers.
+
+Each server:
+* Is fronted by nginx.
+* Controls connections from Jenkins agents via iptables rules.
+* Is connected to our grafana server via a telegraf agent.
+* Run a script to sync the Jenkins job configurations to GitHub via `crontab`
+(see [Recovering Jenkins configuration] below).
+
+If setting up replacement servers:
+* `/var/lib/jenkins` should reside on a disk with 300GB capacity.
+* The IP address will need to be updated in the DNS settings at CloudFlare.
+* The Ansible inventory and [Jenkins host playbooks] updated.
+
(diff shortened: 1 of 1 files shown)


richardlau:
Ah also forgot about the telegraf agent for grafana (which I copied over from the old server, https://github.com/nodejs/build/issues/2626#issuecomment-825032593) -- maybe again rsnapshot all of `/etc/telegraf`?.

AshCripps:
is the telegraf setup something we could ansible-ise? at least for `infra` machines?

richardlau:
> is the telegraf setup something we could ansible-ise? at least for `infra` machines?

Potentially, yes.

rvagg:
Yeah, so the thing I was surprised by was that our backups were only for /var/lib/jenkins. If we hadn't been able to sync the items from /etc/ that we needed from the recovery-mode old server then it'd have been much harder. If we can make sure /etc/ is backed up from each of the servers we're backing up then that would go a long way to making sure this isn't too hard when we need to recover.

Also, making sure the backups continue to run is critical! I have very low confidence in that process and was surprised we had a recent backup available. We often run out of space on the backup server and only notice when ci fills up and stops working because the backup server is responsible for cleaning that up! Maybe the new telegraf thing has alerting for this that I haven't tuned in to? But that seems like a critical piece of this puzzle to get right that we've been on cruise-mode with.

richardlau:
> Yeah, so the thing I was surprised by was that our backups were only for /var/lib/jenkins. If we hadn't been able to sync the items from /etc/ that we needed from the recovery-mode old server then it'd have been much harder. If we can make sure /etc/ is backed up from each of the servers we're backing up then that would go a long way to making sure this isn't too hard when we need to recover.
> 

Proposed changes to backup `/etc` in https://github.com/nodejs/build/pull/2637

> Also, making sure the backups continue to run is critical! I have very low confidence in that process and was surprised we had a recent backup available. We often run out of space on the backup server and only notice when ci fills up and stops working because the backup server is responsible for cleaning that up! Maybe the new telegraf thing has alerting for this that I haven't tuned in to? But that seems like a critical piece of this puzzle to get right that we've been on cruise-mode with.

I'm not aware of whether we have alerting enabled/set up for grafana yet, or if the backup server is being monitored. cc @jbergstroem

AshCripps:
I spoke to Johan about alerting recently - its not setup yet but is totally feasible and is something I want to explore when I get some time with a key focus on alerting to the disk filling up on the ci machines.

mhdawson (approved):
This is a good start, LGTM

PR #2638, opened 2021-04-29
Upgrade playbooks for GCC 10 on IBM i and add new host

ThePrez:
Also change current host to ibmi73 from ibmi72.

Opening as draft until we can verify that it works with the CI properly, but the playbook currently runs cleanly against the newly-added rzkh host

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -45,7 +45,7 @@ hosts:
         rhel7-s390x-1: {ip: 148.100.86.101, user: linux1}
 
     - iinthecloud:
-        ibmi72-ppc64_be-1: {ip: 65.183.160.62, user: nodejs}
+        ibmi73-ppc64_be-1: {ip: 65.183.160.62, user: nodejs}
 
     - joyent:
         smartos17-x64-2: {ip: 147.75.88.60}
@@ -267,8 +267,14 @@ hosts:
             user: administrator
 
     - iinthecloud:
-        ibmi72-ppc64_be-1: {ip: 65.183.160.52, user: nodejs}
-        ibmi72-ppc64_be-2: {ip: 65.183.160.59, user: nodejs}
+        ibmi73-ppc64_be-1: {ip: 65.183.160.52, user: nodejs}
+        ibmi73-ppc64_be-2: {ip: 65.183.160.59, user: nodejs}
+
+    - rzkh:
+        ibmi73-ppc64_be-1:
+            ip: 185.113.4.148
+            user: nodejs
+            server_jobs: 32
 
     - rackspace:
         centos7-x64-1: {ip: 119.9.27.82}
--- a/ansible/roles/package-upgrade/tasks/partials/yum.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/yum.yml
@@ -5,6 +5,6 @@
 #
 
   - name: upgrade installed packages
-    yum: name=* state=latest use_backend=yum
+    yum: name=* state=latest use_backend=yum expire-cache=yes
 # If difficulty recognizing yum is encountered, see https://github.com/ansible/ansible/pull/69484, it is possible
 # to apply that patch manually to a local install of ansible (if necessary).
--- a/ansible/MANUAL_STEPS.md
+++ b/ansible/MANUAL_STEPS.md
@@ -497,6 +497,10 @@ another job.
 ### Install open source ecosystem
 See http://ibm.biz/ibmi-rpms (see "Installation" section)
 
+### Create Nodejs user
+```
+/QOpenSys/usr/bin/system -kpib 'CRTUSRPRF USRPRF(NODEJS) PASSWORD() USRCLS(*SECOFR) SPCAUT(*USRCLS) PWDEXPITV(*NOMAX)'
+```
 ### Create Nodejs user's home directory
 ```
 mkdir -p /home/NODEJS
@@ -517,10 +521,11 @@ echo 'export PATH' >> /QOpenSys/etc/profile
 After that is completed, copy to the `.bashrc` file for the nodejs user
 `cp /QOpenSys/etc/profile /home/NODEJS/.bashrc`
 
-### Use bash as the default shell (maintainer convenience only)
+### Use bash as the default shell for your user (maintainer convenience) and the nodejs user
 ```
-yum install chsh
-chsh -s /QOpenSys/pkgs/bin/bash
+/QOpenSys/pkgs/bin/yum install chsh
+/QOpenSys/pkgs/bin/chsh -s /QOpenSys/pkgs/bin/bash
+/QOpenSys/pkgs/bin/chsh -s /QOpenSys/pkgs/bin/bash nodejs
 ```
 
 ## z/OS
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -43,6 +43,12 @@ become_method             = sudo
 [hosts:iinthecloud]
 ansible_become             = false
 ansible_python_interpreter = /QOpenSys/pkgs/bin/python2
+ansible_shell_executable   = /QOpenSys/pkgs/bin/bash
+
+[hosts:rzkh]
+ansible_become             = false
+ansible_python_interpreter = /QOpenSys/pkgs/bin/python2
+ansible_shell_executable   = /QOpenSys/pkgs/bin/bash
 
 [hosts:marist]
 ansible_become             = false
--- a/ansible/plugins/inventory/nodejs_yaml.py
+++ b/ansible/plugins/inventory/nodejs_yaml.py
@@ -49,7 +49,7 @@ valid = {
     'provider': ('azure', 'digitalocean', 'ibm', 'iinthecloud', 'joyent',
                  'linuxonecc', 'macstadium', 'marist', 'mininodes', 'msft',
                  'nearform', 'orka', 'osuosl', 'packetnet', 'rackspace',
-                 'requireio', 'scaleway', 'softlayer', 'voxer')
+                 'requireio', 'rzkh', 'scaleway', 'softlayer', 'voxer')
 }
 DECRYPT_TOOL = "gpg"
 INVENTORY_FILENAME = "inventory.yml"
--- a/ansible/roles/baselayout/tasks/partials/repo/ibmi73.yml
+++ b/ansible/roles/baselayout/tasks/partials/repo/ibmi73.yml
@@ -1,0 +1,7 @@
+---
+
+- name: ibmi73 | install IBM i 7.3 repo
+  get_url:
+    url: https://public.dhe.ibm.com/software/ibmi/products/pase/rpms/ibm-7.3.repo
+    dest: /QOpenSys/etc/yum/repos.d/ibm-7.3.repo
+    mode: 0644
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -53,7 +53,7 @@ packages: {
   ],
 
   ibmi: [
-    'autoconf,automake,ca-certificates-mozilla,ccache,cmake,coreutils-gnu,gcc,gcc-cplusplus,gcc-cpp,git,libstdcplusplus-devel,m4-gnu,openssl-devel,python3,python3-pip,sed-gnu,zlib-devel',
+    'autoconf,automake,ca-certificates-mozilla,tar-gnu,gzip,ccache,cmake,coreutils-gnu,gcc10,gcc10-cplusplus,gcc10-cpp,git,libstdcplusplus6-10-devel,m4-gnu,openssl-devel,python3,python3-pip,sed-gnu,zlib-devel,icu-devel,c-ares-devel,brotli-devel,grep-gnu,pkg-config,xz',
   ],
 
   debian8: [
--- a/ansible/roles/java-base/tasks/main.yml
+++ b/ansible/roles/java-base/tasks/main.yml
@@ -36,8 +36,8 @@
 # as appropriate -- try to use generic os family if available.
 
 - name: Check if java is already installed
-  raw: java -version
   check_mode: no
+  raw: "{{ java_path[os] | default(\"java\") }} -version"
   register: java
   ignore_errors: yes
 
@@ -45,7 +45,6 @@
   when:
     - java.rc > 0
     - not os|startswith("zos")
-    - (arch != "ppc64" or os|startswith("ibmi"))
     - not os|startswith("macos")
   package:
     name: "{{ java_package_name }}"
--- a/ansible/roles/jenkins-worker/tasks/main.yml
+++ b/ansible/roles/jenkins-worker/tasks/main.yml
@@ -41,6 +41,15 @@
   args:
     creates: "/QSYS.LIB/IOJS.USRPRF"
 
+- name: setup user home directory for ibmi
+  when: os|startswith("ibmi")
+  file:
+    path: "{{ home }}/{{ server_user|upper }}"
+    state: directory
+    owner: "{{ server_user }}"
+    mode: 0755
+    recurse: yes
+
 - name: add ::1 to /etc/hosts for ipv6 compat
   when: not os|startswith("zos") and not os|startswith("ibmi")
   lineinfile:
--- a/ansible/roles/jenkins-worker/tasks/partials/tap2junit/ibmi.yml
+++ b/ansible/roles/jenkins-worker/tasks/partials/tap2junit/ibmi.yml
@@ -5,7 +5,5 @@
 #
 
 - name: install tap2junit
-  pip:
-    name: tap2junit
-    state: present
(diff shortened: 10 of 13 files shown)


ansible/inventory.yml
```diff
@@ -45,7 +45,10 @@ hosts:
         rhel7-s390x-1: {ip: 148.100.86.101, user: linux1}
 
     - iinthecloud:
-        ibmi72-ppc64_be-1: {ip: 65.183.160.62, user: nodejs}
+        ibmi73-ppc64_be-1: {ip: 65.183.160.62, user: nodejs}
+
+    - rzkh:
+        ibmi73-ppc64_be-4: {ip: 185.113.4.148, user: nodejs}
```
richardlau, line 51:
```suggestion
```

The new host should either be test (in the section below) or release (as is here) but not both.

ansible/inventory.yml
```diff
@@ -267,8 +270,11 @@ hosts:
             user: administrator
 
     - iinthecloud:
-        ibmi72-ppc64_be-1: {ip: 65.183.160.52, user: nodejs}
-        ibmi72-ppc64_be-2: {ip: 65.183.160.59, user: nodejs}
+        ibmi73-ppc64_be-1: {ip: 65.183.160.52, user: nodejs}
+        ibmi73-ppc64_be-2: {ip: 65.183.160.59, user: nodejs}
+
+    - rzkh:
+        ibmi73-ppc64_be-4: {ip: 185.113.4.148, user: nodejs}
```
richardlau, line 277:
```suggestion
        ibmi73-ppc64_be-1: {ip: 185.113.4.148, user: nodejs}
```
The hostnames we use in Jenkins follow `<type>-<provider>-<os>-<arch>-<num>` and we usually start numbering from 1, e.g. `test-rzkh-ibmi73-ppc64_be-1`.

ansible/roles/package-upgrade/tasks/partials/yum.yml
```diff
@@ -5,6 +5,6 @@
 #
 
   - name: upgrade installed packages
-    yum: name=* state=latest use_backend=yum
+    yum: name=* state=latest use_backend=yum expire-cache=yes
```
richardlau, line 8: ? `expire-cache` isn't an option on the `yum` task: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html
Perhaps the intent is `update_cache`?

ThePrez: `expire-cache` is an alias to `update_cache`. I figured it was more readable since the yum command equivalent is `yum clean expire-cache`

richardlau: 🤦 I missed the smallprint!

richardlau:
Testing a build 🤞: https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/nodes=ibmi73-ppc64/2/
(This will be without a populated ccache so likely an indicator of the slowest expected build time.)

richardlau:
I've done a few runs and it looks like with a populated ccache we can build the node binary in just over 30 minutes but then spend another hour building/running the test suite (of which 17 mins is building the tests, 3 minutes to build the docs). This compares to a ballpark 2 hours on the existing IBM i machines, so it is faster.

Some of the test suites are able to be run in parallel and benefit from the 32 cpus but there are some large ones that have to run sequentially (sequential, benchmark and pummel being the main ones). 

I've added extra parameters to the new temp job to allow the test suites run to be customized -- mainly to see how we fare if we don't run benchmark and pummel. Unfortunately I've yet to get a clean test run so far in that configuration (in the most recent, https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/nodes=ibmi73-ppc64/15/consoleFull, `wpt/test-encoding` appears to fail and break the test runner 😞 ).

I also tried building with
```
CONFIG_FLAGS=--shared-openssl --with-intl=system-icu --shared-cares --shared-brotli
```
but that failed because we didn't have `pkg-config` installed (needed for building with system ICU). 
https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/nodes=ibmi73-ppc64/6/console
I pushed a commit to install it and reansibled the machine but haven't rerun this configuration yet.

Trying with 
```
CONFIG_FLAGS=--shared-openssl --shared-cares --shared-brotli
```
appears to have two consistent test failures. The overall job time didn't seem to be much different from the default build configuration.

<details>

https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/nodes=ibmi73-ppc64/7/
https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/nodes=ibmi73-ppc64/8/
```
18:39:04 not ok 623 parallel/test-dns
18:39:04   ---
18:39:04   duration_ms: 2.206
18:39:04   severity: fail
18:39:04   exitcode: 1
18:39:04   stack: |-
18:39:04     (node:701833) [DEP0118] DeprecationWarning: The provided hostname "" is not a valid hostname, and is supported in the dns module solely for compatibility.
18:39:04     (Use `node --trace-deprecation ...` to show where the warning was created)
18:39:04     node:internal/errors:456
18:39:04         ErrorCaptureStackTrace(err);
18:39:04         ^
18:39:04     
18:39:04     Error: queryAny ENOMEM example.org
18:39:04         at QueryReqWrap.onresolve [as oncomplete] (node:internal/dns/promises:169:17) {
18:39:04       errno: undefined,
18:39:04       code: 'ENOMEM',
18:39:04       syscall: 'queryAny',
18:39:04       hostname: 'example.org'
18:39:04     }
18:39:04   ...
18:39:05 not ok 634 parallel/test-dns-resolveany
18:39:05   ---
18:39:05   duration_ms: 2.39
18:39:05   severity: fail
18:39:05   exitcode: 1
18:39:05   stack: |-
18:39:05     node:internal/errors:456
18:39:05         ErrorCaptureStackTrace(err);
18:39:05         ^
18:39:05     
18:39:05     Error: queryAny ENOMEM example.org
18:39:05         at QueryReqWrap.onresolve [as oncomplete] (node:internal/dns/promises:169:17) {
18:39:05       errno: undefined,
18:39:05       code: 'ENOMEM',
18:39:05       syscall: 'queryAny',
18:39:05       hostname: 'example.org'
18:39:05     }
18:39:05   ...
```
</details>

richardlau:
1 hour 13 mins without benchmark and pummel:
https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/nodes=ibmi73-ppc64/16/

ThePrez:
@richardlau , thanks for all the work on this! 
I did fix a possible DNS configuration issue on the system, so the next run may be cleaner (not betting my house on it, but..). Regardless, persistently failing tests is good. 

the new system is roughly 25% faster than the current system, which is good. Of course, was hoping for more of a difference. This machine has a lot more horsepower. 

Maybe we could try 16 concurrent jobs for the build and compare the results, but I don't expect a silver bullet there. I'll take a peek at system performance data for any low-hanging fruit also

ThePrez:
The overall performance signature of the builds all look pretty similar:
image
(the line is CPU %, the dark red is dispatched CPU time, the green is disk time). 
At a glance, it looks like:

- the 30-odd minutes of Node build look pretty good in terms of utilizing resources
- the 20 minutes of building tests and doc hovers pretty low for disk and CPU usage
- the first 10 minutes of running the tests pegs a single CPU (expected because of no concurrency)
- there's 5 minutes where it looks like two CPUs get pegged (parallel tests?)
- ... and then there's the weird 30-45 minutes of trickling use of resources.

I also checked the disk response times and they are sub-millisecond, except during peak Node build times, when we do see response times as high as 1.5ms with a bit of disk contention (probably ok and expected). 

Bottom line: no obvious performance issues spotted. :(

ThePrez:
On a whim I created a temporary in-memory filesystem for the Node.js builds. I know this has helped performance of internal builds (we don't use it, though, because we do internal builds in containers and the two techs don't jive). 

That should help performance, albeit an unknown amount.

However, in the process, I accidentally blasted the jenkins workspace. :(

ThePrez:
I also tweaked the ccache settings and moved the ccache directory into the in-memory filesystem. When you have time, please kick off two more runs to see how she fares. Thanks!

richardlau:
I requested a build but the agent/machine appears to have gone offline (agent is disconnected and I get time outs if I try to ssh into the machine).

ThePrez:
Currently connectable for me, and I restarted the agent job (I forgot I killed it earlier, and assumed it would get auto-restarted via `start-jenkins.sh`)

ThePrez:
FYI, just doubled the memory on the system to 64G

richardlau:
I started another build: https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/30/nodes=ibmi73-ppc64/

I did start a couple a few days ago, but these seem worse (!). (The first one took 4hrs 31mins, probably due to ccache misses but the second one took 2hrs 6mins which is a bit longer than we were seeing and around the same ballpark as the iinthecloud machines):
- https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/28/nodes=ibmi73-ppc64/
- https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/29/nodes=ibmi73-ppc64/

I also looked at the runs on the iinthecloud machines, e.g. https://ci.nodejs.org/job/node-test-commit-ibmi/nodes=ibmi72-ppc64/378/consoleFull. This built in 36 mins, 22 mins building tests and docs, then 1 hour 8 mins running tests. When compared to the timings we were seeing on the rzkh machine, it looks like:
- 30~mins for building in the best case (populated ccache, mostly cache hits). This time is reading the cache and linking? I expect make to be running parallel jobs here, perhaps down to single threads/processes for linking. 
- For building tests/docs the tests are built in parallel, but also utilise ccache (like when building the main node binary). The docs are built serially -- I think because there was a race condition where building the docs also relinked (believed to be unnecessary but would require trying to unpick how gyp generated the makefile) the node binary.
- Testing. This is where we're spending a lot of time. We currently run ~3500 tests. The tests are split into buckets (majority of the individual directories in https://github.com/nodejs/node/tree/master/test). Most of the tests are in `parallel`, which as the name suggests the test runner attempts to run in parallel (so should scale with the number of CPUs). There are still quite a lot of tests that are in, e.g. `sequential`, `benchmark`, `pummel`, that run serially for reasons including avoiding contention for resources (e.g. ports). 

One of the things I've been looking at is how we run CI on Windows and on the Raspberry Pi's. There we build and test in separate jobs ("fanned"), mainly so that for the tests we slice the buckets into smaller bits and run those bits on separate hosts (effectively increasing parallelism). It's not quite the same as the all-in-one build/test job (for example, I don't think the fanned jobs build the docs). I did a few manually simulations of running the tests on the rzkh machine in slices and came up with these timings:
```
-bash-4.4$ time make -j 32 test-ci-native
...
real    21m24.956s
user    6m12.226s
sys     0m35.025s
-bash-4.4$ time make -j 32 test-ci-js TEST_CI_ARGS="--run 0,4"
...
real    8m11.751s
user    4m52.718s
sys     1m13.078s
-bash-4.4$ time make -j 32 test-ci-js TEST_CI_ARGS="--run 1,4"
...
real    8m13.911s
user    4m25.491s
sys     0m43.475s
-bash-4.4$ time make -j 32 test-ci-js TEST_CI_ARGS="--run 2,4"
...
real    7m33.347s
user    3m10.493s
sys     0m49.875s
-bash-4.4$ time make -j 32 test-ci-js TEST_CI_ARGS="--run 3,4"
...
real    6m23.970s
user    2m50.388s
sys     0m48.779s
```

`test-ci-native` is the equivalent of building the tests (native addons) and running those (plus benchmark as one of the benchmark tests requires a native addon). The other `--run` invocations are the slicing of the remaining test buckets and do not rely on native addons (so are not dependent on `test-ci-native`). In a fanned scenario each of those bits could be running on a different host at the same time, so if for example we have the rzkh and two iinthecloud machines we could potentially be running three of those slices at once. 
e.g. theoretically (if hosts 1,2 and 3 were similar to the rzkh machine -- even if some were a bit slower we'd probably still save overall elapsed time?)
&nbsp; | host 1 | host 2 | host 3
---|---|---|---
&nbsp; |test-ci-native, 22 mins | run 0,4 8 mins 11s | run 1,4 8 mins 13s
&nbsp; | &nbsp;  | run 2,4 7mins 33s | run 3,4 6 mins 23 s
elapsed time | 22mins | 15mins 44s | 14 mins 33s

The cost of fanned is the job configuration is a bit more complicated (two jobs instead of one and we have to spend some time uploading the compiled binaries from the first job and then downloading them into the test jobs). Also any time savings get reduced if multiple builds are run at the same time (which is likely e.g. if we started to test IBM i on every pull request). 
@ThePrez thoughts?

richardlau:
Job times are now consistently worse than before
https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/31/nodes=ibmi73-ppc64/consoleFull took 2hrs 13 mins
- 48 mins to compile node
- 30 mins to compile tests and docs
- 55 mins to run tests&sup1;
 
https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/32/nodes=ibmi73-ppc64/consoleFull took 2hrs 9 mins
- 45 mins to compile node
- 30 mins to compile tests and docs
- 54 mins to run test&sup1;

&sup1; test failures. Investigating.

richardlau:
> ¹ test failures. Investigating.

Most of the test failures were due to case sensitive path comparisons in the tests (I believe we're on a case-insensitive file system).
e.g. https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/37/nodes=ibmi73-ppc64/console
```
18:31:54 not ok 3309 sequential/test-child-process-execsync
18:31:54   ---
18:31:54   duration_ms: 8.585
18:31:54   severity: fail
18:31:54   exitcode: 1
18:31:54   stack: |-
18:31:54     /bin/sh: iamabadcommand:  not found.
18:31:54     node:assert:123
18:31:54       throw new AssertionError(obj);
18:31:54       ^
18:31:54     
18:31:54     AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
18:31:54     + actual - expected
18:31:54     
18:31:54     + '/home/iojs/build/.tmp.3308'
18:31:54     - '/home/IOJS/build/.tmp.3308'
18:31:54              ^
```
I renamed the existing `/home/iojs` directory on the machine and reran the Ansible playbook, which has recreated `/home/IOJS` (note the upper casing) and this has cleared up most of the test failures. The remaining failure (which we don't, AFAIK, see on the existing iinthecloud machines) is wasi/test-wasi, e.g. https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/42/nodes=ibmi73-ppc64/testReport/(root)/test/wasi_test_wasi/ -- perhaps a real limitation/issue on UDFS?

Job runtimes are stubbornly in the 2hr 13mins ballpark (https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/nodes=ibmi73-ppc64/buildTimeTrend).

ThePrez:
@richardlau thanks for all your work on this. It's disheartening to see the performance numbers stubbornly "stuck". I thought the UDFS would help a lot, especially with the extra memory available. I will take another look at system performance data but am losing hope of a silver bullet. 

The fanned test jobs seem like a good thing to chase, but I fear we're still going to be on the slow end (and of course, now there are more points of failure). 

A crazy idea: Do you think the WG would let IBM i be in the regular CI without tests? If we could get the compile itself to 30 minutes, and could ensure commits don't break our build, that would still be valuable.

ThePrez:
@richardlau, we moved storage to VIOS internal SSD, so that may help performance. Mind giving it a whirl when you're free, to see if that made any impact?

richardlau:
New build: https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/43/nodes=ibmi73-ppc64/
I had to restart the Jenkins agent as it wasn't running -- I think the agent start script wipes out the .ccache dir so this build will have no ccache hits.

richardlau:
CI with populated ccache: https://ci.nodejs.org/job/node-test-commit-ibmi-richardlau/45/nodes=ibmi73-ppc64/

ThePrez:
OPE! I think we're also going to need to explicitly choose gcc 10. Where is that best done?
```
export CC=gcc-10
export CXX=g++-10
```

richardlau (approved):

richardlau:
I think at this point we should merge this and fix up remaining issues afterwards. I've rebased the PR, fixed the merge conflicts and squashed the commits and then run the playbook against `test-iinthecloud-ibmi72-ppc64_be-1` (which has now become `test-iinthecloud-ibmi73-ppc64_be-1`) and updated the two libuv CI jobs to run on `ibmi73-ppc64` (test builds passed). 

For the Node.js CI I'll need to merge this PR as we need the select-compiler.sh change. But the current Node.js CI for IBM i is broken without the newer GCC anyway (https://github.com/nodejs/node/issues/39479).

AshCripps (approved):

Issue #2123, opened 2020-01-03
Broken nodejs.org webhook since December 17th


nschonni:
The last deploy in https://nodejs.org/github-webhook.log was for https://github.com/nodejs/nodejs.org/commit/a1196dddf8a0db9787fe68970743cd7db3bba40a
The next commit in this repo was https://github.com/nodejs/build/commit/9df1fc4544dcfa8b17e4e5098836ac48cef9231d but it doesn't look like it should affect the hook so maybe something manual happened.

/cc @Trott

Trott:
I don't know much about the webhook, although I can learn I suppose? But hopefully someone else on @nodejs/build knows what to do?

XhmikosR:
May I also add that it might be better to rotate the log more frequently since it gets quite huge very fast?

phillipj:
Here's a couple of pointers I've learned after working with the @nodejs-github-bot which has a similar GitHub webhook setup:

- [github-webhook](https://github.com/rvagg/github-webhook) is a web server that gets installed globally on the server where nodejs.org is hosted
- [github-webhook systemd service](https://github.com/nodejs/build/blob/master/ansible/www-standalone/resources/config/github-webhook.service.j2) ensures the above web server is running and restarted it accordingly when needed
- [github-webhook.json](https://github.com/nodejs/build/blob/master/ansible/www-standalone/resources/config/github-webhook.json.j2) configures the webhook server to listen for certain GitHub webhook events and execute certain commands
- [build-site.sh nodejs](https://github.com/nodejs/build/blob/master/ansible/www-standalone/resources/scripts/build-site.sh) gets executed by github-webhook when commits lands on master

Hopefully the above gives some hints about what might be wrong or where to look 🤷‍♂

rvagg:
weird error messages in the system logs for github-webhook:

```
Jan 14 14:10:48 infra-digitalocean-ubuntu1604-x64-1 github-webhook[2647]: Emitted 'error' event at:
Jan 14 14:10:48 infra-digitalocean-ubuntu1604-x64-1 github-webhook[2647]:     at lazyFs.open (internal/fs/streams.js:273:12)
Jan 14 14:10:48 infra-digitalocean-ubuntu1604-x64-1 github-webhook[2647]:     at FSReqWrap.oncomplete (fs.js:141:20)
Jan 14 14:11:19 infra-digitalocean-ubuntu1604-x64-1 github-webhook[2956]: events.js:167
[...]

PR #2639, opened 2021-04-29
www: use `docker:lts` in `build-site.sh`

richardlau:
The nodejs.org website build uses `fibers` which is broken on Node.js 16
due to an incompatibility with the version of V8 that it contains.
Switch the Docker image used to build the website from `node:latest` to
`node:lts` to buy some time for the website to migrate off `fibers`.

---

This is live -- I edited the script on the www server, tested it and then synced the changes into this PR via:
```
ssh www cat /home/nodejs/build-site.sh > build-site.sh
```

Fixes: https://github.com/nodejs/build/issues/2123#issuecomment-829551740


richardlau:
> 🤦 I don't even want to know why `fibers` is getting involved here but using `lts` is a good move regardless

FWIW I recommended (https://github.com/nodejs/nodejs.org/issues/3828#issuecomment-829641558) that the website drops use of fibers.

AshCripps (approved):

PR #2642, opened 2021-05-04
jenkins: stop testing 16.x on AIX71

AshCripps:
I like to propose we stop testing 16.x on AIX71 - we list support as for AIX72 only and we only build the 16.x releases on AIX 72 meaning the binaries wont even run on AIX71.

cc @richardlau @mhdawson

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -92,7 +92,7 @@ def buildExclusions = [
 
   // AIX PPC64 ---------------------------------------------
   [ /aix71/,                          anyType,     lt(10)  ],
-  [ /aix71/,                          releaseType, gte(15) ],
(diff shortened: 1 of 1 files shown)


richardlau:
I'm in favour -- we have never released binaries for Node.js 16 on AIX 7.1. Now with Node.js 10 End-of-Life all of the in support versions of Node.js (12, 14, 16) are only supported on AIX 7.2, although we are still building binaries for 12 and 14 on AIX 7.1 as we did not have AIX 7.2 machines at the time they were initially released. I'd be open to the idea of planning whether we can drop AIX 7.1 entirely from our infra sooner than End-of-Life of Node.js 14. 

I think in the past we did switch release machines (from AIX 6.3 to AIX 7.1) mid-release line?

mhdawson (approved):
LGTM

richardlau (approved):

PR #2652, opened 2021-05-13
ansible: fixes for freebsd

richardlau:
- Update `python` to point to `python2` (for consistency with other
platforms in our CI; use `python3` for Python 3).
- Use `py38-pip` to match the version of Python 3 installed.
- Create a `pip3` symlink as Ansible's `pip` task fails without it.
- `-slaveLog` Jenkins agent parameter has been obsoleted. Use the
`-o` parameter to `daemon` to redirect stdout/stderr to the log file.

Refs: https://github.com/nodejs/build/issues/2531

### DIFF
--- a/ansible/roles/baselayout/tasks/main.yml
+++ b/ansible/roles/baselayout/tasks/main.yml
@@ -134,6 +134,13 @@
     name: python3
     path: "/usr/bin/python3.9"
 
+- name: freebsd | update python package alternatives
+  when: os == "freebsd11"
+  file:
+    dest: "/usr/local/bin/python"
+    state: link
+    src: "/usr/local/bin/python2"
+
 - name: smartos17 | update gcc symlinks
   when: os == "smartos17"
   file:
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -73,7 +73,7 @@ packages: {
   ],
 
   freebsd: [
-    'ccache,git,gmake,sudo,python3,py36-pip'
+    'ccache,git,gmake,sudo,python3'
   ],
 
   'macos10.10': [
--- a/ansible/roles/jenkins-worker/tasks/partials/tap2junit/freebsd.yml
+++ b/ansible/roles/jenkins-worker/tasks/partials/tap2junit/freebsd.yml
@@ -1,11 +1,18 @@
 ---
 
 #
-# freebsd: python2.7
+# freebsd: python3.8
 #
 
 - name: install pip
-  package: name=py27-pip state=present
+  package: name=py38-pip state=present
+
+- name: freebsd | update pip3 symlink
+  when: os == "freebsd11"
+  file:
+    dest: "/usr/local/bin/pip3"
+    state: link
+    src: "/usr/local/bin/pip-3.8"
 
 - name: install tap2junit
   pip: name=tap2junit state=present
--- a/ansible/roles/jenkins-worker/templates/freebsd.initd.j2
+++ b/ansible/roles/jenkins-worker/templates/freebsd.initd.j2
@@ -47,8 +47,8 @@ monitor_pidfile="/var/run/${name}/${name}_monitor.pid"
 required_files="${procname} ${jenkins_jar}"
 
 command="/usr/sbin/daemon"
-command_args="-f -r -p ${pidfile} -P ${monitor_pidfile} \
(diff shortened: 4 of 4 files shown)


Trott:
@nodejs/platform-freebsd sd

richardlau:
> This is an attempt to fix the `Failed 'jenkins-worker : install tap2junit'` error running the ansible work create playbook against `test-digitalocean-freebsd11-x64-1`. The good news is that running ansible against the freebsd now succeeds. Unfortunately the Jenkins service on the machines doesn't appear to be working and I've not yet worked out why 😞. `sudo service jenkins status` shows the status running/not running -- I haven't been able to work out why its failing to start properly as nothing gets written to `jenkins_console.log`.

Eventually figured it out by editing [`/usr/local/etc/rc.d/jenkins`](https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/templates/freebsd.initd.j2) to add a `-o` parameter to `daemon` so I could get at the output from the Java command, which shows an invalid argument. Jenkins renamed an agent parameter (from `-slaveLog` to `-agentLog`) and running the playbook updated the agent to a version where the parameter we were using has been dropped. Updated the PR to use the replacement parameter.

Refs: https://github.com/jenkinsci/remoting/pull/416

<details>

```
"-slaveLog" is not a valid option
java -jar agent.jar [options...]
 -agentLog FILE                  : Local agent error log destination (overrides
                                   workDir)
 -auth user:pass                 : If your Jenkins is security-enabled, specify
                                   a valid user name and password.
 -cert VAL                       : Specify additional X.509 encoded PEM
                                   certificates to trust when connecting to
                                   Jenkins root URLs. If starting with @ then
                                   the remainder is assumed to be the name of
                                   the certificate file to read.
 -connectTo HOST:PORT            : make a TCP connection to the given host and
                                   port, then start communication.
 -cp (-classpath) PATH           : add the given classpath elements to the
                                   system classloader.
 -failIfWorkDirIsMissing         : Fails the initialization if the requested
                                   workDir or internalDir are missing ('false'
                                   by default) (default: false)
 -help                           : Show this help message (default: false)
 -internalDir VAL                : Specifies a name of the internal files
                                   within a working directory ('remoting' by
                                   default) (default: remoting)
 -jar-cache DIR                  : Cache directory that stores jar files sent
                                   from the master
 -jnlpCredentials USER:PASSWORD  : HTTP BASIC AUTH header to pass in for making
                                   HTTP requests.
 -jnlpUrl URL                    : instead of talking to the master via
                                   stdin/stdout, emulate a JNLP client by
                                   making a TCP connection to the master.
                                   Connection parameters are obtained by
                                   parsing the JNLP file. (default:
                                   https://ci.nodejs.org/computer/test-digitaloc
                                   ean-freebsd11-x64-1/slave-agent.jnlp)
 -loggingConfig FILE             : Path to the property file with
                                   java.util.logging settings
 -noKeepAlive                    : Disable TCP socket keep alive on connection
                                   to the master. (default: false)
 -noReconnect                    : Doesn't try to reconnect when a
                                   communication fail, and exit instead
                                   (default: false)
 -proxyCredentials USER:PASSWORD : HTTP BASIC AUTH header to pass in for making
                                   HTTP authenticated proxy requests.
 -secret HEX_SECRET              : Agent connection secret to use instead of
                                   -jnlpCredentials. (default: 3ed3fb76127f63501
                                   0d4e8d1094b11414fb675d509b5f14124d2a4d1a9b7e1
                                   13)
 -tcp FILE                       : instead of talking to the master via
                                   stdin/stdout, listens to a random local
                                   port, write that port number to the given
                                   file, then wait for the master to connect to
                                   that port.
 -text                           : encode communication with the master with
                                   base64. Useful for running agent over 8-bit
                                   unsafe protocol like telnet
 -version                        : Shows the version of the remoting jar and
                                   then exits (default: false)
 -workDir FILE                   : Declares the working directory of the
                                   remoting instance (stores cache and logs by
                                   default)
```
</details>

Trott (approved):

richardlau:
> Updated the PR to use the replacement parameter.

For clarity (I'd updated the description but thought I'd also posted this comment) I decided not to use `-agentLog` as if you don't use `-agentLog` the output is sent to stdout/stderr -- we can capture that via the `-o` option for `daemon` and that's a better way of redirecting the output as it captures output in the case where the agent fails to start.

mhdawson (approved):
LGTM

jbergstroem (approved):
LGTM

PR #2653, opened 2021-05-13
ansible: remove openrc init from jenkins-worker

richardlau:
The openrc script appears to be a leftover from when we had standalone
Alpine CI hosts. All of our Alpine hosts are now Docker images so the
openrc script isn't being used (and wouldn't work in its current form
as it still uses the obsoleted `-slaveLog` Jenkins agent parameter).

Refs: https://github.com/nodejs/build/commit/bfd7b15e90c26920a8826126904ad7911662b8fd#diff-49397bb7595f8edbe823704d91a9f3daf07caa81d2ad573e56aad837b9ef4608


mhdawson (commented):
LGTM

AshCripps (approved):

PR #2659, opened 2021-05-25
ansible: use python3 in nodejs_yaml

AshCripps:
python3 is the default for python3 now with python pointing to python2.
This script fails on machines without python2 installed but as its
now EOL we should prefer python3


richardlau (approved):

rvagg (approved):

richardlau:
> This script fails on machines without python2 installed

Technically it fails on machines without a binary named `python` (without version suffixes). Some OS distributions symlink `python` to Python 3 but many others either only symlink `python` to Python 2 or do not create a `python` binary at all.

mhdawson (approved):
LGTM

Issue #2445, opened 2020-09-28
Action plan for 16.x upgrades


AshCripps:
This issue is the tracker for the updates and upgrades we need to undertake before the release date of 16.x in April 2021.

#

### Toolchain updates

These machines will need the new versions installed and added to the ansible scripts

- [x] Linux x86
- [x] Linux arm
- [ ] Linux cross compiling
- [x] MacOS
- [ ] Windows
- [x] AIX
- [x] zLinux
- [x]  Linux on P
- [ ] IBMi
- [ ] zOS
- [ ] smartos

# 

### Scripts

- [ ] Select-Compiler
- [ ] Possible some other jenkins scripts?

#

rvagg:
From https://github.com/nodejs/build/issues/2423, so far there's agreement that we need to get to _at least_ devtoolset-7 for all the Linux platforms. V8 will force this on us. If we can get to a later one then even better but this is probably going to be dictated by IBM folks.

AshCripps:
We would like to go to devtoolset-8 (GCC8) - A lot of our systems already have that available and ones we have in the pipleline are targeting 8 aswell.

richardlau:
We don't release on it, but we still test on smartos -- I assume we'll need to update those too?

AshCripps:
ah I forgot to mention feel free to update the list as I knew I wouldve forgotten something - Ill add smartos now

rvagg:
yeah, and we should also look at what we should retire when we do these, we often don't clean up old combos as we move forward.

AshCripps:
@rvagg 100% agree if we get on top of this early enough we can clear a lot of cruft from the buildfarm especially with 10.x going EOL roughly around the same time.

anlexN:
i think we don't need more talking, many and many os already have newer toolchain that nodejs need. so please quickly upgrade. i need many modern C++ syntax (for example: smart pointer) to write nodejs, i think if so, nodejs source code can be decreased very much and lighter. can we go further? action plan for 15.x, not wait 16.x.

AshCripps:
@anlexN Unfortunatly we can't push the update to 15.x as not all the platforms we support have GCC8 compatability and won't have it until near 16.x's release date. Also V8 won't require c++17 support until mid 2021 - https://chromium.googlesource.com/chromium/src/+/master/styleguide/c++/c++.md#modern-c_features

anlexN:
but at least, we must remove python 2. it is died.

AshCripps:
[...]

PR #2660, opened 2021-05-25
ansible: add gcc-8 to Ubuntu 18.04 hosts/containers

richardlau:
Baseline gcc/g++ compiler version for Node.js 16 and above is 8. The
default `gcc` and `g++` packages on Ubuntu 18.04 install version 7.

Moves `gcc-6` and `g++-6` out of the common `ubuntu` packages into
`ubuntu1604` -- it doesn't seem worth installing a lower version of
GCC to the default for Ubuntu 18.04.

Update `select-compiler.sh` to select GCC 8 for Node.js 16 and later.

Refs: https://github.com/nodejs/build/issues/2445

### DIFF
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -144,14 +144,20 @@ packages: {
   ],
 
   ubuntu: [
-    'ccache,g++,gcc,g++-6,gcc-6,git,libfontconfig1,sudo,python3-pip',
+    'ccache,g++,gcc,git,libfontconfig1,sudo,python3-pip',
   ],
 
   ubuntu1404: [
     'ntp,gcc-8,g++-8,gcc-6,g++-6,g++-4.8,gcc-4.8,g++-4.9,gcc-4.9,binutils-2.26',
   ],
 
+  # Default gcc/g++ package is 5.
   ubuntu1604: [
-    'python3.9,python3.9-distutils',
+    'gcc-6,g++-6,python3.9,python3.9-distutils',
+  ],
+
+  # Default gcc/g++ package is 7.
+  ubuntu1804: [
+    'gcc-6,g++-6,gcc-8,g++-8',
   ],
 }
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -144,7 +144,7 @@ packages: {
   ],
 
   ubuntu: [
-    'ccache,g++,gcc,git,libfontconfig1,sudo,python3-pip',
+    'ccache,git,libfontconfig1,sudo,python3-pip',
   ],
 
   ubuntu1404: [
@@ -153,7 +153,7 @@ packages: {
 
   # Default gcc/g++ package is 5.
   ubuntu1604: [
-    'gcc-6,g++-6,python3.9,python3.9-distutils',
+    'gcc-8,g++-6,gcc-6,g++-6,python3.9,python3.9-distutils',
   ],
 
   # Default gcc/g++ package is 7.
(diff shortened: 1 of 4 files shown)


ansible/roles/docker/templates/ubuntu1804.Dockerfile.j2
```diff
@@ -17,6 +17,8 @@ RUN apt-get update && apt-get dist-upgrade -y && apt-get install -y \
     ccache \
     g++ \
     gcc \
+    g++-8 \
+    gcc-8 \
```
richardlau, line 21: FWIW I'm not convinced the containers created by `ansible/roles/docker/templates/ubuntu1804.Dockerfile.j2` are actually being used in our CI. 
i.e. 
- https://ci.nodejs.org/computer/test-digitalocean-ubuntu1804_container-x64-1/builds
- https://ci.nodejs.org/computer/test-digitalocean-ubuntu1804_container-x64-2/builds
- https://ci.nodejs.org/computer/test-softlayer-ubuntu1804_container-x64-1/builds

are all empty (compare to an actual non-container e.g. https://ci.nodejs.org/computer/test-packetnet-ubuntu1804-x64-1/builds)

richardlau:
I haven't deployed this anywhere yet.

ansible/roles/baselayout/vars/main.yml
```diff
@@ -144,14 +144,20 @@ packages: {
   ],
 
   ubuntu: [
-    'ccache,g++,gcc,g++-6,gcc-6,git,libfontconfig1,sudo,python3-pip',
+    'ccache,g++,gcc,git,libfontconfig1,sudo,python3-pip',
```
rvagg, line 147: maybe we should just remove `g++,gcc` from here too?

ansible/roles/docker/templates/ubuntu1804.Dockerfile.j2
rvagg: yeah, probably not used, they were at one stage for some temporary testing work and they also exist as a way of experimenting with offloading VM tests to container tests, but other than that they're probably idling (which really only costs us the memory impact of the jenkins nodes)

rvagg:
What do you think about being more explicit? Remove `g++` and `gcc` from both baselayout and the Dockerfiles, add in `g++-6` and `gcc-6` and make select-compiler.sh use those instead of YOLO plain `gcc` and `g++`? As long as select-compiler.sh is involved it should be fine then. I'm not sure if it is on the bare 18.04 VMs but we could easily do that if needed.

richardlau:
Pushed https://github.com/nodejs/build/pull/2660/commits/ca9dcab588854f6cdbe5f5bd14885c446e97629e to remove `gcc`/`g++` and explicitly install `gcc-6`/`g++-6` (in addition to `gcc-8`/`g++-8`). I think (based on a modified local copy of the sharedlibs Dockerfile) we still end up with the `gcc`/`g++` package installed through some dependency chain.

richardlau:
I've deployed this onto the Docker containers hosted on
- test-digitalocean-ubuntu1804_docker-x64-1
- test-digitalocean-ubuntu1804_docker-x64-2
- test-softlayer-ubuntu1804_docker-x64-1

and the two test non-containered Ubuntu 18.04 hosts:
- test-digitalocean-ubuntu1804-x64-1 
- test-joyent-ubuntu1804-x64-1

I've also updated the [node-test-commit-linux-containered](https://ci.nodejs.org/job/node-test-commit-linux-containered/) job to call the `select-compiler.sh` script (https://github.com/nodejs/jenkins-config-test/commit/96bd09b15124cdfc9f985a947d94091b325d9287 and https://github.com/nodejs/jenkins-config-test/commit/741f1f8553e411adbea24acfbad8591d3003da67) in preparation for when this PR is merged (until then the builds fall through the existing `select-compiler.sh` script and default to `gcc`/`g++`). The [node-test-commit-linux](https://ci.nodejs.org/job/node-test-commit-linux/) job is already calling `select-compiler.sh`.

rvagg (approved):

PR #2670, opened 2021-06-10
ansible: provide absolute path in nodejs_yaml.py

AshCripps:
For AWX we need to use the nodejs_yaml.py script to parse the inventory
however the file expects to be run from within the `ansible`
subdirectory.

As Adoptium use AWX and the same format script as us this PR uses their
solution to fix the problem.

refs: https://github.com/adoptium/infrastructure/pull/1730

### DIFF
--- a/ansible/plugins/inventory/nodejs_yaml.py
+++ b/ansible/plugins/inventory/nodejs_yaml.py
@@ -29,6 +29,7 @@ import json
 import os
 import subprocess
 import sys
+from os import path
 
 import yaml
 try:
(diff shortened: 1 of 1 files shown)


richardlau (approved):

sxa (approved):
Looks suspiciously similar to an equivalent fix I made in another repository, so I approve ;-)

ansible/plugins/inventory/nodejs_yaml.py
```diff
@@ -74,7 +75,9 @@ def main():
     config.read('ansible.cfg')
 
     # load public inventory
-    export = parse_yaml(load_yaml_file(INVENTORY_FILENAME), config)
+    basepath = path.dirname(__file__)
+    inventory_path = path.abspath(path.join(basepath, "..", "..", "inventory.yml"))
```
bl-ue, line 79:
```suggestion
    basepath = os.path.dirname(__file__)
    inventory_path = os.path.abspath(os.path.join(basepath, "..", "..", "inventory.yml"))
```

ansible/plugins/inventory/nodejs_yaml.py
```diff
@@ -29,6 +29,7 @@
 import os
 import subprocess
 import sys
+from os import path
```
bl-ue, line 32:
```suggestion
```

bl-ue (changes requested):
It's generally considered better practice to use `os.path.<function>` rather than importing `os.path` directly.  Optional though :)

mhdawson (approved):
LGTM

PR #2675, opened 2021-06-16
ansible: build gn from source on Linux PPC/s390x

richardlau:
Add `build-test-v8` meta-role for installing prerequisites for V8
building/testing. Add roles to build `gn` and `ninja` binaries from
source and call those from the `build-test-v8` role for Linux on
ppc64/s390x.

This removes a previously manual step from the Linux PPC/s390x setup (docs also updated).

I've deployed this onto
- test-ibm-rhel7-s390x-1
- test-ibm-rhel7-s390x-2
- test-ibm-rhel7-s390x-3
- test-ibm-rhel7-s390x-4
- test-osuosl-centos7-ppc64_le-1
- test-osuosl-centos7-ppc64_le-2

to update the version of `gn` (the version we'd previously manually compiled is now too old, see https://github.com/nodejs/node/pull/38990#issuecomment-858764907). 
We're now updated to:
```
[iojs@test-osuosl-centos7-ppc64--le-1 build-tools]$ ./ninja --version
1.10.2
[iojs@test-osuosl-centos7-ppc64--le-1 build-tools]$ ./gn --version
1916 (d2dce75)
[iojs@test-osuosl-centos7-ppc64--le-1 build-tools]$
```

### DIFF
--- a/ansible/MANUAL_STEPS.md
+++ b/ansible/MANUAL_STEPS.md
@@ -7,7 +7,6 @@
     * [Full Xcode](#full-xcode)
     * [Signing certificates](#signing-certificates)
 * [RHEL7-S390X](#rhel7-s390x)
-  * [V8 build-tools](#v8-build-tools)
   * [devtoolset-6 install](#devtoolset-6-install)
 * [macOS](#macos)
 * [AIX 7.2 Install](#aix-72-install)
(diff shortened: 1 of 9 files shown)


mhdawson (approved):
LGTM

PR #2677, opened 2021-06-18
ansible: add ICU to sharedlibs containers

richardlau:
Add versions of ICU to the sharedlibs containers. Intended to be used
to build and test Node.js against a system ICU:
e.g. for ICU 65
```
$ export PKG_CONFIG_PATH=$ICU65DIR/lib/pkgconfig
$ export LD_LIBRARY_PATH=$(pkg-config --variable=libdir icu-i18n)
$ export CONFIG_FLAGS='--with-intl=system-icu --verbose'
$ make run-ci
```

### DIFF
--- a/ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
+++ b/ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
@@ -23,6 +23,7 @@ RUN apt-get update && apt-get install apt-utils -y && \
       gcc-8 \
       git \
       openjdk-8-jre-headless \
+      pkg-config \
       curl \
       python-pip \
       python3-pip \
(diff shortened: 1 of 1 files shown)


richardlau:
This is deployed onto the containers on
- test-digitalocean-ubuntu1804_docker-x64-1
- test-digitalocean-ubuntu1804_docker-x64-2
- test-softlayer-ubuntu1804_docker-x64-1

I've not made any CI job changes yet.

rvagg (approved):
there's some opportunities for some clever bash in here, but maybe keeping it simple-ish is good enough

are we going to add regular jobs for each of these? we're going to need to have a hard look at capacity I think

AshCripps (approved):

richardlau:
> there's some opportunities for some clever bash in here, but maybe keeping it simple-ish is good enough
> 

Admittedly I didn't try but the docs for `RUN` (https://docs.docker.com/engine/reference/builder/#run) says Docker runs `/bin/sh -c` which I believe is `dash` (and not `bash`) on Ubuntu systems. We could use `SHELL` instead to specifically execute `bash`, but I think we'll end up with something less readable as that needs to be in JSON form. 

> are we going to add regular jobs for each of these? we're going to need to have a hard look at capacity I think

My initial thought was to add one job (well subjob of the linux-containered job) that would parse out the minimum ICU version from [`tools/icu/icu_versions.json`](https://github.com/nodejs/node/blob/4330fb757f76182365ad1ac1b67a6170d052c4b1/tools/icu/icu_versions.json) and setup the "system" version of ICU to build based on that. The various versions of ICU in this PR is because each release line has a different minimum ICU -- I added the current minimum ICU for the release lines (12.x, 14.x, 16.x and master -- for 14.x we're planning on restoring the minimum ICU back to 65 but it's currently 67, https://github.com/nodejs/node/pull/39068) plus ICU 69 (V8 9.2, https://github.com/nodejs/node/pull/38990).

The other way we could do it is to test as many versions of ICU as possible (i.e. a new subjob for each ICU version and then VersionSelector.groovy to filter them by release line) but that will be more maintenance for the Build team (keeping VersionSelector.groovy updated, managing several icu* labels in Jenkins) and would be several new (sub)jobs which I would then share your concerns over capacity.

mhdawson (approved):
LGTM

PR #2682, opened 2021-06-22
ansible: allow running playbook in check mode

richardlau:
In some places we use the Ansible `raw` task to execute a command on
the remote machine. Typically this is done to check for the presence
of a file. Ansible itself has no way of knowing if the command we are
executing makes modifications to the system so does not run `raw` tasks
when `ansible-playbook` is run with `--check`. Some of our subsequent
tasks rely on the `raw` task having previous run (to create a variable)
so the playbook currently errors with messages such as:
`'dict object' has no attribute ...`
when run in `--check` mode.

Mark `raw` tasks that do not make modifications with `check_mode: no`
(which means "don't check if `--check` was set and just execute the
command"). This allows us to run the playbook with `--check` first to
see what Ansible thinks would change without actually making the
changes.

### DIFF
--- a/ansible/roles/bootstrap/tasks/partials/centos.yml
+++ b/ansible/roles/bootstrap/tasks/partials/centos.yml
@@ -1,5 +1,6 @@
 - name: check for firewalld
   raw: stat /usr/sbin/firewalld
+  check_mode: no
   register: has_firewalld
   failed_when: has_firewalld.rc > 1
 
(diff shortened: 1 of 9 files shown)


AshCripps (approved):

rvagg (approved):
🤷 ok, I didn't know about this feature but trust you know what you're doing!

mhdawson (approved):
LGTM

Issue #2681, opened 2021-06-22
Update GCC on  benchmark-ubuntu1604-intel-64


targos:
See https://ci.nodejs.org/job/node-test-commit-v8-linux/4078/nodes=benchmark-ubuntu1604-intel-64,v8test=v8test/console

Since we increased our minimum supported GCC version to >=8.3, V8 already removed hacks that they had to support old versions. Now V8 CI fails on the benchmark machine because GCC is too old:

```
15:19:43 + ./configure --verbose
15:19:45 Node.js configure: Found Python 3.9.4...
15:19:45 Detected C++ compiler (CXX=g++-6) version: 6.5.0
15:19:45 WARNING: C++ compiler (CXX=g++-6, 6.5.0) too old, need g++ 8.3.0 or clang++ 8.0.0
15:19:45 Detected C compiler (CC=gcc-6) version: 6.5.0
```

```
15:41:19 ../deps/v8/src/compiler/node.h:53:29: error: enclosing class of constexpr non-static member function ‘v8::internal::compiler::IrOpcode::Value v8::internal::compiler::Node::opcode() const’ is not a literal type
15:41:19    constexpr IrOpcode::Value opcode() const {
15:41:19                              ^~~~~~
15:41:19 ../deps/v8/src/compiler/node.h:42:7: note: ‘v8::internal::compiler::Node’ is not literal because:
15:41:19  class V8_EXPORT_PRIVATE Node final {
15:41:19        ^~~~
15:41:19 ../deps/v8/src/compiler/node.h:42:7: note:   ‘v8::internal::compiler::Node’ is not an aggregate, does not have a trivial default constructor, and has no constexpr constructor that is not a copy or move constructor
```

rvagg:
Perhaps a duplicate of #2656, it really just needs an upgrade. /cc @aduh95

We can do this, but need to find a time slot when it won't matter if we screw it up and have to scramble to get nearForm to help us revive it.

richardlau:
I agree that we need to look at updating the benchmark machines to Ubuntu 20.04. 

In the meantime our Ansible scripts do install gcc/g++ 8 on Ubuntu 16.04 https://github.com/nodejs/build/blob/3aaa7232294686ceb184942616696bb2ad0a1992/ansible/roles/baselayout/vars/main.yml#L155-L157 but it's been a while since we ran the scripts on the benchmarking machines (but we have done so in the past (https://github.com/nodejs/build/pull/1975#issuecomment-545237763)).

[...]

PR #2684, opened 2021-06-22
jenkins: use gcc 8 on the benchmark machines

richardlau:
Selects gcc/g++ 8 on the benchmark machines when building Node.js 16
or later.

Fixes: https://github.com/nodejs/build/issues/2681

### DIFF
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -164,15 +164,7 @@ elif [ "$SELECT_ARCH" = "X64" ]; then
       export CXX="ccache g++"
       echo "Compiler set to GCC" `$CXX -dumpversion`
       ;;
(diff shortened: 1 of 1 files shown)


rvagg (approved):
we're going to need to make sure we have ubuntu-toolchain-r onto 16.04 machines for this to work I think ... or get rid of them entirely

AshCripps (approved):

richardlau:
> we're going to need to make sure we have ubuntu-toolchain-r onto 16.04 machines for this to work I think ... or get rid of them entirely

For reference, this happens in https://github.com/nodejs/build/blob/3aaa7232294686ceb184942616696bb2ad0a1992/ansible/roles/baselayout/tasks/partials/repo/ubuntu1604.yml#L3-L12

PR #2698, opened 2021-07-02
Delete stale.yml

vtjnash:
This bot seems rude: it is closing active (non-stale) issues.


AshCripps:
It seems it didnt pick up the comment will look into fixing that, Closing this though as the bot is still useful.

PR #2713, opened 2021-07-16
ansible: require a limit when running build helper tasks

AshCripps:
Ensure a limit is set when running this potentially dangerous playbooks
which could cause a lot of damage if run against every machine in the
cluster


mhdawson (approved):
LGTM

rvagg (approved):
whoaboy, yeah

I assume also that we're limiting this to `test-*` hosts?

AshCripps:
@rvagg the only credentials Ive put on AWX is the test but I could also add a step here to only allow on test hosts.

PR #2722, opened 2021-08-04
ansible: clear yum packages cache

richardlau:
Clear the yum packages cache to prevent `/var` filling up.

Refs: https://github.com/nodejs/build/issues/2718#issuecomment-891126582

Ansible doesn't provide a task/option to do this so we need to directly run `yum clean`, see https://github.com/ansible/ansible/pull/31450#issuecomment-352889579.

### DIFF
--- a/ansible/roles/package-upgrade/tasks/partials/yum.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/yum.yml
@@ -3,8 +3,10 @@
 #
 # Updates all packages for centos-based distributions
 #
+- name: clean yum packages cache
+  ansible.builtin.command: yum clean packages
 
-  - name: upgrade installed packages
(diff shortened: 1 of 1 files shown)


AshCripps (approved):

mhdawson (approved):
LGTM

mhdawson:
@richardlau if this occurs between ansible runs we should consider if it could be one of the AWX bulid helper functions to clean up.

richardlau:
@mhdawson `/var/cache` is written to when `yum` is run -- that would usually happen as part of an Ansible run, although it's always possible for someone with access to log in and manually run `yum`. Without the change here the cache isn't cleared (unless done manually) which means it slowly grows every time a new package is added or updated. 

I'm not opposed to adding `yum clean` to an AWX template but that should probably happen separately.

Issue #2729, opened 2021-08-19
Set up new Ampere Altra machines in the build farm to replace older packetnet systems


sxa:
Equinix (formerly packet) are replacing their older aarch64 hardware and consolidating us on new 160-core Ampere Altra systems. This issue will track progress/deployments/miration on the new machines.

My intention is to prototype creating multiple docker containers underneath a base OS of Ubuntu 20.04 (to allow a wider range of OSs to be used) to allow a wider range of testing. Unlike the ThunderX system we had previously these also support the 32-bit armv7 instruction set, so we can also look at running docker containers that could build and test our `armv7l` builds on these systems which could allow us to reduce our reliance on the raspberry pi systems.

(Ref https://github.com/nodejs/build/issues/2708 as the machine that will likely be the first to be decommissioned)

sxa:
Successful test-commit run on two docker images on the machine, however there were issues running the playbooks:
- [The hostname cannot be changed](https://github.com/nodejs/build/blob/fcb12f9d2a72fcb4db7e437b0d95f184e9ac0561/ansible/roles/baselayout/tasks/main.yml#L18)
- [The systemd-timesyncd service](https://github.com/nodejs/build/blob/master/ansible/roles/baselayout/tasks/partials/ntp/systemd.yml) cannot be started (Though I expect having it running on the host will be adequate!)
- `::1` cannot be added to `/etc/hosts` for [ipv6 compat](https://github.com/nodejs/build/blob/fcb12f9d2a72fcb4db7e437b0d95f184e9ac0561/ansible/roles/jenkins-worker/tasks/main.yml#L53)
- On Ubuntu 20.04 there is no packet python-pip, only python-pip3 so [this task fails](https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/tasks/partials/tap2junit/ubuntu.yml)
- We cannot easily start the jenkins agent as service [the task](https://github.com/nodejs/build/blob/fcb12f9d2a72fcb4db7e437b0d95f184e9ac0561/ansible/roles/jenkins-worker/tasks/main.yml#L411) says the service doesnt' exist, so I've initiated them manually via nohup for now.

richardlau:
> * On Ubuntu 20.04 there is no package with the name `python-pip`, only `python-pip3` so [this task fails](https://github.com/nodejs/build/blob/master/ansible/roles/jenkins-worker/tasks/partials/tap2junit/ubuntu.yml)

[...]

PR #2731, opened 2021-08-24
ansible: add first equinix Altra as docker host

sxa:
This defines the first of the Ampere Altra systems (160 core, loads of RAM) as a docker host which can be used for building and testing.

This is in draft as it will need another change for the docker package, which may need to be changed to `docker.io` instead of the standalone `docker-ce` package on Ubuntu 20.04.

Part of https://github.com/nodejs/build/issues/2729


ansible/inventory.yml
```diff
@@ -14,22 +14,18 @@ hosts:
         ubuntu1604-x64-1: {ip: 138.197.224.240, alias: www}
         ubuntu1804-x64-1: {ip: 178.128.202.158, alias: gzemnid}
         ubuntu1804-x64-2: {ip: 45.55.45.227, alias: unofficial-builds}
-        ubuntu1804-x64-3: {ip: 157.245.7.159, alias: metrics}
-
-    - ibm:
-        ubuntu1804-x64-1: {ip: 169.45.166.50, alias: ci-release}
-        ubuntu2004-x64-1: {ip: 169.60.150.91, alias: ansible}
```
richardlau, line 21: Could you check the large number of changes/deletions of non-arm64 hosts? 🙂

sxa: Sorted - my rebase I'd done didn't seem to have the desired effect for some reason ...

ansible/plugins/inventory/nodejs_yaml.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env /usr/local/bin/python3
```
richardlau, line 1: Was this necessary?

ansible/roles/docker/templates/ubuntu2004_sharedlibs.Dockerfile.j2
```diff
@@ -0,0 +1,108 @@
+FROM ubuntu:18.04
```
richardlau, line 1:
```suggestion
FROM ubuntu:20.04
```

PR #2738, opened 2021-08-31
build,test: Add CentOS7 Dockerfile (arm64 only for now)

sxa:
"arm64 only" is due to the special way that I'm currently installing the devtoolset-6 packages in here for now. I feel this needs a better solution ...


mhdawson (approved):
LGTM as this documents the current install even if we are going to change in the future.

PR #2744, opened 2021-09-02
build,test: Add libuv prereqs to centos7 dockerfile

sxa:
[libuv-test-commit-linux](https://ci.nodejs.org/job/libuv-test-commit-linux) requires some additional packages which were not in the original CentOS7 dockerfile which is in use for the new aarch64 systems ([Ref issue](https://github.com/nodejs/build/issues/2729))

This has been deployed on the machines and those jobs can now run without prereq issues.

### DIFF
--- a/ansible/roles/docker/templates/centos7.Dockerfile.j2
+++ b/ansible/roles/docker/templates/centos7.Dockerfile.j2
@@ -25,6 +25,8 @@ RUN yum -y update && \
       python3-pip \
       fontconfig-devel
 
+RUN yum -y install libtool automake
+
 RUN pip3 install tap2junit
 
 RUN groupadd --gid {{ server_user_gid.stdout_lines[0] }} {{ server_user }}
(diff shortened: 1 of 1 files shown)


ansible/roles/docker/templates/centos7.Dockerfile.j2
```diff
@@ -25,6 +25,8 @@ RUN yum -y update && \
       python3-pip \
       fontconfig-devel
 
+RUN yum -y install libtool automake
```
richardlau, line 28: Is there a particular reason (layer caching?) to keep this separate from the earlier package installs?

richardlau (approved):

ansible/roles/docker/templates/centos7.Dockerfile.j2
sxa: Only that it was for a separate purpose from the main node build - happy to unify if preferred

sxa: Adjusted :-)

richardlau (approved):

AshCripps (approved):

PR #2750, opened 2021-09-06
www: run sync-benchmarking once every four hours

richardlau:
Fix the crontab so that `sync-benchmarking.sh` is only run once
every four hours and not every minute of every fourth hour.


rvagg (approved):
yeah, good catch!

AshCripps (approved):

mhdawson (approved):
LGTM

PR #2700, opened 2021-07-05
ansible: move `roles` and `plugins` under `playbooks`

AshCripps:
Wanted to float this here, basically if we want to expand AWX to be able to use our regular playbooks we need to move `role` to `playbooks`. This is because AWX runs from the root directory therefore doesnt pick up our `ansible.cfg` and cant find the roles, by moving roles under playbooks this allows for AWX to run the playbooks (currently need to find a solution for secrets) without impacting our usual workflow of running from within the `ansible` directory. So this should **not** make any difference to us running the playbooks manually.

thoughts?

PR #2767, opened 2021-09-22
ansible: move `ansible.cfg`

AshCripps:
replacement of #2700.

After consulting with some ansible experts at Red Hat the best method to get AWX working is to move the `ansible.cfg` file to the root of the repo where AWX expects it.

What does this change? Nothing except for now we need to run our playbooks from the root directory instead - e.g. `ansible-playbook ansible/playbooks/playbook --limit "target"`

I think ive updated the doc but let me know if it appears elsewhere

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,12 +1,12 @@
 [defaults]
-filter_plugins      = plugins/filter
+filter_plugins      = ansible/plugins/filter
 gathering           = explicit
 host_key_checking   = False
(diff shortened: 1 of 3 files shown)


mhdawson (approved):
LGTM

AshCripps:
ping  @nodejs/build as this changes the location you run ansible manually from (albeit minorly)

jbergstroem (approved):
SGTM

richardlau (approved):

PR #2772, opened 2021-09-24
ansible: use Java 17 on AIX

richardlau:
Update the JRE used to run the Jenkins agent on AIX to Java ~11~17.
Also adds `/opt/freeware/bin` to the `PATH` for the agent to
pick up binaries from yum packages that no longer create symbolic
links in `/usr/bin`.

Refs: https://github.com/nodejs/build/issues/2718

Trying this out on `test-osuosl-aix72-ppc64_be-3` to see if we still hit issues with agent disconnects and/or zombie processes when the Jenkins agent is run with Java ~11~17.

### DIFF
--- a/ansible/roles/jenkins-worker/templates/aix.rc2.j2
+++ b/ansible/roles/jenkins-worker/templates/aix.rc2.j2
@@ -16,6 +16,7 @@ start )
               export HOME=/home/{{server_user}}; \
               export NODE_TEST_DIR=$HOME/tmp; \
               export NODE_COMMON_PIPE="$HOME/test.pipe"; \
+              export PATH=$PATH:/opt/freeware/bin \
               export OSTYPE=AIX72; \
               export JOBS={{ jobs_env }}; \
         /usr/bin/java -Xmx128m -Dorg.jenkinsci.plugins.gitclient.Git.timeOut=30 -jar "$HOME/slave.jar" \
--- a/ansible/roles/java-base/tasks/main.yml
+++ b/ansible/roles/java-base/tasks/main.yml
@@ -89,7 +89,7 @@
     body_format: json
     return_content: yes
     status_code: 200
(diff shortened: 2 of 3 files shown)


mhdawson (approved):
LGTM

ansible/roles/jenkins-worker/templates/aix.rc2.j2
```diff
@@ -16,6 +16,7 @@ start )
               export HOME=/home/{{server_user}}; \
               export NODE_TEST_DIR=$HOME/tmp; \
               export NODE_COMMON_PIPE="$HOME/test.pipe"; \
+              export PATH=/opt/freeware/bin:$PATH; \
```
sxa, line 19:
```suggestion
              export PATH=$PATH:/opt/freeware/bin \
```

sxa (commented):
Is there a reason to add `/opt/freeware/bin` to the START of the `PATH`? In general I'd consider it preferable to not override the default system utilities from `/usr/bin` where possible but there may be a good reason for doing it in this case so I won't block this if the comment can be dismissed :-)

richardlau:
> Is there a reason to add `/opt/freeware/bin` to the START of the `PATH`? In general I'd consider it preferable to not override the default system utilities from `/usr/bin` where possible but there may be a good reason for doing it in this case so I won't block this if the comment can be dismissed :-)

It might not need to be at the start of the PATH. I'll try moving it to the end.

sxa (approved):

AshCripps (approved):

sxa:
Maybe it should be java 17 now ;-)

richardlau:
> Maybe it should be java 17 now ;-)

Tried. Can't find a Java 17 JRE for AIX via the api.adoptopenjdk.net rest API we're using in the Ansible scripts.

sxa:
> > Maybe it should be java 17 now ;-)
> Tried. Can't find a Java 17 JRE for AIX via the api.adoptopenjdk.net rest API we're using in the Ansible scripts.

That is a fair point. We haven't been able to release AIX17 formally yet although there are nightly builds that are available if you use `ea` instead of `ga` and go for `hotpost` instead of `openj9`  but they should obviously not necesarily be considered production ready. But they might be worth trying to see if they do show the same problems.

richardlau:
Updated to try Java 17 ea. (FWIW it appears that `openj9` still worked 🤷.)

richardlau:
Superseded by https://github.com/nodejs/build/pull/3047.

PR #2776, opened 2021-09-29
Exclude ubuntu2004-armv7l execution on Node<12

sxa:
Current Ubuntu armv7l docker image is 20.04 and so needs to be excluded from Node 12.
I did this initially with another conditional but have modified it to try and catch all ubuntu2004 - can anyone see any issue with this (I don't think the 32/64 bit is necessarily significant?


richardlau (approved):

AshCripps (commented):
Does this need a wildcard character to catch all?

sxa:
> Does this need a wildcard character to catch all?

If you mean at the end of th string I don't believe so - the regexs have `^` and `$` characters if it has to be explicitly the start/end of the field if reuqired (see the Alpine line folowing the one I've edited. I have not included the `$` to indicate the end so any trailing stuff would be permissible

AshCripps (approved):

mhdawson (commented):
LGTM

PR #2778, opened 2021-09-30
ansible: add macos11-x64 macs

AshCripps:


richardlau (approved):

ansible/roles/jenkins-worker/vars/main.yml
```diff
@@ -92,6 +92,7 @@ java_path: {
   'macos10.14': 'java',
   'macos10.15': 'java',
   'macos10.16': 'java',
+  'macos11': 'java',
```
richardlau, line 95: @AshCripps You'll probably need a `macos11` equivalent of https://github.com/nodejs/build/blob/664fabdb7b6e6990d78bffe450b5fad2dfe2421e/ansible/roles/java-base/vars/main.yml#L16 as well?

AshCripps: yep thats what Ive got locally will push up will all working

AshCripps: we do - I added checks for level and arch now but may need to rethink this part - although it may be a none issue with adopts arm builds being available. also will need to tweak the ci labels

AshCripps:
Right this PR should serve to combine the javas for intel and arm as adoptium now has there build available for the arm macs. Before this lands I will need to add a new `macos11-x64` label to the test ci and probably change the arm one to `macos11-arm`

PR #2813, opened 2021-11-25
metrics: add index-generator file

AshCripps:
This is the small file that makes the simple website with links for
our metrics. It runs every time a new summary file is created
and replaces the `index.html` file at the root of the bucket


mhdawson (approved):
LGTM

PR #2819, opened 2021-12-06
update redirect for security page

Trott:
Refs: https://github.com/nodejs/nodejs.org/pull/4280

### DIFF
--- a/ansible/www-standalone/resources/config/nodejs.org
+++ b/ansible/www-standalone/resources/config/nodejs.org
@@ -347,7 +347,7 @@ server {
     }
 
     rewrite ^/security.txt$                                       https://$server_name/.well-known/security.txt permanent;
-    rewrite ^/about/security/?$                                   https://$server_name/en/security/ permanent;
+    rewrite ^/about/security/?$                                   https://github.com/nodejs/node/blob/HEAD/SECURITY.md#security permanent;
     rewrite ^/contribute/?$                                       https://$server_name/en/get-involved/ permanent;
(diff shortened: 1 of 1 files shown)


Trott:
@nodejs/security-triage

danielleadams (approved):

Trott:
What would be the process for this? If I land it, does it automatically deploy? Or does someone with the right permissions need to run the ansible script?

Trott:
(I goofed and forgot to wait for this to land before landing the change to nodejs.org, so landing this promptly would be a good thing.)

Trott:
> What would be the process for this? If I land it, does it automatically deploy? Or does someone with the right permissions need to run the ansible script?

@nodejs/build

richardlau:
> What would be the process for this? If I land it, does it automatically deploy? Or does someone with the right permissions need to run the ansible script?

The short answer is no, it's not an automatic deployment.

https://github.com/nodejs/build/blob/master/ansible/www-standalone/README.md has this disclaimer:
> Not actively tested against the live server so may not be idempotent.

which is basically saying that it might not be safe to run the playbook against the live server. So what more likely needs to happen is someone from @nodejs/build-infra (including myself) needs to make an equivalent edit on the `nodejs.org` nginx config file on the server and then restart nginx. I'll do that now 🤞.

Trott:
> So what more likely needs to happen is someone from @nodejs/build-infra (including myself) needs to make an equivalent edit on the `nodejs.org` nginx config file on the server and then restart nginx. I'll do that now 🤞.

Seems to be working to me. Thanks!

Issue #2687, opened 2021-06-23
#node-build IRC channel


richardlau:
The project as a whole appears to be migrating from Freenode: https://github.com/nodejs/TSC/issues/1031
References to #node-build were removed from the core repo in https://github.com/nodejs/node/pull/39047

Our README still references the channel:
> Chat with us! We use IRC: #node-build at Freenode.

We should discuss whether to follow suite. I understand Freenode did a mass channel purge just over a week ago (https://www.devever.net/~hl/freenode_suicide) and all registered channels and nicknames were dropped.

I think our options are:
1. Stay on #node-build on Freenode. The rest of the project is moving/has moved away. 
2. Move the channel to libera.chat. This is where the #node.js channel went. If we moved here it would be good to re-establish the IRC bridge.
3. Move to some other IRC network. I'm not sure why we'd do this over Freenode (status quo) or libera (where the project is moving to).
4. Move off IRC altogether. Options here include:
a. #build on node-js.slack.com. This still exists. This Slack is a free tier Slack so has limited preserved history (is that an issue?).
b. Move to the OpenJS Foundation Slack (openjs-foundation.slack.com). This is a paid tier Slack so preserves history. A number of other working groups (e.g. Release who are probably one of the constituents most likely to want to talk to someone from Build) have a channel in this Slack. It might be possible to use Slack Connect to bridge the channel between openjs-foundation.slack.com and node-js.slack.com so you would not need to be logged into two different Slack workspaces. 
c. Something else (Discord/Gitter/Matrix/etc). Same as 3 -- I'm not sure what the benefits here would be (it would be yet another thing to have to log into).
5. Drop all "realtime" chat and force communication to GitHub issues (maybe enable discussions?).

richardlau:
I personally haven't used Freenode since the bridge to #build on node-js.slack.com was set up -- I'm not even sure if the bridge is still working. Would be good to hear opinions of anyone still using IRC.

AshCripps:
ive personally never once logged into the IRC so im in favour of slack, dont mind either instance but I have set up an alerting channel from our grafana on the open-js slack

mhdawson:
I'm on slack and keep up better there. Have only been able to intermittently keep up on IRC. Long way of saying my preference is slack.

ljharb:
The bridge was working until recently when freenode dropped all nick and channel data.

[...]

PR #2838, opened 2021-12-28
chore: IRC -> Slack

Trott:
We don't use IRC anymore (do we?).


ljharb:
LGTM

rvagg (approved):

richardlau:
> We don't use IRC anymore (do we?).

We do not. FWIW there was a discussion in https://github.com/nodejs/build/issues/2687.

Issue #2760, opened 2021-09-10
webhook "build-site.sh already running"


richardlau:
Reported as happened with 16.9.0 and now has happened with 16.9.1 -- publishing the release blog post to https://github.com/nodejs/nodejs.org is triggering the webhook but the logs show it is discarded as the website is currently being built

https://nodejs.org/github-webhook.log
```
event="push", match="ref == "refs/heads/master" && repository.full_name == "nodejs/nodejs.org"", exec="/home/nodejs/build-site.sh nodejs"
Wed Sep 08 2021 08:02:24 GMT+0000 (Coordinated Universal Time)
Took 295248 ms
stdout: build-site.sh already running
event="push", match="ref == "refs/heads/master" && repository.full_name == "nodejs/nodejs.org"", exec="/home/nodejs/build-site.sh nodejs"
Fri Sep 10 2021 19:54:04 GMT+0000 (Coordinated Universal Time)
Took 147 ms
stdout: build-site.sh already running
event="push", match="ref == "refs/heads/master" && repository.full_name == "nodejs/nodejs.org"", exec="/home/nodejs/build-site.sh nodejs"
Fri Sep 10 2021 19:54:04 GMT+0000 (Coordinated Universal Time)
Took 23 ms
```

richardlau:
https://github.com/nodejs/nodejs.org/pull/4105 was my original publish (not picked up as OP).
I've just pushed https://github.com/nodejs/nodejs.org/pull/4106 and it appears to be working 🤷.

richardlau:
Possibly a race with [`check-build-site.sh`](https://github.com/nodejs/build/blob/master/ansible/www-standalone/resources/scripts/check-build-site.sh) triggering a website rebuild and the webhook coming in while it's still in progress?

XhmikosR:
Probably unrelated but it seems that the live site hasn't been updated with the latest patch from the repo. Could the `https://nodejs.org/github-webhook.log` be rotated more frequently? Now it's huge making it hard to view/search it.

Trott:
https://nodejs.org/github-webhook.log is giving me a 403. Is that to be expected and I need a login or something?

richardlau:
> https://nodejs.org/github-webhook.log is giving me a 403. Is that to be expected and I need a login or something?

No, not expected. Maybe a side effect of https://github.com/nodejs/build/pull/2840? cc @rvagg

rvagg:
[...]

PR #2840, opened 2022-01-03
www,webhook: add logrotate for github-webhook

rvagg:
https://github.com/nodejs/build/issues/2760#issuecomment-1002242478

Long overdue. I've manually run this and it's 👌.

The main problem here is that I've put in a service stop and start. It's quite quick but this service is the parent of a docker build for the site. So if we happen to his this while the site is being built then there could be problems, so I'm not sure this is the best idea or approach.

I don't have time to go deeper into this but some thoughts:

* The service may not need to be stop & started, I don't recall if the log file is kept open or not, maybe this is a systemd redirect or maybe the service writes directly - someone could investigate this.
* We could add some kind of check to see whether it's running, not sure the best way to do this but we already have internal signalling in the service (iirc) so it doesn't double-run so maybe we can piggy back that somehow.

But for the time being, I think the risk is very low that this is going to be run (monthly) at exactly the same time the site is being built, it's not built _that_ often from the webhook. It's also not super critical (usually) and can be easily re-triggered from GitHub if needed. So IMO this is a quick fix that's worth doing now and improving later.

### DIFF
--- a/ansible/www-standalone/resources/config/logrotate-github-webhook
+++ b/ansible/www-standalone/resources/config/logrotate-github-webhook
@@ -1,0 +1,23 @@
+/home/nodejs/github-webhook.log
+{
+        monthly
+        missingok
+        compresscmd /usr/bin/xz
+        uncompresscmd /usr/bin/unxz
+        compressext .xz
+        compress
+        notifempty
+        create 0640 nodejs nodejs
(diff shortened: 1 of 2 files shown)


mhdawson (approved):
LGTM

Issue #2540, opened 2021-02-05
Official Windows arm64 binaries


gengjiawen:
@nodejs/releasers It's tier2 build now. Maybe make it in the official release process ?

_Originally posted by @gengjiawen in https://github.com/nodejs/node/issues/36658#issuecomment-767246793_

rvagg:
it does say "Tier 2 (compiling) / Experimental (running)", it's got two different tiers for good reason

gengjiawen:
> it does say "Tier 2 (compiling) / Experimental (running)", it's got two different tiers for good reason

Make it more available will get it more tested. Some of our user just want to use it. Building it from source will be a heavy task for most Node.js users. I think this will bring more value to the community in the long term.

joaocgreis:
Ref https://github.com/nodejs/build/issues/2450 and https://github.com/nodejs/node/issues/25998

There are some releases in https://unofficial-builds.nodejs.org/ . Since it's Windows and can't be built from the docker system that builds for other platforms, I have to manually trigger and publish the builds, so not all versions are available.

To be able to bring Windows ARM64 fully into tier 2, we'll need a better solution for testing. We're essentially waiting for Windows ARM64 machines to become available in Azure.

pb5050:
@joaocgreis  @gengjiawen  you guys should reach out to Microsoft engineers ive seen  where people have on occasion and the engineers actually chiming in  randomly on git hub, and the outcome 80% was very favorable and seemed like it was a good idea they did.

atishay:
@joaocgreis Is it possible to get node 16 unofficial binaries on Windows ARM64. It seems Azure arm64 is still further away.

jasnell:
While we are waiting on Azure to provide Windows arm64 images, if the project/foundation is willing to provide the Windows Arm64 hardware, we (Nearform) would be willing to host them alongside the mac minis we currently host so that we can at least have them always available. /cc @mcollina

mhdawson:
@jasnell do you have cost estimates/configs for the hardware.

@joaocgreis I assume if these were added you'd an eye on them, reconnecting jenkins agents, compiler upgrades etc. like you do for other windows systems ?

joaocgreis:
@atishay working on the releases, they should be done shortly. Will post on https://github.com/nodejs/build/issues/2450 when done.

@mhdawson yes I can keep them functional.

[...]

PR #2860, opened 2022-01-31
Update Windows Inventory

joaocgreis:
We changed the Azure subscription used for the CI machines, this PR updates the Ansible inventory. The new Windows ARM machines hosted by NearForm are also included (ref https://github.com/nodejs/build/issues/2540).

Two important changes were made:
- Removed VCBT2015 (not VS2015) machines and configurations, as the install package no longer works. Already removed from Jenkins as well. I believe testing with VS2015 provides enough coverage for a version so old.
- Only server images are supported on the new subscription, so I was not able to create Windows 10 machines. For now, I replaced it with Windows Server 2022. The server names don't yet reflect this everywhere, at least the VersonSelectorScript needs to be updated before the server names can change.

cc @nodejs/libuv, I swapped the VCBT2015 configuration to VS2019.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -110,20 +110,16 @@ hosts:
   - test:
 
     - azure:
-        msft-win10_vcbt2015-x64-1: {ip: nodejs.eastus2.cloudapp.azure.com}
-        msft-win10_vcbt2015-x64-2: {ip: nodejs.westus2.cloudapp.azure.com}
-        msft-win10_vcbt2015-x64-3: {ip: nodejs.eastus2.cloudapp.azure.com}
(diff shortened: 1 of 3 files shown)


rvagg (approved):
> replaced it with Windows Server 2022

In your assessment is this a _good enough_ replacement? How much of a gap are we leaving open here for possible problems on 10 that don't show up on 2022?

richardlau (approved):

mmarchini (approved):

mcollina (approved):
lgtm

joaocgreis:
@rvagg Windows Server 2022 is between Windows 10 and 11 (but closer to 10). It's always possible to have issues specific to a single version/build/combination, but I think in this case it's not very likely to happen. We still test on 2012 and 2016.

PR #2874, opened 2022-02-19
doc: s/Node.js Foundation/Node.js Project/g

mmarchini:
Node.js Foundation doesn't exist anymore. We could replace instances of it in our README with OpenJS Foundation, but that wouldn't be accurate since we only provide infrastructure for Node.js Projects and libuv (afaik libuv isn't under OpenJSF either, which would also make our README inaccurate).


mhdawson (approved):
LGTM

PR #2882, opened 2022-03-02
fix: make github-webhook.log on www publicly readable

rvagg:
Ref: https://github.com/nodejs/build/issues/2760#issuecomment-1056027274


rvagg:
live on www https://nodejs.org/github-webhook.log

Issue #2815, opened 2021-11-29
Platform requirements for Node.js 18


richardlau:
Node.js 18 is due out in April 2022 and is planned to last until April 2025 (https://github.com/nodejs/Release/blob/main/schedule.json). We need to plan out any required toolchain/hardware changes we want/need to make.

cc @nodejs/build 

## What we build and test on

From https://github.com/nodejs/node/blob/master/BUILDING.md#official-binary-platforms-and-toolchains

Binary package | Platform and Toolchain | Decision | done
-- | -- | -- | --
aix-ppc64 | AIX 7.2 TL04 on PPC64BE with GCC 8 | move minimum hardware to Power 8
darwin-x64 | macOS 10.15, Xcode Command Line Tools 11 with -mmacosx-version-min=10.13
darwin-arm64 (and .pkg) | macOS 11 (arm64), Xcode Command Line Tools 12 with -mmacosx-version-min=10.13
linux-arm64 | CentOS 7 with devtoolset-8 / GCC 8 | Need to migrate (https://github.com/nodejs/build/issues/2741)
linux-armv7l | Cross-compiled on Ubuntu 18.04 x64 with custom GCC toolchain
linux-ppc64le | CentOS 7 with devtoolset-8 / GCC 8 | Need to migrate (https://github.com/nodejs/build/issues/2741) | ✔️ 
linux-s390x | RHEL 7 with devtoolset-8 / GCC 8 | Need to migrate (https://github.com/nodejs/build/issues/2741) | ✔️ 
linux-x64 | CentOS 7 with devtoolset-8 / GCC 8 | Need to migrate (https://github.com/nodejs/build/issues/2741) | ✔️ 
win-x64 and win-x86 | Windows 2012 R2 (x64) with Visual Studio 2019

## Compiler levels

Operating System | Compiler Versions
-- | --
Linux | GCC >= 8.3
Windows | Visual Studio >= 2019 with the Windows 10 SDK on a 64-bit host
macOS | Xcode >= 11 (Apple LLVM >= 11)

## Actions

- [x] Check with IBM if remaining on Power 8 until 2025 is viable (we'll need additional machines if not).
- [ ] Any reason to move the minimum deployment targets for macOS on from macOS 10.13?
- [x] CentOS 7 gets updates until 30 June 2024 which is within the expected lifetime of Node.js 18, so we'll need to move (probably to RHEL 8, https://github.com/nodejs/build/issues/2741). 
[...]

Issue #2741, opened 2021-09-01
Consider moving from Centos7 to RHEL8 for 18.x


AshCripps:
Whilst looking at https://github.com/nodejs/build/issues/2740 ive discovered that Centos7 goes EOL by the end of 18.x, so we should consider moving off of centos7 before 18.x is cut next year.

A suitable option would be to upgrade to RHEL8 - Red Hat is providing a scheme for open source communities to gain free access to RHEL subscriptions but wanted to pose the question to the wider group to see if anyone had any objections?

cc @nodejs/build

richardlau:
@AshCripps I don’t think we have enough ppc Linux hosts for both CentOS7 (existing workloads) and the next (e.g. RHEL8).

AshCripps:
we are nowhere near our capacity for ppc linux machines atm (provided we are allowed to use everything given to us)

mhdawson:
Confirming @AshCripps assessment that we should have no problem adding more ppc machines. We should look at the same for other architectures as well.

Either way I think it's something we have to figure out as I don't think we want to be building on an EOL Operating system.  One of the main concerns is that it will main an increase in the minimum glibc levels...

richardlau:
One thing we'd need to check for the x64 platforms is what is offered on the various cloud providers and if RHEL 8 is an out of the box offering (i.e. existing selectable image).

richardlau:
| Provider | RHEL 8 image available |
|---|---|
| DigitalOcean | No |

# DigitalOcean
![image](https://user-images.githubusercontent.com/5445507/131831558-7b80f1fc-e499-407a-b84b-fe75d3353b9e.png)

richardlau:
So it looks like currently only IBM offer an OOTB RHEL image (of any version). On DigitalOcean we can use [custom images](https://docs.digitalocean.com/products/images/custom-images/) but that's going to be some work to set up. Or we run RHEL in a container via Docker under a different host OS.

AshCripps:
I assume all the other OS are free version? RHEL probably doesnt show up as it requires a subscription

richardlau:
I've started adding RHEL8 s390x (LinuxONE) in https://github.com/nodejs/build/pull/2859.

[...]

PR #2886, opened 2022-03-04
ansible: add RHEL 8 x64 instances

richardlau:
Add RHEL 8 x64 instances, including repurposing some obsolete CentOS 6
instances which are no longer used by any CI jobs.

Refs: https://github.com/nodejs/build/issues/2741


mhdawson (approved):
LGTM

PR #2890, opened 2022-03-09
ansible: add ppc64le RHEL 8 instances

richardlau:
Add new ppc64le RHEL 8 instances. Repurpose test-osuosl-centos7-ppc64_le-4
as an additional RHEL 8 instance.

Also ensure time synchronization via `chrony` on RHEL 8, and add a handler
to run when packages are updated to sync with Red Hat Subscription Manager.

Refs: https://github.com/nodejs/build/issues/2741


mhdawson (approved):
LGTM

PR #2892, opened 2022-03-12
ansible: decrypt secrets earlier in playbook

richardlau:
Decrypt stuff from secrets much earlier in the playbook to avoid
gpg agent timeouts.


mhdawson (approved):
LGTM

Issue #2893, opened 2022-03-12
s390x is offline


tniessen:
`node-test-commit-linuxone` jobs have been pending for about 12 hours. Jenkins shows this for all s390x platforms:

![Jenkins shows all nodes are offline](https://user-images.githubusercontent.com/3109072/158020200-19aa2146-fb26-4783-84ce-039ce5ccd027.png)

richardlau:
See https://github.com/nodejs/build/issues/2869

tniessen:
Thanks for the info @richardlau! s390x has come back online but rhel8-s390x seems to always fail now.

richardlau:
hmm it looks like despite writing iptables rules to `/etc/sysconfig/iptables` in https://github.com/nodejs/build/blob/master/ansible/roles/bootstrap/tasks/partials/rhel8-s390x.yml they don't appear to be in effect after a system restart/reboot. Running `sudo iptables restart` does appear to be reloading the rules, so it's not like the settings were "lost". I think the machines should be working now (I ran a test build) but we'll need to figure out how to get this to persist across reboots.

richardlau:
This is very strange. I've tracked down the issue to the presences of `/etc/sysconfig/iptables.save`. When this file is there (as it is on all of our RHEL 8 s390x systems), a system reboot doesn't appear to be loading the rules in `/etc/sysconfig/iptables` (but `sudo systemctl restart iptables` does). This file doesn't appear to be anything we have created via our Ansible scripts, at least judging by the dates in the comments at the beginning and end of the file. The file is not present on our RHEL 7 instances.

<details>
<summary>/etc/sysconfig/iptables.save</summary>

```
# Generated by iptables-save v1.4.21 on Fri Dec 18 00:25:24 2015
*raw
:PREROUTING ACCEPT [123:8625]
:OUTPUT ACCEPT [65:6680]
-A PREROUTING -i lo -j NOTRACK
-A OUTPUT -o lo -j NOTRACK
COMMIT
*filter
:INPUT ACCEPT [0:0]
:FORWARD DROP [0:0]
:OUTPUT ACCEPT [63:6376]
-A INPUT -p all -s localhost -d localhost -j ACCEPT
-A INPUT -p tcp -m conntrack --ctstate NEW -m tcp --dport 22 -j ACCEPT
-A INPUT -p icmp -j ACCEPT
-A INPUT -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
[...]

PR #2896, opened 2022-03-14
ansible: remove `iptables.save` on rhel8-s390x

richardlau:
The presence of `/etc/sysconfig/iptables.save` on our RHEL 8 s390x
instances is interfering with the loading of our iptables rules
when the system reboots.

Refs: https://github.com/nodejs/build/issues/2893#issuecomment-1067054317


mhdawson (approved):
LGTM

PR #2902, opened 2022-03-18
ansible: fix path to ccache symlinks on M1 macs

richardlau:
Homebrew installs into a different path on M1 macs compared to x64.
Adjust the path to the ccache symlinks accordingly.

Refs: https://formulae.brew.sh/formula/ccache

cc @AshCripps


AshCripps (approved):

richardlau:
@AshCripps I pushed another commit to account for the administrator account being named differently between the macStadium and NearForm hosted macs. PTAL.

(Also I don't appear to be able to reach the macstadium arm64 test macs (but can reach the release one at macStadium) 🤷)

mhdawson (approved):
LGTM

PR #2903, opened 2022-03-21
jenkins: stop testing on centos7 for Node.js 18 onwards

richardlau:
We're starting to get upstream changes that require later glibc
than is available on CentOS 7. In addition, our nightly (and
eventually Node.js 18 onwards releases) binaries require later
glibc (2.28) which will not run on CentOS 7.


mhdawson (approved):
LGTM

PR #2905, opened 2022-03-23
ansible: update RHEL 8 s390x inventory

richardlau:
Moving our LinuxONE self-provisioned machines into a new "Node.js"
project. The previous ones were under a project named with Richard's
email address and the project cannot be renamed.


mhdawson (approved):
LGTM

PR #2908, opened 2022-03-25
ansible: add initial AIX 73 machine

mhdawson:
Signed-off-by: Michael Dawson <mdawson@devrus.com>

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -146,6 +146,9 @@ hosts:
         aix71-ppc64_be-4:  
             ip: 169.48.22.51
             server_jobs: 6
+        aix73-ppc64_be-1:
+            ip: 163.68.64.114
+            server_jobs: 6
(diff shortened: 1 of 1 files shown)


ansible/inventory.yml
```diff
@@ -146,6 +146,9 @@ hosts:
         aix71-ppc64_be-4:  
             ip: 169.48.22.51
             server_jobs: 6
+        aix73-ppc64_be-1:
+            ip: 163.68.64.114
+            server_jobs: 6
```
richardlau, line 151: I don't suppose we know if this is a sensible value for the new machine? I guess we can fix up later if not.

richardlau (approved):

ansible/inventory.yml
mhdawson: +1

mhdawson: The machine should be similar to our other machines as we asked for the same specs.

PR #2910, opened 2022-03-30
ansible: updates for Linux arm64

richardlau:
Adds two `c2.large.arm` Ampere eMag machines configured to run two
containers each:
  - CentOS 7 container intended for Node.js earlier than 18
  - RHEL 8 container intended for Node.js 18 and later

Add one RHEL 8 container to each of the Equinix test servers.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -40,6 +40,8 @@ hosts:
 
     - equinix:
         centos7-arm64-1: {ip: 139.178.85.110, server_jobs: 32}
+        ubuntu2004_docker-arm64-1: {ip: 145.40.80.210}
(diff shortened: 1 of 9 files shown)


richardlau:
FYI I've taken https://ci-release.nodejs.org/computer/release-equinix-centos7-arm64-1/ offline (since yesterday) and the nightlies have been built in the new CentOS 7 containers. Once this lands I'll add the `rhel8-arm64-release` to iojs+release which will (in combination with the VersionSelectorScript change) switch building the release binaries for Node.js 18 and later to RHEL 8 (as we've done on the other Linux archs).

mhdawson (approved):
LGTM

Issue #2775, opened 2021-09-28
Try running arm32 test jobs in containers on aarch64 hosts


sxa:
Unlike the old ThunderX systems, the new Altra machines which we recently set up are capable of running 32-bit ARM code, so can run our armv7l releases on the hardware.

Since we are already using multiple docker containers on those systems, we can also deploy armv7l containers running the distribution of our choice, and try running a non-fanned test job on those systems in an expedient manner.

sxa:
We had previously discussed adding some [raspberry pi4 systems](https://github.com/nodejs/build/issues/1851) into the jenkins farm, but that seemed to go stale. This would probably provide a more stable, and faster, option at zero additional cost and hopefully lower maintenance overhead.

sxa:
Initial testing with a copy of the node-test-commit-linuxone job:

Job | Time for v17 | Comment
--|--|--
node-test-commit-linuxone | 12-15 minutes | variation probably ccache dependent on different PRs
sxa-test-commit-arm32 | 14 minutes | NOTE: This currently runs without ccache enabled

On the basis of this, I'm going to propose adding this into the main group of jobs that run under [node-test-commit](https://ci.nodejs.org/job/node-test-commit/) and run it alongside [node-test-commit-arm-fanned](https://ci.nodejs.org/job/node-test-commit-arm-fanned/) for a while to see how it works out.

For comparison with the above figures, [node-test-binary-arm-12+](https://ci.nodejs.org/job/node-test-binary-arm-12+/buildTimeTrend) has been taking over an hour for the most recent runs (and always over half an hour) so there are significant performance avantages to be had here. This would also remove the use of a cross-compiler for this platform if we wished to do so.

Questions I see with the general approach:
- Are we happy that running on a container on an armv8l host would be ok (Technically the pi3s are armv8 but I'm guessing most are not running the RaspiOS 64-bit kernel)
- We'd need to be comfortable that the version of glibc that's being targetted is suitable for running on all target platforms we'd be interested in. My current tests have been with libc 2.31 as I'm using [an Ubuntu 20.04.3 container](https://ci.nodejs.org/computer/test-equinix-ubuntu2004_container-armv7l-1/), but we could use something else.

mhdawson:
Great to hear it seems to be working. Would be great to be able to use docker images versus physical machines.

rvagg:
[...]

PR #2911, opened 2022-03-31
ansible: add armv7l on aarch64 docker files

sxa:
This dockerfile will be used on the aarch64 Altras to provide a faster way to build natively for arm32. We're already using a deployment of this for https://ci.nodejs.org/job/node-test-commit-arm/nodes=ubuntu2004-armv7l after the experiments for #2775  and this will make it formal to make it easier to expand onto other machines.


richardlau (approved):

mhdawson (approved):
LGTM

PR #2912, opened 2022-03-31
ansible: arm cross compiler updates for Node.js 18

richardlau:
Add a new RHEL 8 based container for cross compilation of armv7l
binaries for Node.js 18 using the recently added cross compiler
toolchain based on gcc 8 with glibc 2.28.

Adds a Debian 10 (Buster) container to the Pi's -- even though the Pi's
are running Debian 10 the tests are run in containers and prior to this
commit were running in Debian 9, which is on glibc 2.24.

---

The plan would be, at least initially, to continue testing older versions of Node.js in the stretch container but use the buster container for Node.js 18 onwards. Or do we think stretch is too old now we should move everything?

### DIFF
--- a/ansible/roles/docker/files/cc-selector.sh
+++ b/ansible/roles/docker/files/cc-selector.sh
@@ -56,7 +56,6 @@ function run {
   fi
   export CC="ccache ${!base_varname}gcc ${!flags_varname}"
   export CXX="ccache ${!base_varname}g++ ${!flags_varname}"
-  env | grep ccache
 }
 
 run $1
--- a/ansible/roles/docker/files/cc-selector.sh
+++ b/ansible/roles/docker/files/cc-selector.sh
@@ -7,33 +7,36 @@
 # cross-compiler-ubuntu1604-armv7-gcc-6
 # cross-compiler-ubuntu1804-armv7-gcc-6
 # cross-compiler-ubuntu1804-armv7-gcc-8
+# cross-compiler-rhel8-armv7-gcc-8-glibc-2.28
 
 rpi_newer_tools_base="/opt/raspberrypi/rpi-newer-crosstools/"
 base_4_9_4="${rpi_newer_tools_base}x64-gcc-4.9.4-binutils-2.28/arm-rpi-linux-gnueabihf/bin/arm-rpi-linux-gnueabihf-"
 base_6="${rpi_newer_tools_base}x64-gcc-6.5.0/arm-rpi-linux-gnueabihf/bin/arm-rpi-linux-gnueabihf-"
 base_8="${rpi_newer_tools_base}x64-gcc-8.3.0/arm-rpi-linux-gnueabihf/bin/arm-rpi-linux-gnueabihf-"
+base_8_glibc_2_28="${rpi_newer_tools_base}x64-gcc-8.3.0-glibc-2.28/arm-rpi-linux-gnueabihf/bin/arm-rpi-linux-gnueabihf-"
+
 flags_armv6="-march=armv6zk"
 flags_armv7="-march=armv7-a"
 
 function run {
   local label="$1"
 
-  export arm_type=$(echo $label | sed -E 's/^cross-compiler-ubuntu1[68]04-(armv[67])-gcc-.*$/\1/')
-  export gcc_version=$(echo $label | sed -E 's/^cross-compiler-ubuntu1[68]04-armv[67]-gcc-(4\.9\.4|6|8)/\1/')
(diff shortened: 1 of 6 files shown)


ansible/roles/docker/files/cc-selector.sh
```diff
@@ -44,10 +47,16 @@ function run {
 
   export ARCH="${arm_type}l"
   export DESTCPU=arm
-  export CC_host="ccache gcc-${gcc_host_version} -m32"
-  export CXX_host="ccache g++-${gcc_host_version} -m32"
+  if [ "$host_os" = "rhel8" ]; then
+    export CC_host="ccache gcc -m32"
```
rvagg, line 51: you're sure this is going to be OK? do we not have the option to version this? IIRC we added the host version because it was picking the wrong version because it happened to be the default on the machine, so we went with the version specifier. But maybe that's not an option with rhel in this setup?

rvagg:
Yeah .. the out of date Pi's .. that's a problem and unfortunately it'll be on me to deal with that. I haven't had to touch them for _so_ long, apparently they've been pretty stable. But I really do need to get around to giving them a cleanup and update and I guess setting up the new containers should be on the TODO list since it's so overdue. Leave this one with me, I'm pretty backlogged at the moment but maybe I'll find time to do this in the next ~week as a distraction from everything else.

ansible/roles/docker/files/cc-selector.sh
richardlau: The RHEL 8 option would be GCC toolsets (the equivalent of devtoolsets in CentOS/RHEL 7). In those cases it's the path to the toolset that changes, not the name of the binary. I could add a check if the gcc version is not 8 (the default in RHEL 8) and attempt to load a GCC toolset, although the container hasn't installed any extra ones.

Refs: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/developing_c_and_cpp_applications_in_rhel_8/additional-toolsets-for-development_developing-applications

mhdawson (approved):
LGTM

richardlau:
I'm currently experimenting with armv7l containers running on the Altras as an extra/alternative/fallback means of testing the cross compiled armv7l binaries. I've set up https://ci.nodejs.org/job/node-test-binary-armv7l/ (based on node-test-binary-arm-12+ but fans onto our armv7l containers instead of the Pi's)  and hooked it into my test job,https://ci.nodejs.org/job/richardlau-node-test-commit-arm-fanned/.

If this works out (I'm retesting as ran into https://github.com/nodejs/node/issues/27862) then I propose adding this new job as an addition to run in parallel with the existing node-test-binary-arm-12+. We could temporarily put a condition in the fanned job to prevent running node-test-binary-arm-12+ for Node.js 18 until the Pi's get upgraded.

richardlau:
https://ci.nodejs.org/job/node-test-binary-armv7l/ looks to be working (tried Node.js 12, 14, 16 and master). I've gone ahead and added it to https://ci.nodejs.org/job/node-test-commit-arm-fanned/ and will keep an eye on the job. We are down to just two online Pi 2's in the CI (https://github.com/nodejs/build/issues/2914) which is causing jobs to back up so having the containerized armv7l testing in place would be a good backup in case we'd need to temporarily disable https://ci.nodejs.org/job/node-test-binary-arm-12+/ in general (and not just for 18 until the containers on the Pi's are updated).

Marking this as ready to review on the basis that this could land and we switch over the cross compiler and test in the new armv7l container job now for Node.js 18 with the view to also test on the Pi's later when they are updated (i.e. we wouldn't be blocked by the Pi's).

ansible/roles/docker/files/cc-selector.sh
richardlau: I've updated this to attempt to load a gcc-toolset if the host gcc version doesn't match.

richardlau:
Based on https://github.com/nodejs/build/issues/2914#issuecomment-1090983168, I'm going to go ahead with the backup plan (https://github.com/nodejs/build/pull/2912#issuecomment-1086160551) of merging this so we're building with the updated toolchain but temporarily prevent https://ci.nodejs.org/job/node-test-binary-arm-12+/ from running on Node.js 18/master until the Pi's get updated. We'll have armv7l coverage in the https://ci.nodejs.org/job/node-test-binary-armv7l/ (armv7l containers on the Ampere Altras) which is already running on PR's.

I'll do this tomorrow (Friday 8 April) unless I head objections.

richardlau:
FTR this is the condition I've put on https://ci.nodejs.org/job/node-test-binary-arm-12+/ in https://ci.nodejs.org/job/node-test-commit-arm-fanned/:

image

We can remove this condition when the Pi's are updated.

Issue #2837, opened 2021-12-22
PSA - added node-test-commit-arm-debug to node-test-commit


mhdawson:
Added  node-test-commit-arm-debug job to node-test-commit as next step in: https://github.com/nodejs/node/issues/41209.

It's configured to only run the build/test > 17, and otherwise just pass as a job.

Will close this once we turn off the x86 debug builds and all has run ok for a few days.

mhdawson:
I should have also noted that it will show as yellow until we investigate: https://github.com/nodejs/node/issues/41204

mhdawson:
Updated configuration for 
https://ci.nodejs.org/job/node-test-commit-linux-containered/

so that the debug build only runs if major version is less than or equal to 17

richardlau:
We usually do exclusions based on Node.js release lines in https://github.com/nodejs/build/blob/master/jenkins/scripts/VersionSelectorScript.groovy.

mhdawson:
As far as I understand the groovy selector allows us to not run based on machine tag (ex ubuntu2004), not sure how that translates to this case.  

If you have  suggestion for an alternate change to the groovy script that would help illustrate what you are suggesting.

richardlau:
So for example https://github.com/nodejs/build/blob/1018f9f415cff394b627c89c3c7744b9fcd8a359/jenkins/scripts/VersionSelectorScript.groovy#L103 means that the dynamically linked OpenSSL 3 bit of linux-containered doesn't run for Node.js versions earlier than 15.
https://github.com/nodejs/build/blob/1018f9f415cff394b627c89c3c7744b9fcd8a359/jenkins/scripts/VersionSelectorScript.groovy#L107 says to not run dynamically linked OpenSSL 1.0.2 for Node.js 10 and later.

So we can add a rule for `sharedlibs_debug` or `sharedlibs_debug_x64` if you want to be more specific (it's a regexp match afterall).

mhdawson:
Ok we can update the groovy script for the part related to the x86 containers. Interestingly if you look at the actual contained job, there was already a check for > version 8. I just added the second check for <= 17

![image](https://user-images.githubusercontent.com/9373002/147163346-bf9bc080-045b-4c15-96ab-92f699c4f428.png)

richardlau:
I'd imagine the existing condition predates the introduction of the groovy script.

richardlau:
[...]

PR #2925, opened 2022-04-13
jenkins: only run sharedlibs_debug_x64 pre-18

richardlau:
Full debug x64 builds have been unreliable in the CI in the sharedlibs
containers and were recently "turned off" by making the job a no-op.
However the job was still being started even if it didn't do anything
but pass -- this change prevents Jenkins from starting 
`sharedlibs_debug_x64` runs.

Refs: https://github.com/nodejs/build/issues/2837

---

cc @mhdawson 

We may end up reintroducing a "lighter" debug build (suggested in https://github.com/nodejs/node/issues/41209#issuecomment-1001261679) later but that would probably be better in its own job and not part of node-test-commit-linux-containered.


mhdawson (approved):
LGTM

Issue #2984, opened 2022-07-01
Jenkins will require Java 11 from September LTS release


richardlau:
We probably need to start planning for this:

https://www.jenkins.io/blog/2022/06/28/require-java-11/
> Beginning with Jenkins 2.357 (released on June 28, 2022) and the forthcoming September LTS release, Jenkins requires Java 11. 

Both CI servers are currently running Java 8. 

cc @nodejs/build-infra

richardlau:
Looking at the release schedules (https://github.com/nodejs/Release/issues/737, https://github.com/nodejs/Release/issues/658, https://github.com/nodejs/Release/issues/567) I'm going to plan to do the updates on 25 Aug 2022, with the proviso that if a security release is needed/happens during August that requires a CI lockdown I would shift doing the updates to between after the security release and before lifting the CI lockdown.

richardlau:
Maintenance has started. I've shutdown the test Jenkins CI, powered off the server and taken a snapshot in DigitalOcean (`infra-digitalocean-ubuntu14-x64-1-1661424784910`).

richardlau:
As a (self-)reminder: just before shutting down the server I unselected all "build" (can start builds) permissions to prevent builds from being started during the updates and will need to put them back when the maintenance is over.

richardlau:
I've managed to upgrade the server from Ubuntu 16.04->18.04->20.04->22.04.1. Had a nameserver issue, which we've run into before with DO droplets and required the same fix as https://github.com/nodejs/build/pull/2745#issuecomment-911842863. Working on switching to Java 17 now.

richardlau:
Also the initial update to Ubuntu 18.04 still had the server on a very old kernel:
```
Welcome to Ubuntu 18.04.6 LTS (GNU/Linux 3.13.0-57-generic x86_64)
```
There's a "Kernel" page for the droplet in DigitalOcean which I switched to "Grubloader" and after a restart the server then used the expected kernel for the Ubuntu release. We're now on:
```
Welcome to Ubuntu 22.04.1 LTS (GNU/Linux 5.15.0-46-generic x86_64)
```
![image](https://user-images.githubusercontent.com/5445507/186712610-ed064072-2bdb-46c3-8810-3bb837c1d76f.png)

richardlau:
Have had to apply https://github.com/afonsof/jenkins-material-theme/issues/193. It looks like the theme were using has some issues with the latest LTS Jenkins release 😞 .

[...]

PR #3019, opened 2022-08-25
jenkins: add workaround for groovy script & Java 11+

richardlau:
The Matrix Groovy Execution Strategy plugin appears to have an issue
on Java 11+ where string parameters are appearing as byte arrays
instead of strings. Use the new String() constructor instead of
toString() to ensure we get string repesentation of the parameter.

Refs: https://github.com/jenkinsci/matrix-groovy-execution-strategy-plugin/issues/20
Refs: https://github.com/nodejs/build/issues/2984#issuecomment-1227537524

We need this to unblock the Java update.


BethGriggs (approved):
Not in build, but an approval if it helps unblock the update

AshCripps (approved):

Issue #3023, opened 2022-08-29
citgm-smoker-pipeline broken by Jenkins update


targos:
Example:
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/190/
https://ci.nodejs.org/job/citgm-smoker-nobuild/1365/console

```
08:12:19 Building remotely on [test-packetnet-ubuntu1804-x64-1](https://ci.nodejs.org/computer/test-packetnet-ubuntu1804-x64-1) (Ubuntu amd64-Ubuntu jenkins-workspace 18.04 amd64-Ubuntu-18.04 Ubuntu-18.04 amd64) in workspace /home/iojs/build/workspace/citgm-smoker-nobuild
08:12:19 using NODE_VERSION: [118, 49, 52]
08:12:19 sliced: 118, 49, 52]
08:12:19 split : [118,, 49,, 52]]
08:12:19 FATAL: For input string: "118,"
08:12:19 java.lang.NumberFormatException: For input string: "118,"
08:12:19 	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
```

targos:
@nodejs/jenkins-admins

sxa:
https://github.com/nodejs/build/pull/3019/files was the fix used to fix what I believe was the same issue elsewhere so it's likely we'll need to change something to `new String(...)` here too.

sxa:
Fix ready. I've modified the offending job to run from my branch in the meantime so that it hopefully won't constantly run red, but I'm still running some tests to verify that it does what it should.

PR #3025, opened 2022-09-01
Workaround for groovy script & Java 11+ in `citgm-smoker-nobuild`

sxa:
This is a follow on to commit dce00d089e6266dd838eabec38d75e4f8dcd627e which doesn't work with the citgm-smoker-nobuild job where the default value is `v18` instead of `19.0.0` so it doesn't parse properly with the original solution. This almost certainly needs a bit of a rethink to get the value passed in in a consistent format if possible.

Fixes https://github.com/nodejs/build/issues/3023 (Subject to disclaimer in the last line of this description)

Looks to have broken the ability to run the citgm jobs. Once I fixed another issue in the citgm-smoker-nobuild job:
```
// SXA build#3023 // def verString = parameters['NODE_VERSION'].toString()
  def verString = new String(parameters['NODE_VERSION'])
```
I hit this. I've patched the job to run from my branch just now.

NOTE: It is unclear yet whether this fix is now breaking the parsing of the file and stopping it from only selecting the correct systems, but it's likely preferable as a workaround to the job failing to run at all.

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -158,7 +158,11 @@ def canBuild = { nodeVersion, builderLabel, buildType ->
 
 int nodeMajorVersion = -1
 if (parameters['NODEJS_MAJOR_VERSION'])
-  nodeMajorVersion = new String(parameters['NODEJS_MAJOR_VERSION']).toInteger()
+  if (parameters['NODEJS_MAJOR_VERSION'] instanceof Integer) {
+    nodeMajorVersion = parameters['NODEJS_MAJOR_VERSION']
+  } else {
+    nodeMajorVersion = (new String(parameters['NODEJS_MAJOR_VERSION'])).toInteger()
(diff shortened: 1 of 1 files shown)


sxa:
This initial patch breaks [node-test-commit-linux](https://ci.nodejs.org/job/node-test-commit-linux), but fixes c[itgm-smoker-nobuild](https://ci.nodejs.org/job/citgm-smoker-nobuild/) so I'm going to continue to experiment with it ...

sxa:
OK I've got copies of both jobs working now (Well, getting to actually running jobs - will wait for any approvals/objections then push it up. I've modified [citgm-smoker-nobuild](https://ci.nodejs.org/job/citgm-smoker-nobuild/) to run from this branch for now because, frankly, it's better to have it at least attempt to build something than constantly run red.

sxa:
Friday afternoon merge - what could go wrong? ;-)
I'll update citgm-smoker-nobuild to point to the `main` branch again

PR #3033, opened 2022-09-17
add redirects for release schedule to website

Trott:
Refs: https://github.com/nodejs/nodejs.org/pull/4821

### DIFF
--- a/ansible/www-standalone/resources/config/nodejs.org
+++ b/ansible/www-standalone/resources/config/nodejs.org
@@ -368,7 +368,8 @@ server {
     rewrite ^/(20\d\d/\d\d/\d\d/.*)$                              http://blog.nodejs.org/$1 permanent;
 
     rewrite ^/about/?$                                            https://$server_name/en/about/ permanent;
-    rewrite ^/about/releases/?$                                   https://$server_name/en/about/releases/ permanent;
+    rewrite ^/about/releases/?$                                   https://github.com/nodejs/release#release-schedule permanent;
+    rewrite ^/en/about/releases/?$                                https://github.com/nodejs/release#release-schedule permanent;
     rewrite ^/about/resources/?$                                  https://$server_name/en/about/resources/ permanent;
     rewrite ^/about/security/?$                                   https://github.com/nodejs/node/blob/HEAD/SECURITY.md#security permanent;
     rewrite ^/about/trademark/?$                                  https://$server_name/en/about/trademark/ permanent;


ansible/www-standalone/resources/config/nodejs.org
```diff
@@ -368,7 +368,8 @@ server {
     rewrite ^/(20\d\d/\d\d/\d\d/.*)$                              http://blog.nodejs.org/$1 permanent;
 
     rewrite ^/about/?$                                            https://$server_name/en/about/ permanent;
-    rewrite ^/about/releases/?$                                   https://$server_name/en/about/releases/ permanent;
+    rewrite ^/about/releases/?$                                   https://github.com/nodejs/release#release-schedule permanent;
+    rewrite ^/en/about/releases/?$                                https://github.com/nodejs/release#release-schedule permanent;
```
nschonni, line 372: is there a why to catch the i18n versions too, instead of just /en/

Trott: I looked through the rest of the config file and saw lots of `en` redirects but nothing that looked like it would catch other languages. I agree it would be a good thing if that were a feature in this.....

nschonni: I think most of the ones in here were for before we had any i18n stuff in the original structure. I don't know the nginx syntax, but my basic guess would be some sort of regex like `rewrite ^/*/about/releases/?$`

bcoe (approved):

GeoffreyBooth (approved):
RSLGTM

Trott:
@nodejs/build If this lands, who would push it to production? I'm guessing @richardlau but maybe someone else? @mhdawson perhaps? I'd want to get an approval on this change from at least one person who might be responsible for pushing it to production.

ansible/www-standalone/resources/config/nodejs.org
Trott: That's my gut too, but I'd want to consult the docs. Since there are already a bunch of `en`-specific redirects in the file, I'm inclined to consider "work for other languages" to be an enhancement that can go in a separate pull request. (Feel free to open it if you feel pretty sure about the syntax!)

rvagg: are you sure the `#` is going to be happy in here unescaped? might be worth double-checking that's OK

richardlau (approved):

richardlau:
I've deployed this -- it appears to be working (may need to clear your browser cache but I did purge the CloudFlare cdn cache).

Issue #3030, opened 2022-09-14
Jenkins agents need newer Java (11+)


richardlau:
It looks like after updating to the latest Jenkins LTS (https://www.jenkins.io/security/advisory/2022-09-09/) several agents are failing to connect to the server.
e.g. example failure from one of the agent logs:
```
WARNING: LinkageError while performing UserRequest:hudson.slaves.SlaveComputer$SlaveVersion@435cd082
java.lang.UnsupportedClassVersionError: Failed to load hudson.slaves.SlaveComputer$SlaveVersion
...
Caused by: java.lang.UnsupportedClassVersionError: JVMCFRE003 bad major version; class=hudson/slaves/SlaveComputer$SlaveVersion, offset=6
...
Sep 14, 2022 11:38:51 AM hudson.remoting.jnlp.Main$CuiListener status
INFO: Terminated
```

Looks like now that Jenkins has dropped support for Java 8 (https://github.com/nodejs/build/issues/2984) the agents also require Java 11+.

richardlau:
For the time being I've rolled back the Jenkins update by backing up the `jenkins.war` file on the test CI server and copying over the one from the release server (which hadn't been updated yet). We'll need to update eventually but this will buy some time to get Java updated on the test machines.

richardlau:
Via `ansible --list-hosts "test-**" | cut -d "-" -f 3 | sort -u` these are the OS's we currently have in the test matrix:

os | current Java | most recent Java from distro | Notes
---|---|---|---
aix71 | 8 | - | IBM Semuru Java 11/17 fails to run. Needs further investigation.
aix72 | ~8~ ➡️ 17 | - | IBM Semuru
aix73 | ~8~ ➡️ 17 | - | IBM Semuru
centos7 | ~8~ ➡️  11 | 11
debian10 | 11 | 11
debian8 | 8 | 8
debian9 | 8 | 8
fedora32
fedora34
freebsd12
ibmi73 | 11 | - |
macos10.14
macos10.15
macos11
macos11.0
rhel7 | 11 | 11
rhel8 | 17
smartos18
smartos20
ubuntu1404
ubuntu1604
ubuntu1804
ubuntu2004
win10
win2012r2
win2016
zos24
```

richardlau:
Just finished updating the Windows machines. Remaining platforms:
- macos
- smartos
- End-of-Life Linux (e.g. debian 8) which should be removed from CI

targos:
[...]

PR #3047, opened 2022-10-05
ansible: update to Java 11/17

richardlau:
Refs: https://github.com/nodejs/build/issues/3030

This PR moves a subset of out test infrastructure off Java 8 onto either Java 11 or 17 (depending on what is available from the OS package repositories) to allow us to update Jenkins in the future. Remaining platforms will be dealt with later.


ansible/roles/package-upgrade/vars/main.yml
```diff
@@ -8,7 +8,7 @@
     'yum': ['centos', 'rhel7', 'aix71', 'aix72', 'ibmi'],
     'apt': ['debian', 'ubuntu'],
     'dnf': ['aix73', 'fedora', 'rhel8'],
-    'pkg': 'freebsd',
+    'pkgng': 'freebsd',
```
richardlau, line 11: FWIW I needed to change this to get the Jenkins worker create Ansible playbook to run against our FreeBSD 12 machines otherwise I was getting errors:
```
Could not find a module for pkg.
```
from Ansible (2.11.2).

mhdawson (approved):
LGTM

Since we don't build/release using XLC (we use GCC instead) I don't think the update of the runtime will have any effecton the releases.

Issue #3028, opened 2022-09-08
Legacy Equinix Metal data facility closures on November 30th, 2022


richardlau:
We've been informed that Equinix are closing their legacy data facilities on November 30th, 2022. This will affect all of the instances we currently have hosted there and we'll need to move to "go-forward Equinix-based environments".

Currently we're spread over two projects -- Node.js and Works on ARM. The Linux ARM64 release machines are older ARM64 hardware (c2.large.arm) that are only in the legacy data facilities and are most at risk.

### Node.js
host | type | desc
---|---|---
release-equinix-ubuntu2004-docker-arm64-1 | c2.large.arm | Linux ARM64 release machine
release-equinix-ubuntu2004-docker-arm64-2 | c2.large.arm | Linux ARM64 release machine

### Works on ARM:
host | type | desc
---|---|---
test-equinix-ubuntu2004-arm64-1 | x.large.arm | Linux ARM64 test machine
test-equinix-ubuntu2004-arm64-3 | x.large.arm | Linux ARM64 test machine
test-packetnet-ubuntu1804-x64-1 | c1.small.x86 | jenkins-workspace-4
test-packetnet-ubuntu1804-x64-2 | c1.small.x86 | jenkins-workspace-5

## Actions

- [x] Contact Equinix to discuss migration plan for the ARM hardware --  Email has been sent to @vielmetti
- [ ] Migrate ARM64 release machines
- [ ] Migrate ARM64 test machines
- [ ] Migrate the two Jenkins workspace machines from Works on ARM into Node.js

vielmetti:
Thanks @richardlau . Here's what I know so far.

The Jenkins machines should be the easiest to move of this bunch. We have a "c3.small" machine type which is of comparable size which should do the trick.

For the x.large.arm systems, the team is working on a migration process that will likely result in us doing physical machine migration on your behalf, moving the systems from DFW2 to one of the DA data centers. Some of those details are TBD, so as of right now I an awaiting instructions. In an ideal situation you would see a little bit of downtime for the physical move, but no need to reconfigure.

For the two release machines (c2.large.arm) my recommendation at the moment is to migrate those services to another provider. We are decommissioning the Ampere eMag hardware these live on, and those machines are not being moved to new facilities.  Our newer c3.large.arm are very similar to the x.large.arm systems you currently are using through Works on Arm - large bare metal systems with high core counts - which makes them a poor fit to be deployed as release-only machines.

[...]

PR #3051, opened 2022-10-11
ansible: add new OSUOSL arm64 release machines

richardlau:
Add two new OSUOSL hosted arm64 release machines to replace the Equinix ones that are going away at the end of November 2022.

Updates the playbook for centos7 to drop devtoolset-6 which is no longer required since Node.js 12 went End-of-Life (Node.js 14 and 16 use devtoolset-8 and Node.js 18 and later are not built on centos7).

Refs: https://github.com/nodejs/build/issues/3028


mhdawson (approved):
LGTM

PR #3067, opened 2022-11-03
template: update zlib to 1.2.13

RafaelGSS:
`curl -sL https://zlib.net/zlib-1.2.12.tar.gz | tar zxv --strip=1` is 404


richardlau (approved):

AshCripps (approved):

mhdawson (approved):
LGTM

richardlau:
@sxa has pointed me to https://www.zlib.net/fossils/ which appears to have all of the releases -- zlib appear to remove the previous version from https://zlib.net/ when a new one is released. We could point to the tarball from https://www.zlib.net/fossils/ so that the Dockerfile doesn't break from underneath us in the future. (We should still aim to periodically update zlib and the other things in the container, but at our own pace when we are ready to.)

mhdawson (approved):
LGTM

PR #3076, opened 2022-11-13
ansible: add new OSUOSL arm64 docker test host

richardlau:
Refs: https://github.com/nodejs/build/issues/3028

---

Existing arm64 docker test hosts at Equinix are being migrated to a new data facility. This host at OSUOSL will hopefully tide us over until the migration completes, and afterwards be available as a fallback.


richardlau:
The first commit is a follow up to https://github.com/nodejs/build/pull/3067 -- we need to update the version in the ubuntu2004_sharedlibs container as well.

mhdawson (approved):
LGTM

PR #3062, opened 2022-10-31
ansible: pin Python 3 to 3.10.8

richardlau:
Node.js 14 is incompatible with Python 3.11. Pin Python 3 to 3.10.8.
Run Jenkins agent on Java 17.

---

Updated Windows CI hosts:
- [x]    release-rackspace-win2012r2-x64-1
- [x]    release-rackspace-win2012r2-x64-2
- [x]    release-rackspace-win2012r2_vs2019-x64-1
- [x]    release-rackspace-win2012r2_vs2019-x64-2
- [x]    test-azure_msft-win10_vs2019-x64-1
- [x]    test-azure_msft-win10_vs2019-x64-2
- [x]    test-azure_msft-win10_vs2019-x64-3
- [x]    test-azure_msft-win10_vs2019-x64-4
- [x]    test-azure_msft-win2016_vs2017-x64-1
- [x]    test-azure_msft-win2016_vs2017-x64-2
- [x]    test-azure_msft-win2016_vs2017-x64-3
- [x]    test-azure_msft-win2016_vs2017-x64-4
- [x]    test-azure_msft-win2016_vs2017-x64-5
- [x]    test-azure_msft-win2016_vs2017-x64-6
- [ ]    test-msft-win10_vs2017-arm64-1
- [ ]    test-msft-win10_vs2017-arm64-2
- [ ]    test-nearform_arm-win10_vs2019-arm64-1
- [ ]    test-nearform_arm-win10_vs2019-arm64-2
- [x]    test-rackspace-win2012r2_vs2013-x64-1
- [x]    test-rackspace-win2012r2_vs2013-x64-2
- [x]    test-rackspace-win2012r2_vs2015-x64-1
- [x]    test-rackspace-win2012r2_vs2015-x64-2
- [x]    test-rackspace-win2012r2_vs2017-x64-1
- [x]    test-rackspace-win2012r2_vs2017-x64-2
- [x]    test-rackspace-win2012r2_vs2017-x64-3
- [x]    test-rackspace-win2012r2_vs2017-x64-4
- [x]    test-rackspace-win2012r2_vs2019-x64-1
- [x]    test-rackspace-win2012r2_vs2019-x64-2
- [x]    test-rackspace-win2012r2_vs2019-x64-3
- [x]    test-rackspace-win2012r2_vs2019-x64-4
- [x]    test-rackspace-win2012r2_vs2019-x64-5
- [x]    test-rackspace-win2012r2_vs2019-x64-6

<details>
[...]

PR #3085, opened 2022-11-20
ansible: install temurin JDK on macOS

targos:
Refs: https://github.com/nodejs/build/issues/3030

### DIFF
--- a/ansible/roles/package-upgrade/tasks/partials/brew.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/brew.yml
@@ -13,7 +13,7 @@
     script: files/install-xcode.sh
     when: xcode.rc > 1
 
-  - name: Check if Homebrew is already installed (Apple Sillicon)
+  - name: Check if Homebrew is already installed (Apple Silicon)
     stat:
       path: /opt/homebrew/bin/brew
     register: armbrew
--- a/ansible/roles/package-upgrade/tasks/partials/brew.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/brew.yml
@@ -45,6 +45,7 @@
     become_user: "{{ ansible_user }}"
     homebrew_tap:
       name: buo/cask-upgrade
+
   - name: Install brew cask-versions
     become_user: "{{ ansible_user }}"
     homebrew_tap:
--- a/ansible/roles/package-upgrade/tasks/partials/brew.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/brew.yml
@@ -45,6 +45,11 @@
     become_user: "{{ ansible_user }}"
     homebrew_tap:
       name: buo/cask-upgrade
+  
+  - name: Install brew cask-versions
+    become_user: "{{ ansible_user }}"
+    homebrew_tap:
+      name: homebrew/cask-versions
 
   - name: Update Casks
     become_user: "{{ ansible_user }}"
--- a/ansible/roles/package-upgrade/tasks/partials/brew.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/brew.yml
@@ -45,7 +45,6 @@
     become_user: "{{ ansible_user }}"
     homebrew_tap:
       name: buo/cask-upgrade
-  
   - name: Install brew cask-versions
     become_user: "{{ ansible_user }}"
     homebrew_tap:
--- a/ansible/roles/package-upgrade/tasks/partials/brew.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/brew.yml
@@ -8,6 +8,7 @@
     raw: xcode-select --print-path &> /dev/null
     register: xcode
     ignore_errors: yes
+    check_mode: no
 
   - name: Install xcode-tools
     script: files/install-xcode.sh
@@ -44,11 +45,6 @@
     become_user: "{{ ansible_user }}"
     homebrew_tap:
       name: buo/cask-upgrade
-
-  - name: Add AdoptOpenJDK Java Repo
-    become_user: "{{ ansible_user }}"
-    homebrew_tap:
-      name: AdoptOpenJDK/openjdk
 
   - name: Update Casks
     become_user: "{{ ansible_user }}"
--- a/ansible/roles/java-base/tasks/main.yml
+++ b/ansible/roles/java-base/tasks/main.yml
@@ -51,31 +51,12 @@
     # Package manager mapping in ansible/roles/package-upgrade/vars/main.yml.
     use: "{{ os|match_key(pm)|default(omit) }}"
 
-- name: install java tap (macOS)
-  become_user: administrator
-  when: java.rc > 0 and os|startswith("macos") and arch == "x64"
-  homebrew_tap:
-        name: AdoptOpenJDK/openjdk
-        state: present
-
 - name: install java (macOS)
   become_user: administrator
-  when: java.rc > 0 and os|startswith("macos") and arch == "x64"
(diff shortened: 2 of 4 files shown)


targos:
I have never used Ansible before and I don't to break everything.
What's the safest way to test this?

Also, since this is completely changing how the package is installed, are we supposed to uninstall the previous stuff? How is that done?

richardlau:
> I have never used Ansible before and I don't to break everything. What's the safest way to test this?

Ansible has a [`--check` mode](https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html) which will attempt to run the playbook without making changes. The caveat here is whether we're using raw commands in our playbook and then depending on the output of those commands -- Ansible doesn't know whether those commands have side effects so doesn't run them in check mode and then the things that depend on them subsequently fail. I've fixed up some of these in the past (https://github.com/nodejs/build/pull/2682) but it's possible more have crept in and would need to be fixed up.

Also [`ansible-playbook` has ` --limit` (or `-l`) option](https://docs.ansible.com/ansible/latest/user_guide/intro_patterns.html#patterns-and-ansible-playbook-flags) to restrict which host the playbook is running against. So, for example, 
```
ansible-playbook --limit "test-nearform-macos10.15-x64-1" ansible/playbooks/jenkins/worker/create.yml
```
would only run the playbook against `test-nearform-macos10.15-x64-1`. Limiting to one host first will contain the changes in case something goes awry -- once you're confident the changes are correct you can then rerun with a wider glob for `--limit` (or do the other machines individually).

> Also, since this is completely changing how the package is installed, are we supposed to uninstall the previous stuff? How is that done?

You can run one-off individual Ansible tasks with `ansible`, 
e.g. here's one of the commands I was using for https://github.com/nodejs/build/pull/3062 to uninstall the python chocolatey package on a subset of the Windows machines prior to running the playbook:
```
ansible -m win_chocolatey -a "name=python state=absent remove_dependencies=yes" -v test-rackspace-win2012r2_vs2015-x64-?
```
You can also just ssh into the machine directly and remove the old package -- the idea of the Ansible scripts is that we can run them against a new/fresh image (which presumably will not have the old package) and the script should put the image into a usable state.

targos:
I took https://ci.nodejs.org/computer/test-nearform-macos11.0-arm64-1/ offline to update it.

targos:
Failed:

```
TASK [java-base : Check if java is already installed] ******************************************************************
[WARNING]: raw module does not support the environment keyword
fatal: [test-nearform-macos11.0-arm64-1]: FAILED! => {"changed": false, "msg": "non-zero return code", "rc": 1, "stderr": "Shared connection to 83.147.191.75 closed.\r\n", "stderr_lines": ["Shared connection to 83.147.191.75 closed."], "stdout": "The operation couldn’t be completed. Unable to locate a Java Runtime.\r\nPlease visit http://www.java.com for information on installing Java.\r\n\r\n", "stdout_lines": ["The operation couldn’t be completed. Unable to locate a Java Runtime.", "Please visit http://www.java.com for information on installing Java.", ""]}
...ignoring

TASK [java-base : install java] ****************************************************************************************
skipping: [test-nearform-macos11.0-arm64-1]

TASK [java-base : install java (macOS)] ********************************************************************************
fatal: [test-nearform-macos11.0-arm64-1]: FAILED! => {"msg": "Failed to set permissions on the temporary files Ansible needs to create when becoming an unprivileged user (rc: 1, err: chmod: Invalid file mode: A+user:administrator:rx:allow\n}). For information on working around this, see https://docs.ansible.com/ansible-core/2.11/user_guide/become.html#becoming-an-unprivileged-user"}
```

targos:
Oh I see, the problem is that on this mac, the administrator account is called `admin`, not `administrator`. Is it possible to `become_user` with a variable somehow?

targos:
Can I use `become_user: {{ ansible_user }}`? I'm wondering why it was hardcoded in this place.

AshCripps:
you are going to hit a caveat on some machines where the user is sometimes `admin` and other times `administrator` IIRC any nearform machine we didnt setup is `admin` and any machine I created is `administrator`. 

My work around was to just change the user to admin run the ansible and then change it back.

targos:
Install looks good!

```
admin@test-nearform-macos11 ~ % java -version
openjdk version "19.0.1" 2022-10-18
OpenJDK Runtime Environment Temurin-19.0.1+10 (build 19.0.1+10)
OpenJDK 64-Bit Server VM Temurin-19.0.1+10 (build 19.0.1+10, mixed mode)
```

targos:
I also updated `test-nearform-macos10.15-x64-1` and `test-nearform-macos10.15-x64-2`.
Now before we go further, are we OK with installing the latest Java or do we want to stay on LTS? For that, I'd have to tap `homebrew/cask-versions` and install `temurin17` instead of `temurin`.

richardlau:
> Can I use `become_user: "{{ ansible_user }}"`? I'm wondering why it was hardcoded in this place.

Yes, that should work. We use it elsewhere, e.g. https://github.com/nodejs/build/blob/b02d18741a07418eaeb53763b0ac6b5df9d7f936/ansible/roles/package-upgrade/tasks/partials/brew.yml#L28-L29

sxa:
> > Now before we go further, are we OK with installing the latest Java or do we want to stay on LTS?
> @sxa thoughts?

I'd stick with LTS (17) personally for this.

targos:
I have an unrelated problem with `test-orka-macos11-x64-2`:

```
failed: [test-orka-macos11-x64-2] (item=cmake,python,ccache) => {"ansible_loop_var": "package", "changed": false, "msg": "Warning: python@3.9 3.9.13_1 already installed.", "package": "cmake,python,ccache"}
```

targos:
# Update status

## Updated to Java 17

- release-nearform-macos10.15-x64-1
- release-nearform-macos11.0-arm64-1
- release-orka-macos10.15-x64-1
- test-nearform-macos10.15-x64-1
- test-nearform-macos10.15-x64-2
- test-nearform-macos11.0-arm64-1
- test-orka-macos10.14-x64-1
- test-orka-macos10.14-x64-2
- test-orka-macos10.14-x64-3
- test-orka-macos11-x64-1
- test-orka-macos11-x64-2

## Impossible to connect via SSH

- release-macstadium-macos11.0-arm64-1
- test-macstadium-macos11.0-arm64-3
- test-macstadium-macos11.0-arm64-4

## Offline

- test-nearform-macos10.15-x64-3
- test-orka-macos10.15-x64-1
- test-orka-macos10.15-x64-2

targos:
@AshCripps What am I missing to connect to the macstadium machines?
This is what I get when I try to ssh:

```
$ ssh test-macstadium-macos11.0-arm64-3
Connection closed by 207.254.38.86 port 22
```

AshCripps:
@targos I think theyve been down for a while - Ill raise a ticket with macstadium to investigate.

AshCripps:
SERVICE-146218 raised. It seems they have updated their portal and we can do things like hard reboots etc bu tthat requires a setting to be turned that I dont know if we have so Ive just raised a ticket for now.

ansible/roles/package-upgrade/tasks/partials/brew.yml
```diff
@@ -44,11 +45,11 @@
     become_user: "{{ ansible_user }}"
     homebrew_tap:
       name: buo/cask-upgrade
-
-  - name: Add AdoptOpenJDK Java Repo
+
```
targos, line 48:
```suggestion

```

richardlau (approved):

PR #3086, opened 2022-11-22
fix: Anchored ubuntu version to support Python 2.7 in Github Actions

UlisesGascon:
### Main changes

- Update Github actions runner from `ubuntu-latest` to `ubuntu-20.04`

### Context
- [See discussion around ubuntu versions](https://github.com/actions/setup-python/issues/543)
- [Fail detected in nodejs/build](https://github.com/nodejs/build/actions/runs/3519497549/jobs/5899537570)

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -2,7 +2,7 @@ name: Lint
 on: [push, pull_request]
 jobs:
   Python:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       max-parallel: 2
       matrix:


targos:
Can we try first with setup-python@v4?

UlisesGascon:
@targos I created an alternative version at #3090 by upgrading the actions, but seems like [is using the newer ubuntu version under the hood](https://github.com/nodejs/build/actions/runs/3521134233/jobs/5902687305)

targos:
Ok, thanks for trying.

targos (approved):

richardlau:
> @targos I created an alternative version at #3090 by upgrading the actions, but seems like [is using the newer ubuntu version under the hood](https://github.com/nodejs/build/actions/runs/3521134233/jobs/5902687305)

Yes, GH is in the middle of migrating `ubuntu-latest` to 22.04: https://github.com/actions/runner-images/issues/6399.

We should probably be updating to setup-python@v4 regardless (maybe a separate PR?).

richardlau (approved):
Let's merge this and fix the workflows.

We might subsequently consider dropping Python 2.7 from the workflow matrix. I think https://github.com/nodejs/build/pull/2659 effectively means you need to be running Python 3 to run our playbooks anyway.

UlisesGascon:
I will do the merge and then two separate PRs for:
- Upgrade to `setup-python@v4`
- Remove support for Python2

PR #3094, opened 2022-11-22
chore: add Dependabot for GitHub Actions

nschonni:
Add Dependabot to update the Actions files. They are already mostly pinned to major versions, so there shouldn't be much noise from this.
Branch filter on the one job is to prevent double builds for the Dependabot branches


UlisesGascon (approved):
LGTM!

richardlau (approved):

PR #3113, opened 2022-12-13
jenkins: handle no matches in download-test.sh

richardlau:
If no matches for `NODE_VERSION` are found, bail out of the script. Previously if no matches were found the following rsync would attempt to download the entire `downloadDir`, e.g., for `release`, every release we've done!

---

See https://ci.nodejs.org/job/validate-downloads/2011/console which ran out of disk space trying to rsync everything in `release` because mirroring from www to unencrypted was broken (due to the IP change for unencrypted -- I've now updated the firewall on www for the unencrypted's new IP address) and the new versions were not mirrored to unencrypted.

### DIFF
--- a/jenkins/download-test.sh
+++ b/jenkins/download-test.sh
@@ -30,6 +30,11 @@ fi
 # If the provided NODE_VERSION is inexact, e.g. `v8`, get the most recent matching version.
 LINK=`rsync rsync://unencrypted.nodejs.org/nodejs/$downloadDir/ | grep $NODE_VERSION | sort -t. -k 1,1n -k 2,2n -k 3,3n | tail -1 | awk '{print $5}'`
 
+if [ -z "$LINK" ]; then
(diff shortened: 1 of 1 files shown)


richardlau:
[Failing CI](https://github.com/nodejs/build/actions/runs/3688155776/jobs/6242625020) is due to a new version of [`shellcheck` (0.9.0)](https://github.com/koalaman/shellcheck/blob/v0.9.0/CHANGELOG.md#v090---2022-12-12) being released two days ago which has introduced [a new check](https://github.com/koalaman/shellcheck/commit/642ad8612597b28af4026bde289985583fddb69d) which an existing script unrelated to this one fails. I've opened https://github.com/nodejs/build/pull/3115 to fix it.

UlisesGascon (approved):

mhdawson (approved):
LGTM

PR #3115, opened 2022-12-15
jenkins: fix shellcheck linter CI

richardlau:
Recently released shellcheck 0.9.0 adds a new check:
```
SC2317 (info): Command appears to be unreachable. Check usage (or
ignore if invoked indirectly).
```

Fix the one instance of this that is causing the CI to now fail with the updated version of shellcheck.

Refs: https://github.com/koalaman/shellcheck/commit/642ad8612597b28af4026bde289985583fddb69d

### DIFF
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -60,7 +60,6 @@ case $NODE_NAME in
         return
         ;;
     esac
-    return
     ;;
 esac
 


AshCripps (approved):

nschonni:
Might be good to land https://github.com/nodejs/build/pull/2833 so some of the other rules can start to be cleaned up too

Issue #3108, opened 2022-12-07
Migrate `*-joyent` machines to new Equinix system. OUTAGE required on 8th December


sxa:
Spinning this out of https://github.com/nodejs/build/issues/3104 since the existing machines are currently back online temporarily.

In February 2021 some of our systems were migrated to from Joyent's data centers to Equinix using an account managed by Joyent team members which was separate from our existing one at Equinix.

Recently it became apparent that some of those were hosted on the Equinix data centers which were due to be shut down at the end of November. After a call today between myself, @richardlau and @bahamat we now have a good understanding of where we area and how to move forward.

To summarise where we currently are: There are two systems hosted on the account managed by Joyent. Both are SmartOS hosts with virtual images inside them. One of these hosts is called `nc-backup-01` and only contains one VM - the backup server which is SmartOS 15.4 and is in the DFW2 data center.

The second host is called `nc-compute-01` and it contains all the other systems referenced in #3104 - some are KVM instances, some are SmartOS zones and one is an lx-branded zone. The details and breakdown are as follows:

```
[root@nc-compute-01 ~]# vmadm list
UUID                                  TYPE  RAM      STATE             ALIAS
0f85685d-0150-4f8f-e211-9ecee63e8b61  KVM   3840     running           test-joyent-ubuntu1604_arm_cross-x64-1
1cf77dcc-8a17-6c35-9132-83f55a8e058f  KVM   3840     running           test-joyent-ubuntu1804-x64-1
49f0a164-4e86-4fda-de73-abcf257587a0  KVM   3840     running           release-joyent-ubuntu1604_arm_cross-x64-1
356655a2-12e6-e1d7-ac7b-b5188ad37cb0  OS    4096     running           test-joyent-smartos20-x64-3
49089cfe-915f-c226-c697-a9faca6041f2  OS    4096     running           release-joyent-smartos20-x64-2
94f76b46-6d20-612c-84e1-92c0dc3bae69  OS    4096     running           release-joyent-smartos18-x64-2
c6e3d47a-1421-ee11-c52d-c3c80c198e95  OS    4096     running           test-joyent-smartos20-x64-4
d894f3c6-d09a-c9df-d7ae-b6f613d9b413  OS    4096     running           test-joyent-smartos18-x64-3
db3664d7-dd31-c233-cafb-df79efb9d069  OS    4096     running           test-joyent-smartos18-x64-4
[...]

PR #3117, opened 2022-12-16
ansible: update IP addresses for "joyent" machines

richardlau:
Updates "joyent" labeled machines with their updated IP addresses.

Refs: https://github.com/nodejs/build/issues/3108

Also renames the test machines with new label "equinix_mnx" to more accurately reflect where they are now hosted. We'll need to rename the release and infra machines too but I won't have time to do that this year.

---

Currently re-ansibling machines to pick up the name change (this list has the old names):
- [x] test-joyent-smartos18-x64-3
- [x] test-joyent-smartos18-x64-4
- [x] test-joyent-smartos20-x64-3
- [x] test-joyent-smartos20-x64-4
- [x] test-joyent-ubuntu1804-x64-1

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -167,7 +167,7 @@ hosts:
         ubuntu1804-x64-1: {ip: 52.117.26.14, alias: jenkins-workspace-6}
         ubuntu1804-x64-2: {ip: 50.97.245.9}
 
-    - joyent:
+    - equinix_mnx:
         smartos18-x64-3: {ip: 147.28.162.102}
         smartos18-x64-4: {ip: 147.28.162.103}
         smartos20-x64-3: {ip: 147.28.162.107}
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -21,9 +21,9 @@ hosts:
         ubuntu2004-x64-1: {ip: 169.60.150.91, alias: ansible}
 
     - joyent:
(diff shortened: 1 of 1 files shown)


mhdawson (approved):
LGTM

ansible/inventory.yml
```diff
@@ -51,9 +51,9 @@ hosts:
         ibmi73-ppc64_be-1: {ip: 65.183.160.62, user: nodejs}
 
     - joyent:
```
targos, line 53: does this have to stay "joyent" ?

ansible/inventory.yml
```diff
@@ -21,9 +21,9 @@ hosts:
         ubuntu2004-x64-1: {ip: 169.60.150.91, alias: ansible}
 
     - joyent:
```
targos, line 23: ditto

ansible/inventory.yml
richardlau: No, it should change, but I ran out of time before going on leave for Christmas and the New Year:

> We'll need to rename the release and infra machines too but I won't have time to do that this year.

I want to keep the names in the inventory and Jenkins in sync. Renaming the hosts in Jenkins will result in a new secret which means updating the secrets repository and reansibling the machines.

richardlau:
I updated the IP address for the backup server as well. Going to merge this so we have the current IP addresses in the inventory. We can rename the remaining release and infra machines from "joyent" to "equinix_mnx" in a follow up PR.

Issue #3119, opened 2022-12-17
Ansible Jenkins macos10.x machines challenges


UlisesGascon:
When We were working on #3112, I discovered that Ansible scripts are not working properly for `macos10.14` and `macos10.15`.

The first time I tried to run the script I got an error related to brew cask:
```
TASK [package-upgrade : Add AdoptOpenJDK Java Repo] *********************************************************************************************************
ok: [test-orka-macos10.14-x64-2]

TASK [package-upgrade : Update Casks]  

...
Deleted Formulae\npython\n\nYou have 12 outdated formulae and 1 outdated cask installed.\nYou can upgrade them with brew upgrade\nor list them with brew outdated.\n\nError: `brew cask` is no longer a `brew` command. Use `brew <command> --cask` instead."}
```

In the next attempt the script stops at this stage:

![Captura de Pantalla 2022-12-17 a las 11 05 24](https://user-images.githubusercontent.com/5110813/208240949-bf1f4c8a-b90d-402e-b446-2f7b1d0a557e.png)

I can confirm that the playbook works perfect for `macos11` machines.

targos:
I'm pretty sure that's because of https://github.com/nodejs/build/issues/3087

PR #3130, opened 2022-12-20
fix: deprecation errors in macos 10.x

UlisesGascon:
### Main changes:

Duplicated brew Upgrade installed packages step in order to allow failure due deprecation (`You are using macOS 10.14.\nWe (and Apple) do not provide support for this old version.`) in macos10.x but keep macos11 the same

### Context: 
- Fix https://github.com/nodejs/build/issues/3119#issuecomment-1359162400

### Notes

I am not sure if there is a more elegant way to condition `ignore_errors` rather than duplicate tasks 😅

### DIFF
--- a/ansible/roles/package-upgrade/tasks/partials/brew.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/brew.yml
@@ -41,19 +41,9 @@
     become_user: "{{ ansible_user }}"
     homebrew:
       upgrade_all: yes
-    when: 
+    ignore_errors: 
       - os == "macos10.15" 
       - os == "macos10.14"
-    ignore_errors: true
-
-
-  - name: Upgrade installed packages
-    become_user: "{{ ansible_user }}"
-    homebrew:
-      upgrade_all: yes
-    when: 
-      - os != "macos10.15" 
-      - os != "macos10.14"   
 
   - name: Install brew cu
     become_user: "{{ ansible_user }}"
--- a/ansible/roles/package-upgrade/tasks/partials/brew.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/brew.yml
@@ -41,7 +41,9 @@
     become_user: "{{ ansible_user }}"
     homebrew:
       upgrade_all: yes
-    when: os == "macos10.15" || os == "macos10.14"
+    when: 
+      - os == "macos10.15" 
+      - os == "macos10.14"
     ignore_errors: true
 
 
@@ -49,7 +51,9 @@
     become_user: "{{ ansible_user }}"
     homebrew:
       upgrade_all: yes
-    when: os != "macos10.15" || os != "macos10.14"   
+    when: 
+      - os != "macos10.15" 
+      - os != "macos10.14"   
 
   - name: Install brew cu
     become_user: "{{ ansible_user }}"
--- a/ansible/roles/package-upgrade/tasks/partials/brew.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/brew.yml
@@ -41,9 +41,7 @@
     become_user: "{{ ansible_user }}"
     homebrew:
       upgrade_all: yes
-    ignore_errors: 
-      - os == "macos10.15" 
-      - os == "macos10.14"
+    ignore_errors: os == "macos10.15" or os == "macos10.14"
 
   - name: Install brew cu
     become_user: "{{ ansible_user }}"
--- a/ansible/roles/package-upgrade/tasks/partials/brew.yml
+++ b/ansible/roles/package-upgrade/tasks/partials/brew.yml
@@ -36,10 +36,20 @@
     script: files/install-homebrew.sh
     when: os == "macos11.0" and not armbrew.stat.exists
 
+  # Ignoring deprecation errors in MacOS 10.x
+  - name: Upgrade installed packages 
+    become_user: "{{ ansible_user }}"
+    homebrew:
+      upgrade_all: yes
+    when: os != "macos11.0"
+    ignore_errors: true
+
+
   - name: Upgrade installed packages
     become_user: "{{ ansible_user }}"
     homebrew:
       upgrade_all: yes
+    when: os == "macos11.0"    
 
   - name: Install brew cu
     become_user: "{{ ansible_user }}"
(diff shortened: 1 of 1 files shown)


ansible/roles/package-upgrade/tasks/partials/brew.yml
```diff
@@ -36,10 +36,20 @@
...
+      upgrade_all: yes
+    when: os != "macos11.0"
+    ignore_errors: true
+
+
   - name: Upgrade installed packages
     become_user: "{{ ansible_user }}"
     homebrew:
       upgrade_all: yes
+    when: os == "macos11.0"
```
targos, line 52: Can we make the condition less specific to avoid having to change it when we add macOS 12 and 13 hosts?

UlisesGascon: Great suggestion @targos! I fix in [1ea3173](https://github.com/nodejs/build/pull/3130/commits/1ea31738e0a8df10008eb13ac53de416c1e662fc)

ansible/roles/package-upgrade/tasks/partials/brew.yml
```diff
@@ -36,10 +36,20 @@
     script: files/install-homebrew.sh
     when: os == "macos11.0" and not armbrew.stat.exists
 
+  # Ignoring deprecation errors in MacOS 10.x
+  - name: Upgrade installed packages
```
targos, line 40:
```suggestion
  - name: Upgrade installed packages (EoL macOS)
```

ansible/roles/package-upgrade/tasks/partials/brew.yml
```diff
@@ -36,10 +36,20 @@
     script: files/install-homebrew.sh
     when: os == "macos11.0" and not armbrew.stat.exists
 
+  # Ignoring deprecation errors in MacOS 10.x
+  - name: Upgrade installed packages 
+    become_user: "{{ ansible_user }}"
+    homebrew:
+      upgrade_all: yes
+    when: os == "macos10.15" || os == "macos10.14"
+    ignore_errors: true
```
richardlau, line 45: Instead of creating a duplicate block, you could just add
```suggestion
    ignore_errors: os == "macos10.15" || os == "macos10.14"
```
to the original block.

UlisesGascon: Great suggestion @richardlau! I fix it in [0d7a28a](https://github.com/nodejs/build/pull/3130/commits/0d7a28ad2c0c483ec0fabbc47535d5a31459a212)

ansible/roles/package-upgrade/tasks/partials/brew.yml
UlisesGascon: We kept one step at the end

ansible/roles/package-upgrade/tasks/partials/brew.yml
```diff
@@ -36,10 +36,14 @@
...
 
-  - name: Upgrade installed packages
+  # Ignoring deprecation errors in MacOS 10.x
+  - name: Upgrade installed packages 
     become_user: "{{ ansible_user }}"
     homebrew:
       upgrade_all: yes
+    ignore_errors: 
+      - os == "macos10.15" 
+      - os == "macos10.14"
```
richardlau, line 46: I think this syntax means "and" and not "or".

UlisesGascon: You are right @richardlau. I fix it in [68c8489](https://github.com/nodejs/build/pull/3130/commits/68c8489c4ec2c7a248c0326682bc2b35214053eb)

richardlau (approved):

ansible/roles/package-upgrade/tasks/partials/brew.yml
```diff
@@ -36,10 +36,12 @@
     script: files/install-homebrew.sh
     when: os == "macos11.0" and not armbrew.stat.exists
 
-  - name: Upgrade installed packages
+  # Ignoring deprecation errors in MacOS 10.x
+  - name: Upgrade installed packages 
     become_user: "{{ ansible_user }}"
     homebrew:
       upgrade_all: yes
+    ignore_errors: os == "macos10.15" or os == "macos10.14"
```
targos, line 44: ansible tells me that the value is invalid:

`TASK [package-upgrade : Upgrade installed packages] ********************************************************************
fatal: [release-nearform-macos11.0-arm64-1]: FAILED! => {"msg": "the field 'ignore_errors' has an invalid value (os == \"macos10.15\" or os == \"macos10.14\"), and could not be converted to an bool.The error was: The value 'os == \"macos10.15\" or os == \"macos10.14\"' is not a valid boolean.  Valid booleans include: 'true', 1, 0, 'no', 'false', 'yes', 'n', 'y', '0', 'on', '1', 'off', 't', 'f'\n\nThe error appears to be in '/home/mzasso/git/nodejs/build/ansible/roles/package-upgrade/tasks/partials/brew.yml': line 40, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  # Ignoring deprecation errors in MacOS 10.x\n  - name: Upgrade installed packages\n    ^ here\n"}
...ignoring`

UlisesGascon: what version are you using? 🤔

targos: ansible [core 2.14.2]

richardlau: Does 
```suggestion
    ignore_errors: "{{ os == 'macos10.15' or os == 'macos10.14' }}"
```
work?

targos: yes

targos: I wonder if we should skip the step instead of ignoring errors. It takes ages to run because Homebrew doesn't support macOS <11 and it has to compile packages from source.

richardlau: > I wonder if we should skip the step instead of ignoring errors. It takes ages to run because Homebrew doesn't support macOS <11 and it has to compile packages from source.

The Ansible scripts are supposed to be able to both:
- Set up a new VM/machine from scratch.
- Update an existing VM/machine.

For Orka we could possibly get away with the first by relying on snapshots, but it wouldn't help for the Nearform VMs (I guess we could snapshot those but we'd have to find storage for the snapshots and we already have disk space issues on those machines).

I would suggest though that we plan to move off macOS 10.15, at least for Node.js 20 and beyond.

PR #3163, opened 2023-01-30
ansible: update V8 builders to Python 3.9

richardlau:
Recent upstream V8 changes now require at least Python 3.8 to run the V8 test runner.

Refs: https://github.com/nodejs/node-v8/issues/244#issuecomment-1401830840
Refs: https://chromium-review.googlesource.com/c/v8/v8/+/4181030

---

I've deployed this onto all of the RHEL 8 ppc64le, s390x and x64 test machines. (We don't currently run the V8 CI on the RHEL8 x64 machines because the builds on those take a lot longer than on the Nearform-hosted Intel machine(s).)

### DIFF
--- a/ansible/roles/build-test-v8/tasks/partials/rhel8-ppc64.yml
+++ b/ansible/roles/build-test-v8/tasks/partials/rhel8-ppc64.yml
@@ -5,9 +5,10 @@
 #
 
 # V8 builds still require Python 2.
+# Newer V8 builds require Python 3.8, or later.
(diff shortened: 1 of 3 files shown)


richardlau:
FWIW I did start a V8 canary CI but I think the build is broken on s390x for other reasons and didn't get to the test running part: https://ci.nodejs.org/job/node-test-commit-v8-linux/5146/

mhdawson (approved):
LGTM

PR #3169, opened 2023-02-02
tools: add web certificate expiry checker

richardlau:
Add a workflow that checks the certificates that are served from the public facing web sites that are hosted on our own infrastructure. In issue will be created if any of the certificates checked are due to expire within the next 30 days.

---

We do get reminder/warning emails from our SSL certificate vendor but this workflow is checking the actual deployed certificates (on the servers and in CloudFlare (nodejs.org is proxied by CloudFlare CDN)).

An example of the issue created: https://github.com/richardlau/build/issues/2
And yes, we have certificates expiring soon. I will reissue our certificate and update them on the servers before they expire.


UlisesGascon (approved):
I Love the implementation, seems clear, effective and readable :)

.github/workflows/certificate-check.yml
```diff
@@ -0,0 +1,43 @@
...
+  issues: write
+
+env:
+  NODE_VERSION: lts/*
+
+jobs:
+  check-certificates:
+    runs-on: ubuntu-latest
+    outputs:
+      report: ${{ steps.check_certificates.outputs.report }}
```
targos, line 19: This seems unused?

tools/check-certificates/check-certificates.mjs
```diff
@@ -0,0 +1,53 @@
+import tls from 'node:tls';
+import { randomUUID } from 'node:crypto';
+
+const daysThreshold = 30;
+const now = Date.now();
+const secondsPerDay = 86_400_000;
```
targos, line 6: Nit: `msPerDay`

targos (approved):

.github/workflows/certificate-check.yml
richardlau: Yes, this is unused. I used https://github.com/nodejs/nodejs-dependency-vuln-assessments/blob/main/.github/workflows/check-vulns.yml as a starting point and missed tidying up this line when I realized I didn't need to spawn a second matrixed job. Thanks.

richardlau: Now removed.

mhdawson (approved):
LGTM

PR #3173, opened 2023-02-03
doc: update certificate renewal/reissue docs

richardlau:
Clarify that the `.crt` file needs to contain three certificates (the certificate plus the intemediate CA certificates).


rvagg (approved):
Very nice. Much better than my approach of forgetting exactly how this works each year then fumbling through, banging keys, until it's working each renewal.

AshCripps (approved):

Issue #3194, opened 2023-02-13
Truncated downloads from Cloudflare resulting in checksum mismatches


richardlau:
@bnb reported that Electron were seeing downloads from the headers server not matching the checksums in the SHA256SUMS.txt file:
```console
curl --compressed https://nodejs.org/download/release/v14.17.5/win-x86/node.lib |sha256sum
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  845k    0  845k    0     0  1242k      0 --:--:-- --:--:-- --:--:-- 1250k
49a6c038e4e53657b1b7a173c601e8ba1585d0f6dbfa88ac789dba107d596812  -
```
https://nodejs.org/download/release/v14.17.5/SHASUMS256.txt
```console
1078be47b9315c81aa3bb989c4bba8ee23e0da9e4854a44006decf45d578833e  win-x86/node.lib
```

We had something similar reported four days ago with the node.lib file from a different release: https://github.com/nodejs/node/issues/46594

In both cases the files on our two servers (www and unencrypted) still matched the checksum -- further investigation showed the file downloaded/cached from the nodejs.org URL was truncated. Purging CloudFlare's CDN cache[1] fixed both issues, although I'm concerned we've now seen this twice.

[1]: @nodejs/build-infra To purge the cache, log into www and run
```console
/home/nodejs/queue-cdn-purge.sh nodejs <reason>
```
where `<reason>` is something like, e.g. 
```console
/home/nodejs/queue-cdn-purge.sh nodejs gh-build-3194
```
(The `<reason>` gets logged to `/home/nodejs/cdn-purge.log`.) Once queued the purge should happen within five minutes. The queue script will write `/tmp/cdnpurge.nodejs` and this file will be removed when the purge happens.

richardlau:
CDN purge has fixed the reported file:
```console
curl --compressed https://nodejs.org/download/release/v14.17.5/win-x86/node.lib |sha256sum
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1359k    0 1359k    0     0  2127k      0 --:--:-- --:--:-- --:--:-- 2127k
1078be47b9315c81aa3bb989c4bba8ee23e0da9e4854a44006decf45d578833e  -
```
Note that the lib file is 1349k now when previously only 845k was received.

mikeharder:
[...]

PR #3195, opened 2023-02-14
ansible: add lib as application/octet-stream type

richardlau:
Serve `.lib` files with `application/octet-stream` content-type.

Refs: https://github.com/nodejs/build/issues/3194

### DIFF
--- a/ansible/www-standalone/tasks/nginx.yaml
+++ b/ansible/www-standalone/tasks/nginx.yaml
@@ -84,6 +84,7 @@
     line: '{{ item }}'
     insertafter: '^types.*'
   with_items:
+    - 'application/octet-stream lib;'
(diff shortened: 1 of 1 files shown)


targos (approved):

nschonni:
Was going to ask if the `exe` needed to be added, but looks like it's part of the default set https://github.com/nginx/nginx/blob/master/conf/mime.types

richardlau:
Forgot to mention that I have already deployed the change in this PR on our two servers.

> The only other one that might be missing is https://www.iana.org/assignments/media-types/text/tab-separated-values for the `.tab` files. Don't think that has any functional change like this change actually does

I'm inclined for now to leave that unless someone runs into a problem with the `.tab` files.

Issue #3125, opened 2022-12-19
Disk full on downloads server


targos:
Found out because release builds fail to push the binaries:
```
07:13:16 mkdir: cannot create directory ‘nodejs/nightly/v20.0.0-nightly202212191d6fda0856’: No space left on device
```

targos:
```
3.5T    /home/dist/nodejs
```

targos:
```
150G    ./chakracore-nightly
823M    ./chakracore-rc
4.6G    ./chakracore-release
34M     ./docs
4.0K    ./next-nightly
2.0T    ./nightly
144G    ./rc
309G    ./release
78G     ./test
822G    ./v8-canary
```

targos:
@nodejs/build-infra 

What should we do? Has this already happened before?

richardlau:
> @nodejs/build-infra
> 
> What should we do? Has this already happened before?

Yes, for example:
- https://github.com/nodejs/build/issues/2162
- https://github.com/nodejs/build/issues/1202

sxa:
I think what @targos suggested in https://github.com/nodejs/build/issues/2162#issuecomment-584629845 is reasonable and I've got some scripts that I use elsewhere which can likely enforce that if we're happy with not retaining all 2Tb of nightlies that we're storing at present.

As a "quick fix" I've moved all the v9 nightlies to `/root/nightly` rather than deleting them which will give us a bit of grace pending any objections to implementing something similar to the above.

The obvious concern would be that we might want to do it separately for each version since there were overlapping major versions in the nightlies. 

Do we want to run the deletion of old stuff past the TSC for approval before implementing anything here?

sxa:
@rvagg Do you know the history of the `/mnt/nodejs_org_home_old` file system? As the `old` would imply it doesn't seem to have been touched in almost three years so I guess it's an original file system before you created the existing larger one to replace it?

mhdawson:
I agree as well that what @targos suggested in https://github.com/nodejs/build/issues/2162#issuecomment-584629845 makes sense.

I think if we reach consensus an FYI to the TSC in advance of implementing would make sense in case there are any concerns.

sxa:
I've got a prototype script that I've tested which will:
- Retain everything for the last two months (So November and December at the time of writing)
- Remove everything with a date ending in 1 from the previous two calendar years (So 2021/2022 just now) so roughly one every 10 days
- Removing everything except for the nightlies on the first of the month earlier than that

[...]

PR #3199, opened 2023-02-17
www: create script to free space by removing old builds

sxa:
Script as described in https://github.com/nodejs/build/issues/3125

This is the script which will clear up the nightlies and v8-canary directories so they do not constantly increase the disk space. If we are happy with this logic we can determine how to run it - either
- As a part of the release process before or after copying new builds there
- Via a regular cron job
- Something else?

### DIFF
--- a/tools/wwwspace/prune.sh
+++ b/tools/wwwspace/prune.sh
@@ -1,0 +1,55 @@
+#!/bin/sh
+#
+# Node.js www server space pruning script.
+# Runs against a subdirectory of /home/dist/nodejs and removes old builds:
+# - For anything over 2 calendar years ago, retain only the build dated first of the month
+# - For anything in last 2 calendar years but not the last two months, retain date numbers ending in 1
+# - Keep everything from the last two months.
+#
+
+[ $# -lt 2 ] && echo Usage: $0 '[nightly|v8-canary]' '[show|delete]' && exit 1
+PRODUCT=$1
+[ "$1" != "nightly" -a "$1" != "v8-canary" ] && echo Parameter mist be one of: nightly v8-canary && exit 1
+if [ "$2" = "show" ]; then
+  RMCOMMAND="echo rm -rv"
+elif [ "$2" = "delete" ]; then
+  RMCOMMAND="rm -rv"
+else
+  echo Second parameter must be \"show\" or \"delete\"
+  exit 1
+fi
+# Location to start the search for candidates for removal
+cd "/home/dist/nodejs/$PRODUCT" || exit 1
+
+THISYEAR=`date +%Y`
+LASTYEAR=`date --date="1 year ago" +%Y`
+THISMONTH=`date +%Y%m`
+LASTMONTH=`date --date="1 month ago" +%Y%m`
+case $PRODUCT in
+	v8-canary) STARTVERSION=9;;
+	nightly) STARTVERSION=5;;
+esac
+for VERSION in `seq $STARTVERSION 20`; do
+
+   # Determine the latest nightly for this version so it is never removed
+   LATEST=`ls -1d v${VERSION}.* | tail -1`
+   [ -z "$LATEST" -o ! -d "$LATEST" ] && echo Could not validate latest version "$LATEST" for Node v${VERSION} && exit 1
+
+   # Versions that aren't from the last two calendar years - keep 1st of the month
+   SELECTED=`ls -1d v${VERSION}.* | egrep -v "${PRODUCT}$THISYEAR|${PRODUCT}$LASTYEAR" | grep -v ${PRODUCT}20....01 | grep -v "${LATEST}"`
+   SELECTEDCOUNT=`echo $SELECTED | wc -w`
+   echo V$VERSION: Selected $SELECTEDCOUNT of `ls -1d v${VERSION}.* | wc -l` from two calendar years ago
+   if [ $SELECTEDCOUNT -ne 0 ]; then
+     ${RMCOMMAND} $SELECTED
+   fi
+
+   # Versions from the last two calendar years except last two months
+   # Keep every date number ending in 1 (So 01 11 21 31)
+   SELECTED=`ls -1d v${VERSION}.* | egrep    "${PRODUCT}$THISYEAR|${PRODUCT}$LASTYEAR" | egrep -v "${PRODUCT}${THISMONTH}|${PRODUCT}${LASTMONTH}" | grep -v ${PRODUCT}20.....1 | grep -v "${LATEST}"`
+   SELECTEDCOUNT=`echo $SELECTED | wc -w`
+   echo V$VERSION: Selected $SELECTEDCOUNT of `ls -1d v${VERSION}.* | wc -l` from the last two calendar years
+   if [ $SELECTEDCOUNT -ne 0 ]; then
+     ${RMCOMMAND} ${SELECTED}
+   fi
+done
+
--- a/tools/wwwspace/prune.sh
+++ b/tools/wwwspace/prune.sh
@@ -33,7 +33,7 @@ for VERSION in `seq $STARTVERSION 20`; do
 
    # Determine the latest nightly for this version so it is never removed
    LATEST=`ls -1d v${VERSION}.* | tail -1`
-   [ -z "$LATEST" -o ! -d "$LATEST" ] && echo Could not validate latest version "$LATEST" for Node v${VERSION} && exit 1
+   [ -z "$LATEST" -o ! -d "$LATEST" ] && echo "Could not validate latest version ${LATEST} for Node v${VERSION}" && exit 1
 
    # Versions that aren't from the last two calendar years - keep 1st of the month
    SELECTED=`ls -1d v${VERSION}.* | egrep -v "${PRODUCT}${THISYEAR}|${PRODUCT}${LASTYEAR}" | grep -v ${PRODUCT}20....01 | grep -v "${LATEST}"`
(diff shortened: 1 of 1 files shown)


targos:
Most releasers don't have the permissions to change old files anyway.

sxa:
@mhdawson @targos (and @richardlau just because he's been involved) are you ok with the logic in the script? I don't want to make it live with automation until it's been eyeballed by others for potential issues.

targos:
I'm sorry I'm not familiar enough with shell scripts to review this with confidence.

tools/wwwspace/prune.sh
```diff
@@ -0,0 +1,55 @@
...
+# Runs against a subdirectory of /home/dist/nodejs and removes old builds:
+# - For anything over 2 calendar years ago, retain only the build dated first of the month
+# - For anything in last 2 calendar years but not the last two months, retain date numbers ending in 1
+# - Keep everything from the last two months.
+#
+
+[ $# -lt 2 ] && echo Usage: $0 '[nightly|v8-canary]' '[show|delete]' && exit 1
+PRODUCT=$1
+[ "$1" != "nightly" -a "$1" != "v8-canary" ] && echo Parameter mist be one of: nightly v8-canary && exit 1
+if [ "$2" = "show" ]; then
```
rvagg, line 13: :thumbsup: nice feature, although this would normally be `"dry-run"` in most commands that do this

tools/wwwspace/prune.sh
```diff
@@ -0,0 +1,55 @@
...
+# Node.js www server space pruning script.
+# Runs against a subdirectory of /home/dist/nodejs and removes old builds:
+# - For anything over 2 calendar years ago, retain only the build dated first of the month
+# - For anything in last 2 calendar years but not the last two months, retain date numbers ending in 1
+# - Keep everything from the last two months.
+#
+
+[ $# -lt 2 ] && echo Usage: $0 '[nightly|v8-canary]' '[show|delete]' && exit 1
+PRODUCT=$1
+[ "$1" != "nightly" -a "$1" != "v8-canary" ] && echo Parameter mist be one of: nightly v8-canary && exit 1
```
rvagg, line 12:
```suggestion
[ "$1" != "nightly" -a "$1" != "v8-canary" ] && echo Parameter mist be one of: nightly v8-canary && exit 1
PRODUCT=$1
```

not important, just aesthetic

tools/wwwspace/prune.sh
```diff
@@ -0,0 +1,55 @@
...
+  echo Second parameter must be \"show\" or \"delete\"
+  exit 1
+fi
+# Location to start the search for candidates for removal
+cd "/home/dist/nodejs/$PRODUCT" || exit 1
+
+THISYEAR=`date +%Y`
+LASTYEAR=`date --date="1 year ago" +%Y`
+THISMONTH=`date +%Y%m`
+LASTMONTH=`date --date="1 month ago" +%Y%m`
```
rvagg, line 27:
```suggestion
THISYEAR=`date -u +%Y`
LASTYEAR=`date -u --date="1 year ago" +%Y`
THISMONTH=`date -u +%Y%m`
LASTMONTH=`date -u --date="1 month ago" +%Y%m`
```

it should be set on the system, bit being explicit is probably worthwhile too

tools/wwwspace/prune.sh
```diff
@@ -0,0 +1,55 @@
...
+	nightly) STARTVERSION=5;;
+esac
+for VERSION in `seq $STARTVERSION 20`; do
+
+   # Determine the latest nightly for this version so it is never removed
+   LATEST=`ls -1d v${VERSION}.* | tail -1`
+   [ -z "$LATEST" -o ! -d "$LATEST" ] && echo Could not validate latest version "$LATEST" for Node v${VERSION} && exit 1
+
+   # Versions that aren't from the last two calendar years - keep 1st of the month
+   SELECTED=`ls -1d v${VERSION}.* | egrep -v "${PRODUCT}$THISYEAR|${PRODUCT}$LASTYEAR" | grep -v ${PRODUCT}20....01 | grep -v "${LATEST}"`
```
rvagg, line 39:
```suggestion
   SELECTED=`ls -1d v${VERSION}.* | egrep -v "${PRODUCT}${THISYEAR}|${PRODUCT}${LASTYEAR}" | grep -v ${PRODUCT}20....01 | grep -v "${LATEST}"`
```

not gong to break, but good practice to do this when variables bump up against other things

tools/wwwspace/prune.sh
```diff
@@ -0,0 +1,55 @@
...
+LASTMONTH=`date --date="1 month ago" +%Y%m`
+case $PRODUCT in
+	v8-canary) STARTVERSION=9;;
+	nightly) STARTVERSION=5;;
+esac
+for VERSION in `seq $STARTVERSION 20`; do
+
+   # Determine the latest nightly for this version so it is never removed
+   LATEST=`ls -1d v${VERSION}.* | tail -1`
+   [ -z "$LATEST" -o ! -d "$LATEST" ] && echo Could not validate latest version "$LATEST" for Node v${VERSION} && exit 1
```
rvagg, line 36:
```suggestion
   [ -z "$LATEST" -o ! -d "$LATEST" ] && echo "Could not validate latest version ${LATEST} for Node v${VERSION}" && exit 1
```

there's a lot of unquoted echoing in this script, the one on line 50 is even more egregious; not a blocking change request but personally I'd be much more comfortable fully quoting most or all of the echoes, particularly where they use variables.

tools/wwwspace/prune.sh
rvagg: minor point also, instead of the weird `egrep` and `grep` alternations in this script, you could just `\|` to make `|` work properly

tools/wwwspace/prune.sh
```diff
@@ -0,0 +1,55 @@
...
+   SELECTED=`ls -1d v${VERSION}.* | egrep -v "${PRODUCT}$THISYEAR|${PRODUCT}$LASTYEAR" | grep -v ${PRODUCT}20....01 | grep -v "${LATEST}"`
+   SELECTEDCOUNT=`echo $SELECTED | wc -w`
+   echo V$VERSION: Selected $SELECTEDCOUNT of `ls -1d v${VERSION}.* | wc -l` from two calendar years ago
+   if [ $SELECTEDCOUNT -ne 0 ]; then
+     ${RMCOMMAND} $SELECTED
+   fi
+
+   # Versions from the last two calendar years except last two months
+   # Keep every date number ending in 1 (So 01 11 21 31)
+   SELECTED=`ls -1d v${VERSION}.* | egrep    "${PRODUCT}$THISYEAR|${PRODUCT}$LASTYEAR" | egrep -v "${PRODUCT}${THISMONTH}|${PRODUCT}${LASTMONTH}" | grep -v ${PRODUCT}20.....1 | grep -v "${LATEST}"`
```
rvagg, line 48: these regexes are fairly lose, is it worth being specific with `^` and/or `$` to bound them properly?

tools/wwwspace/prune.sh
```diff
@@ -0,0 +1,55 @@
...
+   SELECTEDCOUNT=`echo $SELECTED | wc -w`
+   echo V$VERSION: Selected $SELECTEDCOUNT of `ls -1d v${VERSION}.* | wc -l` from the last two calendar years
```
rvagg, line 50:
```suggestion
   echo v$VERSION: Selected $SELECTEDCOUNT of `ls -1d v${VERSION}.* | wc -l` from the last two calendar years
```

(& lack of quoting is making my eye twitch on this one)

rvagg (approved):
Functionally this seems fine to me, my critique is mostly just personal preference; so that that as you want. I would have leaned on:

1. Heavier use of quoting
2. No backticks, `$()` only
3. More use of `find` with its ability to do date stuff; but what you've got here seems perfectly reasonable since we're doing nice date things in file/dir names.

nschonni:
I'm not sure if some the Shellcheck rules will catch those https://github.com/nodejs/build/blob/main/.shellcheckrc. Never took another pass at cleaning any more of them up

tools/wwwspace/prune.sh
sxa: `egrep` is standard across more *IX systems so I'd prefer to keep that for this case (The syntax you're suggesting wouldn't work on all systems, although I do appreciate that in this case it won't be run on anything other than Linux)

tools/wwwspace/prune.sh
sxa: Was very useful when I was being paranoid during testing ;-) Yep happy to rename it

tools/wwwspace/prune.sh
sxa: Resolved (Various places)

tools/wwwspace/prune.sh
sxa: The issue is that with names like `v20.0.0-nightly20230222a3211e18b4` it means regexping the whole thing which I'd suggest makes it more complex than it needs to be. `$PRODUCT` should be an adequate start position (I could use `^v${VERSION}.*` at the start and `[0-9a-e]$` at the end though I guess although I'm not sure we'll hit a situation where that would make much of a difference.

Maybe if someone renamed a file to have a `-KEEPME` suffix or similar that they wanted to retain?

sxa:
Verified that this is still behaving as it should after all the changes from the comments

mhdawson:
@sxa I would assume that.

sxa:
Merging since there have been no objections

PR #3230, opened 2023-03-16
www: update checked html page

richardlau:
After the switch to Next.js, `/home/www/nodejs/en/index.html` no longer exists.

### DIFF
--- a/ansible/www-standalone/resources/scripts/check-build-site.sh
+++ b/ansible/www-standalone/resources/scripts/check-build-site.sh
@@ -11,6 +11,6 @@ indexjson=/home/dist/${site}/release/index.json
 indexhtml=/home/www/${site}/en/index.html
 buildsite=/home/nodejs/build-site.sh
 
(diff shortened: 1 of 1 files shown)


richardlau:
Shellcheck 😞

richardlau:
Since `/home/www/nodejs/en/index.html` no longer exists, 
```
[ /home/dist/nodejs/release/index.json -nt /home/www/nodejs/en/index.html ]
```
was always true and the website was being rebuilt every 5 minutes (`check-build-site.sh` is run every 5 mins on a cron timer). Which is likely to be contributing to the server instability (cpu/mem usage and more frequent CDN purges).

domdomegg:
This script isn't used for iojs anymore since that got turned into just a redirect. We can probably just strip that out to make this much simpler.

domdomegg (approved):
Fine for now. Can remove iojs bits as a separate PR.

targos (approved):

PR #3249, opened 2023-03-22
ansible,jenkins: use gcc 10 on RHEL 8 for Node.js 20

richardlau:
Refs: https://github.com/nodejs/node/pull/46806
Refs: https://github.com/orgs/nodejs/discussions/45892

### DIFF
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -134,7 +134,7 @@ packages: {
   ],
 
   rhel8: [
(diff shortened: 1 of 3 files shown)


richardlau:
I've deployed the Ansible changes onto all of the test and release RHEL8 systems (including the three Ubuntu 2004 arm64 docker hosts which are running RHEL8 containers). Builds won't use gcc10 until the `select-compiler.sh` script changes in this PR land.

This doesn't address the cross-compiler for armv7l -- that will need to be updated separately.

mhdawson (approved):
LGTM

targos (approved):

PR #3262, opened 2023-03-28
Add Windows ARM64 assets to release

StefanStojanovic:
This change brings Windows ARM64 one step closer to being a tier 2 supported platform, as described in https://github.com/nodejs/node/pull/47233.

With these changes, Windows ARM64 assets will get picked when releasing the following versions of NodeJS. Please note that these changes are partial, the rest of the required changes (eg. modifying the release job) will be done by @joaocgreis.


targos (approved):

joaocgreis:
I updated the two files on the www server.

I did not update the metrics server because the file present in the server is significantly different than the file that was used as base for this PR, so I would need more context or time to investigate and fix issues.

Issue #3268, opened 2023-03-30
tap2junit needs to be updated


MoLow:
tap2junit has been released with various fixes: https://github.com/nodejs/tap2junit/releases/tag/v0.1.6, and needs to be updated
Ref: https://github.com/nodejs/node/issues/47048
Ref: https://github.com/nodejs/tap2junit/pull/45

I will start with gradually updating some of the docker hosts since https://github.com/nodejs/node/issues/47048 mainly affects them

MoLow:
I have updated `test-softlayer-ubuntu1804_docker-x64-1`  (using https://github.com/nodejs/build/pull/3269)

MoLow:
I am running https://ci.nodejs.org/job/node-test-commit-linux-containered/36877/ to see the results

MoLow:
updated:
`test-digitalocean-ubuntu1804_docker-x64-1` 
`test-digitalocean-ubuntu1804-docker-x64-2`
`test-softlayer-ubuntu1804_docker-x64-1`

PR #3269, opened 2023-03-30
ansible: use explicit python version on containers

MoLow:
the default installed python version for these environments is 3.6.8, 
the minimal supported version for the latest `tap2junit` is 3.7
this forces the installation of a later version

ref: https://github.com/nodejs/build/issues/3268

### DIFF
--- a/ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
+++ b/ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
@@ -24,7 +24,7 @@ RUN apt-get update && apt-get install apt-utils -y && \
       git \
       pkg-config \
       curl \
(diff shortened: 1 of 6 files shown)


richardlau (approved):

ansible/roles/docker/templates/ubuntu1804_sharedlibs.Dockerfile.j2
```diff
@@ -24,7 +24,7 @@ RUN apt-get update && apt-get install apt-utils -y && \
       git \
       pkg-config \
       curl \
-      python-pip \
+      python 3.8 \
```
richardlau, line 27: This doesn't look correct. Either `python3.8` or removed if the default Python 3 on Ubuntu 18.04 is new enough.

PR #3271, opened 2023-03-30
ansible: use explicit tap2junit version on containers

MoLow:
Due to docker layer caching, it makes more sense to pin to a specific version than rebuild the entire image

### DIFF
--- a/ansible/roles/docker/templates/alpine311.Dockerfile.j2
+++ b/ansible/roles/docker/templates/alpine311.Dockerfile.j2
@@ -42,7 +42,7 @@ RUN apk add --no-cache --virtual .build-deps \
         libtool \
         autoconf
 
-RUN pip3 install tap2junit
+RUN pip3 install tap2junit=={{ tap2junit_version }}
 
 RUN addgroup -g {{ server_user_gid.stdout_lines[0] }} {{ server_user }}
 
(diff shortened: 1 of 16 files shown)


ansible/roles/docker/templates/alpine311.Dockerfile.j2
```diff
@@ -42,7 +42,7 @@ RUN apk add --no-cache --virtual .build-deps \
         libtool \
         autoconf
 
-RUN pip3 install tap2junit
+RUN pip3 install tap2junit==0.1.6
```
richardlau, line 45: We could maybe use a variable so we don't have to update every template when we update in the future. 
To do this:
1. Add a new variable to `ansible/roles/docker/vars/main.yml` for the tap2junit version.
2. Replace in the templates using handlebars, e.g.
```suggestion
RUN pip3 install tap2junit=={{ tap2junit_version }}
```

nschonni:
What about adding a `requirements.txt` or Pipfile? Then Dependabot could PR updates when new releases come out? Maybe this way is less maintenance from an Ansible approach though.

ansible/roles/docker/templates/ubuntu1804.Dockerfile.j2
```diff
@@ -29,7 +30,11 @@ RUN apt-get update && apt-get dist-upgrade -y && apt-get install -y \
 # something in the list of package above when installed at the same time.
 RUN apt-get install -y openjdk-17-jre-headless
 
-RUN pip3 install tap2junit
+RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.6 1 &&\
```
richardlau, line 33: Minor nit, can you add a space before the line continuation characters? I don't know if it causes any problems, but at least GitHub renders it oddly without the space.
```suggestion
RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.6 1 && \
```

PR #3301, opened 2023-04-11
ci: restrict Python linting runs

nschonni:
This workflow really only checks for python issues, so renamed the file and restricted it to only run when the files are impacted.

### DIFF
--- a/.github/workflows/lint-python.yml
+++ b/.github/workflows/lint-python.yml
@@ -1,24 +1,29 @@
-name: Lint
+name: Lint Python
 on:
   push:
     branches-ignore:
       - "dependabot/*"
+    paths:
+      - '**/*.py'
+      - '.flake8'
+      - .github/workflows/lint-python.yml
   pull_request:
+    paths:
+      - '**/*.py'
+      - '.flake8'
+      - .github/workflows/lint-python.yml
 jobs:
   Python:
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Python 3.7
+    - name: Set up Python
       uses: actions/setup-python@v4
-      with:
-        python-version: 3.7
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        # pip install -r requirements.txt  # TODO: Add tap2junit
+        pip install flake8
     - name: Lint with flake8
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-python.yml
```diff
@@ -1,24 +1,29 @@
...
 jobs:
   Python:
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Python 3.7
+    - name: Set up Python
       uses: actions/setup-python@v4
-      with:
-        python-version: 3.7
```
nschonni, line 15: @cclauss I didn't think the Python need to be pinned for the linting job, but correct me if i'm wrong

cclauss:
Let's instead replace flake8 with ruff as we have done on node-gyp, gyp-next, tap2junit.

[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

The ruff Action uses minimal steps to run in ~5 seconds, rapidly providing intuitive GitHub Annotations to contributors.

image

.github/workflows/lint-python.yml
```diff
@@ -1,24 +1,29 @@
...
   push:
     branches-ignore:
       - "dependabot/*"
+    paths:
+      - '**/*.py'
+      - '.flake8'
+      - .github/workflows/lint-python.yml
   pull_request:
+    paths:
+      - '**/*.py'
```
cclauss, line 12: Node.js has Python files that do not have a .py suffix.

nschonni: In this repo?

nschonni:
Closing since the ruff PR landed upstream, so it likely makes more sense to use the other PR

PR #3302, opened 2023-04-11
ci: replace Python linter flake8 with ruff

cclauss:
[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

The ruff Action uses minimal steps to run in ~5 seconds, rapidly providing intuitive GitHub Annotations to contributors.

image

@nschonni


.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,14 @@
+# https://beta.ruff.rs
+name: lint_python
+on:
```
nschonni, line 3: I'd suggest copying the filters from https://github.com/nodejs/build/pull/3301/files#diff-14721317565b7c66355f276ec27762054b12f390afd6c35fba85cfaf77d311c7R2, but adding `pyproject.toml` to the lists

pyproject.toml
```diff
@@ -0,0 +1,52 @@
...
+  "SIM",  # flake8-simplify
+  "SLF",  # flake8-self
+  "T10",  # flake8-debugger
+  "TCH",  # flake8-type-checking
+  "TID",  # flake8-tidy-imports
+  "W",    # pycodestyle
+  "YTT",  # flake8-2020
+]
+ignore = [
+  "B904",
```
nschonni, line 35: Are this ignores TODOs or forever things? Sometimes useful to document why/how the might be fixed if someone has an interest to try and address them later

pyproject.toml
```diff
@@ -0,0 +1,52 @@
...
+  "PLR2004",
+  "RET505",
+  "RUF005",
+  "S110",
+  "S701"
+]
+line-length = 223
+target-version = "py37"
```
nschonni, line 43: If the target is going to stay 3.7 for now, maybe the CI should keep the pinned version that I dropped in the other PR

pyproject.toml
```diff
@@ -0,0 +1,52 @@
...
+]
+ignore = [
+  "B904",
+  "PLR2004",
+  "RET505",
+  "RUF005",
+  "S110",
+  "S701"
+]
+line-length = 223
```
nschonni, line 42: Guessing this was to get it passing, rather than a preferred line length like the old 250 value

pyproject.toml
cclauss: I will let others determine if they want to reduce the ignore list

Rule information is easy to acquire: `ruff rule B904`
# raise-without-from-inside-except (B904)

Derived from the **flake8-bugbear** linter.

Message formats:
* Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling

pyproject.toml
cclauss: Just like Node.js, Python has a release cycle: https://devguide.python.org/versions

Py37 will reach EOL in two months.

pyproject.toml
cclauss: 88 is the preferred value for Python code.  The goal is to set high water marks so that if a contributor wants to go beyond that level they will need to justify their actions in pull requests.  The same is true of `max-args`, `max-branches`, `max-complexity`, etc.

pyproject.toml
cclauss: Ruff does not run on Python so it changes in similar ways to flake8 but only based on `--target-version=pyXX`.

richardlau (approved):

.github/workflows/lint_python.yml
nschonni: It's not about tool speed, it's about not running jobs that can have unrelated job failures on a PR

Issue #3382, opened 2023-06-13
website build broken during Docker image preparation


richardlau:
@RafaelGSS has pointed out from the [webhook log](https://nodejs.org/github-webhook.log) that the build did not succeed:

```console
stderr: Previous HEAD position was 58704a9... chore: updated issue templates descriptions
stderr: HEAD is now at 7a4e29c... blog: add june 2023 security releases (#5434)
stdout: lts: Pulling from library/node
stdout: c98dda1b0e97: Pulling fs layer
stdout: 9a463ac54ed3: Pulling fs layer
stdout: b70766638915: Pulling fs layer
stdout: 4bcc83cd0070: Pulling fs layer
stdout: 217770c726a2: Pulling fs layer
stdout: cec0e089046a: Pulling fs layer
stdout: 6fedcf8a3041: Pulling fs layer
stdout: 471e0a0d1303: Pulling fs layer
stdout: 4bcc83cd0070: Waiting
stdout: 217770c726a2: Waiting
stdout: cec0e089046a: Waiting
stdout: 471e0a0d1303: Waiting
stdout: 9a463ac54ed3: Download complete
stdout: c98dda1b0e97: Verifying Checksum
stdout: c98dda1b0e97: Download complete
stdout: 217770c726a2: Verifying Checksum
stdout: 217770c726a2: Download complete
stdout: b70766638915: Verifying Checksum
stdout: b70766638915: Download complete
stdout: 6fedcf8a3041: Verifying Checksum
stdout: 6fedcf8a3041: Download complete
stdout: 471e0a0d1303: Verifying Checksum
stdout: 471e0a0d1303: Download complete
stdout: cec0e089046a: Verifying Checksum
stdout: cec0e089046a: Download complete
stdout: 4bcc83cd0070: Verifying Checksum
stdout: 4bcc83cd0070: Download complete
stdout: c98dda1b0e97: Pull complete
stdout: 9a463ac54ed3: Pull complete
stdout: b70766638915: Pull complete
stdout: 4bcc83cd0070: Pull complete
stdout: 217770c726a2: Pull complete
stdout: cec0e089046a: Pull complete
stdout: 6fedcf8a3041: Pull complete
stdout: 471e0a0d1303: Pull complete
stdout: Digest: sha256:d244ade7d0b2ad97002102880142c8b9a4b016ea2308af87e9b150c43987b301
stdout: Status: Downloaded newer image for node:lts
stdout: docker.io/library/node:lts
[...]

PR #3383, opened 2023-06-13
www: use node:lts-bullseye in build-site.sh

richardlau:
Temporarily switch to `node:lts-bullseye` while we investigate why the website build is broken with `node:lts`/`node:lts-bookworm`.

Refs: https://github.com/nodejs/build/issues/3382


mhdawson (approved):
LGTM

Issue #3270, opened 2023-03-30
Terraforming Cloudflare


UlisesGascon:
Seems like there is traction to use Terraform to manage changes in Cloudflare. https://github.com/nodejs/build/issues/3220.

I would love to discuss this idea further and take the lead on the initiative if we decide to move forward.

I am facing a similar challenge with my personal project, so I can create a minimum viable product and share it with the team to see how it would work in practice.

In the past, I have utilized Terraform in conjunction with Azure services and discovered it to be an exceptional solution for implementing infrastructure modifications through pull requests, thereby increasing transparency within the infrastructure.

UlisesGascon:
As agreed in #3299 I will work on a personal POC that I will present to the team so we can understand the complexity and challenges in our organization.

The most critical part is to avoid any disruption in the downloads related settings in Cloudflare.

UlisesGascon:
I finished the POC and it is working fine, if you are interested in review it... I can grant you access to the private repo and a short demo 15m (terraform basics, github actions changes, etc..) in video format.

Based on the discussion in https://github.com/nodejs/build/issues/3362

### Next steps
- [ ] I will ask for Cloudflare permissions (read only) in the admin repo.
- [ ] I will ask for a new repository creation (initially private) in the admin repo.
- [ ] I will check how to setup Terrafom Cloud account with the @nodejs/build and document the process 
- [ ] I will fetch all the Cloudflare resources into the new repository, prepare the the Github pipelines and documentation (Read only token, no risk for Cloudflare current setup). 

**After that we can decide**
- If we can make the repository open (as the DNS information seems not to be sensitive)
- If we are happy with this solution and then promote the usage including a Write Token for trigger Cloudflare changes from Terraform files using Github Actions.

richardlau:
Do we need a Terraform Cloud account (vs running Terraform locally)?

UlisesGascon:
Once #3391 is merged.

**Next Steps:**
- [ ] Clean the DNS records that are no longer relevant, see: 
  - https://github.com/nodejs/build/pull/3391#discussion_r1236284669, 
  - https://github.com/nodejs/build/pull/3391#discussion_r1236285111 and 
  - https://github.com/nodejs/build/pull/3391#discussion_r1236435278
- [ ] Update the documentation with better examples, see: 
  - https://github.com/nodejs/build/pull/3391#discussion_r1236639196

UlisesGascon:
This is currently stopped until R2 migration is completed.

[...]

Issue #3370, opened 2023-05-26
Setup Terraform Cloud account


UlisesGascon:
Based on the discussion in https://github.com/nodejs/build/issues/3270#issuecomment-1562947464. cc: @nodejs/build

I will need help to setup the [Terraform Cloud Account](https://portal.cloud.hashicorp.com/sign-in) for Node.js Org. This account will be used to create a Team in Terraform Cloud and provide access to individual accounts. Can you create the account using the `*@iojs.org` email? 

**Additional steps**
- [x] Add my personal account to the team as well as `Workspace Administrator level`. [This guide can help](https://developer.hashicorp.com/terraform/tutorials/cloud/cloud-permissions) or I can do it if you share the credentials for Terraform Cloud
- [x] I will generate a token for the GitHub actions that will sync the terraform state

targos:
Reopening until the credentials have been somehow shared with @nodejs/build-infra

UlisesGascon:
There is no option to scope the Terraform Cloud. So any personal token might have access to all the workspaces in all the organizations as the user.

I suggest to create an account with a build team email or another alias and generate the personal token from that account for the Github Actions part 🤔

targos:
>  create an account with a build team email or another alias

Isn't it what this issue was about?

UlisesGascon:
> Isn't it what this issue was about?

Yes, that is correct. I expected to generate a token on behalf of the organization, but the current Terraform Cloud settings only allow me to generate tokens as myself. The next step will be to add a new member to the organization who is not a personal account and generate the token from that account. Is there a better way to manage this scenario?

targos:
The [API Tokens page](https://app.terraform.io/app/nodejs/settings/authentication-tokens) recommends to create Team Tokens for CI/CD. What's wrong with that?

UlisesGascon:
🤦 I didn't show that in the UI. Thanks @targos! 

I created a team API Token and it is working as expected.
![Captura de pantalla 2023-06-13 a las 9 26 31](https://github.com/nodejs/build/assets/5110813/0db7e4b7-1163-4416-974f-2e3fabca27a8)

I was not so sure about the expiration that we need, so I used a short one for now.

targos:
> Reopening until the credentials have been somehow shared with https://github.com/orgs/nodejs/teams/build-infra

Done in the secrets repo.

PR #3391, opened 2023-06-20
Terraforming Cloudflare for DNS records

UlisesGascon:
### Main Changes

- Added support for Terraform in the repository (setup, remote backend, dependencies, etc)
- Imported DNS records for nodejs.org
- Imported DNS records for iojs.org
- Added dedicated pipeline for Terraform that plan (PR) and apply (merged to `main`) the changes when changes are done in the `terraform` folder.
- Added basic documentation, including examples of how to review PRs
- Added specific git ignore rules

### Context

- related https://github.com/nodejs/build/issues/3270
- related https://github.com/UlisesGascon/poc-nodejs-cloudflare-terraform
- related https://github.com/nodejs/admin/issues/804#issuecomment-1597498465

#### Cloudflare API Token

The Cloudflare API Token is stored in the Terraform Cloud workspace. The token is used to authenticate the Terraform Cloud agent to the Cloudflare API. The token is stored in the `TF_API_TOKEN` environment variable in the Terraform Cloud workspace.

This token currently is in READ ONLY mode so it can not be used to create or update resources in Cloudflare. This is a security measure to prevent accidental changes in the Cloudflare account until we are familiar with Terraform as a team. See: https://github.com/nodejs/build/issues/3370

#### CI Settings

I need to include a secret `TF_API_TOKEN` in the github actions settings in this repository. Not sure how to do that as I am not an admin in this repository.

Captura de pantalla 2023-06-20 a las 14 56 30

### Changelog

- 5e21bef feat: added specific git ignore rules for terraform folder by @UlisesGascon
- 4b5f770 feat: added terraform initialization with Cloudflare support by @UlisesGascon
- 65d528e feat: added dedicated pipeline for Terraform by @UlisesGascon
- a7eb6f9 feat: imported DNS records for iojs.org by @UlisesGascon
- accc9da feat: imported DNS records for nodejs.org by @UlisesGascon
- f18bfde docs: added basic Terraform documentation by @UlisesGascon


terraform/dns_nodejs_org.tf
```diff
@@ -0,0 +1,348 @@
...
+  value   = "logs.libuv.org"
+  zone_id = "1206c4f949d69993ae55d9d015804406"
+}
+
+resource "cloudflare_record" "terraform_managed_resource_d0dbc6d7694d2445b23e84f5fdc8e6cd" {
+  name    = "modules"
+  proxied = false
+  ttl     = 1
+  type    = "CNAME"
+  value   = "npmjs.org"
```
nschonni, line 186: Aside: that seems like something that should transfer to GitHub/NPM @MylesBorins

terraform/dns_nodejs_org.tf
```diff
@@ -0,0 +1,348 @@
...
+  value   = "nodejs.org"
+  zone_id = "1206c4f949d69993ae55d9d015804406"
+}
+
+resource "cloudflare_record" "terraform_managed_resource_e30d4a05a180489fa391d40baee0b95c" {
+  name    = "packages"
+  proxied = false
+  ttl     = 1
+  type    = "CNAME"
+  value   = "npmjs.org"
```
nschonni, line 213: Also this one

terraform/dns_nodejs_org.tf
```diff
@@ -0,0 +1,348 @@
...
+  value   = "nodejs.github.io"
+  zone_id = "1206c4f949d69993ae55d9d015804406"
+}
+
+resource "cloudflare_record" "terraform_managed_resource_e7385cf24ad8660f929ad64132c56c72" {
+  name    = "logs"
+  proxied = true
+  ttl     = 1
+  type    = "CNAME"
+  value   = "logs.libuv.org"
```
nschonni, line 177: Doesn't appear to resolve

terraform/.gitignore
```diff
@@ -0,0 +1,34 @@
...
+#
+# !example_override.tf
+
+# Include tfplan files to ignore the plan output of command: terraform plan -out=tfplan
+# example: *tfplan*
+
+
+# OTHER
+*.env
+IGNORE/
```
MoLow, line 34:
```suggestion
IGNORE/

```

terraform/README.md
```diff
@@ -0,0 +1,48 @@
...
+### How to review changes in a PR?
+
+The terraform plan is generated automatically in the PR. You can review the changes in the `terraform plan` section in the PR.
+
+There are two possible scenarios:
+- No changes
+- Changes required
+
+When changes are required you can see the details in the `terraform plan` section. Here you have several isolated examples of the changes:
+- [Add resource](https://github.com/UlisesGascon/poc-nodejs-cloudflare-terraform/pull/5)
```
MoLow, line 25: NIT, add a TODO to replace links to PRS in this repo?

UlisesGascon: It is a great suggestion, right now `UlisesGascon/poc-nodejs-cloudflare-terraform` is archived, so the PRs still visible. We can replace them in the future when we have the first changes made from Terraform in this repository.

I created a subtask for it in https://github.com/nodejs/build/issues/3270#issuecomment-1600513990 👍

UlisesGascon:
@nschonni I am not planning to make DNS record changes in this PR because we are only migrating the current state from the Cloudflare UI to Terraform in this PR, and we don't have writing permissions yet in the GitHub Action.

I think we can move that discussion to a separate PR. In the meantime, I created a subtask for it on [#3270 (comment)](https://github.com/nodejs/build/issues/3270#issuecomment-1600513990).

Let me know if this approach works for you 👍.

MoLow (approved):

nschonni:
Random comment on https://github.com/nodejs/build/issues/3223 reminded me that this can't close #3270 because it doesn't cover things like the firewall rules (yet)

mhdawson:
@UlisesGascon my only question is if it would make sense for us to break this into 2 PRs, one to add the records for the iojs.org and  s second one for nodejs.org.  The idea being that  landing the iojs.org one would be lower risk and once we see that all working we land the second. Not sure if it helps but was wonding.

terraform/README.md
```diff
@@ -0,0 +1,48 @@
...
+- [Delete resource](https://github.com/UlisesGascon/poc-nodejs-cloudflare-terraform/pull/4)
+
+Note that any change can combine several of the previous examples. For example, you can see a change that add a resource and update another one.
+
+Once the PR is merged, the changes will be applied in the Cloudflare account via Github Action.
+
+### Local development
+
+#### IMPORTANT ⚠️
+If you apply any plan in you local machine, you can trigger changes in the Cloudflare account. So, you must be careful with the changes you are applying in your local environment. In order to review/create changes you are not require to do local development as the github pipeline will generate a plan with the changes.
```
mhdawson, line 36: @UlisesGascon I would have thought you would need Cloudflare access for anything you do locally to have an effect?

terraform/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+#### Requirements
+- Terraform version `Terraform v1.4.5` used to build this infrastructure. [How to install Terraform](https://developer.hashicorp.com/terraform/tutorials/aws-get-started/install-cli)
+- You will need a `TF_API_TOKEN` in you environment with a valid Terraform cloud API Key. [How to create a Terraform cloud API Key](https://www.terraform.io/docs/cloud/users-teams-organizations/api-tokens.html#creating-an-api-token)
```
mhdawson, line 41: I guess this is related to my question above. I assume we don't expect anybody to be doing this. Intead there should be a PR and the action which has the TF_API_TOKEN will apply any changes.

terraform/dns_nodejs_org.tf
mhdawson: Is this meant to make packages.nodejs.org point to nphjs.org as I don't see that I just get to the main website when I go to packages.nodejs.org.

But I agree that we should deal with changes in follow on PRs

nschonni: @mhdawson the 2 npmjs.org ones were fixed in https://github.com/nodejs/build/issues/3395, which is why you are just seeing the normal site now. I don't see the "Resolve" button to hide these, but they'd also go away the next time the sync is run, since the records no longer exist

UlisesGascon:
@mhdawson @nschonni let me know if this can help solve some of the questions from your previous comments

### General overview

Captura de pantalla 2023-06-29 a las 8 58 37


**Green path**

This is the default way (and recommended) when we want to make any changes in Cloudflare once this PR is merged. Any user can create a PR and add/update/remove resources from Cloudflare. Once the PR is merged, the GitHub actions will notify this change directly to Terraform Cloud. Terraform Cloud will apply the changes in Cloudflare.

**Blue path**

Any user who has access to Terraform Cloud (can use or generate a token). This token can be used in the same way as the GitHub actions do in the green path. So, even if you don't have access to Cloudflare, when you trigger a new plan directly from Terraform Cloud, it will be applied as Terraform Cloud can authenticate against Cloudflare.

This option is quite interesting as it is the best way to import resources. For example, as @nschonni suggested with the firewall rules.

**Red path**

As far I know is what we do today, the user has write access to Cloudflare and can make changes via web UI or CLI. This option remains possible even when this PR is merged.

Only a few people in the org have this access level, which will be used, for example, to modify Cloudflare settings for items that are not covered with Terraform.

### Specific concerns

> @UlisesGascon my only question is if it would make sense for us to break this into 2 PRs, one to add the records for the iojs.org and s second one for nodejs.org. The idea being that landing the iojs.org one would be lower risk and once we see that all working we land the second. Not sure if it helps but was wonding.

We can split this PR into two separate ones, but currently Terraform cannot apply any changes in Cloudflare as we have a _read only_ token.

I think that we can keep this approach (_read only tokens_) until we feel confident with this jump. I believe that just by keeping these Terraform files in sync with the Cloudflare state, we are already providing value to the org members as now our infrastructure is much more transparent than before and it is easier to propose changes, even if we make manual changes via Cloudflare CLI or web UI. at the end.

> Random comment on #3223 reminded me that this can't close #3270 because it doesn't cover things like the firewall rules (yet)

I think it is a good idea to include the rules in this PR. Should I import them in this PR or do we prefer a separate one?

nschonni:
> I think it is a good idea to include the rules in this PR. Should I import them in this PR or do we prefer a separate one?

I think either this PR or another is ok, whichever you prefer. I was just flagging that this PR is set to autoclose that other issue, but I think there are one or 2 more things that might still be needed like the firewall stuff.

UlisesGascon:
> I think either this PR or another is ok, whichever you prefer. I was just flagging that this PR is set to autoclose that other issue, but I think there are one or 2 more things that might still be needed like the firewall stuff.

Great point, I just removed the autoclose reference

mhdawson:
@UlisesGascon thanks for the nice picture.  We might want to include that in the doc somewhere. I probably was not quite clear enough in my questions, but your answer confirms it anway. Generally people should not have the cloudflare token and updates will be through the GitHub action.  As more of an exceptional case (doing initial reource imports etc.) people can get a token an use it locally.

mhdawson:
> I think that we can keep this approach (read only tokens) until we feel confident with this jump

I do think we want to make the jump sooner than later. I'd prefer we make a step of managing some of the DNS entries to start versus waiting longer. I may be making more work, but for example if we added 5 at a time, then if anything when wrong the manual work to fixup would be lower. @richardlau does doing them in smaller batches make any sense to you?

UlisesGascon:
I will close this PR as we will move this initiative to a different repository. Follow up details in  https://github.com/nodejs/admin/issues/848.

Issue #3259, opened 2023-03-27
Investigate increasing open file limit


MattIPv4:
A follow-on from https://github.com/nodejs/nodejs.org/issues/5184 and https://github.com/nodejs/nodejs.org/issues/5149, we should look at increasing the number of files that NGINX can have open at any one time.

This could either be done by increasing the system ulimit, or by setting `worker_rlimit_nofile` in NGINX.

targos:
> or by setting `worker_rlimit_nofile` in NGINX.

Are we allowed to do that without increasing the system limit?

blimmer:
We're starting to see a lot more errors that look like this today, which I think is related.

> HTTP/2 stream 0 was not closed cleanly: INTERNAL_ERROR

I found this ticket from https://github.com/nodejs/nodejs.org/issues/5184

MattIPv4:
> Are we allowed to do that without increasing the system limit?

I've been testing this locally with NGINX in Docker Compose...

<details>
<summary>Soft limit set low, no NGINX limit: errors ❌</summary>

`docker-compose.yml`:

```yaml
  ulimits:
    nofile:
      soft: 50
      hard: 2048
```

`nginx.conf`:

```
# worker_rlimit_nofile 65535;
```

Getting back errors:

```
ab -n 100 -c 50 -v 2 localhost/en/blog.html | grep "Non-2xx"

Non-2xx responses:      2
```
</details>

<details>
<summary>Soft limit set low, NGINX limit set high: errors ❌</summary>

`docker-compose.yml`:

```yaml
  ulimits:
    nofile:
      soft: 50
      hard: 2048
```

`nginx.conf`:

```
worker_rlimit_nofile 65535;
```

Getting back errors:

```
ab -n 100 -c 50 -v 2 localhost/en/blog.html | grep "Non-2xx"

Non-2xx responses:      2
```
</details>

<details>
<summary>Soft limit set high, no NGINX limit set: works ✅</summary>

`docker-compose.yml`:

```yaml
  ulimits:
    nofile:
      soft: 200
      hard: 2048
```

`nginx.conf`:

```
# worker_rlimit_nofile 65535;
```

No errors:

```
ab -n 100 -c 50 -v 2 localhost/en/blog.html | grep "Non-2xx"

[no output]
```
</details>

<details>
<summary>Soft limit set high, NGINX limit set high: works ✅</summary>

`docker-compose.yml`:

```yaml
  ulimits:
    nofile:
      soft: 200
[...]

PR #3406, opened 2023-07-03
www: raise file limit

richardlau:
Refs: https://github.com/nodejs/build/issues/3259

---

I've manually updated the `nginx.conf` on the server, restarted nginx and confirmed via `cat /proc/<pid>/limits` that the soft limit is now raised.

Let's see if it makes any difference to the server stability.

### DIFF
--- a/ansible/www-standalone/tasks/nginx.yaml
+++ b/ansible/www-standalone/tasks/nginx.yaml
@@ -72,6 +72,14 @@
     regexp: "worker_connections"
   tags: nginx
 
+- name: nginx | Increase number of files
+  lineinfile:
+    line: "worker_rlimit_nofile 4096;"
+    dest: "/etc/nginx/nginx.conf"
+    insertafter: "worker_processes"
+    regexp: "worker_rlimit_nofile"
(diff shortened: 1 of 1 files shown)


MoLow (approved):
LGTM. 
Question (might be a naive one): why do we serve files from a standalone server instead of a managed file storage such as s3 or Azure blob storage behind cloudflare?

Linkgoron (approved):

richardlau:
> LGTM. Question (might be a naive one): why do we serve files from a standalone server instead of a managed file storage such as s3 or Azure blob storage behind cloudflare?

I don't know the original reason. At least in the time I have been involved with Build nobody has proposed how that might work/be managed.

mhdawson (approved):
LGTM

richardlau:
While we haven't seen the load balancer notifications from Cloudflare since raising the file limit, we have been seeing reports of slow downloads (https://github.com/nodejs/build/issues/3259#issuecomment-1621640847) which has also impacted the release process. I've reverted the changes to the file limit for now.

PR #3428, opened 2023-07-19
fix(website): optimize and fix build-site.sh

targos:
Ignore npm install scripts. This makes the dependencies install faster
and avoids a crash that happens when Playwright attempts to install
browsers.


targos:
Change is already on the server and I executed it manually once.

MoLow (approved):

Trott (approved):

PR #3454, opened 2023-08-16
meta: add MoLow to infra admins

MoLow:
following discussion on the latest TSC meeting, I desire to get more involved in researching for a permanent solution for the situation with https://nodejs.org/dist/
I currently want to better learn the current infrastructure before doing enything.
I sent @mhdawson a gpg public key for him to add to the secrets repo (as well as replacing my existing one that does not work well)

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ Above list is manually synced with the [gpg member list](https://github.com/node
 * [@jbergstroem](https://github.com/jbergstroem) - Johan Bergström
 * [@joaocgreis](https://github.com/joaocgreis) - João Reis
 * [@mhdawson](https://github.com/mhdawson) - Michael Dawson
+* [@MoLow](https://github.com/MoLow) - Moshe Atlow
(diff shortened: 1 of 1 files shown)


mhdawson:
+1 from me

UlisesGascon:
+1 for me too

mhdawson:
I think at this point we should consider this approved.  @richardlau could you add his key?  If not I'll take a look but it's been a while since I added somebody's keys.  I'm happy to answer any additional questions/onboarding that @MoLow might need after that.

richardlau:
> I think at this point we should consider this approved. @richardlau could you add his key? If not I'll take a look but it's been a while since I added somebody's keys. I'm happy to answer any additional questions/onboarding that @MoLow might need after that.

I can do that tomorrow.

richardlau:
gpg has been added to the remaining parts of the secrets repo.

MoLow:
thanks @richardlau!

PR #3484, opened 2023-09-13
jenkins,doc: update supported Visual Studio for v21

targos:
Refs: https://github.com/nodejs/node/pull/49051

/cc @StefanStojanovic I think the test CI is ready for this, but the release CI isn't. Can you help with that?

### DIFF
--- a/doc/windows-visualstudio-supported-versions.md
+++ b/doc/windows-visualstudio-supported-versions.md
@@ -10,10 +10,9 @@ Supported versions for running the Node.js installer and executable as released.
 
 | Node.js Version | Windows Version            |
 |-----------------|----------------------------|
-| v16             | 10  / 2012 R2              |
 | v18             | 10 / 2016                  |
-| v19             | 10 / 2016                  |
 | v20             | 10 / 2016                  |
+| v21             | 10 / 2016                  |
 
 ## For building Node.js Core
 
@@ -21,10 +20,9 @@ Supported versions for building Node.js from source.
 
 | Node.js Version | Visual Studio Version               |
 |-----------------|-------------------------------------|
-| v16             | 2019 <sup>[1]</sup>                 |
 | v18             | 2019                                |
-| v19             | 2019                                |
 | v20             | 2019                                |
+| v21             | 2022                                |
 
 ## For building Node.js Addons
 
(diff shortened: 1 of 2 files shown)


MoLow (approved):

targos:
@StefanStojanovic We're coming closer to Node.js 21, and I don't know who else to ask.

mhdawson:
Adding mention to @joaocgreis in case he can help out or know when @StefanStojanovic might be around.

targos:
We are now only two weeks before the release of Node.js 21. I think this is urgent.

StefanStojanovic:
Sorry for jumping in late, I had some issues that are now over and I'm available to help here going forward. Hey @targos just so I do not lose any action item out of sight - what tasks do you need help with in the CI exactly?

In test CI we have 6 Win2022-VS2022 machines which are fully integrated into the compilation job. Here I think we are good to go.

In release CI we have 3 Win2022-VS2022 machines. They are fully configured (with the `vs2022` labels), but [iojs+release](https://ci-release.nodejs.org/job/iojs+release/) is still not changed (I'll move this forward). Besides changes to that job and changes in the version selector script, do you think something else will be needed?

targos:
Thanks for coming back to this!

> In test CI we have 6 Win2022-VS2022 machines which are fully integrated into the compilation job. Here I think we are good to go.

Agreed.

> In release CI we have 3 Win2022-VS2022 machines. They are fully configured (with the vs2022 labels), but [iojs+release](https://ci-release.nodejs.org/job/iojs+release/) is still not changed (I'll move this forward).

Changing `iojs+release` is what I had in mind.

> Besides changes to that job and changes in the version selector script, do you think something else will be needed?

I don't think so.

StefanStojanovic:
I've created [a test job](https://ci-release.nodejs.org/job/vs2022_test_iojs+release/) to try these changes in the CI. From what I see VS2019 is excluded from v21 onward. However, we need to add `[ /vs2022/, releaseType, lt(21) ]` to prevent breaking the rule of a single VS version building each Node.js version.

Other than this, we can remove all `VS2015/VS2017` and also make the test CI match the release CI (currently V20 is compiled on both VS2019 and VS2022 which I made as a temporary thing until VS2022 is enabled in release CI). However, all of these other changes are not that important right now, I can do it all in a following PR after this lands.

@targos if there's something else you could use my help with, please let me know.

targos:
@StefanStojanovic I added a skip of vs2022 for lt(21). Can you explicitly review the PR?

StefanStojanovic (approved):
LGTM. Release builds now pick the correct VS versions. If I was nitpicking, I'd say `Windows Version` for ` Official Releases` should be changed from `2012 R2` to `2019`, but since I'll be doing a follow on PR anyway I'll change that and you can go ahead and land this as is @targos 👍

StefanStojanovic:
@targos FYI I've changed [iojs+release](https://ci-release.nodejs.org/job/iojs+release/) to use `vs2022` labels now. VS2022 should be used for v21+ from now on.

PR #3489, opened 2023-09-18
jenkins: remove v16 from VersionSelectorScript

targos:


mhdawson:
I would be good to keep this around a bit longer if possible. I don't think it hurts anything, right?

targos:
That's why I made this a draft. I'll come back to it in 6 months.

MoLow (approved):

richardlau (approved):

StefanStojanovic (approved):
LGTM, Removing VS2015 and VS2017 from compilation jobs is highly optional.

EDIT: I was thinking about maybe removing the vs2015 and vs2017 compilation configurations completely, but that is not very urgent. I may open a follow-up PR for that at some point.

PR #3490, opened 2023-09-18
ansible,jenkins: rm centos7-release-sources host

targos:
Since v16 went EoL, RHEL8 is used for all release lines.

TODO:
- Remove from secrets repo
- Delete DO droplet

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -35,7 +35,6 @@ hosts:
   - release:
 
     - digitalocean:
-        centos7-x64-1: {ip: 138.68.12.105}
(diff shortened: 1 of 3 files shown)


richardlau:
It's written down that we keep stuff around for 6 months after a release line goes End-of-Life before removing. 

https://github.com/nodejs/build/blob/2e7c763cc6b02eee6edbe1f5199a54350b273e2a/README.md?plain=1#L294-L296

FWIW I can only remember one time when we decided to do an additional release after End-of-Life, which IIRC was within the  month after the End-of-Life.

targos:
We're more than 6 months later now

richardlau (approved):

PR #3505, opened 2023-09-29
release: copy cloudflare assets when promoting

MoLow:
While I believe assets can be copied directly from Jenkins in many cases, I think it is better to start with this approach where everything is promoted from within the www server.

this still doesn't handle the latest soft links. see discussion https://github.com/nodejs/release-cloudflare-worker/issues/33

### DIFF
--- a/ansible/www-standalone/tools/promote/promote_release.sh
+++ b/ansible/www-standalone/tools/promote/promote_release.sh
@@ -37,6 +37,8 @@ while true; do
   if [ "X${yorn}" == "Xy" ]; then
       . ${__dirname}/_promote.sh $site $2
       nodejs-latest-linker /home/dist/${site}/release/ /home/dist/${site}/docs/
+      dstdir=/home/dist/${site}/docs
+      . ${__dirname}/upload_to_cloudflare.sh $site $2
     break
   fi
 done
--- a/ansible/www-standalone/tools/promote/promote_release.sh
+++ b/ansible/www-standalone/tools/promote/promote_release.sh
@@ -37,8 +37,6 @@ while true; do
   if [ "X${yorn}" == "Xy" ]; then
       . ${__dirname}/_promote.sh $site $2
       nodejs-latest-linker /home/dist/${site}/release/ /home/dist/${site}/docs/
-      dstdir=/home/dist/${site}/docs
-      . ${__dirname}/upload_to_cloudflare.sh $site $2
     break
   fi
 done
--- a/ansible/www-standalone/tools/promote/upload_to_cloudflare.sh
+++ b/ansible/www-standalone/tools/promote/upload_to_cloudflare.sh
@@ -39,3 +39,5 @@ relativedir=${dstdir/$dist_rootdir/"$site/"}
 tmpversion=$2
 
 aws s3 cp $dstdir/$tmpversion/ $destination_bucket/$relativedir/$tmpversion/ --endpoint-url=$cloudflare_endpoint --profile $cloudflare_profile --recursive
+aws s3 cp $dstdir/index.json $destination_bucket/index.json --endpoint-url=$cloudflare_endpoint --profile $cloudflare_profile --recursive
+aws s3 cp $dstdir/index.tab $destination_bucket/index.tab --endpoint-url=$cloudflare_endpoint --profile $cloudflare_profile --recursive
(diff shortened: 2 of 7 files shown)


ansible/www-standalone/tools/promote/upload_to_cloudflare.sh
richardlau, line 1: If this file needs things to be set from the caller (e.g. `$dist_rootdir`) then it should probably check and error if those are not set.

ansible/www-standalone/tools/promote/promote_release.sh
```diff
@@ -37,6 +37,8 @@ while true; do
   if [ "X${yorn}" == "Xy" ]; then
       . ${__dirname}/_promote.sh $site $2
       nodejs-latest-linker /home/dist/${site}/release/ /home/dist/${site}/docs/
+      dstdir=/home/dist/${site}/docs
+      . ${__dirname}/upload_to_cloudflare.sh $site $2
```
richardlau, line 41: Why is this needed for releases but not the non-release (e.g. nightlies, v8-canary) builds?

richardlau:
> While I realize we do not run the playbook for the www server against the live www server, for my own benefit I'm assuming this needs something similar to #3501 to get the aws client and secrets onto the server?

Actually we probably need to make sure we're okay giving the `dist` user access to the secrets to publish to R2 storage as that is who the releasers log into the www machine as to do the promotion step. 
Refs: https://github.com/nodejs/node/blob/7b624c30b2902375c3c81885fc964b81da0891db/tools/release.sh#L14

MoLow:
> Actually we probably need to make sure we're okay giving the dist user access to the secrets to publish to R2 storage as that is who the releasers log into the www machine as to do the promotion step.
Refs: [nodejs/node@7b624c3/tools/release.sh#L14](https://github.com/nodejs/node/blob/7b624c30b2902375c3c81885fc964b81da0891db/tools/release.sh#L14)

Why would that not be ok? these credentials can only access this r2 bucket and nothing else

MoLow:
> While I realize we do not run the playbook for the www server against the live www server, for my own benefit I'm assuming this needs something similar to https://github.com/nodejs/build/pull/3501 to get the aws client and secrets onto the server?

I already uploaded that manually for the initial file sync, but added to the code for the sake of documentation

ansible/www-standalone/tools/promote/promote_release.sh
MoLow: this line will actually become redundant once https://github.com/nodejs/release-cloudflare-worker/pull/39 is merged

MoLow:
I'v deployed all files manually besides `ansible/www-standalone/tools/promote/resha_release.sh`, manually re-created `.done` files for the last nightly build and promotion was performed to R2: https://r2.nodejs.org/download/nightly/v21.0.0-nightly20231001092fb9f541/

MoLow:
https://r2.nodejs.org/download/nightly/v21.0.0-nightly2023100285c09f178c/ was deployed, 100% automatically now

richardlau (approved):

richardlau:
> > Actually we probably need to make sure we're okay giving the dist user access to the secrets to publish to R2 storage as that is who the releasers log into the www machine as to do the promotion step.
> > Refs: [nodejs/node@7b624c3/tools/release.sh#L14](https://github.com/nodejs/node/blob/7b624c30b2902375c3c81885fc964b81da0891db/tools/release.sh#L14)
> 
> Why would that not be ok? these credentials can only access this r2 bucket and nothing else

I may be being a tad cautious but most releasers (who have access to `dist` and run the promotion script to promote releases) are not members of build. We currently prevent releasers from changing existing downloads after one week (see https://github.com/nodejs/build/pull/2020) -- I've no idea if that's possible with R2.
https://github.com/nodejs/build/blob/726a7ca3a4724b3de72083ae43118f0317fc7d71/ansible/www-standalone/resources/scripts/dist-perms#L1-L10

While we do generally trust releasers we have had one incident recently where a releaser lost their laptop which meant we had to revoke their ssh keys to `dist`. I guess maybe it'd be okay if we're able to rotate the secret used to publish to R2 in the future if we had a similar potential compromise.

MoLow:
updated files manually

Issue #3504, opened 2023-09-27
`test-rackspace-win2022_vs2022-x64-6` systematic build timeout for `win-vs2022-arm64`


targos:
See https://ci.nodejs.org/computer/test-rackspace-win2022_vs2022-x64-6/builds

All `win-vs2022-arm64` systematically fail with:

```
00:56:43 Cross-compilation to ARM64 detected. We'll use the x64 Node executable for license2rtf.
01:56:43 Build timed out (after 60 minutes). Marking the build as failed.
01:56:44 Build was aborted
```

Examples:
https://ci.nodejs.org/job/node-compile-windows/nodes=win-vs2022-arm64/52608/console
https://ci.nodejs.org/job/node-compile-windows/nodes=win-vs2022-arm64/52601/console
https://ci.nodejs.org/job/node-compile-windows/nodes=win-vs2022-arm64/52597/console

Other machines don't have this problem so I removed the `win-vs2022-arm64` label from that one in https://ci.nodejs.org/computer/test-rackspace-win2022_vs2022-x64-6/configure to prevent CI failures.

richardlau:
> 
> ```
> 00:56:43 Cross-compilation to ARM64 detected. We'll use the x64 Node executable for license2rtf.
> ```
> 

This message is immediately before an attempt is made to download a x64 Node.js Windows binary (to convert the license to RTF). https://github.com/nodejs/build/issues/3477#issuecomment-1701611087 says that on the test CI that there should be a caching mechanism to prevent the builds getting stuck here.

targos:
Two win2019 machines are also affected:
https://ci.nodejs.org/computer/test-rackspace-win2019_vs2019-x64-1/builds
https://ci.nodejs.org/computer/test-rackspace-win2019_vs2019-x64-2/builds

StefanStojanovic:
I took a look. There is indeed a cache there but it can get corrupted (that's what happened here). I'll improve the logic to prohibit this. There is also an occasional ongoing problem with release CI where nodes get stuck trying to download the x64 node.exe for days (no caching there). I'll try to make a standardized approach that will work in both CI environments.

targos:
Label added back to https://ci.nodejs.org/computer/test-rackspace-win2022_vs2022-x64-6/configure

targos:
@StefanStojanovic It seems the bug is still here: https://ci.nodejs.org/job/node-compile-windows/53041/nodes=win-vs2022-arm64/console

[...]

PR #3315, opened 2023-04-20
jenkins: cache x64 node.exe for arm64 license file

StefanStojanovic:
After promoting Windows on ARM64 to a tier 2 platform, as a part of the compilation job in the CI, the ARM64 MSI installer is created. The first node-test-commit-windows-fanned build that did this was 54061. Since then, I've been monitoring all ARM64 builds, and until now, there's been a bit over 750 (the last build checked is 54823).

I noticed that 15 builds, roughly 2%, failed when downloading the x64 node.exe used for generating the license file. Changes proposed here tend to minimize, if not remove errors by caching the x64 node.exe and updating it weekly.

I've already tested these changes in my [job that tests migrating to VS2022](https://ci.nodejs.org/view/All/job/mefi-node-compile-windows-vs2022/) (another topic for another day) and saw no issues.

PR #3514, opened 2023-10-09
jenkins: improve x64 node.exe caching for arm64

StefanStojanovic:
It was noticed on a few occasions that x64 node.exe caching is error-prone. This happens due to issues with downloading the new node.exe which, if it goes bad, makes the machine unusable for arm64 compilation for a week. This improvement in caching logic removes that possibility, by validating the downloaded file's SHASUM and replacing the old one with it, only if it matches the SHASUM found in the official `SHASUMS256.txt` file.

Fixes: https://github.com/nodejs/build/issues/3504
Refs: https://github.com/nodejs/build/pull/3315

### DIFF
--- a/jenkins/scripts/windows/compile.cmd
+++ b/jenkins/scripts/windows/compile.cmd
@@ -1,3 +1,5 @@
+SETLOCAL EnableDelayedExpansion
+
 :: Use Python 2 up to Node 12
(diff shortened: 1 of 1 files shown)


StefanStojanovic:
Can someone from @nodejs/build review please? After landing this, the same code will be added to the release CI since not having caching there caused machines to be stuck on ARM64 builds indefinitely until manual action is taken to unblock it.

richardlau (approved):

Issue #3385, opened 2023-06-14
Action required by Apple: Transition to the notarytool command-line utility


targos:
Build got this by email:

> We’re reaching out because you recently used the altool command-line utility to notarize your macOS software with Apple. As announced last year at [WWDC22](https://developer.apple.com/videos/play/wwdc2022/10109), if you’re still using altool with the Apple notary service, you should transition to the notarytool command-line utility as soon as possible. Notarizing software with altool was deprecated in Xcode 13, and the Apple notary service will no longer accept uploads from altool as of November 1, 2023. Existing notarized software will continue to function properly.

> For information on notarizing your apps, read [TechNote TN3147: Migrating to the latest notarization tool](https://developer.apple.com/documentation/technotes/tn3147-migrating-to-the-latest-notarization-tool?changes=_3_3).

targos:
According to the above TechNote:

> If xcrun can’t find the tool, make sure you have Xcode 13 or later selected:

This means we will have to migrate to Xcode>=13 for all releases by the end of the year. v16.x will be EoL but v18.x won't and is currently built on macOS 10.15 (Xcode 13 doesn't support it)

targos:
We use [`gon`](https://github.com/mitchellh/gon) to notarize the pkg. The tool doesn't support `notarytool` and doesn't seem to be maintained anymore.

https://github.com/nodejs/node/blob/b0e08d178ed3fcb4442324f70154d5f36f8dfcc2/tools/osx-notarize.sh#L3

richardlau:
For background, we implemented notarization using [gon](https://github.com/mitchellh/gon) in https://github.com/nodejs/node/pull/31459 and https://github.com/nodejs/build/pull/2199, with discussion in https://github.com/nodejs/node/issues/29216.

Another tool, [electron-notarize-cli](https://www.npmjs.com/package/electron-notarize-cli), was mentioned in the [context of OpenJDK](https://github.com/nodejs/node/issues/29216#issuecomment-546932966) -- this one looks like it [has an option to use notarytool](https://github.com/electron/notarize/issues/137#issuecomment-1578500573).

I guess another question is whether we need another tool at all or whether we can directly use `notarytool`. https://developer.apple.com/videos/play/wwdc2021/10261/ suggests that `notarytool` can now wait for the notarization to succeed (or I guess fail) whereas previously with `altool` there needed to be a polling loop (which for us was provided by `gon`).

@UlisesGascon is this something you're interested in/have the time to look at?

[...]

PR #3531, opened 2023-10-19
jenkins: remove macOS 10.x release machines

UlisesGascon:
### Main changes

This will remove macOS 10.x release machines in Jenkins pipelines.

### Context

On November 1, 2023 macOS 10.x won't be able to notarize the binaries.

Related: https://github.com/nodejs/build/issues/3385#issuecomment-1729281269
Related: https://github.com/nodejs/node/pull/50291

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -72,10 +72,6 @@ def buildExclusions = [
   [ /sharedlibs_openssl102/,          anyType,     gte(16) ],
   [ /sharedlibs_fips20/,              anyType,     gte(16) ],
 
-  // OSX ---------------------------------------------------
-  [ /osx11-x64-release-tar/,          releaseType, gte(18) ],
-  [ /osx11-x64-release-pkg/,          releaseType, gte(18) ],
-
   // Source / headers / docs -------------------------------
   [ /^centos7-release-sources$/,      releaseType, gte(18) ],
   [ /^rhel8-release-sources$/,        releaseType, lt(18)  ],
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -73,8 +73,8 @@ def buildExclusions = [
   [ /sharedlibs_fips20/,              anyType,     gte(16) ],
 
   // OSX ---------------------------------------------------
-  [ /osx1015/,                        anyType,     gte(21) ],
-  [ /osx11-x64-release-tar/,          releaseType, lt(20)  ],
+  [ /osx11-x64-release-tar/,          releaseType, gte(18) ],
+  [ /osx11-x64-release-pkg/,          releaseType, gte(18) ],
 
   // Source / headers / docs -------------------------------
   [ /^centos7-release-sources$/,      releaseType, gte(18) ],
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -75,8 +75,6 @@ def buildExclusions = [
   // OSX ---------------------------------------------------
   [ /osx1015/,                        anyType,     gte(21) ],
   [ /osx11-x64-release-tar/,          releaseType, lt(20)  ],
-  [ /osx1015-release-pkg/,            releaseType, gte(16) ],
-  [ /osx1015-release-tar/,            releaseType, gte(20) ],
 
   // Source / headers / docs -------------------------------
   [ /^centos7-release-sources$/,      releaseType, gte(18) ],
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -72,6 +72,9 @@ def buildExclusions = [
   [ /sharedlibs_openssl102/,          anyType,     gte(16) ],
   [ /sharedlibs_fips20/,              anyType,     gte(16) ],
 
+  // OSX ---------------------------------------------------
+  [ /osx1015/,                        anyType,     gte(21) ],
+
   // Source / headers / docs -------------------------------
   [ /^centos7-release-sources$/,      releaseType, gte(18) ],
   [ /^rhel8-release-sources$/,        releaseType, lt(18)  ],


jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -75,8 +75,6 @@ def buildExclusions = [
   // OSX ---------------------------------------------------
   [ /osx1015/,                        anyType,     gte(21) ],
   [ /osx11-x64-release-tar/,          releaseType, lt(20)  ],
-  [ /osx1015-release-pkg/,            releaseType, gte(16) ],
-  [ /osx1015-release-tar/,            releaseType, gte(20) ],
```
richardlau, line 79: We'll need to adjust `osx11-x64-release-tar` so that Node.js 18 runs on those instead of `osx1015-release-tar` otherwise we'll have no Node.js 18 macOS x64 tarballs.

UlisesGascon: Not sure how to do that, can you guide me or provide an example? I was using an old PR as a reference for this change 🤔

targos: If only one `release-tar` remains, you can remove the previous line, which currently excludes Node.js 18 from `osx11-x64-release-tar`.
Also, if there are no longer any `osx1015` configs, you can delete the first line.
In other words, if after this change all Node.js versions are supposed to be built on the same machine label, you can remove all osx-related lines from this file.

richardlau: @UlisesGascon Quick overview of how this works:
- Job has all possible things it might run on added as configuration matrix selections.
- Job also has reference to this groovy script which places addition restrictions. The table at the top of this file lists where to exclude (i.e. do not run) configurations matching the regular expressions. If nothing excludes the configuration being matched against (i.e. no entry in this table) it is allowed.

As an aside, please don't merge/make any change to the set up until after tomorrow's Node.js 20 LTS transition release is done.

UlisesGascon: Thanks for the explanations and the feedback @targos @richardlau 👍 

I made some changes in 146d9cf:
- I removed the references to `osx1015`
- I added support for Node18 in `osx11-x64-release-tar`
- I added `osx11-x64-release-pkg` so the pkg notarization can be done also (not sure if this is correct). As far I know we want to target pkg rather than tarball currently.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -73,8 +73,8 @@ def buildExclusions = [
   [ /sharedlibs_fips20/,              anyType,     gte(16) ],
 
   // OSX ---------------------------------------------------
-  [ /osx1015/,                        anyType,     gte(21) ],
-  [ /osx11-x64-release-tar/,          releaseType, lt(20)  ],
+  [ /osx11-x64-release-tar/,          releaseType, gte(18) ],
+  [ /osx11-x64-release-pkg/,          releaseType, gte(18) ],
```
richardlau, line 77: `gte(18)` will mean that these will be excluded (i.e. not run) on all our currently supported versions of Node.js, which isn't what we want. 

Highlighting again @targos 's comment in https://github.com/nodejs/build/pull/3531#discussion_r1367893562 
> In other words, if after this change all Node.js versions are supposed to be built on the same machine label, you can remove all osx-related lines from this file.

UlisesGascon: ok, I see it now much clear :)

I removed the references to macOS in [ee12477](https://github.com/nodejs/build/pull/3531/commits/ee124777935dcdf229a6972f4e1afd8dd009b381) because we will only have macOS11 machines available and all the Node.js versions (18-21) should use them. So there is no need to restrict any machine

UlisesGascon:
My last question is.. if this remove the restrictions. How do I remove the machine tags for macos10.x? Is it done in the UI? 🤔

targos:
You can either remove the label from the machine (in the "configure" tab) or remove the machine. When no machine has the label anymore, the label disappears.

richardlau:
> You can either remove the label from the machine (in the "configure" tab) or remove the machine. When no machine has the label anymore, the label disappears.

Please remove the labels from the CI jobs first otherwise they'll try to run on machines with that label but none will exist leading to the job never completing.

UlisesGascon:
> Please remove the labels from the CI jobs first otherwise they'll try to run on machines with that label but none will exist leading to the job never completing.

I added this as a requirement in https://github.com/nodejs/build/pull/3532

targos (approved):
LGTM but only merge **after** osx1015 is removed from release CI

UlisesGascon:
> LGTM but only merge after osx1015 is removed from release CI

Can I do it already? I don't see any proposal release on going https://github.com/nodejs/node/pulls?q=is%3Apr+release+is%3Aopen+proposal

richardlau (approved):

UlisesGascon:
Both Jenkins agents are removed, so I will merge the pr. 

Note: I was not able to directly remove/delete the tags as they are dynamic

richardlau:
> Both Jenkins agents are removed, so I will merge the pr.
 
@UlisesGascon https://ci-release.nodejs.org/job/iojs+release/ is still showing 
image

UlisesGascon:
This is strange. Even if you only select them, the `iojs+release` is skipping them (see: https://ci-release.nodejs.org/job/iojs+release-ulises-experimental/5/). Therefore, there is no error even if no machines are associated with this label. I did this test in `iojs+release-ulises-experimental` just few minutes ago.


Maybe the label is there due historical/previously use?

richardlau:
> This is strange. Even if you only select them, the `iojs+release` is skipping them (see: https://ci-release.nodejs.org/job/iojs+release-ulises-experimental/5/). Therefore, there is no error even if no machines are associated with this label. I did this test in `iojs+release-ulises-experimental` just few minutes ago.
> 
> Maybe the label is there due historical/previously use?

@UlisesGascon https://ci-release.nodejs.org/job/iojs+release-ulises-experimental/5/ ran for `main` (e.g. Node.js 22) which means that it will skip macOS 10.15 due to https://github.com/nodejs/build/blob/a7cc7402837340d90b3a1c858da6af5ea8fe89ce/jenkins/scripts/VersionSelectorScript.groovy#L76. I've requested a test build on v18.x-staging, https://ci-release.nodejs.org/job/iojs+release-ulises-experimental/6/, and you can see from https://ci-release.nodejs.org/job/iojs+release-ulises-experimental/6/console that it is trying to run on the (now removed) osx1015 labels.
```console
19:15:50 Node.js major version: 18
19:15:50 Node.js version: 18.18.3
19:15:50 Running nodes
19:15:50 Triggering [iojs+release-ulises-experimental » osx11-x64-release-tar](https://ci-release.nodejs.org/job/iojs+release-ulises-experimental/nodes=osx11-x64-release-tar/)
19:15:50 Triggering [iojs+release-ulises-experimental » osx11-arm64-release-tar](https://ci-release.nodejs.org/job/iojs+release-ulises-experimental/nodes=osx11-arm64-release-tar/)
19:15:50 Triggering [iojs+release-ulises-experimental » osx1015-release-tar](https://ci-release.nodejs.org/job/iojs+release-ulises-experimental/nodes=osx1015-release-tar/)
19:15:50 Triggering [iojs+release-ulises-experimental » osx11-release-pkg](https://ci-release.nodejs.org/job/iojs+release-ulises-experimental/nodes=osx11-release-pkg/)
19:15:50 Triggering [iojs+release-ulises-experimental » osx1015-release-pkg](https://ci-release.nodejs.org/job/iojs+release-ulises-experimental/nodes=osx1015-release-pkg/)
```

Normally you would edit the job, e.g. https://ci-release.nodejs.org/job/iojs+release/configure, and remove the labels from "Configuration Matrix". I asked in https://github.com/nodejs/build/pull/3531#issuecomment-1779565236 to do that first before removing the machines since if you edit the jobs now it won't show the osx1015 labels as no machines are available with them so you cannot deselect them.

targos:
IIRC you should just open the edit page, save it, and the config won't contain the labels anymore

targos:
I did that. osx1015 isn't here anymore:
CleanShot 2023-10-27 at 08 11 54

PR #3542, opened 2023-10-25
ansible: don't copy symlinks to R2 buckets

flakey5:
As of now, symlinks are currently being copied into R2 as two completely separate directories. The release worker handles the symlinks manually though as of nodejs/release-cloudflare-worker#39. This means we can just ignore any symlinks when syncing with R2 since they should never be read from (e.g. the worker should never be trying to read nodejs/release/latest/ from the bucket, instead nodejs/release/v21.0.0).

cc @MoLow

### DIFF
--- a/ansible/www-standalone/tools/promote/upload_to_cloudflare.sh
+++ b/ansible/www-standalone/tools/promote/upload_to_cloudflare.sh
@@ -38,6 +38,6 @@ fi
 relativedir=${dstdir/$dist_rootdir/"$site/"}
 tmpversion=$2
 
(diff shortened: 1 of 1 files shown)


MoLow (approved):
LGTM. will deploy manually

flakey5:
Should I keep this open to be merged or close it?

MoLow:
Once the upgrade of the www server is complete I will deploy this, then I will merge

targos:
The server is back up and running

MoLow:
File deployed

Issue #3499, opened 2023-09-22
Modernize Debian in CI


targos:
Our Debian hosts are not in a very good shape:
- 2 x64 hosts with Debian 8 (EoL, offline for outdated Java)
- 2 x64 hosts with Debian 9 (EoL, offline for outdated Java)
- 2 x64 hosts and 3 armv7l containers with Debian 10 (EoL in 9 months, GCC 10 not available)

![CleanShot 2023-09-22 at 15 24 37](https://github.com/nodejs/build/assets/2352663/f3e6a904-916c-4705-bc0f-057a6df7f324)

https://endoflife.date/debian

targos:
My suggestions:
- Repurpose Debian 8 x64 to Debian 11
- Repurpose Debian 9 x64 to Debian 12
- Keep Debian 10 x64 for now, especially since it's one of the Docker distributions for Node.js 20.
- Skip Debian 10 for testing Node.js 21 and later

I don't know what to do about the container setup.

richardlau:
> I don't know what to do about the container setup.

We stopped testing on Debian 10 armv7l for Node.js 20 due to the available version of libstc++ (https://github.com/nodejs/build/pull/3274, https://github.com/rvagg/rpi-newer-crosstools/pull/2). At least for armv7l I'd be okay moving the Debian 10 container to Debian 11.

targos:
Status:
- debian8 x64: migrated to debian11 and enabled in CI jobs
- debian9 x64: wip (help needed): https://github.com/nodejs/build/pull/3560
- debian10 x64: disabled on v21+
- debian10 armv7l: migrated to debian11

richardlau:
Looks like [test-rackspace-debian11-x64-1](https://ci.nodejs.org/computer/test%2Drackspace%2Ddebian11%2Dx64%2D1/) only has 2GB of memory:
```console
$ ssh test-rackspace-debian11-x64-1 free -h
               total        used        free      shared  buff/cache   available
Mem:           1.9Gi       320Mi       1.5Gi       1.0Mi       134Mi       1.5Gi
Swap:             0B          0B          0B
$
```

richardlau:
I'll try to add swap to the machine in the meantime. We have a lot of space:
```console
$ ssh test-rackspace-debian11-x64-1 df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            972M     0  972M   0% /dev
tmpfs           198M  504K  198M   1% /run
[...]

PR #3560, opened 2023-11-06
ansible: replace DO debian9-x64-1 with debian12-x64-1

targos:
Refs: https://github.com/nodejs/build/issues/3499


targos:
Draft because the Python setup changed from previous versions and I don't know how to install `tap2junit` with it.

```
  cmd: [
    '/usr/bin/python3',
    '-m',
    'pip.__main__',
    'install',
    'tap2junit==0.1.6'
  ],

:stderr: error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

cclauss:
`usr/bin/env python3 -m pip install tap2junit==0.1.6`

The `.__main__` thing seems like something to delete.

targos:
`/usr/bin/env python3 -m pip install tap2junit==0.1.6` prints the same error.

Would it be ok to install with `--break-system-packages` ?

ansible/roles/baselayout/vars/main.yml
```diff
@@ -75,6 +75,10 @@ packages: {
     'gcc-10,g++-10,ccache,git,curl,libfontconfig1,apt-transport-https,ca-certificates,sudo,python3-pip',
   ],
 
+  debian12: [
+    'gcc-12,g++-12,ccache,git,curl,libfontconfig1,apt-transport-https,ca-certificates,sudo,python3-pip',
```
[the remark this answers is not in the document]
cclauss: I think `python3 -m pip install --user tap2junit==0.1.6` is much safer than `--break-system-packages`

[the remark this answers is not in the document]
targos: The ansible config used to install tap2junit is https://github.com/nodejs/build/blob/main/ansible/roles/jenkins-worker/tasks/partials/tap2junit/pip.yml

I'm not sure how to change it so it works.

targos:
With the last commit, `tap2junit` is installed in a virtualenv:

```
# ls /home/iojs/venv/bin
activate      activate.fish  activate.ps1      pip	 pip3	  python   python3.11  wheel	   wheel3
activate.csh  activate.nu    activate_this.py  pip-3.11  pip3.11  python3  tap2junit   wheel-3.11  wheel3.11
```

Now what's the best way to have it activated when jenkins runs?

targos:
`/home/iojs/venv/bin` is a Python script with hashbang: `#!/home/iojs/venv/bin/python`

ansible/roles/baselayout/vars/main.yml
```diff
@@ -75,6 +75,10 @@ packages: {
     'gcc-10,g++-10,ccache,git,curl,libfontconfig1,apt-transport-https,ca-certificates,sudo,python3-pip',
   ],
 
+  debian12: [
+    'gcc-12,g++-12,ccache,git,curl,libfontconfig1,apt-transport-https,ca-certificates,sudo,python3-pip,python3-virtualenv',
```
cclauss, line 79: Can we use Python's built-in [`venv`](https://docs.python.org/3/library/venv.html) via `python3-venv` and `python3-full` instead?

targos: Sure, let me try.

targos: What's the command to run in this case? Ansible defaults to `virtualenv` which doesn't exist with `python3-venv`: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pip_module.html#parameter-virtualenv_command

targos: I went with `virtualenv_command: python3 -m venv`

targos:
Updated to use `venv` instead of `virtualenv`. Question above is still open.

ansible/roles/baselayout/vars/main.yml
cclauss: Install the venv: `python3 -m venv name_of_my_venv`
Activate the venv (optional): `source name_of_my_venv/bin/activate`

targos:
I can think of solutions but can't say if they're good:
- Add the location of the venv/bin directory to the PATH in https://github.com/nodejs/build/blob/main/ansible/roles/jenkins-worker/templates/systemd.service.j2#L21
- Add a step that creates a symlink of `tap2junit` in `/usr/bin` or `/usr/local/bin`
- Source `venv/bin/activate` if it exists before starting Jenkins

cclauss:
* [x] Add the location of the venv/bin directory to the PATH

targos:
Subsidiary question: at the start or the end of the PATH?

Another question: the current setup installs the venv as root, so maybe we want to do it in another place than `/home/iojs`, or we want to install it as the `iojs` user?

cclauss:
https://packages.debian.org/stable/pipx also brings along `python3-venv` and friends.

targos:
Running with `pipx` sounds much more involved because we would need to do it conditionally until all CI hosts are setup with it.

cclauss (approved):

targos:
Test build: https://ci.nodejs.org/job/node-test-commit-linux-targos/nodes=debian12-x64/1/

targos:
https://ci.nodejs.org/job/node-test-commit-linux-targos/3/nodes=debian12-x64/

targos:
Test run is green.

richardlau (approved):

Issue #3127, opened 2022-12-20
ansible: "include" is deprecated


targos:
```
[DEPRECATION WARNING]: "include" is deprecated, use include_tasks/import_tasks instead. This feature will be removed in version 2.16. Deprecation warnings can be disabled by setting deprecation_warnings=False
 in ansible.cfg.
```

targos:
The latest version of ansible now quickly stops with an error:

```
 ansible-playbook --limit test-digitalocean-ubuntu2204-x64-1 ansible/playbooks/jenkins/worker/create.yml

PLAY [test,release,!*-win*] **************************************************************************************************************************

TASK [check if secret is properly set] ***************************************************************************************************************
ok: [test-digitalocean-ubuntu2204-x64-1]

TASK [read-secrets : Validating arguments against arg spec 'main' - Sets facts based on data from secrets.] ******************************************
ok: [test-digitalocean-ubuntu2204-x64-1]

TASK [read-secrets : decrypt group secrets] **********************************************************************************************************
...
ok: [test-digitalocean-ubuntu2204-x64-1] => (item=None)
ok: [test-digitalocean-ubuntu2204-x64-1]
ERROR! [DEPRECATED]: ansible.builtin.include has been removed. Use include_tasks or import_tasks instead. This feature was removed from ansible-core in a release after 2023-05-16. Please update your playbooks.
```

PR #3580, opened 2023-12-04
ansible: use include_tasks instead of include

targos:
Closes: https://github.com/nodejs/build/issues/3127

Note that I'm not sure at all that it's the correct fix (also see inline comments).

### DIFF
--- a/ansible/roles/package-upgrade/handlers/main.yml
+++ b/ansible/roles/package-upgrade/handlers/main.yml
@@ -3,10 +3,9 @@
 # Package upgrade handler
 
 - name: package updated
-  include: "{{ handler_include }}"
+  include_tasks: "{{ handler_include }}"
   loop_control:
     loop_var: handler_include
-  notify: package updated
   with_first_found:
     - files:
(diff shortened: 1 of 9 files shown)


ansible/roles/package-upgrade/handlers/main.yml
```diff
@@ -3,10 +3,9 @@
 # Package upgrade handler
 
 - name: package updated
-  include: "{{ handler_include }}"
+  include_tasks: "{{ handler_include }}"
   loop_control:
     loop_var: handler_include
-  notify: package updated
```
targos, line 9: `notify` is not a valid option with `include_tasks`

ansible/roles/package-upgrade/tasks/main.yml
```diff
@@ -6,5 +6,4 @@
 
   - name: include package manager tasks
     when: not os|startswith("zos")
-    include: "partials/{{ template }}.yml"
-    failed_when: template is False
```
targos, line 10: ditto

targos:
@nodejs/build

PR #3582, opened 2023-12-07
ansible: upgrade armv7l containers to Debian 11

targos:
Refs: https://github.com/nodejs/build/issues/3499

### DIFF
--- a/ansible/roles/docker/templates/debian11_armv7l.Dockerfile.j2
+++ b/ansible/roles/docker/templates/debian11_armv7l.Dockerfile.j2
@@ -1,4 +1,4 @@
-FROM arm32v7/debian:10
+FROM arm32v7/debian:11
 
 ENV LC_ALL C
 ENV USER {{ server_user }}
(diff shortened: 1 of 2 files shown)


richardlau (approved):

targos:
Test build: https://ci.nodejs.org/view/All/job/node-test-commit-linux-targos/7/nodes=debian11-armv7l/

richardlau:
> Why do we have "fanned" jobs for armv7l? Compiling directly in the container seems to work fine?

In the release CI the armv7l builds are cross compiled so the fanned jobs are more representative of what we release. If we want to change the release builds to compile natively on armv7l then we'd need to pick a Linux distro to use -- we use RHEL 8 elsewhere for Linux but there's is no armv7l support in RHEL (only arm64)).

targos:
Ok I see. Is the test job I did enough to validate that it will (probably) work in cross-compiled tests?

richardlau:
> Ok I see. Is the test job I did enough to validate that it will (probably) work in cross-compiled tests?

I think so... the only thing I cannot remember is whether we need to add the host key for the jenkins workspace machine that hosts the temporary binary git repository to `known_hosts`.

targos:
> the only thing I cannot remember is whether we need to add the host key for the jenkins workspace machine that hosts the temporary binary git repository to `known_hosts`.

Fanned job passed, so I guess it's good?

PR #3601, opened 2023-12-28
jenkins: fix vs2019 exclusion for native suites job

StefanStojanovic:
While working on adding new Windows 10 and 11 machines to the CI I noticed that some configurations were getting excluded in the native add-on tests when they should be included. After some investigation, I narrowed it down to VS2019 machines. This change fixes that behavior as seen from tests done in this [temporary copied job](https://ci.nodejs.org/view/All/job/mefi-node-test-binary-windows-native-suites/).


targos (approved):

UlisesGascon (approved):

MoLow (approved):

mhdawson (approved):
LGTM

Issue #3630, opened 2024-02-13
Update compiler on Linux s390x


richardlau:
@miladfarca has found that there is a nasty gcc bug on s390x with `std::optional has_value()` (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106355) during review of https://chromium-review.googlesource.com/c/v8/v8/+/5246687 (where V8 is proposing to drop their custom `v8::base::Optional` for `std::optional`).

While the upstream gcc bug has supposedly been fixed in gcc 10.5.0, that isn't available via `dnf` on RHEL 8 as `gcc-toolset-10` was "retired" in November 2022 (FWIW `gcc-toolset-11` is also considered "retired"), which probably means it is unlikely to get updates (we'll check): https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle

What we know so far:

> Tested the bug (using the example in bugzilla) on ubi8:
> - gcc-toolset-10 with `cc 10.3.1`: broken
> - gcc-toolset-11 with `cc 11.2.1`: broken
> - gcc-toolset-12 with `cc 12.2.1`: ok

A quick scan of the Node.js codebase shows we are already using `std::optional` outside of V8, although I don't know if we're hitting the bug with those usages.

So we're probably going to have to move to at least `gcc-toolset-12` for Linux on s390x, and the only question is when and whether we make the switch for existing release lines to be on the safe side. Switching to `gcc-toolset-12` should be safe from a runtime compatibility perspective as the code will still be linked against the glibc and libstdc++ versions in base RHEL 8.

targos:
FWIW I'm always up for upgrading compilers anywhere 😄

richardlau:
> @richardlau is the V8 code with the bug already in existing release lines or if not, likely to be backported ?

Current V8 in Node.js uses V8's custom `optional`. I don't know if the change will end up in the version of V8 we'll end up using in Node.js 22.

mhdawson:
Ok so the question is if we need to have a new gcc for Node.js 22 or Node.js 23, right?

mhdawson:
Thinking about it, we do often pull in a version of v8 after 22 is cut, so it may not be easy to know that until close to the LTS promotion.

It may be best to assume it will be needed for 22 so that we don't need to update compilers after 22 is cut.

richardlau:
Another s390x gcc bug has been identified, this time in gcc 12 and 13 when c++20 is enabled: 
[...]

PR #3649, opened 2024-03-12
ansible: add gcc-toolsets 12 and 13 to RHEL 8

richardlau:
Add gcc 12 and 13 to our RHEL 8 machines.

---

Semi-related to https://github.com/nodejs/build/issues/3630, this PR will add gcc-toolset-12 and gcc-toolset-13 to our RHEL 8 machines, which gives us the ability to do test builds with gcc 12 and/or gcc 13. It does not switch the compiler selection to build Node.js -- that will be done in a follow up PR.

### DIFF
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -104,7 +104,7 @@ packages: {
   ],
 
   rhel8: [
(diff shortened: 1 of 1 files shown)


targos (approved):

richardlau:
I've updated the RHEL 8 x64 test and release machines, and the test RHEL 8 ppc64le machines. Currently (re)Ansibling the release RHEL 8 ppc64le machine.

Issue #3335, opened 2023-05-03
Unpin or update GN version


targos:
Node.js 14 is EoL.

https://github.com/nodejs/build/blob/612dcfead3720c178a76cfd827cd7ca408167987/ansible/roles/gn/vars/main.yml#L9-L12

PR #3659, opened 2024-03-21
use gcc 12 on rhel8-s390x for Node.js 22

richardlau:
- ansible: add libatomic-devel to gcc toolsets
- ansible: build newer version of gn
- jenkins: use gcc 12 on rhel8-s390x for Node.js 22

Refs: https://github.com/nodejs/build/issues/3630
Fixes: https://github.com/nodejs/build/issues/3335

### DIFF
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -104,7 +104,7 @@ packages: {
   ],
 
   rhel8: [
-    'ccache,cmake,gcc-c++,gcc-toolset-10,gcc-toolset-10-libatomic-devel,gcc-toolset-11,gcc-toolset-12,gcc-toolset-13,git,make,python3',
+    'ccache,cmake,gcc-c++,gcc-toolset-10,gcc-toolset-10-libatomic-devel,gcc-toolset-11,gcc-toolset-12,gcc-toolset-12-libatomic-devel,gcc-toolset-13,gcc-toolset-13-libatomic-devel,git,make,python3',
   ],
 
   rhel9: [
--- a/ansible/roles/build-test-v8/files/gcc-bug-113960.patch
+++ b/ansible/roles/build-test-v8/files/gcc-bug-113960.patch
@@ -1,0 +1,57 @@
+--- a/usr/include/c++/12/bits/stl_algobase.h
++++ b/usr/include/c++/12/bits/stl_algobase.h
+@@ -1778,11 +1778,14 @@
+     }
+
+ #if __cpp_lib_three_way_comparison
+-  // Iter points to a contiguous range of unsigned narrow character type
+-  // or std::byte, suitable for comparison by memcmp.
+-  template<typename _Iter>
+-    concept __is_byte_iter = contiguous_iterator<_Iter>
+-      && __is_memcmp_ordered<iter_value_t<_Iter>>::__value;
++  // Both iterators refer to contiguous ranges of unsigned narrow characters,
++  // or std::byte, or big-endian unsigned integers, suitable for comparison
++  // using memcmp.
++  template<typename _Iter1, typename _Iter2>
++    concept __memcmp_ordered_with
++      = (__is_memcmp_ordered_with<iter_value_t<_Iter1>,
++                                 iter_value_t<_Iter2>>::__value)
++         && contiguous_iterator<_Iter1> && contiguous_iterator<_Iter2>;
+
+   // Return a struct with two members, initialized to the smaller of x and y
+   // (or x if they compare equal) and the result of the comparison x <=> y.
+@@ -1832,20 +1835,20 @@
+       if (!std::__is_constant_evaluated())
+        if constexpr (same_as<_Comp, __detail::_Synth3way>
+                      || same_as<_Comp, compare_three_way>)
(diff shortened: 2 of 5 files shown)


richardlau:
I've deployed this onto test-ibm-rhel8-s390x-4 and am running a test build with https://github.com/nodejs/node/pull/52183: https://ci.nodejs.org/job/richardlau-node-test-commit-v8-linux/546/nodes=rhel8-s390x,v8test=v8test/

ansible/roles/baselayout/vars/main.yml
```diff
@@ -104,7 +104,7 @@ packages: {
   ],
 
   rhel8: [
-    'ccache,cmake,gcc-c++,gcc-toolset-10,gcc-toolset-10-libatomic-devel,gcc-toolset-11,gcc-toolset-12,gcc-toolset-13,git,make,python3',
+    'ccache,cmake,gcc-c++,gcc-toolset-10,gcc-toolset-10-libatomic-devel,gcc-toolset-11,gcc-toolset-12,gcc-toolset-12-libatomic-devel,gcc-toolset-13-libatomic-devel,git,make,python3',
```
targos, line 107: I guess we don't need `gcc-toolset-13` explicitly because `gcc-toolset-13-libatomic-devel` depends on it?

targos (approved):

richardlau:
> I've deployed this onto test-ibm-rhel8-s390x-4 and am running a test build with [nodejs/node#52183](https://github.com/nodejs/node/pull/52183): https://ci.nodejs.org/job/richardlau-node-test-commit-v8-linux/546/nodes=rhel8-s390x,v8test=v8test/

Looks like we need https://github.com/v8/v8/commit/f8d5e576b814c92c39ec0cea80c21e4162270e12 (https://chromium-review.googlesource.com/c/v8/v8/+/5280535) as well.

ansible/roles/baselayout/vars/main.yml
richardlau: ah that's an oversight. I'd rather be specific -- I'll add it back.

richardlau: Add back explicit gcc-toolset-13.

richardlau:
Updated https://github.com/nodejs/node/pull/52183 to include https://chromium-review.googlesource.com/c/v8/v8/+/5280535.

Retry: https://ci.nodejs.org/job/richardlau-node-test-commit-v8-linux/547/nodes=rhel8-s390x,v8test=v8test/

richardlau:
> Retry: https://ci.nodejs.org/job/richardlau-node-test-commit-v8-linux/547/nodes=rhel8-s390x,v8test=v8test/

This is failing two tests:
- [v8tests.inspector/runtime/es6-module](https://ci.nodejs.org/job/richardlau-node-test-commit-v8-linux/547/nodes=rhel8-s390x,v8test=v8test/testReport/junit/(root)/v8tests/inspector_runtime_es6_module/)
- [v8tests.inspector/debugger/evaluate-on-call-frame-in-module](https://ci.nodejs.org/job/richardlau-node-test-commit-v8-linux/547/nodes=rhel8-s390x,v8test=v8test/testReport/junit/(root)/v8tests/inspector_debugger_evaluate_on_call_frame_in_module/)

This is expected while we wait for https://issues.redhat.com/browse/RHEL-29952 to backport https://gcc.gnu.org/bugzilla/show_bug.cgi?id=113960.

FWIW the Red Hat team maintaining the upstream s390x V8 port are currently manually patching one of the libstdlibc++ headers as per https://gcc.gnu.org/bugzilla/show_bug.cgi?id=113960#c11 in the container they use to build/test V8:
https://github.com/nodeshift/v8-build/blob/85b7ce18cb95cacd68ba49adae2ed7bc1f23be2c/Dockerfile.base#L42-L47

richardlau:
> We don't build Node.js with C++20 (yet) so it's OK?

Yes, wrt https://gcc.gnu.org/bugzilla/show_bug.cgi?id=113960. My understanding is that V8 is building with C++20, hence this showing up in V8 build -- presumably the V8 CI is inheriting C++20 from some part of the gn build process. 

I was hoping to avoid needing to patch the libstdc++ headers on the s390x machines ourselves and if possible wait for the gcc-toolset-12 packages to be updated, but patching the headers ourselves is an (temporary) option if we want to move to C++20 for Node.js for 22.

anonrig (approved):

richardlau:
> 
> I was hoping to avoid needing to patch the libstdc++ headers on the s390x machines ourselves and if possible wait for the gcc-toolset-12 packages to be updated, but patching the headers ourselves is an (temporary) option if we want to move to C++20 for Node.js for 22.

Latest from RH for updating gcc-toolset-12 to patch https://gcc.gnu.org/bugzilla/show_bug.cgi?id=113960 is "around May". I've added a commit to temporarily patch the libstdc++ header file for gcc-toolset-12, which we'll revert when the toolset is updated.

Started a test build with the patched header:
https://ci.nodejs.org/job/richardlau-node-test-commit-v8-linux/577/nodes=rhel8-s390x,v8test=v8test/

mhdawson (approved):
LGTM

PR #3664, opened 2024-03-27
Hetzner-role-updates

ryanaslett:
This PR updates the ansible to build perf and updates the packages for the benchmarking jobs.

### DIFF
--- a/ansible/playbooks/jenkins/worker/create.yml
+++ b/ansible/playbooks/jenkins/worker/create.yml
@@ -73,9 +73,10 @@
 #
 
 - hosts:
-    - test-*intel-ubuntu*-x*
+    - test-*hetzner-ubuntu*-x*
   roles:
-    - linux-perf
+    - name: linux-perf
+      tags: linux-perf
 
 #
 # Ensure node is not installed anywhere but the linter servers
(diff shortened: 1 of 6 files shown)


ryanaslett:
This ended up with the two commits from https://github.com/nodejs/build/pull/3661 hopefully that will work.

ansible/playbooks/jenkins/worker/create.yml
```diff
@@ -72,9 +73,10 @@
 #
 
 - hosts:
-    - test-*intel-ubuntu*-x*
+    - test-*hetzner-ubuntu*-x*
```
richardlau, line 76: This could be 
```suggestion
    - test-hetzner-ubuntu*-x*
```
or even better if it was similar to the `benchmarking` role where it was conditional on an inventory variable rather than based on the hostname. But I'm fine with landing the PR as-is and iterating on it afterwards.

richardlau (approved):

ansible/playbooks/jenkins/worker/create.yml
ryanaslett: Agreed. that linux-perf role is currently *only* used in benchmarking so I'll adjust that to be the same as the other benchmarking role.

My original changes were "tread as lightly as possible, things are potentially this way for a reason", and also "get it done before the deadline"

PR #3692, opened 2024-04-24
ansible: add 2GB swap to debian x64 hosts

richardlau:
Adds 2GB swap to
- test-digitalocean-debian11-x64-1
- test-digitalocean-debian12-x64-1
- test-softlayer-debian12-x64-1

in an attempt to stabilize the CI after recent V8 updates.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -111,8 +111,8 @@ hosts:
         msft-win2016_vs2017-x64-2: {ip: nodejs.westus3.cloudapp.azure.com}
 
     - digitalocean:
-        debian11-x64-1: {ip: 174.138.79.159}
-        debian12-x64-1: {ip: 159.203.105.159}
+        debian11-x64-1: {ip: 174.138.79.159, swap_file_size_mb: 2048}
+        debian12-x64-1: {ip: 159.203.105.159, swap_file_size_mb: 2048}
         fedora37-x64-1: {ip: 159.65.248.149}
         fedora38-x64-1: {ip: 162.243.187.89}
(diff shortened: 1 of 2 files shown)


richardlau:
Test CI for `refs/heads/v22.0.0-proposal`: https://ci.nodejs.org/job/node-test-commit-linux/57437/nodes=debian12-x64/

targos (approved):
Maybe we should generalize this. Fedora machines also have this issue.

richardlau:
I've added swap to test-digitalocean-debian11-x64-1 as well.
FTR [test-rackspace-debian11-x64-1](https://ci.nodejs.org/computer/test%2Drackspace%2Ddebian11%2Dx64%2D1/) already has 4GB of swap.

I'll look at the Fedora machines next.

targos:
I feel like we should change our strategy and instead of N machines with 2 CPUs and relatively little RAM we should have N/2 machines with 4 CPUs and more RAM 🤔

targos:
ok, I'm probably misremembering.

richardlau:
> I feel like we should change our strategy and instead of N machines with 2 CPUs and relatively little RAM we should have N/2 machines with 4 CPUs and more RAM 🤔

I don't necessarily disagree but we might need to do more investigation. Going from 2 CPUs to 4 CPUs may require even more memory if the build is potentially running twice as many parallel threads at once.

Also N is generally 2 at the moment for the test CI which gives us a little redundancy (i.e. one machine goes down (or is taken down for maintenance) and builds continue on the other). Halving 2 (i.e. 1) would mean no redundancy.

Issue #3710, opened 2024-05-09
Ansible jenkins/worker/create playbook cannot get adoptopenjdk metadata


richardlau:
Attempting to run [`ansible/playbooks/jenkins/worker/create.yml`](https://github.com/nodejs/build/blob/102f630f31a33cf9762e98bfd10da416d472010c/ansible/playbooks/jenkins/worker/create.yml) against an AIX machine (test-ibm-aix72-ppc64_be-2):
```console
TASK [java-base : fetch adoptopenjdk metadata] ******************************************************************************************************************************************************************
fatal: [test-ibm-aix72-ppc64_be-2]: FAILED! => {"alt_svc": "h3=\":443\"; ma=86400", "cf_cache_status": "DYNAMIC", "cf_ray": "8812af18c99f440b-EWR", "changed": false, "connection": "close", "content": "<html>\n<head><title>403 Forbidden</title></head>\n<body>\n<center><h1>403 Forbidden</h1></center>\n<hr><center>Microsoft-Azure-Application-Gateway/v2</center>\n</body>\n</html>\n", "content_type": "text/html", "date": "Thu, 09 May 2024 15:22:15 GMT", "elapsed": 0, "msg": "Status code was 403 and not [200]: HTTP Error 403: Forbidden", "redirected": false, "server": "cloudflare", "status": 403, "transfer_encoding": "chunked", "url": "https://api.adoptopenjdk.net/v3/assets/feature_releases/17/ga?architecture=ppc64&image_type=jre&jvm_impl=openj9&os=aix&project=jdk&heap_size=normal&page_size=1&sort_method=DEFAULT&sort_order=DESC&vendor=adoptopenjdk"}
```

The 403 error is strange -- with a web browser or curl I get redirected instead.

If I change the url from api.adoptopenjdk.net to api.adoptium.net I get a different error:
```console
TASK [java-base : fetch adoptopenjdk metadata] ******************************************************************************************************************************************************************
[...]

PR #3711, opened 2024-05-09
ansible: fix adoptopenjdk metadata fetch

richardlau:
Pass an empty `body` when fetching AdoptOpenJDK metadata to prevent getting a 403 error from CloudFlare.

Fixes: https://github.com/nodejs/build/issues/3710

### DIFF
--- a/ansible/roles/java-base/tasks/main.yml
+++ b/ansible/roles/java-base/tasks/main.yml
@@ -57,6 +57,7 @@
   check_mode: no
   register: adoptopenjdk_metadata
   uri:
+    body: ""
     body_format: json
     return_content: yes
     status_code: 200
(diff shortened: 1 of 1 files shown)


ansible/roles/java-base/tasks/main.yml
```diff
@@ -57,6 +57,7 @@
   check_mode: no
   register: adoptopenjdk_metadata
   uri:
+    body: ""
     body_format: json
```
targos, line 61: Is this necessary? It may be the cause of the null value being sent by default

targos (approved):

ansible/roles/java-base/tasks/main.yml
richardlau: I'll check what happens without it.

richardlau: Looks like it is isn't needed. I've removed both `body` and `body_format` options.

targos (approved):

UlisesGascon (approved):

mhdawson (approved):
LGTM

PR #3730, opened 2024-05-17
jenkins: pack embedtest.exe for Windows builds

vmoroz:
CI runs for PR https://github.com/nodejs/node/pull/52646 fail because the `embedtest.exe` is not packaged along with other binary files and it cannot be found by the test being enabled.

This PR changes Jenkins Windows scripts to treat `embedtest.exe` the same way as `cctest.exe`.
The only difference is that we check that `embedtest.exe` file exists.
It must unblock the PR https://github.com/nodejs/node/pull/52646.

### DIFF
--- a/jenkins/scripts/windows/node-compile-windows.cmd
+++ b/jenkins/scripts/windows/node-compile-windows.cmd
@@ -15,7 +15,7 @@ call %~dp0compile.cmd
 if errorlevel 1 exit /b
 
 :: Select files to pack
-set "BINARY_FILES=config.gypi icu_config.gypi Release/node.exe Release/node.lib Release/openssl-cli.exe Release/cctest.exe Release/node.pdb"
+set "BINARY_FILES=config.gypi icu_config.gypi Release/node.exe Release/node.lib Release/openssl-cli.exe Release/cctest.exe Release/embedtest.exe Release/node.pdb"
 if exist Release\node_test_engine.dll set "BINARY_FILES=%BINARY_FILES% Release/node_test_engine.dll"
 if exist Release\overlapped-checker.exe set "BINARY_FILES=%BINARY_FILES% Release/overlapped-checker.exe"
 
--- a/jenkins/scripts/windows/node-compile-windows.cmd
+++ b/jenkins/scripts/windows/node-compile-windows.cmd
@@ -15,7 +15,8 @@ call %~dp0compile.cmd
 if errorlevel 1 exit /b
 
 :: Select files to pack
-set "BINARY_FILES=config.gypi icu_config.gypi Release/node.exe Release/node.lib Release/openssl-cli.exe Release/cctest.exe Release/embedtest.exe Release/node.pdb"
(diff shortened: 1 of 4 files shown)


jenkins/scripts/windows/node-compile-windows.cmd
```diff
@@ -15,7 +15,7 @@ call %~dp0compile.cmd
 if errorlevel 1 exit /b
 
 :: Select files to pack
-set "BINARY_FILES=config.gypi icu_config.gypi Release/node.exe Release/node.lib Release/openssl-cli.exe Release/cctest.exe Release/node.pdb"
+set "BINARY_FILES=config.gypi icu_config.gypi Release/node.exe Release/node.lib Release/openssl-cli.exe Release/cctest.exe Release/embedtest.exe Release/node.pdb"
```
richardlau, line 18: I'd recommend putting `embedtest.exe` behind an `if exist` check like the lines below otherwise you'll have to wait for the PR that adds `embedtest.exe` to be backported to all supported release lines before this change could land.

vmoroz: I can do it, but the `embedtest.exe` is already being built. It is part of an existing test,
I am just trying to enable it running on Windows.

vmoroz: I have updated the script to make use of `embedtest.exe` conditional.

richardlau (approved):

richardlau:
cc @StefanStojanovic

StefanStojanovic (approved):
The changes look reasonable and good. Just to check, was this tested through CI? Eg. with a copy of the Windows job using this branch instead of the main one?

richardlau:
> Just to check, was this tested through CI? Eg. with a copy of the Windows job using this branch instead of the main one?

I don't believe so.

vmoroz:
> The changes look reasonable and good. Just to check, was this tested through CI? Eg. with a copy of the Windows job using this branch instead of the main one?

No, I did not test it this way. How can I do it?

StefanStojanovic:
> No, I did not test it this way. How can I do it?

You'd need a higher access level to Jenkins to create a job copy and point it to your branch. I will do that for you. and send the run  here.

StefanStojanovic:
Here is [the CI run](https://ci.nodejs.org/view/All/job/mefi-embedtest-node-test-commit-windows-fanned/2/) (run 1 failure is something unrelated to this). I used this branch for jobs and ran the branch from https://github.com/nodejs/node/pull/52646 through CI. Everything seems to be in order.

mhdawson:
Based on what @StefanStojanovic mentioned in terms of testing looks like this is ready to land, landing.

vmoroz:
> Here is [the CI run](https://ci.nodejs.org/view/All/job/mefi-embedtest-node-test-commit-windows-fanned/2/) (run 1 failure is something unrelated to this). I used this branch for jobs and ran the branch from [nodejs/node#52646](https://github.com/nodejs/node/pull/52646) through CI. Everything seems to be in order.

Thank you, @StefanStojanovic , for verifying it!

Issue #3723, opened 2024-05-14
Coverity builds broken


richardlau:
Two most recent [node-daily-coverity](https://ci.nodejs.org/view/Node.js%20Daily/job/node-daily-coverity/) builds have failed. 
Some error about the agent going offline, no obvious other error.

e.g. https://ci.nodejs.org/view/Node.js%20Daily/job/node-daily-coverity/3010/console
```console
make[1]: *** Deleting file '/home/iojs/build/workspace/node-daily-coverity/out/Release/obj.target/v8_compiler/deps/v8/src/compiler/graph.o'
make[1]: *** Deleting file '/home/iojs/build/workspace/node-daily-coverity/out/Release/obj.target/v8_compiler/deps/v8/src/compiler/graph-visualizer.o'
make[1]: *** Deleting file '/home/iojs/build/workspace/node-daily-coverity/out/Release/obj.target/v8_compiler/deps/v8/src/compiler/graph-reducer.o'
make[1]: *** Deleting file '/home/iojs/build/workspace/node-daily-coverity/out/Release/obj.target/v8_compiler/deps/v8/src/compiler/frame-states.o'
FATAL: Unable to delete script file /tmp/jenkins5048327983852336012.sh
java.nio.channels.ClosedChannelException
	at org.jenkinsci.remoting.protocol.NetworkLayer.onRecvClosed(NetworkLayer.java:155)
	at org.jenkinsci.remoting.protocol.impl.NIONetworkLayer.ready(NIONetworkLayer.java:143)
	at org.jenkinsci.remoting.protocol.IOHub$OnReady.run(IOHub.java:789)
	at jenkins.util.ContextResettingExecutorService$1.run(ContextResettingExecutorService.java:28)
	at jenkins.security.ImpersonatingExecutorService$1.run(ImpersonatingExecutorService.java:68)
	at jenkins.util.ErrorLoggingExecutorService.lambda$wrap$0(ErrorLoggingExecutorService.java:51)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
[...]

PR #3752, opened 2024-06-05
ansible: move Coverity Scan build tool install

richardlau:
Install the Coverity Scan build tool on the `benchmark` machines instead of the `jenkins-workspace` machines.

Refs: https://github.com/nodejs/build/issues/3723

---

To run the playbook I also needed to patch the `benchmark` role to use gcc 12 instead of the default gcc 11 to compile `h2load` from `nghttp2` due to https://github.com/nghttp2/nghttp2/issues/2194.


richardlau:
I'll swap the [node-daily-coverity](https://ci.nodejs.org/view/Node.js%20Daily/job/node-daily-coverity/) job to run on the [benchmark](https://ci.nodejs.org/label/benchmark/) machines and run some test builds. They will fail to upload the scan (since we are limited to one submission per day) but we should be able to check that the build completes up to the upload step.

targos (approved):

richardlau:
https://ci.nodejs.org/view/Node.js%20Daily/job/node-daily-coverity/3037/ looks okay, apart from the expected failure to submit/upload due to the one per day limit.

mhdawson (approved):
LGTM

PR #3755, opened 2024-06-11
ansible: ensure Iptables persists between reboots

ryanaslett:
Adds the apt package `iptables-persistent` to preserve iptables between reboots.


ryanaslett:
oops. extra commits.

PR #3758, opened 2024-06-13
ansible: update sharedlibs containers

richardlau:
- [ansible: remove ICU 68 from sharedlibs containers](https://github.com/nodejs/build/commit/d5ee0b3221b68879aed6cd151f8cead92d2f964b) 
The earliest supported Node.js version we still build in the CI is 18.x.
For Node.js 18.x, the minimum supported version of ICU is 69.
- [ansible: add OpenSSL 3.2 to sharedlibs containers](https://github.com/nodejs/build/commit/719daecd26ab07f3bec7e50e6ab8b074ac165dd4)


richardlau:
I've deployed this onto the three x64 docker hosts in the test CI. 

I'm testing the OpenSSL 3.2 addition in https://ci.nodejs.org/job/richardlau-node-test-commit-linux-containered/26/ but there are expected failures (see https://github.com/nodejs/node/issues/53382) -- adding OpenSSL 3.2 to the containers and in a copy of the job (https://ci.nodejs.org/job/richardlau-node-test-commit-linux-containered) will allow easier testing of Node.js compiled with OpenSSL 3.2.

(For the moment, the plan is to stick with OpenSSL 3.0 as long as possible as it's the current OpenSSL LTS version. There's no indication yet what the next OpenSSL LTS version will be, but testing against later OpenSSL should hopefully ease the upgrade path when it happens.)

PR #3773, opened 2024-06-20
ansible: install gcc 12 on AIX

richardlau:


mhdawson (approved):
LGTM

richardlau:
Pushed update to create a `gcc-12` and `g++-12` symlink for `ccache`.

Issue #3799, opened 2024-07-02
metrics: produce_summaries service is not processing `date` parameter


trivikr:
## Describe the bug
The produce_summaries service is not processing `date` parameter

## Steps to reproduce
Refs: https://github.com/nodejs/build/issues/3780#issuecomment-2203049104

#### Observed behavior
The summary is created for `20240701`

#### Expected behavior
The summary should have been created for `20240401`

trivikr:
This happens, since the query params in express are in `req.query` and not `req.params`

Test code
```js
import express from "express";

const app = express();
const port = 3000;
app.use(express.json());

app
  .get("/", (req, res) => {
    res.send(`query: ${JSON.stringify(req.query)}, params: ${JSON.stringify(req.params)}`);
  })
  .listen(port, () => {
    console.log(`Server is running on http://localhost:${port}`);
  });
```

Test output
```console
$ curl http://localhost:3000/\?date\=20240401
query: {"date":"20240401"}, params: {}
```

Docs: https://expressjs.com/en/api.html#req.query

The update in https://github.com/nodejs/build/pull/3791 checks for `req.params`

trivikr:
Alternatively, it'll be a cleaner solution if we write a separate route for processing dates

Test code
```js
import express from "express";

const app = express();
const port = 3000;
app.use(express.json());

app
  .get("/date/:date", (req, res) => {
    res.send(`query: ${JSON.stringify(req.query)}, params: ${JSON.stringify(req.params)}`);
  })
  .listen(port, () => {
    console.log(`Server is running on http://localhost:${port}`);
  });
```

Test output
```console
$ curl http://localhost:3000/date/20240401
query: {}, params: {"date":"20240401"}%
```

Docs: https://expressjs.com/en/api.html#req.params

PR #3800, opened 2024-07-02
metrics: use query for passing date to process summary

trivikr:
Fixes: https://github.com/nodejs/build/issues/3799

Test code in https://github.com/nodejs/build/issues/3799#issuecomment-2203071925


trivikr:
Replaced by a cleaner alternative which contains new route in https://github.com/nodejs/build/pull/3801

PR #3802, opened 2024-07-02
metrics: convert from CommonJS to ES Modules

trivikr:
Converts the scripts in metrics from CommonJS to ES Modules


trivikr:
Closing, as I'll put forward a proposal to have single Cloud Run instance with routes for the following:
* `generate-index`
* `produce-summary`
* `process-log`

PR #3822, opened 2024-07-10
ansible: update ppc64le and s390x machines

richardlau:
Decommission centos7-ppc64le and rhel7-s390x machines. Create new RHEL 9 test CI replacements.

Add 2 GB swap to all ppc64le and s390x machines.


mhdawson (approved):
LGTM

Issue #3686, opened 2024-04-19
Infrastructure for Orka (2024 and beyond)


UlisesGascon:
I plan to work on it during the weekend, so I can provide a good overview on the next build meeting on Tuesday.

**Current status**

- [x] Cleaned up all the macos machines that are available via SSH
- [ ] Open discussion with support for macstadium machines with ssh problems: https://github.com/nodejs/build/issues/3667 and https://github.com/nodejs/build/issues/3685
- [x] Open discussion to apply upgrade in the Firewall: https://github.com/nodejs/build/issues/3642 
- [ ] Review Orka state
- [ ] Plan the final allocation for resources in Orka
- [ ] Provide access for @ryanaslett to the resources: https://github.com/nodejs/build/issues/3658
- [ ] Prepare and deliver and onboarding session for @ryanaslett to the Orka infrastructure

UlisesGascon:
### Next Orka state

_updated on April 19, 2024_

**Intel Nodes**

| SSH port | Node: macpro-4 | Node: macpro-5 | Node: macpro-6 |
|----------|----------|-------------|------|
| 8822 | release-macos11-x64-1  | **test-macos13-x64-2** | test-macos11-x64-1 |
| 8823 | **test-macos13-x64-1**  | **release-macos13-x64-1** | test-macos11-x64-2 |
| 8824 | **release-macos13-x64-2** | test-macos1015-x64-2 | test-macos1015-x64-1 |
| 8825 | _empty_  | _empty_ | _empty_ |

**ARM Nodes**

_We assume that ARM Nodes can handle only 2 VMs and not +4 as Intel in the past due license limitations. This needs to be confirmed with support AFAIK?_

| SSH port | Node: arm-1 | Node: arm-2 |  Node: arm-3 | Node: arm-4 | Node: arm-5 |
|----------|-------------|-------------|-------------|-------------|-------------|
| 8822 | **test-nearform-macos11.0-arm64-1** | **release-macstadium-macos11.0-arm64-1** | **test-macstadium-macos11.0-arm64-4** | **test-macos13-arm64-2** | **release-macos13-arm64-2** |
| 8823 | **release-nearform-macos11.0-arm64-1** | **release-macos13-arm64-1** | **test-macstadium-macos11.0-arm64-3** | **test-macos13-arm64-1** | _empty_ |

targos:
> 	release-macos13-x64-2

I don't think it's necessary to have two identical release machines.

targos:
[...]

PR #3874, opened 2024-08-27
Orka: packer macos11 intel test

UlisesGascon:
### Main Changes

- Add support for SSH Connection using the default credentials.
- Use a custom base image developed by us.
- Add steps for installing Homebrew and basic dependencies.

### Related

- https://github.com/nodejs/build/issues/3686#issuecomment-2278119351

### DIFF
--- a/orka/templates/README.md
+++ b/orka/templates/README.md
@@ -40,6 +40,8 @@ You need to load the environment variables:
     echo $ORKA_AUTH_TOKEN
     echo $SSH_USERNAME
     echo $SSH_PASSWORD
+    echo $SSH_DEFAULT_USERNAME
+    echo $SSH_DEFAULT_PASSWORD
     ```
 
 ## Validate the template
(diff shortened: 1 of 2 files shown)


targos:
Didn't we agree on not migrating macOS to 11 but skipping directly to 13 instead?

UlisesGascon:
> Didn't we agree on not migrating macOS to 11 but skipping directly to 13 instead?

We will skip MacOS 12 for sure, but we need to migrate the current Orka workload (MacOS 10.15 and MacOS 11) to the new cluster, so all the infra in Orka will be ephemeral instances controlled directly by Jenkins.

You have a recap in our last meeting ([minute 33:19](https://www.youtube.com/live/eJSzKTzSMU0?si=zESZ4Iy_wtZycCDR&t=1999))

targos:
Ok, I won't object if you want to maintain 10.15 and 11, but I would be in favor of dropping them entirely. The build policy allows to do it for EoL platforms.

UlisesGascon:
> Ok, I won't object if you want to maintain 10.15 and 11, but I would be in favor of dropping them entirely. The build policy allows to do it for EoL platforms.

Oh! I love the idea a lot. I was purely thinking about the migration. I didn't realize that we are "open" to dropping support. Let me create an issue for better visibility, and let's see if the team agrees on dropping support. 👍

UlisesGascon:
As discussed in https://github.com/nodejs/build/issues/3876. I will close this PR now.

PR #3906, opened 2024-09-13
Orka template updates

UlisesGascon:
### Main Changes
- Ensure Homebrew environment is set up in the shell profile (01f78be)
- Add xz as dependency (8628fe3)
- Add xcode version information (26f4f34)
- Add tap2junit as dependency (a976199)

### Context

Related https://github.com/nodejs/build/issues/3686


targos:
Is there no way to share common parts of the hcl files? It must be annoying to always do the same thing 4 times.

UlisesGascon:
> Is there no way to share common parts of the hcl files? It must be annoying to always do the same thing 4 times.

Agree, I was expecting more differences between the templates. After the migration I will work on this unification :)

Issue #3838, opened 2024-07-17
Promotion to R2's dist-prod bucket never happens


targos:
![CleanShot 2024-07-17 at 18 20 46@2x](https://github.com/user-attachments/assets/ba388ed6-0d3c-4a62-a13a-22c9946a9845)

/cc @flakey5

targos:
It's possible that I'm using the wrong search prefix, but I get results with `nodejs/release/v20.0.0/` for example.

richardlau:
Suspect it'll be the same as we're seeing with the nightlies: https://github.com/nodejs/build/pull/3826#issuecomment-2227580402

flakey5:
> Suspect it'll be the same as we're seeing with the nightlies

+1

targos:
closing as duplicate

richardlau:
Moving conversation from https://github.com/nodejs/build/pull/3826 for visibility as that PR was merged.

Summary:
https://github.com/nodejs/build/blob/533d7e1fbce29f834f4946d83541719152851139/ansible/www-standalone/tools/promote/upload_to_cloudflare.sh#L45 is failing to recursively copy from the staging bucket to prod. It appears that Cloudflare have not fully implemented S3 and `aws s3 cp` is setting `x-amz-tagging-directive` which is one of the unimplemented features.

Refs: https://community.cloudflare.com/t/cloudflare-r2-not-supporting-copy-object/672982

flakey5:
We might be able to make `aws s3api copy-object` work ([docs](https://docs.aws.amazon.com/cli/latest/reference/s3api/copy-object.html)), but I don't think it supports recursion or pattern matching from my tests

Simplest option might just be having two copy calls, one for the smaller files and one for the larger ones that require multipart. Nowhere near ideal however, gonna try to see if there's anything else we could do

flakey5:
Could we try `aws s3 sync s3://dist-staging/<path> s3://dist-prod/<path>`? Or just `aws s3 sync s3://dist-staging s3://dist-prod`

Also still not sure on why the uploads aren't happening, but can take a look later today

richardlau:
I've changed the recursive `aws s3 cp` line in `upload_to_cloudflare.sh` with:
```
aws s3 sync $staging_bucket/$relativedir/$tmpversion/ $dist_bucket/$relativedir/$tmpversion/ --endpoint-url=$cloudflare_endpoint --profile $cloudflare_profile --no-follow-symlinks
```

[...]

PR #3840, opened 2024-07-19
ansible: mirror `upload_to_cloudflare.sh` changes

richardlau:
Refs: https://github.com/nodejs/build/issues/3838#issuecomment-2239458749

PR #3931, opened 2024-10-10
ansible: promote releases to r2 via actions

flakey5:
See https://github.com/nodejs/build/issues/3838#issuecomment-2394864184

Pre-reqs:
 * [ ] Github CLI being installed on the origin server
 * [x] nodejs/release-cloudflare-worker#142 getting merged

cc @ovflowd @richardlau

### DIFF
--- a/ansible/www-standalone/tools/promote/upload_to_cloudflare.sh
+++ b/ansible/www-standalone/tools/promote/upload_to_cloudflare.sh
@@ -42,6 +42,6 @@ fi
 relativedir=${dstdir/$dist_rootdir/"$site/"}
 tmpversion=$2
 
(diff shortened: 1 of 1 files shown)


ovflowd (approved):
LGTM, did you test this works?

UlisesGascon:
We had a Build meeting today ([recording]() and [meeting notes](https://github.com/nodejs/build/pull/3944/)) and the consensus from the team seems to be avoid adding GH credentials in the server due security concerns.

_Note: Relevant discussion starts on 15:00 aprox_

richardlau (changes requested):
To expand on today's Build WG meeting discussion -- I'm not keen on adding GitHub credentials onto the www server if we can avoid it (and I think this is no longer needed after https://github.com/nodejs/build/pull/3840). 

I consider the www server to be one of the top attack targets in terms of Node.js infrastructure, so I would like to minimize the access a potential attacker might gain if they were to somehow get onto the system (obviously we'll do our best that they do not!).

PR #3932, opened 2024-10-11
build,win: compile with clang

StefanStojanovic:
Since https://github.com/nodejs/node/pull/55249 landed, the main branch of the `node` repo supports compilation with ClangCL with PCH. MSVC and ClangCL compilation times are similar now, so we can start compiling (in some time testing, and eventually releasing) with ClangCL.

This PR adds support for compiling with ClangCL from v23. It also stops compiling v20 with VS2022 since it was temporary (we just never removed it before).

The labels we currently use in CI for VS2022 compilation are `win-vs2022` and `win-vs2022-arm64`. The format will stay the same, only `vs2022` will be replaced by `vs2022_clang`. The way I see it, this is the most consistent way of labeling it compared to what we already have in the CI, but if someone thinks differently we can discuss it.

I already added those labels to the CI machines that will be compiling and made a copy of the Windows fanned job in https://ci.nodejs.org/job/mefi-node-test-commit-windows-fanned/, so I can confirm these changes work. Once this PR lands we can edit the original compile jobs to do the same.

The second part of the plan is to make new labels, similar to these for test jobs and once I test it in the CI, open a PR for them as well.

One more notable thing is that clcache doesn't work with ClangCL, so in parallel with all of this, I'll try and see about alternatives eg. ccache for ClangCL compilation.

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -58,10 +58,9 @@ def buildExclusions = [
   [ /vs2015(-\w+)?$/,                 testType,    gte(18)       ],
   [ /vs2017(-\w+)?$/,                 testType,    gte(18)       ],
   [ /vs2019(-\w+)?$/,                 testType,    gte(21)       ],
-  [ /vs2022(-\w+)?$/,                 testType,    lt(20)        ], // Temporarily compile Node v20+ on both VS2019 and VS2022
-  [ /vs2022-x86$/,                    testType,    lt(20)        ], // Temporarily compile Node v20+ arm64 and x86 on both VS2019 and VS2022
-  [ /vs2022-x86$/,                    testType,    gte(23)       ],
-  [ /vs2022-arm64$/,                  testType,    lt(20)        ],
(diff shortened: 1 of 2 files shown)


richardlau (approved):

richardlau:
LGTM. Only question I have is whether this is pushed live before or after the Node.js 23 and 22 releases planned for Tuesday/Monday of next week.

cc @RafaelGSS @aduh95

StefanStojanovic:
> LGTM. Only question I have is whether this is pushed live before or after the Node.js 23 and 22 releases planned for Tuesday/Monday of next week.
> 
> cc @RafaelGSS @aduh95

I do not find it necessarily connected. Adding them will mean each Windows run will have 3 or 4 (depending if we add both x64 and ARM64 at the same time, or one by one) compilations instead of 2 we have now, and there are 6 compilation machines in total (also used for testing), so we may expect a decrease in our throughput. With that in mind, I'll probably wait for releases to go as we're not in a hurry with it, at least as far as I know.

StefanStojanovic:
Although I've merged it, I'll wait with using it in the CI until after the release.

PR #3947, opened 2024-11-04
ansible: don't assume the relative paths for staging & dist are the same

flakey5:
Some non-mainline releases' src dirs differ from their dst dirs (i.e. v8-canary builds upload to `/home/staging/nodejs/custom` and then are promoted to `/home/dist/nodejs/v8-canary`. This wasn't being accounted for when we promoted files from the `dist-staging` bucket to the
    `dist-prod` bucket.

### DIFF
--- a/ansible/www-standalone/tools/promote/upload_to_cloudflare.sh
+++ b/ansible/www-standalone/tools/promote/upload_to_cloudflare.sh
@@ -31,9 +31,10 @@ if [ -z ${staging_bucket+x} ]; then
   exit 1
 fi
 
-relativedir=${dstdir/$dist_rootdir/"$site/"}
+relative_srcdir=${srcdir/$staging_rootdir/"$site/"}
+relative_dstdir=${dstdir/$dist_rootdir/"$site/"}
 tmpversion=$2
 
-rclone copy $staging_bucket/$relativedir/$tmpversion/ $prod_bucket/$relativedir/$tmpversion/
-rclone copyto $staging_bucket/$relativedir/index.json $prod_bucket/$relativedir/index.json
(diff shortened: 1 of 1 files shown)


ansible/www-standalone/tools/promote/upload_to_cloudflare.sh
```diff
@@ -31,9 +31,10 @@ if [ -z ${staging_bucket+x} ]; then
...
+relative_srcdir=${srcdir/$staging_rootdir/"$site/"}
+relative_dstdir=${dstdir/$dist_rootdir/"$site/"}
 tmpversion=$2
 
-rclone copy $staging_bucket/$relativedir/$tmpversion/ $prod_bucket/$relativedir/$tmpversion/
-rclone copyto $staging_bucket/$relativedir/index.json $prod_bucket/$relativedir/index.json
-rclone copyto $staging_bucket/$relativedir/index.tab $prod_bucket/$relativedir/index.tab
+rclone copy $staging_bucket/$relative_srcdir/$tmpversion/ $prod_bucket/$relative_dstdir/$tmpversion/
+rclone copyto $staging_bucket/$relative_srcdir/index.json $prod_bucket/$relative_dstdir/index.json
+rclone copyto $staging_bucket/$relative_srcdir/index.tab $prod_bucket/$relative_dstdir/index.tab
```
targos, line 40: I think we should use `$relative_dstdir` as the source for the index files. 
They are copied to the right folder in https://github.com/nodejs/build/blob/8fc01e083c2382bf86cf0df395dbf46748b55656/ansible/www-standalone/tools/promote/_resha.sh#L41-L49

targos (approved):

richardlau (approved):

targos:
deployed

PR #3951, opened 2024-11-05
ansible: use --log-file option for rclone

flakey5:
Open to suggestion for where the logs should go, this currently writes them to `/home/dist/nodejs/rclone` which requires the creation of that folder

### DIFF
--- a/ansible/www-standalone/tools/promote/settings
+++ b/ansible/www-standalone/tools/promote/settings
@@ -39,3 +39,6 @@ chakracore_release_dirmatch=.*
 
 prod_bucket=r2:dist-prod
 staging_bucket=r2:dist-staging
+
+# format: yyyy-mm-dd-hh-mm-ss.log
+rclone_log=${dist_rootdir}rclone/$(date "+%F-%H-%M-%S.log")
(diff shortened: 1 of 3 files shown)


targos:
Does it actually enable logs? I saw that by default rclone logs nothing (maybe unless there's an error )

flakey5:
Going off of this https://forum.rclone.org/t/how-to-generate-log-file-for-rclone/13053 the log level defaults to `NOTICE`, we can set it to be more verbose however

targos (approved):

ansible/www-standalone/tools/promote/settings
```diff
@@ -39,3 +39,7 @@ chakracore_release_dirmatch=.*
 
 prod_bucket=r2:dist-prod
 staging_bucket=r2:dist-staging
+
+# format: yyyy-mm-dd-hh-mm-ss.log
+rclone_log_level=INFO
+rclone_log=${dist_rootdir}rclone/$(date "+%F-%H-%M-%S.log")
```
targos, line 45:
```suggestion
rclone_log=/home/dist/rclone/$(date "+%F-%H-%M-%S.log")
```

richardlau (approved):

Issue #3958, opened 2024-11-13
Update `check_assets` to also check in the R2 staging bucket


targos:
https://github.com/nodejs/build/blob/main/ansible/www-standalone/tools/promote/check_assets.js

To avoid future issues like https://github.com/nodejs/node/issues/55829

richardlau:
Started looking at this but it opens a [can of worms](https://en.wiktionary.org/wiki/can_of_worms):
- `check_assets.js` has tests, rather unintuitively run with `node check_assets.js test`
https://github.com/nodejs/build/blob/c26ecffc45872cff2fd38a2ad8dad48d949160a9/ansible/www-standalone/tools/promote/check_assets.js#L18-L21
  - These tests appear to fail when run😞. 
- We're still running Node.js 16.20.2(!) on the server.

I'll look at updating our version of Node.js first. We're using the Nodesource distribution and I'll need to point `apt` at the newer repository to pick up later versions of Node.js.

richardlau:
>   * These tests appear to fail when run😞.

For posterity:
<details>
<summary>node check_assets.js test</summary>

```console
$ node check_assets.js test

Test 6 Assertions ???
STDOUT------------------------------------------

... Checking assets
... Expecting a total of 41 assets for v10.x
... 2 assets waiting in staging
... 39 assets already promoted
 ✓  Complete set of expected assets in place for v10.x

------------------------------------------------
AssertionError [ERR_ASSERTION]: '... Checking assets\n' +
  '... Expecting a total of 41 assets for v10.x\n' +
  '... 2 assets waiting in staging\n' +
  '... 39 assets already promoted\n' +
  ' \x1B[32m\x1B[1m✓\x1B[22m\x1B[39m  Complete set of expected assets in place for v10.x\n' == '... Checking assets\n' +
  '... Expecting a total of 40 assets for v10.x\n' +
  '... 2 assets waiting in staging\n' +
  '... 38 assets already promoted\n' +
  ' \x1B[32m\x1B[1m✓\x1B[22m\x1B[39m  Complete set of expected assets in place for v10.x\n'
    at runTest (/home/rlau/sandbox/github/build/ansible/www-standalone/tools/promote/check_assets.js:290:12) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: '... Checking assets\n' +
    '... Expecting a total of 41 assets for v10.x\n' +
    '... 2 assets waiting in staging\n' +
    '... 39 assets already promoted\n' +
[...]

PR #3967, opened 2024-11-22
ansible: update www Node.js version

richardlau:
Update the version of Node.js used on the www server from Node.js 16 (16.20.2) to Node.js 22 (22.11.0).

Refs: https://github.com/nodejs/build/issues/3958#issuecomment-2493955665


richardlau:
Since we don't run the Ansible scripts against the live server, I've run:
```console
$ apt-get remove nodejs
$ rm /etc/apt/sources.list.d/nodesource.list*
$ bash -c 'curl -sL https://deb.nodesource.com/setup_22.x | bash -'
$ apt-get install nodejs
```

to remove the old Node.js version (16.20.2) and Nodesource repositories, and add the new repository and install Node.js from that (which installed Node.js 22.11.0).

```console
root@infra-digitalocean-ubuntu1604-x64-1:~# node -v
v22.11.0
root@infra-digitalocean-ubuntu1604-x64-1:~#
```

targos (approved):

PR #3977, opened 2024-12-09
promote: add check_r2_assets tool

richardlau:
Add tool and tests for checking if assets on R2 look ready for promotion. Relies on `rclone` being available -- there's a test for the case where it is not.

The new tool is based on existing `check_assets.js` but:
- Written in ESM.
- Uses built-in Set objects instead of emulating sets with Arrays.
- Always assumes an asset file exists. `check_assets.js` has some fallback logic, but it's not straightforward and nowadays we can consider a missing asset file to be a process failure.
- Reflects that `.done` files are not used for R2 (basically they don't exist so no handling logic). Also note that for R2, promotion is a copy and not a move so after a partial promotion subsequent checks will result in overwrite warnings for files previously promoted.
- `SHASUMS256.txt` can end up in the staging directory. Ignore for the staging directory as `check_assets.js` does for dist. (The shasums are generated by later parts of the release process.)

Fixes: https://github.com/nodejs/build/issues/3958

---

FYI @nodejs/releasers


ansible/www-standalone/tools/promote/promote_release.sh
```diff
@@ -24,6 +24,11 @@ dirmatch=$release_dirmatch
 
 node --no-warnings /home/staging/tools/promote/check_assets.js $srcdir/$2 $dstdir/$2
 
+relative_srcdir=${srcdir/$staging_rootdir/"$site/"}
+relative_dstdir=${dstdir/$dist_rootdir/"$site/"}
+
+node --no-warnings /home/staging/tools/promote/check_r2_assets.mjs $staging_bucket/$relative_srcdir/$2 $prod_bucket/$relative_dstdir/$2
```
richardlau, line 30: @targos FYI I have deployed this PR onto www and am fairly confident it should work (without doing an actual release). If this does interfere with https://github.com/nodejs/node/pull/56119 then commenting out these additional lines should restore the old behaviour (i.e. check the DO droplet only).

richardlau:
> * Reflects that `.done` files are not used for R2 (basically they don't exist so no handling logic). Also note that for R2, promotion is a copy and not a move so after a partial promotion subsequent checks will result in overwrite warnings for files previously promoted.

To illustrate the difference:
Check on already fully promoted release on DO droplet:
```console
$ node --no-warnings /home/staging/tools/promote/check_assets.js /home/staging/nodejs/release/v22.12.0 /home/dist/nodejs/release/v22.12.0
... Checking assets
... Expecting a total of 47 assets for v22.x
... 0 assets waiting in staging
... 47 assets already promoted
 ✓  Complete set of expected assets in place for v22.x
$
```

Equivalent check on R2:
```console
$ node --no-warnings /home/staging/tools/promote/check_r2_assets.mjs r2:dist-staging/nodejs/release/v22.12.0 r2:dist-prod/nodejs/release/v22.12.0
... Checking R2 assets
... Expecting a total of 47 assets for v22.x
... 47 assets waiting in R2 staging
 ⚠  47 assets already promoted in R2 will be overwritten, is this OK?
 ✓  Complete set of expected assets in place for v22.x
    Promote if you are certain this is the the correct course of action
$
```

ansible/www-standalone/tools/promote/check_r2_assets.mjs
```diff
@@ -0,0 +1,147 @@
...
+  const stagingDistIntersection = stagingAssets.intersection(distAssets);
+  const stagingDistUnion = stagingAssets.union(distAssets);
+  let notInActual = expectedAssets.difference(stagingAssets);
+  let stagingNotInExpected = stagingAssets.difference(expectedAssets);
+  let distNotInExpected = distAssets.difference(expectedAssets);
+
+  console.log('... Checking R2 assets');
+  // No expected asset list available for this line
+  if (expectedAssets.size === 0) {
+    console.log(` \u001b[31m\u001b[1m✖\u001b[22m\u001b[39m  No expected asset list is available for ${line}, does one need to be created?`);
```
targos, line 90: Feel free to ignore, but these could be simplified with [`util.styleText`](https://nodejs.org/docs/latest-v22.x/api/util.html#utilstyletextformat-text-options)

targos (approved):
I'm not very used to `node:test` but generally LGTM. Thanks!

richardlau:
> I'm not very used to `node:test` but generally LGTM. Thanks!

Neither am I 🙂.

mhdawson (approved):
LGTM

ansible/www-standalone/tools/promote/test/fixtures/all-present-v22.12.0.json
aduh95, line 1: How was this generated? How can it be updated?

richardlau: This is output from `rclone lsjson` (i.e. fixture data to simulate https://github.com/nodejs/build/blob/7b42c6a93ca9102be5f34ebaa64c0908e5cfdc0f/ansible/www-standalone/tools/promote/check_r2_assets.mjs#L51).

e.g. 
```sh
~$ rclone lsjson r2:dist-staging/nodejs/release/v22.12.0 --no-modtime --no-mimetype -R --max-depth 2
[
{"Path":"SHASUMS256.txt","Name":"SHASUMS256.txt","Size":3777,"ModTime":"","IsDir":false,"Tier":"STANDARD"},
{"Path":"docs","Name":"docs","Size":0,"ModTime":"","IsDir":true},
{"Path":"node-v22.12.0-aix-ppc64.tar.gz","Name":"node-v22.12.0-aix-ppc64.tar.gz","Size":69853847,"ModTime":"","IsDir":false,"Tier":"STANDARD"},
{"Path":"node-v22.12.0-arm64.msi","Name":"node-v22.12.0-arm64.msi","Size":26984448,"ModTime":"","IsDir":false,"Tier":"STANDARD"},
...
```

Some of the fixture data was copied and then hand edited (e.g. to manually remove or add a `{"Path":...}` to simulate missing or extra files).

PR #4018, opened 2025-02-11
Enable ccache in Windows CI

StefanStojanovic:
This PR enables using `ccache` on Windows VMs in the CI if it is present and not disabled.

This PR also has a fixup for https://github.com/nodejs/build/pull/4016 as `ccache` was only enabled for MSVC (copying as cl.exe) and not for ClangCL (copying as clang-cl.exe)

**IMPORTANT: As `ccache` option landed in `vcbuild.bat` in Node.js recently, we should wait some time before landing this as it may break CI runs on branches without that change**

### DIFF
--- a/ansible/roles/visual-studio/tasks/main.yml
+++ b/ansible/roles/visual-studio/tasks/main.yml
@@ -66,8 +66,10 @@
       version: "4.10.2"
   - name: make ccache directory
     win_command: 'mkdir C:\ccache'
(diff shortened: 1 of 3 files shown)


richardlau (approved):

StefanStojanovic:
Since a lot of time passed since this was opened I think it is safe to merge it now. Potentially there will be some problematic PRs in the node and for those, I'll write comments to rebase on the latest main.

PR #4046, opened 2025-03-21
jenkins: add check for checked out commit

richardlau:
Add a check that the gitref being built has not been updated since the build was requested. Requires `COMMIT_SHA_CHECK` to be set to the SHA of the commit to build/test.

### DIFF
--- a/jenkins/scripts/node-test-commit-pre.sh
+++ b/jenkins/scripts/node-test-commit-pre.sh
@@ -29,7 +29,7 @@ git rev-parse $REBASE_ONTO
 
 # COMMIT_SHA_CHECK needs to be set in the job. Check that it looks like
 # a SHA and not some other git ref (e.g. branch ref)
-if ! expr "${COMMIT_SHA_CHECK}" : [0-9a-fA-F]\\+\$ > /dev/null; then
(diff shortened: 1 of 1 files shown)


jenkins/scripts/node-test-commit-pre.sh
```diff
@@ -27,6 +27,13 @@ git status
 git rev-parse HEAD
 git rev-parse $REBASE_ONTO
 
+# COMMIT_SHA_CHECK needs to be set in the job. Check that the gitref
+# that is checked out hasn't been updated since the job was requested.
+if [ "$(git rev-parse HEAD)" != "$(git rev-parse ${COMMIT_SHA_CHECK})" ]; then
+    echo "HEAD does not match expected COMMIT_SHA_CHECK"
+    exit 1
+fi
```
richardlau, line 35: We might be able to change the checkout on L20 above to checkout `$COMMIT_SHA_CHECK` but I don't know if anything will break if the checkout is not a checkout of a branch.

UlisesGascon: We can deploy the change and monitor for errors.

UlisesGascon (approved):

richardlau:
I've added an extra check that `COMMIT_SHA_CHECK` looks like a SHA and not some other gitref (such as a branch, e.g. `refs/pull/<pr id>/head`).

PR #4057, opened 2025-04-01
build: add expected assets for v24

RafaelGSS:
Refs: https://github.com/nodejs/Release/issues/1081


juanarbol (approved):

richardlau:
> I see we have removed the ppc 32 bits, but it looks like it wasn't in the v23.x expected assets either, so I assume that's correct.

I'm confused, we've never produced 32-bit ppc binaries for Node.js.

richardlau:
> > I see we have removed the ppc 32 bits, but it looks like it wasn't in the v23.x expected assets either, so I assume that's correct.
> 
> I'm confused, we've never produced 32-bit ppc binaries for Node.js.

ah you're referring to https://github.com/nodejs/node/commit/44b0e423dc. That's just a tidy up. As mentioned, we've never produced binaries for that, but a long time ago you could build that configuration from source.

richardlau (approved):

RafaelGSS:
Yes, that's exactly what I was talking about. Thanks for the clarification.

Issue #4059, opened 2025-04-02
Ansible: `build gn` task broken


richardlau:
I'm rebuilding rhel9 ppc64_le and s390x machines and Ansible is failing to build `gn`:

e.g.
```console
TASK [gn : build gn] ********************************************************************************************************************************************************************************************
fatal: [test-ibm-rhel8-s390x-4]: FAILED! => {"changed": true, "cmd": "python3 build/gen.py && . /opt/rh/gcc-toolset-11/enable && /home/iojs/build-tools/ninja -C out && out/gn_unittests\n", "delta": "0:00:00.038250", "end": "2025-04-02 08:59:48.858060", "msg": "non-zero return code", "rc": 1, "start": "2025-04-02 08:59:48.819810", "stderr": "Traceback (most recent call last):\n  File \"build/gen.py\", line 960, in <module>\n    sys.exit(main(sys.argv[1:]))\n  File \"build/gen.py\", line 228, in main\n    WriteGNNinja(os.path.join(out_dir, 'build.ninja'), platform, host, options, args_list)\n  File \"build/gen.py\", line 494, in WriteGNNinja\n    if is_gcc(cxx):\n  File \"build/gen.py\", line 239, in is_gcc\n    capture_output=True)\n  File \"/usr/lib64/python3.6/subprocess.py\", line 423, in run\n    with Popen(*popenargs, **kwargs) as process:\nTypeError: __init__() got an unexpected keyword argument 'text'", "stderr_lines": ["Traceback (most recent call last):", "  File \"build/gen.py\", line 960, in <module>", "    sys.exit(main(sys.argv[1:]))", "  File \"build/gen.py\", line 228, in main", "    WriteGNNinja(os.path.join(out_dir, 'build.ninja'), platform, host, options, args_list)", "  File \"build/gen.py\", line 494, in WriteGNNinja", "    if is_gcc(cxx):", "  File \"build/gen.py\", line 239, in is_gcc", "    capture_output=True)", "  File \"/usr/lib64/python3.6/subprocess.py\", line 423, in run", "    with Popen(*popenargs, **kwargs) as process:", "TypeError: __init__() got an unexpected keyword argument 'text'"], "stdout": "", "stdout_lines": []}
```

I think this is a sequencing issue after https://github.com/nodejs/build/pull/3746 -- we need to install `ninja-build` before the `gn` role (as written the `gn` role is being run as a dependency of `build-test-v8` so before `build-test-v8` attempts to install `ninja-build`).

PR #4060, opened 2025-04-02
ansible: fix build dependencies for `gn` role

richardlau:
Install dependencies needed to build `gn` before attempting to build it.

Fixes: https://github.com/nodejs/build/issues/4059


richardlau:
I've tested this against all of the rhel8 ppc64le and s390x test machines, including
- test-osuosl-rhel8-ppc64_le-2
- test-ibm-rhel8-s390x-4

which have been rebuilt from a clean OS image.

UlisesGascon (approved):

PR #4098, opened 2025-06-25
ansible: remove Docker from www

targos:
It's unused since we stopped building the website on that server.


targos:
I'll uninstall Docker from the machine after this is merged.

Issue #4083, opened 2025-05-21
Add macOS 15 CI images with Xcode 16


targos:
V8 builds now need at least Xcode 16.3, which requires us to bump macOS to at least 15.2.

See https://github.com/nodejs/node/issues/58069
Example failing canary release: https://ci-release.nodejs.org/job/iojs+release/10961/

@ryanaslett Is this something that can be done? I don't know if we would have to use the newer version for all jobs or if can keep macOS 13 and 15 in parallel and select a version in the Jenkins jobs based on the Node.js version.

ryanaslett:
Our current mac infra has 2 intel nodes and 2 arm nodes that are serving both release infra and test infra. 

Each node is capable of having 2 VM's at a time on them, for a total of 4 concurrent intel nodes and 4 concurrent arm nodes.

I have set the test infra to never have more than 3 at a time running so that release always has an available node.

If the node version determines which macOS version we test, then it should be relatively simple to do, as we're not trying to test both versions of xcode with the same test (thus not incurring more tests)

ryanaslett:
Update on this:

We need to upgrade the underlying cluster from our current 3.0 to at least 3.2 in order to have Sequoia images available.

Orka is releasing 3.3 on Jun 11, so we can either wait until the new release is out (to avoid having to upgrade again), or, alternately we can ask to upgrade to 3.2 if we need this sooner.

targos:
Waiting SGTM

mcollina:
@ryanaslett what's the status of this?

targos:
Could you make some progress? It's frustrating to see the deadline come closer while I can't do anything.

ryanaslett:
Sorry for the lack of updates, I have been working on this, just kept hitting snags and kept erroneously imagining I was going to update with "here, its done".

The current status: 
I have the arm64 test image built, and its actually running a build on the ci server now:
https://ci.nodejs.org/job/node-test-commit-osx/66838/nodes=osx15-arm64/console

I'll have the x64 image, and the release variants built tomorrow (I have to do weird userid wrangling to allow the x64 ones to write to the shared ccache).

One thing that we're going to need to sort out is how we manage the osx images.. do we need the osx13 builds? We only have 2 arm and 2 intel machines capable of running 2 vms each (so 4 osx, and 4 x64 runners) -> those are shared with release so, effectively we only have 3 osx of each arch available for builds. If we need 2 per build it'll bottleneck quickly.

targos:
[...]

PR #4154, opened 2025-09-19
jenkins: run on macOS 15 from Node.js 25

targos:
Refs: https://github.com/nodejs/build/issues/4083


targos:
This can land only when all the Orka images are deployed.

richardlau (approved):

PR #4174, opened 2025-10-09
ansible: stop promotion if missing assets

RafaelGSS:
As discussed in the release channel, this should prevent people unfamiliar with this state from wrongly promoting a Node.js version.

Refs: https://github.com/nodejs/node/issues/60176


RafaelGSS:
ping @nodejs/build

PR #4234, opened 2026-02-06
aix: install clang

abmusse:
This PR installs precompiled clang binaries from IBM repo: https://github.com/IBM/llvm-project/releases

These changes install the clang frontend but there are some manual steps needed to install the clang backend as well.

If you don't have the clang backend installed you will run into an error like this:

```sh
$ /opt/clang+llvm-20.1.7-powerpc64-ibm-aix-7.2/bin/clang --version
Could not load program /opt/clang+llvm-20.1.7-powerpc64-ibm-aix-7.2/bin/clang:
        Dependent module /usr/lib/libc++.a(shr2_64.o) could not be loaded.
        Member shr2_64.o is not found in archive
```

Here are some refs on what needs to get installed Manually:

- runtime: https://www.ibm.com/support/pages/fix-list-xl-cc-runtime-aix
- utilities: ttps://www.ibm.com/support/pages/ibm-open-xl-cc-utilities-aix-1713

After installing these packages we will need to update dnf:

```sh
/usr/sbin/updtvpkg
```

### DIFF
--- a/ansible/roles/baselayout/tasks/partials/clang/aix.yml
+++ b/ansible/roles/baselayout/tasks/partials/clang/aix.yml
@@ -1,0 +1,10 @@
+---
+#
+# Downloads and installs clang
+#
+
+- name: "clang : extract tarball - aix"
+  ansible.builtin.unarchive:
+    src: https://github.com/IBM/llvm-project/releases/download/llvmorg-20.1.7/clang+llvm-20.1.7-powerpc64-ibm-aix-7.2.tar.xz
+    dest: /opt
+    remote_src: yes
--- a/ansible/roles/baselayout/tasks/partials/clang/aix.yml
+++ b/ansible/roles/baselayout/tasks/partials/clang/aix.yml
@@ -39,7 +39,7 @@
     src: "{{ clang_local.dest }}"
   when: update_clang == True
 
-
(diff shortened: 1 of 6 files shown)


richardlau (commented):
Ansible changes LGTM but we need to document the manual steps.

abmusse:
I've Updated MANUAL_STEPS.md to include info to install clang backend.

richardlau (approved):

UlisesGascon (approved):

ansible/roles/baselayout/tasks/partials/clang/aix.yml
```diff
@@ -0,0 +1,46 @@
...
+  register: clang_unpacked
+  unarchive:
+    dest: "/opt/"
+    list_files: yes
+    remote_src: yes
+    src: "{{ clang_local.dest }}"
+  when: update_clang == True
+
+
+- name: Check if we have a runnable clang
```
abmusse, line 43: This will if clang is runnable, if we don't have the clang runtime libraries installed we will fail. We should add comment here for a breadcrumb for others.

abmusse:
@richardlau 

I think we are good to merge this one now,

richardlau (approved):

PR #4273, opened 2026-03-13
aix: install rust

abmusse:
Add rust, cargo, and rustfmt to the list of packages to install.

### DIFF
--- a/ansible/roles/baselayout/tasks/partials/postinstall/aix.yml
+++ b/ansible/roles/baselayout/tasks/partials/postinstall/aix.yml
@@ -1,0 +1,15 @@
+#
+# AIX post-installation tasks
+#
+
+- name: "rust : create symlinks for Rust binaries"
+  file:
+    src: /opt/freeware/lib/RustSDK/1.88/bin/{{ item }}
+    dest: /opt/freeware/bin/{{ item }}
+    owner: root
+    group: system
+    state: link
+  loop:
+    - rustc
+    - rustfmt
+    - cargo
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -37,7 +37,7 @@ common_packages: [
 # % ansible -m debug -a "var=os"  HOST
 packages: {
   aix: [
-    'bash,cmake,coreutils,curl,gcc-c++,tar,unzip,git,make,sudo,python3-setuptools,python3,rust-1.86.0,cargo-1.86.0,rustfmt-1.86.0',
+    'bash,cmake,coreutils,curl,gcc-c++,tar,unzip,git,make,sudo,python3-setuptools,python3,rust-1.88.0,cargo-1.88.0,rustfmt-1.88.0',
   ],
 
   # Appears to be some issue with the Ansible dnf task on AIX and gcc10-c++, so handle separately.
--- a/ansible/roles/baselayout/vars/main.yml
+++ b/ansible/roles/baselayout/vars/main.yml
@@ -37,7 +37,7 @@ common_packages: [
 # % ansible -m debug -a "var=os"  HOST
 packages: {
   aix: [
-    'bash,cmake,coreutils,curl,gcc-c++,tar,unzip,git,make,sudo,python3-setuptools,python3,rust-1.88.0,cargo-1.88.0,rustfmt-1.88.0',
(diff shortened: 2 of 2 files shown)


richardlau (approved):

UlisesGascon (approved):

richardlau:
It looks like on AIX, `cargo`/`rustc` are not installed into `/opt/freeware/bin` so are not on the PATH, e.g. https://ci.nodejs.org/job/node-test-commit-aix/61998/nodes=aix72-power9/console

It looks like they are in `/opt/freeware/lib/RustSDK/1.88/bin`. We either need to:
- Add this path to the PATH in the Jenkins start script: https://github.com/nodejs/build/blob/522b933410a21dc4abf940d2c2fa3729c4769eb2/ansible/roles/jenkins-worker/templates/aix.rc2.j2#L20
- Add the path into `select-compiler.sh` for AIX.
- Export `CARGO` and `RUSTC` environment variables in `select-compiler.sh` for AIX with fully qualified paths to `cargo`/`rustc`.
- Create symlinks in `/opt/freeware/bin` to `cargo`/`rustc` (i.e. the symlinks would be on the PATH).

sxa (commented):
As I'm sure you'd agree, none of these options are good :'(

I guess that since the version is hardcoded to 1.88 we could legitimately use any of those. I might lean towards the symlinks (mostly to make any manual building/debugging on the machine easier) although I do wonder if there was a good reason the package maintainer decided to keep it isolated like this and not include a symlink.

abmusse:
> As I'm sure you'd agree, none of these options are good :'(
> 
> I guess that since the version is hardcoded to 1.88 we could legitimately use any of those. I might lean towards the symlinks (mostly to make any manual building/debugging on the machine easier) although I do wonder if there was a good reason the package maintainer decided to keep it isolated like this and not include a symlink.

I'm ok with creating the symlinks, just wondering where would best spot to do so in the playbooks? I guess I could do that in select-compiler.sh or the AIX Jenkins start script.

richardlau:
> > As I'm sure you'd agree, none of these options are good :'(
> > I guess that since the version is hardcoded to 1.88 we could legitimately use any of those. I might lean towards the symlinks (mostly to make any manual building/debugging on the machine easier) although I do wonder if there was a good reason the package maintainer decided to keep it isolated like this and not include a symlink.
> 
> I'm ok with creating the symlinks, just wondering where would best spot to do so in the playbooks? I guess I could do that in select-compiler.sh or the AIX Jenkins start script.

Don't create symlinks in `select-compiler.sh` -- We only want to set env variables in there (e.g. setting `PATH`). If you opt for symlinks, create a new `aix.yml` in [`ansible/roles/baselayout/tasks/partials/postinstall`](https://github.com/nodejs/build/tree/main/ansible/roles/baselayout/tasks/partials/postinstall) which gets run after the package installs by https://github.com/nodejs/build/blob/601c84b51fc65bca1fc2c3f4e64097267b8afae3/ansible/roles/baselayout/tasks/main.yml#L114-L122.

sxa (approved):

richardlau (approved):

targos:
Can we merge this and https://github.com/nodejs/build/pull/4286 ?

richardlau:
FWIW https://ci.nodejs.org/job/node-test-commit-aix/62159/nodes=aix72-power9/consoleFull shows that
- cargo/rustc were picked up
```
15:31:26 python3 ./configure --verbose  --dest-cpu=ppc64
15:31:26 Node.js configure: Found Python 3.12.13...
15:31:26 Detected C++ compiler (CXX=g++-12) version: 12.3.0
15:31:26 WARNING: C++ compiler (CXX=g++-12, 12.3.0) too old, need g++ 13.2.0 or clang++ 19.1.0
15:31:26 Detected C compiler (CC=gcc-12) version: 12.3.0
15:31:26 Detected cargo (CARGO=cargo): 1.88.0
15:31:26 Detected rustc (RUSTC=rustc): 1.88.0
```
- crates built
```
15:37:13    Compiling temporal_rs v0.1.0
...
15:37:18    Compiling temporal_capi v0.1.0
```
- temporal tests passed
```
17:39:32 ok 3787 parallel/test-temporal
17:39:32   ---
17:39:32   duration_ms: 258.07900
17:39:32   ...
...
17:39:32 ok 3789 parallel/test-temporal-with-zoneinfo
17:39:32   ---
17:39:32   duration_ms: 337.77000
17:39:32   ...
```

Issue #4144, opened 2025-09-10
Upgrade `ubuntu2204-64` test machines to Ubuntu 24.04


targos:
These ones: https://ci.nodejs.org/label/ubuntu2204-64/

@nodejs/build Do you think it's ok to just replace them? Adding new VMs would be more conservative but I don't know if we can do that with our credits / agreements with the providers.

targos:
See https://github.com/nodejs/build/pull/4145
I started by creating a new VM for testing. Then we can replace the existing ones.

richardlau:
> [#4144 (comment)](https://github.com/nodejs/build/issues/4144#issuecomment-3278713863)
> 
> [@richardlau](https://github.com/richardlau) Can I delete these two containers and Jenkins hosts?

Yes.

richardlau:
> Related to [#4144 (comment)](https://github.com/nodejs/build/issues/4144#issuecomment-3278723257) and [#3563](https://github.com/nodejs/build/issues/3563), it seems like we have two large DO droplets (https://cloud.digitalocean.com/droplets/164538942, https://cloud.digitalocean.com/droplets/164538944) that are completely unused.

Those are from https://github.com/nodejs/build/pull/1982 (the IP addresses match the Ansible inventory changes there), which was a (now abandoned) attempt to overhaul the way we run containerized builds. They can be deleted.

targos:
Working on the DO hosts now.

ryanaslett:
@targos I do have access to Hetzner. Let me see how to make that more available.

richardlau:
The two Ubuntu 24.04 Hetzner machines have been set up via Ansible. I've taken the two Ubuntu 22.04 machines offline in Jenkins and swapped jobs over to the new ones. Assuming no issues come up in the next few days on the new machines we can then give @ryanaslett the ok to remove the Ubuntu 22.04 ones.

richardlau:
> The two Ubuntu 24.04 Hetzner machines have been set up via Ansible. I've taken the two Ubuntu 22.04 machines offline in Jenkins and swapped jobs over to the new ones. Assuming no issues come up in the next few days on the new machines we can then give [@ryanaslett](https://github.com/ryanaslett) the ok to remove the Ubuntu 22.04 ones.

Coverity build failed on the new machines as we have a newer version of Coverity on the new machines (2024.12.1 vs 2024.6.1):
https://ci.nodejs.org/job/node-daily-coverity/3695/console
```
[...]

PR #4079, opened 2025-05-08
ansible: update RHEL 8 machines to Python 3.11

richardlau:
The V8 CI now needs Python >= 3.10. Update to a newer version of Python on the RHEL 8 machines used to run the V8 CI.

Also add `filecheck` dependency for running V8 tests.

Fixes: https://github.com/nodejs/build/issues/4078
Refs: https://chromium-review.googlesource.com/c/chromium/tools/depot_tools/+/6508214
Refs: https://chromium-review.googlesource.com/c/v8/v8/+/6083883

---

### Deployment status

- [x] [test-osuosl-rhel8-ppc64_le-1](https://ci.nodejs.org/computer/test%2Dosuosl%2Drhel8%2Dppc64%5Fle%2D1/)
- [x] [test-osuosl-rhel8-ppc64_le-2](https://ci.nodejs.org/computer/test%2Dosuosl%2Drhel8%2Dppc64%5Fle%2D2/) 
- [x] [test-osuosl-rhel8-ppc64_le-3](https://ci.nodejs.org/computer/test%2Dosuosl%2Drhel8%2Dppc64%5Fle%2D3/) 
- [x] [test-osuosl-rhel8-ppc64_le-4](https://ci.nodejs.org/computer/test%2Dosuosl%2Drhel8%2Dppc64%5Fle%2D4/)
- [x] [test-ibm-rhel8-s390x-1](https://ci.nodejs.org/computer/test%2Dibm%2Drhel8%2Ds390x%2D1/) 
- [x] [test-ibm-rhel8-s390x-2](https://ci.nodejs.org/computer/test%2Dibm%2Drhel8%2Ds390x%2D2/) 
- [x] [test-ibm-rhel8-s390x-3](https://ci.nodejs.org/computer/test%2Dibm%2Drhel8%2Ds390x%2D3/) 
- [x] [test-ibm-rhel8-s390x-4](https://ci.nodejs.org/computer/test%2Dibm%2Drhel8%2Ds390x%2D4/)

PR #4278, opened 2026-03-18
ansible: update benchmark/v8 roles for Ubuntu 24.04

richardlau:
Add role dependencies to the `benchmarking` role to ensure the user is created and github.com keys are present.

Add V8 build dependencies for Ubuntu 24.04.

---

### Deployment

- [x] [test-hetzner-ubuntu2404-x64-1](https://ci.nodejs.org/computer/test%2Dhetzner%2Dubuntu2404%2Dx64%2D1/)
- [x] [test-hetzner-ubuntu2404-x64-2](https://ci.nodejs.org/computer/test%2Dhetzner%2Dubuntu2404%2Dx64%2D2/)

### DIFF
--- a/ansible/roles/benchmarking/meta/main.yml
+++ b/ansible/roles/benchmarking/meta/main.yml
@@ -1,0 +1,7 @@
+---
+
+dependencies:
+  - role: user-create
+  - role: github
+    vars:
+      user_home_dir: "{{ home }}/{{ server_user }}"
--- a/ansible/roles/build-test-v8/tasks/partials/ubuntu2404.yml
+++ b/ansible/roles/build-test-v8/tasks/partials/ubuntu2404.yml
@@ -1,0 +1,16 @@
+---
+
+#
+# Install packages for V8 builds.
+#
+
+- name: install apt packages required to build V8
+  ansible.builtin.apt:
+    name: ['ninja-build']
+    state: present
+
+- name: install Python packages required to build V8
+  ansible.builtin.pip:
+    name: ['httplib2==0.22.0']
+    virtualenv: /home/{{ server_user }}/venv
+    virtualenv_command: python3 -m venv
--- a/ansible/roles/build-test-v8/tasks/partials/ubuntu2404.yml
+++ b/ansible/roles/build-test-v8/tasks/partials/ubuntu2404.yml
@@ -11,6 +11,6 @@
 
 - name: install Python packages required to build V8
   ansible.builtin.pip:
-    name: ['httplib2==0.22.0']
+    name: ['filecheck', 'httplib2==0.22.0', 'six']
     virtualenv: /home/{{ server_user }}/venv
     virtualenv_command: python3 -m venv


richardlau:
Needs more work.

https://ci.nodejs.org/job/richardlau-node-test-commit-v8-linux/770/nodes=benchmark-ubuntu2404-intel-64,v8test=v8test/console is failing with 
```
18:30:28 Failed to fetch file gs://v8-wasm-spec-tests/daf4c75430122590992650885be9ee4e87692c5b for test/wasm-spec-tests/tests.tar.gz. [Err: Traceback (most recent call last):
18:30:28   File "/home/iojs/build/workspace/richardlau-node-test-commit-v8-linux/deps/v8/third_party/depot_tools/external_bin/gsutil/gsutil_4.68/gsutil/gsutil", line 21, in <module>
18:30:28     gsutil.RunMain()
18:30:28   File "/home/iojs/build/workspace/richardlau-node-test-commit-v8-linux/deps/v8/third_party/depot_tools/external_bin/gsutil/gsutil_4.68/gsutil/gsutil.py", line 121, in RunMain
18:30:28     import gslib.__main__
18:30:28   File "/home/iojs/build/workspace/richardlau-node-test-commit-v8-linux/deps/v8/third_party/depot_tools/external_bin/gsutil/gsutil_4.68/gsutil/gslib/__main__.py", line 36, in <module>
18:30:28     from six.moves import configparser
18:30:28 ModuleNotFoundError: No module named 'six.moves'
18:30:28 ]
```
which was seen in https://github.com/nodejs/build/pull/4079#issuecomment-2863682799 with Python 3.12 and solved by dropping down to Python 3.11.

richardlau:
Refs: https://issues.chromium.org/issues/40283283#comment49

FWIW `depot_tools` commit sha is in https://github.com/nodejs/node/blob/f68824a6805a82bd017004443430cf40c3efaf19/deps/v8/DEPS#L248-L249 and earlier than the gsutils update.

As a data point I'm running https://ci.nodejs.org/job/richardlau-node-test-commit-v8-linux/771/nodes=benchmark-ubuntu2404-intel-64,v8test=v8test/ for https://github.com/nodejs/node/pull/61898 and that appears to be working (or at least building).

We have got a problem though -- the default Python 3 on Ubuntu 24.04 is 3.12 and there does not appear to be a package for 3.11 so we cannot use an earlier version. And we still need to be able to run the V8 CI on e.g. Node.js 22 and 24.

richardlau:
> As a data point I'm running https://ci.nodejs.org/job/richardlau-node-test-commit-v8-linux/771/nodes=benchmark-ubuntu2404-intel-64,v8test=v8test/ for [nodejs/node#61898](https://github.com/nodejs/node/pull/61898) and that appears to be working (or at least building).

FTR this got as far as https://github.com/nodejs/node/pull/61898#issuecomment-4056334147 (that comment mentions GCC < 14 but we're using clang19 here and seeing the same issue).

richardlau:
Overriding `depot_tools` on Node.js 22 with the one from the V8 14.6 update PR allows the CI to pass: https://ci.nodejs.org/job/richardlau-node-test-commit-v8-linux/776/nodes=benchmark-ubuntu2404-intel-64,v8test=v8test/
with
```diff
diff --git a/tools/v8/fetch_deps.py b/tools/v8/fetch_deps.py
index 728e48eb14e..185972a90fd 100755
--- a/tools/v8/fetch_deps.py
+++ b/tools/v8/fetch_deps.py
@@ -24,6 +24,7 @@ GCLIENT_SOLUTION = [
     "deps_file"   : "DEPS",
     "managed"     : False,
     "custom_deps" : {
+      "v8/third_party/depot_tools"            : "https://chromium.googlesource.com/chromium/tools/depot_tools.git@284c5ccb591c3de4e9f71be4a4beb5d1916d5383",
       # These deps are already part of Node.js.
       "v8/base/trace_event/common"            : None,
       "v8/third_party/abseil-cpp"             : None,
```

we could alternatively directly modify [`deps/v8/DEPS`](https://github.com/nodejs/node/blob/859c8c761bee1bd395848b7404d3de47b55b8aea/deps/v8/DEPS#L255). Either option would require us to land patches on v25.x/v24.x/v22.x/v20.x branches (although Node.js 20 will be End-of-Life soon). I guess we could also make the Jenkins job patch the dependency, but I feel like that would be less transparent, and we would need to ensure we do not patch unnecessarily (e.g. we should not be patching for newer V8 versions).

@nodejs/v8-update @nodejs/releasers Any thoughts/preferences?
tl;dr We now have Python version constraints with the V8 CI:
- The [bootstrap depot_tools](https://github.com/nodejs/node/blob/0998c37eb64cc9e147d90f3aa031a0d128b6282a/tools/v8/node_common.py#L19-L43) now requires Python 3.11 or later.
- The version of depot_tools used by the V8 build (bootstrapped by the above) in `deps/V8/DEPS` doesn't work with Python 3.12 for all versions of V8 in the in-support Node.js release lines.
- Ubuntu 24.04 defaults to Python 3.12, with no easy way to use an earlier version.

I don't remember if moving from [Ubuntu 22.04 to 24.04](https://github.com/nodejs/build/issues/4144) was toolchain related (@targos ?) which would make it impractical to stay on Ubuntu 22.04 (note that Python 3.11 there is a `rc` version).

targos:
> I don't remember if moving from https://github.com/nodejs/build/issues/4144 was toolchain related

Yes, it was to get access to Clang 19+

richardlau:
> > I don't remember if moving from #4144 was toolchain related
> 
> Yes, it was to get access to Clang 19+

Right, in which case staying on Ubuntu 22.04 is out of the question.

@targos Do you have thoughts on overriding (either DEPS or fetch_deps.py) depot_tools in either the checked in source code (all release lines) or directly in the Jenkins job? (Or have other suggestions to try?)

@ryanaslett was asking if the machine has to be Ubuntu, and I don't think it does, but would be more work to transition to something else (RHEL?).

targos:
> Do you have thoughts on overriding (either DEPS or fetch_deps.py) depot_tools

That sounds good to me. As long as it works, I would take the easiest to maintain route

richardlau:
I've opened https://github.com/nodejs/node/pull/62344.

targos (approved):

Issue #4245, opened 2026-02-13
Add rust toolchain to CI machines


richardlau:
Refs: https://github.com/nodejs/node/issues/58730

To support Temporal in Node.js we will need:
- V8 [14.4](https://github.com/nodejs/node/pull/61002) (or later, e.g. [14.5](https://github.com/nodejs/node/pull/61681)).
- An appropriate Rust toolchain (including [cargo](https://github.com/nodejs/node/blob/main/deps/crates/crates.gyp)) installed on all machines in Jenkins (test and release) used for Node.js 26.

I think ideally we want to aim for Node.js 26.0.0 but technically enabling Temporal in Node.js could be a semver-minor update.

## Tasks

- [x] Discover all machines (and containers) that will need to be updated (FWIW I've started https://ci.nodejs.org/job/node-test-commit/85454/ with `--v8-enable-temporal-support` set as `CONFIG_FLAGS`).
- [ ] Update machines/containers.

### AIX 7.2

- [x] release-ibm-aix72-ppc64_be-1
- [x] test-ibm-aix72-ppc64_be-1
- [x] test-ibm-aix72-ppc64_be-2
- [x] test-ibm-aix72-ppc64_be-3

### Alpine x64

- [x] alpine 3.21 containers
- [x] alpine 3.22 containers

### Debian x64

[Needs 13](https://github.com/nodejs/build/issues/4245#issuecomment-3908238578)
- [x] test-digitalocean-debian13-x64-1
- [x] test-digitalocean-debian13-x64-2

### Fedora x64

- [x] test-digitalocean-fedora41-x64-1
- [x] test-digitalocean-fedora41-x64-2
- [x] test-digitalocean-fedora42-x64-1
- [x] test-digitalocean-fedora42-x64-2

### macOS 15

https://github.com/nodejs/build/issues/4301
- [x] release VMs
- [x] test VMs

### RHEL 8 arm64

- [x] release-osuosl-rhel8-arm64-1
- [x] test-azure-rhel8_container-arm64-1
- [x] test-azure-rhel8_container-arm64-2
- [x] test-osuosl-rhel8_container-arm64-1

### RHEL 8 ppc64le

- [x] release-osuosl-rhel8-ppc64_le-1
- [x] release-osuosl-rhel8-ppc64_le-2
- [x] test-ibm-rhel8-ppc64_le-1
- [x] test-osuosl-rhel8-ppc64_le-1
- [x] test-osuosl-rhel8-ppc64_le-2
[...]

PR #4293, opened 2026-04-09
ansible: add cargo/rust to Alpine containers

richardlau:
Refs: https://github.com/nodejs/build/issues/4245

### DIFF
--- a/ansible/roles/docker/templates/alpine321.Dockerfile.j2
+++ b/ansible/roles/docker/templates/alpine321.Dockerfile.j2
@@ -20,6 +20,7 @@ RUN apk add --no-cache libstdc++
 RUN apk add --no-cache --virtual .build-deps \
         shadow \
         binutils-gold \
+        cargo \
         curl \
         clang19 \
         g++ \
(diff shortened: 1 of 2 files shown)


richardlau:
I've updated all of the containers.
Test build (with `--v8-enable-temporal-support` configure flag): https://ci.nodejs.org/job/node-test-commit-linux/69354/

richardlau:
> I've updated all of the containers. Test build (with `--v8-enable-temporal-support` configure flag): https://ci.nodejs.org/job/node-test-commit-linux/69354/

Debian 12 failed. I'll open another PR to stop running Debian 12 for Node.js 26.

richardlau:
> > I've updated all of the containers. Test build (with `--v8-enable-temporal-support` configure flag): https://ci.nodejs.org/job/node-test-commit-linux/69354/
> 
> Debian 12 failed. I'll open another PR to stop running Debian 12 for Node.js 26.

Actually I don't think we need to do that. The build _will_ fail when explicitly configured with `--v8-enable-temporal-support` since Debian 12 doesn't have `cargo`/`rustc` installed (since they are not a new enough version) but https://github.com/nodejs/node/pull/61806 would detect `cargo`/`rustc` not being present, warn and then disable Temporal by default.

targos (approved):

Issue #4294, opened 2026-04-09
node-test-commit-aix jobs failing to git clone


richardlau:
https://ci.nodejs.org/job/node-test-commit-aix/ jobs have been failing today, all on [test-ibm-aix72-ppc64_be-1](https://ci.nodejs.org/computer/test%2Dibm%2Daix72%2Dppc64%5Fbe%2D1/).

e.g. https://ci.nodejs.org/job/node-test-commit-aix/nodes=aix72-power9/61999/console
```
05:20:48  > git rev-parse --resolve-git-dir /home/iojs/build/workspace/node-test-commit-aix/nodes/aix72-power9/.git # timeout=30
05:20:48 ERROR: Workspace has a .git repository, but it appears to be corrupt.
05:20:48 hudson.plugins.git.GitException: Error performing git command: git rev-parse --resolve-git-dir /home/iojs/build/workspace/node-test-commit-aix/nodes/aix72-power9/.git
05:20:48 	at PluginClassLoader for git-client//org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2864)
...
05:20:48 Caused by: java.io.IOException: Cannot run program "git" (in directory "/home/iojs/build/workspace/node-test-commit-aix/nodes/aix72-power9"): error=0, Failed to exec spawn helper: pid: 31654226, exit value: 127
05:20:48 	at java.base/java.lang.ProcessBuilder.start(Unknown Source)
05:20:48 	at java.base/java.lang.ProcessBuilder.start(Unknown Source)
05:20:48 	at Jenkins v2.541.3//hudson.Proc$LocalProc.<init>(Proc.java:252)
05:20:48 	at Jenkins v2.541.3//hudson.Proc$LocalProc.<init>(Proc.java:221)
05:20:48 	at Jenkins v2.541.3//hudson.Launcher$LocalLauncher.launch(Launcher.java:995)
05:20:48 	at Jenkins v2.541.3//hudson.Launcher$ProcStarter.start(Launcher.java:507)
05:20:48 	at PluginClassLoader for git-client//org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2841)
05:20:48 	... 21 more
05:20:48 Caused by: java.io.IOException: error=0, Failed to exec spawn helper: pid: 31654226, exit value: 127
[...]

PR #4295, opened 2026-04-09
ansible: fix existing Java detection edge case

richardlau:
Add additional check for whether an existing `java` ran successfully as the original check could erroneously match if the Java version being checked for was in an error message (e.g. part of the directory name included in the message).

---

An example (from playbook runs for https://github.com/nodejs/build/issues/4294):
```
TASK [java-base : Check if java is already installed] ***********************************************************************************************************************************************************
[WARNING]: raw module does not support the environment keyword
fatal: [test-ibm-aix72-ppc64_be-1]: FAILED! => {"changed": false, "msg": "non-zero return code", "rc": 6, "stderr": "Shared connection to 169.54.113.142 closed.\r\n", "stderr_lines": ["Shared connection to 169.54.113.142 closed."], "stdout": "Error: dl failure on line 591\r\nError: failed /opt/jdk-17.0.18+8-jre/lib/server/libjvm.so, because Could not load module /opt/jdk-17.0.18+8-jre/lib/server/libjvm.so.\r\nSystem error: Exec format error\r\n", "stdout_lines": ["Error: dl failure on line 591", "Error: failed /opt/jdk-17.0.18+8-jre/lib/server/libjvm.so, because Could not load module /opt/jdk-17.0.18+8-jre/lib/server/libjvm.so.", "System error: Exec format error"]}
...ignoring
```

The original check was passing because `jdk-17.0.18+8` was in stdout from the failing command.

### DIFF
--- a/ansible/roles/java-base/tasks/main.yml
+++ b/ansible/roles/java-base/tasks/main.yml
@@ -50,7 +50,7 @@
 # If we're already using the latest there is no need to do anything.
 - name: check existing adoptopenjdk version is up to date
   set_fact:
-    update_adoptopenjdk: "{{ adoptopenjdk_metadata.json[0].version_data.openjdk_version not in java.stdout }}"
+    update_adoptopenjdk: "{{ java.rc != 0 or adoptopenjdk_metadata.json[0].version_data.openjdk_version not in java.stdout }}"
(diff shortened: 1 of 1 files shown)


sxa (approved):
I'm guessing this has been tested, in which case I approve.

PR #4286, opened 2026-03-30
aix: select clang for Node.26+ builds

abmusse:
Selects clang complier for Node.js 26+ builds

This is expected to land after we merge: https://github.com/nodejs/node/pull/62656

PR #4307, opened 2026-04-22
restrict AIX use of clang to 26+ instead of 25+

sxa:
Draft until tested. Building node 25 with clang results in a build failure after https://github.com/nodejs/build/pull/4286 selected clang for 25+ (like the other platforms) instead of 26+
```
clang++: error: unknown argument '-mno-popcntb'; did you mean '-mno-popcntd'?
clang++: error: unknown argument: '-fno-extern-tls-init'
```
Other option would be to have a separate "if 26 and AIX" section outside the 25+ block but I think this is simpler.

### DIFF
--- a/jenkins/scripts/select-compiler.sh
+++ b/jenkins/scripts/select-compiler.sh
@@ -37,11 +37,14 @@ fi
 if [ "$NODEJS_MAJOR_VERSION" -ge "25" ]; then
   case $NODE_NAME in
     *aix*)
-      echo "Using Clang for Node.js $NODEJS_MAJOR_VERSION"
-      export PATH="/opt/ccache-3.7.4/libexec:/opt/clang+llvm-20.1.7-powerpc64-ibm-aix-7.2/bin/:$PATH"
-      export CC="clang"
-      export CXX="clang++"
-      echo "Compiler set to Clang" `${CXX} -dumpversion`
+      # AIX does not support building v25 with clang so restrict to >25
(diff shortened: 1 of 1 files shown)


jenkins/scripts/select-compiler.sh
```diff
@@ -37,11 +37,14 @@ fi
...
-      echo "Compiler set to Clang" `${CXX} -dumpversion`
+      # AIX does not support building v25 with clang so restrict to >25
+      if [ "$NODEJS_MAJOR_VERSION" -ge "26" ]; then
+        echo "Using Clang for Node.js $NODEJS_MAJOR_VERSION"
+        export PATH="/opt/ccache-3.7.4/libexec:/opt/clang+llvm-20.1.7-powerpc64-ibm-aix-7.2/bin/:$PATH"
+        export CC="clang"
+        export CXX="clang++"
+        echo "Compiler set to Clang" `${CXX} -dumpversion`
+      fi
       return
```
richardlau: You may need the `return` to be inside the `if` block for the fall through to work as expected.

sxa: Done.

richardlau (approved):

abmusse:
Thanks for fixing this up @sxa!

sxa:
https://ci.nodejs.org/job/node-stress-single-test/nodes=aix72-power9/687/console is runnign through ok with 25 so I think this is ok - will also keep an eye on the 26 runs.

Issue #4265, opened 2026-03-05
Upgrade to Clang 20 and Rust 1.88 


miladfarca:
RHEL `llvm-toolset` and `rust-toolset` packages are version‑coupled. For example, installing `llvm-toolset-19.1.7` brings in Rust `1.84.1`. Currently the ppc64 and s390x platforms are using Rust `1.84`, which still ships with `libadler`. Starting with Rust `1.86`, this changes to `libadler2`.

The Chromium build repository [currently has this logic](https://source.chromium.org/chromium/chromium/src/+/e3282c21eb10c23dc6724e63c3b7f61e9b33bd1e:build/rust/std/BUILD.gn;l=58) to work around the mismatch, since our toolchain is older than Chromium upstream:
```
if (rustc_nightly_capability) {
  stdlib_files += [ "adler2" ]
} else {
  stdlib_files += [ "adler" ]
}
```
The issue now is that AIX is also moving to Rust `1.86`, which uses `libadler2`, but AIX still falls into the else branch because it does not use the Chromium toolchain. That means this logic no longer works for AIX.

To avoid having to maintain `libadler` for different Rust versions upstream, we need to upgrade our toolchain to Rust `1.86+` and fully remove that if/else block.

On RHEL, upgrading Rust to `1.88` also means upgrading our llvm-toolset (Clang) to version `20`, since Rust and LLVM are version‑coupled there.

Furthermore Libc++ now only supports Clang 20 and later: https://github.com/llvm/llvm-project/pull/165684
Chromium upstream is also already using Clang 23, so an upgrade may need to be performed down the road regardless.

targos:
I guess that's ok if we do it on specific systems when there's a good reason like this, but I'm not sure we can easily bump the official requirement to Clang 20. For example on macOS, stable Xcode is still on Clang 19:

<img width="598" height="234" alt="Image" src="https://github.com/user-attachments/assets/0a21af68-a314-4656-94e2-70524ecb1651" />

https://en.wikipedia.org/wiki/Xcode#Toolchain_versions

miladfarca:
`adler` is now removed from chromium build and `adler2` is being used unconditionally:
http://crrev.com/c/7704595
For building V8 `rustc >= 1.86` will now be required.

richardlau:
We've landed V8 14.6 on main where the V8 CI passed with Clang 19 and Rust 1.88. 

[...]

PR #4314, opened 2026-04-24
ansible: update gn on Linux ppc64le and s390x

richardlau:
Update to a newer version of `gn` for compatibility with more recent versions of V8.

Refs: https://github.com/nodejs/node/pull/62572#issuecomment-4305823842

---

Had been deployed onto all of the Linux ppc64le and s390x machines used to build V8.

### DIFF
--- a/ansible/roles/gn/tasks/main.yml
+++ b/ansible/roles/gn/tasks/main.yml
@@ -41,7 +41,7 @@
 # Requires a C++17 compiler.
 - name: build gn
   ansible.builtin.shell: |
-    python3 build/gen.py && \
+    python3 build/gen.py --allow-warning && \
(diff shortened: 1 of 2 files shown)


targos (approved):

sxa (approved):

richardlau:
FTR this gets us past the first error in https://github.com/nodejs/node/pull/62572#issuecomment-4305823842, but there is more work to do to get the V8 CI passing on Linux ppc64le and s390x with V8 14.8. I was working on those additional changes but have spent most of this week looking into the macOS Temporal issue so haven't been able to progress (I've a good idea of what [needs to happen](https://github.com/nodejs/build/issues/4265), but FYI I'll be away next week).

Issue #4304, opened 2026-04-21
Update Jenkins to v2.555.3


targos:
https://www.jenkins.io/doc/upgrade-guide/2.555/

Most notable change is that it now requires Java 21 for controllers and agents.
We have **a lot** of agents running on Java 17!

Procedure (as done on ci-release):
- Make sure all agents have Java 25 or 21
- Update plugins from the Jenkins UI and restart Jenkins
- Update system and Jenkins (apt upgrade)
- Update plugins again from the Jenkins UI and restart Jenkins
- Reboot the server

richardlau:
Do we know if it includes the fix for the issue in https://github.com/nodejs/build/issues/4247?

cc @sxa

richardlau:
> Most notable change is that it now requires Java 21 for controllers and agents. We have **a lot** of agents running on Java 17!

JVM versions are listed on https://ci.nodejs.org/computer/

richardlau:
I think the post-17 Long Term Support versions of Java are 21 and 25, so where possible we should upgrade to 25 where available.

sxa:
Yep that's correct - I've got a couple of jenkins agents running 25 so that should be the default choice now with 25 being LTS.
I'll see if I can confirm for sure but yes 2.555.x should include the fix. We'll need to decide if we want to re-enable the restart or not. As I mentioned somewhere else I'm not sure it's strictly required...

sxa:
We still have Debian11/arm32 in the CI (e.g. [test-azure-debian11-container-armv7l-1](https://ci.nodejs.org/computer/test%2Dazure%2Ddebian11%5Fcontainer%2Darmv7l%2D1/builds) which is used for the fanned arm32 jobs. Debian 11 does not have any JDK later than 17 in the repositories.

Options:
1. Remove Debian from the test matrix
2. Update those containers to a later Debian which has a jdk21 available (12 doesn't so it would have to be 13/trixie)
3. I do have what amounts to an "unofficial jdk25" build which was done using the Temurin build scripts and environment (Noting that temurin stopped producing arm32 builds after JDK17)  which we could use for the Debian11 agents, although it wouldn't be automatically updated. That would be a path for allowing ust to continue with the systems we have.

sxa:
jdk-21 is available for IBMi but according to @abmusse is not currently installed on our systems: https://www.ibm.com/support/pages/download-installation-and-usage-java-21-ibm-i-os

sxa:
[...]

PR #4323, opened 2026-05-06
ansible: use adoptium 25 instead of 17

sxa:
Required for the next jenkins update https://github.com/nodejs/build/issues/4304

### DIFF
--- a/ansible/roles/java-base/vars/main.yml
+++ b/ansible/roles/java-base/vars/main.yml
@@ -39,5 +39,5 @@ adoptopenjdk: {
 
 adoptopenjdk_arch: "{{ adoptopenjdk[os+'_'+arch].arch | default(ansible_architecture) }}"
 adoptopenjdk_os: "{{ adoptopenjdk[os+'_'+arch].os | default(ansible_system | lower) }}"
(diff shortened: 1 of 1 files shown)


targos (approved):

sxa:
Appears to have worked in a playbook deployment to one of the (currently offline as discussed with targos in another issue!) AIX 7.3 system.

[test-ibm-aix73-ppc64_be-1](https://ci.nodejs.org/computer/test%2Dibm%2Daix73%2Dppc64%5Fbe%2D1) is running with java 25 for now so I believe this is good and can be deployed to the other AIX machines now. I don't /think/ we're using this anywhere else (I don't believe we need it for any arm32 hosts which would be the other possibility) so this should be clean to land.

sxa:
Having some problems deploying this via the playbooks on machines that don't already have rust on, but that's not related to this PR. If necessary we can deploy manually. The three IBM AIX72 POWER9 machines have been upgraded.

sxa:
Merging since this has already been deployed on a few of the machines and it's failing on others for unrelated reasons which are [being worked separately](https://github.com/nodejs/build/issues/4326).

Issue #4230, opened 2026-02-03
AIX 7.3 testing strategy


sxa:
At the moment we are exclusively testing AIX PRs against AIX 7.2.
AIX 7.3 has been available since December 2021 and we are not performing any testing on it.
This issue is to decide what approach we should use - how many machines we require etc.

I have confirmed that we can build node.js on AIX 7.3 and testing verification will be performed under 

Assuming this works suitably well, we should decide on how many machines we need of AIX 7.2 and 7.3 in order to keep the CI happy.

Performance testing of the builds with different concurrency (`make -j` values) has been performed in https://github.com/nodejs/build/issues/4226

At present we have four AIX 7.2 test machines - two at OSUOSL and two at IBM. There is another 7.2 system at OSUOSL waiting to be reinstalled and I would propose that we get it reinstalled with AIX 7.3.

We currently have two IBM-hosted machines running AIX 7.3. An older one currently with 8GiB of RAM that is already in the CI at [test-ibm-aix73-ppc64_be-1](https://ci.nodejs.org/computer/test%2Dibm%2Daix73%2Dppc64%5Fbe%2D1) and the new one that I have been experimenting with recently at [test-ibm-aix73-ppc64_be-3](https://ci.nodejs.org/computer/test%2Dibm%2Daix73%2Dppc64%5Fbe%2D3)

The other decision to make is whether we need a full set of AIX 7.3 machines to test on or whether we just allow each PR test to run on any AIX machine and let it randomly choose between 7.2 and 7.3. This approach would mean we likely wouldn't need as many systems in the set.

sxa:
With the changes to the playbooks and documentation under https://github.com/nodejs/build/pull/4340 I believe we're in a good shape to add AIX 7.3 testing into the [node-test-commit-aix job](https://ci.nodejs.org/job/node-test-commit-aix). I'm doing some performance tests to ensure that we have a suitable `JOBS` value for these systems so as not to overwhelm them with a plan to make them live tomorrow.

We currently have five machines. My intention is to enable 3 and 5 for now.

Machine | Specification | Notes
---|---|---
test-ibm-aix73-ppc64_be_1 | IBM,9009-22A, 4 PV_9_Compat CPUs, 8GiB | Unknown setup, no clang/rust
test-ibm-aix73-ppc64_be_2 | IBM,9009-22A, 4 PV_9_Compat CPUs, 8GiB | Unknown setup, no clang/rust
[...]

PR #4344, opened 2026-05-20
jenkins: enable node 26+ (clang) builds on AIX 7.3

sxa:
Ref: https://github.com/nodejs/build/issues/4230#issuecomment-4496631570 which is why I don't want to enable this for earlier versions yet.

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -50,6 +50,7 @@ def buildExclusions = [
   // AIX ---------------------------------------------------
   [ /^aix72-ppc64/,                     anyType,     gte(26) ], // Power 8 was dropped in v26
   [ /^aix72-power9/,                    releaseType, lt(26)  ], // Build releases on Power 9 for >=26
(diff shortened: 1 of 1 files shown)


jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -50,6 +50,7 @@ def buildExclusions = [
   // AIX ---------------------------------------------------
   [ /^aix72-ppc64/,                     anyType,     gte(26) ], // Power 8 was dropped in v26
   [ /^aix72-power9/,                    releaseType, lt(26)  ], // Build releases on Power 9 for >=26
+  [ /^aix73-power9/,                    releaseType, lt(26)  ], // Build releases on Power 9 for >=26
```
richardlau, line 53: We don't build any releases on AIX 7.3. Did you mean `testType` (or `anyType`)?

sxa:
Cancelling this PR as I think I have a workaround for the GCC builds (32-bit compiler) on the AIX 7.3 systems so this restrictions is not required

PR #4349, opened 2026-05-26
inventory: add AIX 7.3 machines

sxa:
1 and 2 are not in active use. 4 (which has lower RAM and CPU) currently has some test failures so is not currently live in jenkins (Ref [this issue](https://github.com/nodejs/node/issues/61660))


richardlau (approved):

sxa:
Secrets now in place (for all except aix73-2 which doesn't have one) so merging this.

PR #4364, opened 2026-06-12
ansible: update sharedlibs containers

### LINKED ISSUE #4297: planned ubuntu2404_sharedlibs updates
Once Node.js 20 reaches End-of-Life, we can remove ICU 69 and 71 from the sharedlibs containers as the [minimum ICU for Node.js 22.0.0](https://github.com/nodejs/node/blob/v22.0.0/tools/icu/icu_versions.json) and later is ICU 73.
https://github.com/nodejs/build/blob/d220fb457a8882c1e621a8b0321b271e90afaa4b/ansible/roles/docker/templates/ubuntu2404_sharedlibs.Dockerfile.j2#L44-L46

I was also talking in Slack with @panva about OpenSSL versions. OpenSSL 3.1 and 3.2 are now End-of-Life and we can drop those (keeping 3.0 and 3.5 which are Long Term Support OpenSSL releases). We can also add OpenSSL 4.0 after it releases ([expected later this month, April 2026](https://openssl-library.org/roadmap/index.html)) on the basis that it will be a stepping stone towards whatever LTS OpenSSL comes after 3.5.

richardlau:
Remove obsolete ICU versions that we no longer test with. 
Removed OpenSSL 3.1 and 3.2.
Update versions of OpenSSL 3.0 and 3.5.
Add OpenSSL 4.0.

Fixes: https://github.com/nodejs/build/issues/4297

---

I'll set up testing in the CI with OpenSSL 4.0 after the security releases next week.


targos (approved):
RSLGTM

PR #4428, opened 2026-08-13
ansible: add alpine/musl builds to expected assets

sxa:
Part of the process for adding Alpine into the release builds.
Does the merging of this have to be explicitly co-ordinated with the addition to `iojs+release`?

### DIFF
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -24,6 +24,9 @@ def buildExclusions = [
   [ /rhel8-ppc64le/,                  anyType,     gte(26) ], // Power 8 was dropped in v26
   [ /rhel8-power9le/,                 releaseType, lt(26)  ], // Power 8 was dropped in v26
 
+  // Alpine/musl -------------------------------------------
+  [ /alpine-x64/,                     releaseType, le(22)  ], // Only release on v24+ for now pending test changes
+
   // ARM  --------------------------------------------------
   [ /^cross-compiler-rhel8-armv7-gcc-8-glibc-2.28/,  anyType, gte(22) ],
   [ /^cross-compiler-rhel8-armv7-gcc-10-glibc-2.28/, anyType, gte(24) ],
--- a/jenkins/scripts/VersionSelectorScript.groovy
+++ b/jenkins/scripts/VersionSelectorScript.groovy
@@ -25,7 +25,7 @@ def buildExclusions = [
   [ /rhel8-power9le/,                 releaseType, lt(26)  ], // Power 8 was dropped in v26
 
   // Alpine/musl -------------------------------------------
-  [ /alpine-x64/,                     releaseType, le(22)  ], // Only release on v24+ for now pending test changes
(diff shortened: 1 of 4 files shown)


richardlau (approved):

richardlau:
> Part of the process for adding Alpine into the release builds. Does the merging of this have to be explicitly co-ordinated with the addition to `iojs+release`?

It needs to happen before any release that would include the additional assets. If it doesn't releasers would need to bypass the warning from the promotion script about extra unexpected assets.

sxa:
> > Part of the process for adding Alpine into the release builds. Does the merging of this have to be explicitly co-ordinated with the addition to `iojs+release`?
> 
> It needs to happen before any release that would include the additional assets. If it doesn't releasers would need to bypass the warning from the promotion script about extra unexpected assets.

OK - so wouldn't affect the nightlies and other non-release builds - only when release promotion occurs.

sxa:
Putting back into draft pending https://github.com/nodejs/node/pull/65282 being merged back into the release lines

sxa:
The checks are failing but it looks like it's checking older releases which won't have the artifacts. It's also complaining about `foo.tar.gz` and `bar.tar.xz` 
```
          ⚠  The following assets are expected for v22.x but are currently missing from R2 staging:
            • node-v22.12.0-linux-x64-musl.tar.gz
            • node-v22.12.0-linux-x64-musl.tar.xz
         ✖  The following assets were already promoted in R2 but are not expected for v22.x:
            • foo.tar.gz
            • bar.tar.xz
```

sxa:
I feel it's better to skip the v22.x change temporarily for now, otherwise there will be a lot of updating of the tests to handle the "unexpected" missing alpine changes which will make most of the testing completely pointless. Perhaps once we've done one release with Alpine we can modify the tests to use a newer release and have the expected numbers and do the changes for v22.x. Should be ok as long as we do one release on 26 or 24 before 22.

sxa:
Going back into draft as I think I'm going to disable v22.x for Alpine in this PR to make sure it doesn't cause any problems.

sxa:
v22.x excluded from the artefact updates and excluded from the version selector.

targos (approved):

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -24,6 +24,9 @@ def buildExclusions = [
   [ /rhel8-ppc64le/,                  anyType,     gte(26) ], // Power 8 was dropped in v26
   [ /rhel8-power9le/,                 releaseType, lt(26)  ], // Power 8 was dropped in v26
 
+  // Alpine/musl -------------------------------------------
+  [ /alpine-x64/,                     releaseType, le(22)  ], // Only release on v24+ for now pending test changes
```
richardlau: We don't have a `le` function (see above in this file).
```suggestion
  [ /alpine-x64/,                     releaseType, lt(24)  ], // Only release on v24+ for now pending test changes
```

richardlau (approved):

juanarbol (approved):

PR #4440, opened 2026-08-26
use v24.20.0 instead of v22.12.0 for checks. Enable musl on v22

sxa:
Update tests to use v24.20.0 which has been published with musl and update v22.x so it expects the musl versions too.
This is required because the currently published v22.12.0 used for the tests [caused problems](https://github.com/nodejs/build/pull/4428#issuecomment-5330049128) when the tests were inconsistent with the version being checked. Now that a version with musl is published we can make this change.

Follow up to https://github.com/nodejs/build/pull/4428

Noting that the file sizes in the files are not correct, but they were not correct for v22.12.0 so are not verified by the checks.


sxa:
FYI @juanarbol since we should aim to release musl versions on v22.x as well as v24 and v26 as it will make the node-docker scripts less complex if we don't have to handle different conditions, and https://github.com/nodejs/node/pull/63737 did not intend to restrict it to a subset of the active release lines.

jenkins/scripts/VersionSelectorScript.groovy
```diff
@@ -25,7 +25,7 @@ def buildExclusions = [
   [ /rhel8-power9le/,                 releaseType, lt(26)  ], // Power 8 was dropped in v26
 
   // Alpine/musl -------------------------------------------
-  [ /alpine-x64/,                     releaseType, lt(24)  ], // Only release on v24+ for now pending test changes
+  [ /alpine-x64/,                     releaseType, lt(22)  ],
```
targos: Everything below 22 is EoL so this can be deleted.

targos (approved):

jenkins/scripts/VersionSelectorScript.groovy
sxa: I'll be completely honest I left that in purely because after spotting that there was one other `lt(22)` clause in the file on one of the shared library selectors.
But I agree so have now removed it.

sxa:
Now that the 26.8.1 release is out with Alpine this could be updated to that version (and the sizes fixed!) to give longer use of the tests on a live version but that can be done in a separate update and the important thing here is to add the v22 support to the release pipeline.

Issue #4457, opened 2026-09-05
Proposal: run SmartOS and AIX in node-daily-master instead of node-test-pull-request


codebytere:
I pulled a month of `node-test-pull-request` history out of the Jenkins API (450 runs across 167 PRs, 2026-08-06 to 2026-09-05) to see where PR CI time goes. Most of it is waiting for an executor on two platforms rather than running tests, so i'd like to propose running SmartOS and AIX from `node-daily-master` only and dropping them from the PR job.

| job | runs | median | p90 | of which waiting for an executor (median / p90) | slowest configuration's own run (median) |
| --- | ---: | ---: | ---: | ---: | --- |
| node-test-commit-smartos | 362 | 246 min | 453 min | ~215 / ~409 min | 46 min (smartos23-x64) |
| node-test-commit-aix | 407 | 152 min | 334 min | ~87 / ~261 min | 69 min (aix73-power9) |
| node-test-commit-linux (10 configs) | 421 | 94 min | 305 min | ~22 / ~194 min | 62 min (rhel8-x64) |
| node-test-commit-osx | 433 | 91 min | 188 min | ~5 / ~105 min | 65 min (macos15-x64) |
| node-test-commit-linux-containered | 435 | 66 min | 154 min | ~17 / ~91 min | 43 min |
| node-test-commit-windows-fanned | 428 | 58 min | 192 min | | |
| node-test-commit-arm | 431 | 52 min | 112 min | ~0 / ~12 min | 45 min (rhel8-arm64) |
| node-test-commit-plinux | 442 | 35 min | 69 min | ~0 / ~22 min | 33 min |
| node-test-commit-linuxone | 447 | 30 min | 51 min | ~0 / ~6 min | 30 min |
| node-test-linter | 445 | 15 min | 20 min | | |
| **node-test-pull-request, end to end** | 437 | **245 min** | **577 min** | | |

"Waiting for an executor" is the matrix job's duration minus its slowest configuration's execution time (the API doesn't expose queue time for matrix children directly); aborted runs are excluded. Over the same window 13% of PR runs finished SUCCESS and another 18% UNSTABLE, the average PR needed 2.7 runs, and when exactly one child job was non-green in a run it was linuxone 34 times (nearly all yellow from marked-flaky tests), aix 19, osx 17, linux 17, windows 9, smartos 4. Raw per-configuration numbers (exec median/p90, failure rate) are available if useful.

[...]

PR #4458, opened 2026-09-06
jenkins: skip SmartOS and AIX for pull request test runs

codebytere:
Implements the proposal in #4457 in the one place this repo controls, `VersionSelectorScript.groovy`.

`node-test-pull-request` passes `GIT_REMOTE_REF=refs/pull/N/head` down through `node-test-commit` to every sub-job, while `node-daily-*`, `node-release-candidate` and the other upstream triggers pass a branch ref (checked against the parameters of recent `node-test-commit-smartos` builds from both). This adds a `prTestType` predicate keyed on that and excludes the `smartos*` and `aix7*` labels for PR-triggered test runs only. `node-test-commit-smartos` and `-aix` are still triggered by the multijob but select no configurations, which the matrix strategy reports as SUCCESS within seconds ([`BaseMES.run`](https://github.com/jenkinsci/matrix-groovy-execution-strategy-plugin/blob/master/src/main/groovy/org/jenkinsci/plugins/BaseMES.groovy#L33-L91) starts from SUCCESS and only folds in configuration results); daily, staging-daily and release runs are unchanged.

Dry-ran the script with emulated `parameters` / `combinations` bindings: a v27 PR run keeps `rhel8-x64`, `macos15-x64`, `rhel9-s390x` and drops `smartos23-x64`, `aix72-power9`, `aix73-power9`; the same labels with `refs/heads/main`, with no ref, or with `buildType = 'release'` keep everything. What i couldn't check from outside is whether either job has a parent-level post-build step that objects to zero configurations (a JUnit publisher, say); if so that wants "do not fail on empty results" alongside this.

Refs: https://github.com/nodejs/build/issues/4457

---

Disclosure: the code and this description were written by Claude Code, directed and reviewed by @codebytere.


richardlau (changes requested):
I object to removing AIX from pull request test jobs (since we release binaries for it), but putting that to one side, this isn't the best way to exclude platforms from PR CIs as the parent jobs will still be run (this only exclude the matrix axes).

For e.g. IBM i (which is not run on PRs), we do not include the job in the `node-test-pull-request` Multijob and instead include it in `node-daily-master`.

PR #4472, opened 2026-09-16
jenkins,win: move clcache cache out of user profile

### LINKED ISSUE #4471: Windows CI: clcache fails with "[WinError 5] Access is denied" on stats.txt
`node-compile-windows` has been failing since 2026-09-11 with:

```
Error: [WinError 5] Access is denied: 'C:\Users\Administrator\clcache\stats.txt.new' -> 'C:\Users\Administrator\clcache\stats.txt'
```

### Scope

Per the reliability reports this first appears on 2026-09-11 and has recurred every day since — 20 PRs over 6 days, on all six `test-rackspace-win2022_vs2022-x64-*` machines and all four compile configs (`win-vs2022`, `-x86`, `-arm64`, and `node-compile-windows-debug`). The v22.23.3 proposal (nodejs/node#65448) is among the affected runs.

Reports: [09-11](https://github.com/nodejs/reliability/blob/main/reports/2026-09-11.md) · [09-12](https://github.com/nodejs/reliability/blob/main/reports/2026-09-12.md) · [09-13](https://github.com/nodejs/reliability/blob/main/reports/2026-09-13.md) · [09-14](https://github.com/nodejs/reliability/blob/main/reports/2026-09-14.md) · [09-15](https://github.com/nodejs/reliability/blob/main/reports/2026-09-15.md) · [09-16](https://github.com/nodejs/reliability/blob/main/reports/2026-09-16.md)

### Where it fails

```python
# clcache/__main__.py
@contextlib.contextmanager
def atomicWrite(fileName):
    tempFileName = fileName + '.new'
    with open(tempFileName, 'w') as f:
        yield f
    os.replace(tempFileName, fileName)   # <- WinError 5
```

### Known upstream bug, and upstream is archived

frerich/clcache#342 reports the identical error. It is still open, and the repository was archived on 2019-12-11, so no upstream fix is coming. All 81 forks are stale too.

### The trigger, though, is on our side

That thread identifies the cause as the cache living under the user profile, and the confirmed workaround is to point `CLCACHE_DIR` elsewhere:

> you should move clcache dir, not clcache binary. It can be configured through `CLCACHE_DIR` environment

> It works with threads > 4 after I set `CLCACHE_DIR=C:\clcache_dir`

`jenkins/scripts/windows/compile.cmd` sets `CLCACHE_BASEDIR` (path normalisation) but never `CLCACHE_DIR`, so the cache defaults to `%USERPROFILE%\clcache` = `C:\Users\Administrator\clcache` — exactly the path in the error.

Possibly related: clcache creates its lock in the session-local namespace (`'Local\\' + mutexName`) while the cache directory is shared machine-wide.

### Options

1. `DISABLE_CLCACHE=1` — immediate stop-gap; the opt-out already exists in `compile.cmd`.
2. Set `CLCACHE_DIR` outside the profile — the upstream-verified workaround. PR: #4472
3. Longer term, move MSVC builds to ccache. The pieces look like they may already be in place: the ansible role installs `C:\ccache\cl.exe` "for MSVC", `vcbuild.bat ccache <path>` is not clang-specific, and `tools/msvs/props_4_ccache.props` exists — `compile.cmd` just gates ccache behind `if %not_clang% equ 0`. Is there a reason MSVC was kept on clcache? You'll know better than me.

### Not verified

I can't read ci.nodejs.org console logs (anonymous access is denied), so this is reconstructed from the reliability reports and the clcache sources — I haven't seen a full traceback. I also don't know how many executors run per Windows machine, which would matter for the concurrency theory.

P4suta:
Fixes #4471

`compile.cmd` sets `CLCACHE_BASEDIR` but not `CLCACHE_DIR`, so clcache stores its cache in `%USERPROFILE%\clcache`. That is the configuration reported in frerich/clcache#342 to fail with:

```
Error: [WinError 5] Access is denied: '...\clcache\stats.txt.new' -> '...\clcache\stats.txt'
```

and setting `CLCACHE_DIR` to a path outside the profile is the workaround confirmed by several people in that thread. Upstream is archived, so this has to be handled on our side.

Notes:

* The path is arbitrary — anything outside the user profile should do. `C:\clcache_dir` just matches the value used upstream. Happy to rename it; it is a little close to `C:\clcache`, which holds the installation.
* clcache creates the directory itself, but if the Jenkins account can't create it at the drive root, the ansible role could pre-create it instead.
* This abandons the existing cache, so the first builds on each machine will be cold. I think that's an acceptable trade against the current failure rate, but say the word if you'd rather migrate the directory.
* **Untested** — I have no access to the Windows workers or to CI console logs. Please treat CI as the source of truth.
* Assisted by Claude Code for the investigation; I checked the diff and the upstream issue myself.


StefanStojanovic (approved):
Approving this so it can be landed and tested in the CI. If it fails, we'll set `DISABLE_CLCACHE` to disable clcache completely.

richardlau (approved):

